PR #19, opened 2020-01-21
Define system and shared error code; remove unnecessary runtime abort helpers.

anorth:
FYI @jzimmerman @frrist


actors/runtime/runtime.go
```diff
@@ -32,23 +32,16 @@ type Runtime interface {
...
-	// Calls Abort with InvalidArguments_User.
-	AbortArgMsg(msg string)
+	// Halts execution upon an error from which the actor cannot recover. This method does not return.
+	// State changes will be rolled back, including any made by the caller.
+	// The message and args are for diagnostic purposes and do not persist on chain. They should be suitable for
+	// passing to fmt.Errorf(msg, args...).
+	Abort(errExitCode exitcode.ExitCode, msg string, args ...interface{})
 
 	// Calls Abort with InconsistentState_User.
+	// TODO: replace all call sites with Abort(exitcode, msg, ...)
```
zixuanzh: looks like this has been replaced

actors/runtime/exitcode/common.go
```diff
@@ -0,0 +1,18 @@
...
+	ErrIllegalArgument = FirstActorErrorCode + iota
+	// Indicates a requested resource does not exist.
+	ErrNotFound
+	// Indicates an action is disallowed.
+	ErrForbidden
+	// Indicates a balance of funds is insufficient.
+	ErrInsufficientFunds
+
+	// An error code intended to be replaced by different code structure or a more descriptive error.
+	ErrPlaceholder = ExitCode(1000)
```
zixuanzh: most ErrPlaceholder use sites are related state not found in `ipld.Get`, replaced with ErrNotFound?

actors/runtime/runtime_util.go
```diff
@@ -64,7 +65,7 @@ func RT_MinerEntry_ValidateCaller_DetermineFundsLocation(rt Runtime, entryAddr a
 		return ownerAddr
 	} else {
 		if entrySpec == MinerEntrySpec_MinerOnly {
-			rt.AbortArgMsg("Only miner entries valid in current context")
+			rt.Abort(exitcode.ErrPlaceholder, "Only miner entries valid in current context")
```
zixuanzh: use ErrForbidden instead?

actors/runtime/exitcode/reserved.go
```diff
@@ -0,0 +1,65 @@
...
+	SysErrOutOfGas = ExitCode(7)
+
+	// Indicates a message execution is forbidden for the caller.
+	SysErrForbidden = ExitCode(8)
+
+	// Indicates actor code performed a disallowed operation. Disallowed operations include:
+	// - mutating state outside of a state acquisition block
+	// - failing to invoke caller validation
+	// - aborting with a reserved exit code (including success or a system error).
+	SysErrorIllegalActor = ExitCode(9)
```
zixuanzh: Is this the equivalent of `RuntimeAPIError` before the change?

actors/runtime/exitcode/reserved.go
```diff
@@ -0,0 +1,65 @@
...
+	// to ignore errors).
+	SysErrorSubcallAbort = ExitCode(11)
+
+	// Reserved exit codes, do not use.
+	SysErrorReserved1 = ExitCode(12)
+	SysErrorReserved2 = ExitCode(13)
+	SysErrorReserved3 = ExitCode(14)
+	SysErrorReserved4 = ExitCode(15)
+)
+
```
zixuanzh: do we not need `RuntimeAssertFailure` because `Assert` will just halt the VM and not return?

actors/runtime/runtime.go
anorth: No there are many more of these

actors/runtime/exitcode/common.go
anorth: THey're all actor state loading, which I intend to clean up entirely.

actors/runtime/runtime_util.go
anorth: This one indicates a programming error, not a caller error. I intend to rewrite it to be unnecessary, i.e make the error impossible.

actors/runtime/exitcode/reserved.go
anorth: Yeah, pretty close

actors/runtime/exitcode/reserved.go
anorth: It was un-used, and yes assert failures don't go on chain.

PR #22, opened 2020-01-21
Add minimal tracing endpoint to Runtime interface

anorth:
This is sufficient for all current tracing in Lotus actors. We can extend it if necessary.

The Span interface may be implemented directly by the opencensus Span type.

### DIFF
--- a/actors/runtime/runtime.go
+++ b/actors/runtime/runtime.go
@@ -92,10 +92,13 @@ type Runtime interface {
 	// Provides the system call interface.
 	Syscalls() Syscalls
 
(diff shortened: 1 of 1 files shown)


actors/runtime/runtime.go
```diff
@@ -89,6 +89,16 @@ type Runtime interface {
 
 	// Provides the system call interface.
 	Syscalls() Syscalls
+
+	// Starts a new tracing span. The span must be End()ed explicitly, typically with a deferred invocation.
+	StartSpan(name string) TraceSpan
+}
```
frrist: Is the idea here that we will keep the context of spans internal to the vm?

PR #26, opened 2020-01-21
Add exit code to Runtime.Send() and rationalize the interface down to one send method.

### LINKED ISSUE #20: Change send semantics to allow actor to intercept failure of subcall
At present, any subcall failure aborts the whole tree: the sender has no means to trap a failure.

We'll change this to propagate the subcall exit code to the caller.

anorth:
This is mostly a mechanical change, preserving the existing semantics, with the exception of the multisig actor. The final multisig approval will now succeed even if the message that is sent fails.

Closes #20


anorth:
cc @icorderi

PR #27, opened 2020-01-21
feature: add cbor gen to multisig actor

frrist:
This will remain in draft until cbor-gen supports int64.


gen/gen.go
```diff
@@ -0,0 +1,26 @@
...
+		multisig.MultiSigActorState{},
+		multisig.MultiSigTransaction{},
+		// method params
+		multisig.ConstructorParams{},
+		multisig.ProposeParams{},
+		multisig.TxnIDParams{},
+		multisig.ChangeNumApprovalsThresholdParams{},
+		multisig.SwapAuthorizedPartyParams{},
+		// method returns
+		multisig.ProposeReturn{},
```
anorth: @hannahhoward you somewhere had a handy-looking Go annotation that put this information in the code itself. Is that something we could use here?

actors/util/assert.go
```diff
@@ -15,7 +15,7 @@ func Assert(b bool) {
 }
 
 func AssertNoError(e error) {
-	AssertMsg(e == nil, e.Error())
+	AssertMsg(e == nil, fmt.Sprintf("%v", e))
```
frrist: required else we get a NPE when `err == nil` since `err.Error()` is evaluated.

gen/gen.go
frrist: I think this is it - github.com/hannahhoward/cbor-gen-for. I was unable to get it working, this might be due to some env vars being required: https://github.com/hannahhoward/cbor-gen-for/blob/master/main.go#L15.

go.mod
```diff
@@ -8,5 +8,6 @@ require (
 	github.com/ipfs/go-hamt-ipld v0.0.14
 	github.com/libp2p/go-libp2p-core v0.3.0
 	github.com/multiformats/go-multihash v0.0.10
-	github.com/whyrusleeping/cbor-gen v0.0.0-20191216205031-b047b6acb3c0
+	github.com/whyrusleeping/cbor-gen v0.0.0-20200122015334-52eaf73beaf5
+	golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543
```
anorth: Don't import xerrors unnecessarily.

frrist: Its required by the generated code https://github.com/filecoin-project/specs-actors/pull/27/files#diff-5703d064128c2ab440d6eed480850a73R12 :disappointed:

PR #28, opened 2020-01-21
Re-order runtime methods to make more sense to me.

anorth:
No semantic change, just makes me happier.

### DIFF
--- a/actors/runtime/runtime.go
+++ b/actors/runtime/runtime.go
@@ -15,27 +15,48 @@ import (
 // Runtime is the VM's internal runtime object.
 // this is everything that is accessible to actors, beyond parameters.
 type Runtime interface {
+	// The current chain epoch number. The genesis block has epoch zero.
 	CurrEpoch() abi.ChainEpoch
-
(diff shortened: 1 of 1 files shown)


actors/runtime/runtime.go
```diff
@@ -47,24 +68,6 @@ type Runtime interface {
...
-	// Look up the current values of several system-wide economic indices.
-	CurrIndices() indices.Indices
-
-	// Look up the code ID of a given actor address.
-	GetActorCodeID(addr addr.Address) (ret abi.ActorCodeID, ok bool)
-
-	// Run a (pure function) computation, consuming the gas cost associated with that function.
-	// This mechanism is intended to capture the notion of an ABI between the VM and native
-	// functions, and should be used for any function whose computation is expensive.
-	Compute(ComputeFunctionID, args []interface{}) interface{}
```
zixuanzh: Assuming Compute is now in Syscalls?

anorth: That's right.

PR #36, opened 2020-01-22
refactor: add reward actor params and code gen

frrist:


actors/builtin/reward/reward_actor.go
```diff
@@ -50,17 +50,22 @@ func (r *Reward) AmountVested(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 
 // ownerAddr to a collection of Reward
 // TODO: AMT
-type RewardBalanceAMT map[addr.Address][]Reward
+// type RewardBalanceAMT map[addr.Address][]Reward
 
 type RewardActorState struct {
-	RewardMap RewardBalanceAMT
+	// TODO AMT
+	RewardBalanceAMT cid.Cid
```
frrist: Eventually the actor state will look something like this with methods to modify its AMT. Until then we need to work around https://github.com/polydawn/refmt/issues/35. cc @anorth is this strategy okay with you or should we do the work to convert this to an AMT now?

anorth: Given the amount of code that needed to be commented out, and hence outside the scope of refactoring tools, I think we should wait until we can put an AMT here.

I don't think numbers as keys are a problem here, but maybe addresses? In the interim, we could use Address.Bytes() as the key.

frrist:
Will revisit after the H/AMT TODO's addressed, ideally https://github.com/filecoin-project/specs-actors/pull/39 will help with this.

PR #37, opened 2020-01-22
Some fixes while integrating into lotus

whyrusleeping:


actors/builtin/codes.go
```diff
@@ -1,31 +1,30 @@
...
 func init() {
 	builder := cid.V1Builder{Codec: cid.Raw, MhType: mh.IDENTITY}
-	makeBuiltin := func(s string) abi.ActorCodeID {
+	makeBuiltin := func(s string) cid.Cid {
 		c, err := builder.Sum([]byte(s))
 		if err != nil {
 			panic(err)
 		}
-		id := abi.ActorCodeID(c)
+		id := cid.Cid(c)
```
anorth: unnecessary conversion?

actors/builtin/singletons.go
```diff
@@ -8,12 +8,12 @@ import (
...
-	RewardActorAddr        = mustMakeAddress(2)
-	CronActorAddr          = mustMakeAddress(3)
-	StoragePowerActorAddr  = mustMakeAddress(4)
-	StorageMarketActorAddr = mustMakeAddress(5)
+	InitActorAddr          = mustMakeAddress(0)
+	RewardActorAddr        = mustMakeAddress(1)
+	StoragePowerActorAddr  = mustMakeAddress(2)
+	StorageMarketActorAddr = mustMakeAddress(3)
+	CronActorAddr          = mustMakeAddress(4)
+	SystemActorAddr        = mustMakeAddress(5)
```
anorth: This seems like an unnecessary change to me. The proposed ordering is somewhat arbitrary history-dependent order, the original somewhat builds up from foundations to application-level.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -39,7 +39,7 @@ func (a *MultiSigActor) State(rt vmr.Runtime) (vmr.ActorStateHandle, MultiSigAct
 }
 
 type ConstructorParams struct {
-	Signers     []addr.Address
```
frrist: Thanks for reminding me that CI needs a gofmt check. https://github.com/filecoin-project/specs-actors/issues/42

actors/builtin/singletons.go
whyrusleeping: I worry that users in the community already have things set up to interact with system actors. Forcing that change will be disruptive, but I won't push the point any further

actors/builtin/codes.go
whyrusleeping: weird that gofmt -s didnt catch that...

PR #39, opened 2020-01-22
feature: define an implement hamt interface

frrist:

PR #41, opened 2020-01-22
Replace manual state acquire/release with closure-like mechanism

anorth:
This style is equivalent, but easier to code on both sides.

FYI @whyrusleeping


actors/runtime/runtime.go
```diff
@@ -123,8 +124,49 @@ type TraceSpan interface {
...
-	Release(checkStateCID abi.ActorSubstateCID)
-	Take() abi.ActorSubstateCID
-}
\ No newline at end of file
+// ReadonlyStateHandle provides read-only access to an actor's state object.
+type ReadonlyStateHandle interface {
+	// Readonly loads a readonly copy of the state into the argument.
+	//
+	// Any modification to the state is illegal and will result in an abort.
+	Readonly(obj CBORMarshalable)
```
anorth: There's no closure or release here because nothing can detect or prevent subsequent inspection of the state object in any case.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -83,24 +72,26 @@ type ProposeReturn struct {
...
+		a.validateAuthorizedParty(rt, &st, callerAddr)
+		txnID := st.NextTxnID
+		st.NextTxnID += 1
+
+		st.PendingTxns = setPendingTxn(rt, st.PendingTxns, txnID, MultiSigTransaction{
+			To:       params.To,
+			Value:    params.Value,
+			Method:   params.Method,
+			Params:   params.Params,
+			Approved: []addr.Address{},
```
frrist: Do we need to add `calledAddr` to this slice as was done before?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -114,25 +105,31 @@ type TxnIDParams struct {
...
-	h, st := a.State(rt)
-	txn := getPendingTxn(rt, st.PendingTxns, params.ID)
-	proposer := txn.Approved[0]
-	if proposer != callerAddr {
-		rt.AbortStateMsg("Cannot cancel another signers transaction")
-	}
+	var st MultiSigActorState
+	rt.State().Transaction(&st, func() interface{} {
+		a.validateAuthorizedParty(rt, &st, callerAddr)
+		txn := getPendingTxn(rt, st.PendingTxns, params.ID)
```
frrist: Will conflict with https://github.com/filecoin-project/specs-actors/pull/39, I think this should go in first.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -114,25 +105,31 @@ type TxnIDParams struct {
...
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	callerAddr := rt.ImmediateCaller()
-	a._rtValidateSignerOrAbort(rt, callerAddr)
-	a._rtApproveTransactionOrAbort(rt, params.ID)
+	var st MultiSigActorState
+	rt.State().Transaction(&st, func() interface{} {
+		a.validateAuthorizedParty(rt, &st, callerAddr)
+		return nil
+	})
+	a.approveTransaction(rt, params.ID)
```
frrist: I don't like having two transaction to perform this operation. I suppose it was already like this, and your changes have just made it more clear. Can we move away from this pattern to a more performant one in a follow on?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -167,24 +165,25 @@ func (a *MultiSigActor) RemoveSigner(rt vmr.Runtime, params *RemoveSigner) *vmr.
...
 		}
-	}
-	if params.Decrease || int64(len(st.Signers)-1) < st.NumApprovalsThreshold {
-		st.NumApprovalsThreshold = st.NumApprovalsThreshold - 1
-	}
-	st.Signers = newSigners
+		if params.Decrease || int64(len(st.Signers)-1) < st.NumApprovalsThreshold {
+			st.NumApprovalsThreshold = st.NumApprovalsThreshold - 1
+		}
+		st.Signers = newAuthorizedParties
```
frrist: nit: should probably be `newSigners`, both here and the lines above :arrow_up:. I think this was missed while rebasing after https://github.com/filecoin-project/specs-actors/pull/35

actors/builtin/multisig/multisig_actor.go
```diff
@@ -197,26 +196,27 @@ func (a *MultiSigActor) SwapSigner(rt vmr.Runtime, params *SwapSignerParams) *vm
...
-	}
+		if !st.isSigner(params.To) {
+			rt.AbortStateMsg("Party already present")
+		}
 
-	newSigners := make([]addr.Address, 0, len(st.Signers))
-	for _, s := range st.Signers {
-		if s != params.From {
-			newSigners = append(newSigners, s)
+		newAuthorizedParties := make([]addr.Address, 0, len(st.Signers))
```
frrist: `s|newAuthorizedParties|newSigners|g`

actors/builtin/reward/reward_actor.go
```diff
@@ -154,32 +144,29 @@ func (a *RewardActor) AwardBlockReward(
...
-			StartEpoch:      rt.CurrEpoch(),
-			EndEpoch:        rt.CurrEpoch(),
-			Value:           actualReward,
-			AmountWithdrawn: abi.NewTokenAmount(0),
-			VestingFunction: None,
+	var st RewardActorState
+	rt.State().Transaction(&st, func() interface{} {
+		if actualReward.GreaterThan(abi.NewTokenAmount(0)) {
+			// put Reward into RewardMap
+			newReward := &Reward{
```
frrist: not your change but does this need to be a pointer?

actors/builtin/multisig/multisig_actor.go
anorth: This is an explicit change, because approveTransaction will add it. Otherwise it would have aborted always.

actors/builtin/multisig/multisig_actor.go
anorth: Yes, this PR makes it more clear that there are tons of unnecessary state load/store cycles which should be collapsed into many fewer! cc @zixuanzh  @jzimmerman

actors/builtin/multisig/multisig_actor.go
anorth: thanks

PR #43, opened 2020-01-23
add payment channel actor methods

sternhenri:
Add methods for payment_channel actor

### DIFF
--- a/actors/builtin/payment_channel/payment_channel_actor.go
+++ b/actors/builtin/payment_channel/payment_channel_actor.go
@@ -1,0 +1,247 @@
+package payment_channel
+
+import (
+	"bytes"
+
+	addr "github.com/filecoin-project/go-address"
+	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	big "github.com/filecoin-project/specs-actors/actors/abi/big"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	vmr "github.com/filecoin-project/specs-actors/actors/runtime"
+	indices "github.com/filecoin-project/specs-actors/actors/runtime/indices"
+	"github.com/filecoin-project/specs-actors/actors/serde"
+	"github.com/minio/blake2b-simd"
+)
+
+type PaymentChannelActor struct{}
+
+// type PaymentInfo struct {
+// 	PayChActor     address.Address
+// 	Payer          address.Address
+// 	ChannelMessage *cid.Cid
+
+// 	Vouchers []*types.SignedVoucher
+// }
+
+/////////////////
+// Constructor //
+/////////////////
+
+type PCAConstructorParams struct {
+	To addr.Address
+}
+
+func (pca *PaymentChannelActor) Constructor(rt vmr.Runtime, params *PCAConstructorParams) *vmr.EmptyReturn {
+
+	var st PaymentChannelActorState
+	rt.State().Transaction(&st, func() interface{} {
+		st.From = rt.ImmediateCaller()
+		st.To = params.To
+		st.LaneStates = make(map[int64]*LaneState)
+		return nil
+	})
+	return &vmr.EmptyReturn{}
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Payment Channel state operations
+////////////////////////////////////////////////////////////////////////////////
+
+type PCAUpdateChannelStateParams struct {
+	Sv     SignedVoucher
+	Secret []byte
+	Proof  []byte
+}
+
+func hash(b []byte) []byte {
+	s := blake2b.Sum256(b)
+	return s[:]
+}
+
+type PaymentVerifyParams struct {
+	Extra []byte
+	Proof []byte
+}
+
+func (pca *PaymentChannelActor) UpdateChannelState(rt vmr.Runtime, params *PCAUpdateChannelStateParams) *vmr.EmptyReturn {
+
+	var st PaymentChannelActorState
+	rt.State().Transaction(&st, func() interface{} {
+
+		sv := params.Sv
+
+		vb, nerr := sv.SigningBytes()
+		if nerr != nil {
+			rt.Abort(1, "failed to serialize signedvoucher")
+		}
+
+		if !rt.Syscalls().VerifySignature(*sv.Signature, st.From, vb) {
+			rt.Abort(2, "voucher signature invalid")
+		}
+
+		if rt.CurrEpoch() < sv.TimeLock {
+			rt.Abort(3, "cannot use this voucher yet!")
+		}
+
+		if len(sv.SecretPreimage) > 0 {
+			if !bytes.Equal(hash(params.Secret), sv.SecretPreimage) {
+				rt.Abort(4, "incorrect secret!")
+			}
+		}
+
+		if sv.Extra != nil {
+
+			_, code := rt.Send(
+				sv.Extra.Actor,
+				sv.Extra.Method,
+				serde.MustSerializeParams(
+					&PaymentVerifyParams{
+						sv.Extra.Data,
+						params.Proof,
+					},
+				),
+				abi.NewTokenAmount(0),
+			)
+			vmr.RequireSuccess(rt, code, "spend voucher verification failed")
+		}
+
+		ls, ok := st.LaneStates[sv.Lane]
+		// create voucher lane if it does not already exist
+		if !ok {
+			ls = new(LaneState)
+			ls.Redeemed = big.NewInt(0)
+			st.LaneStates[sv.Lane] = ls
+		}
+		if ls.Closed {
+			rt.Abort(5, "cannot redeem a voucher on a closed lane")
+		}
+
+		if ls.Nonce > sv.Nonce {
+			rt.Abort(6, "voucher has an outdated nonce, cannot redeem")
+		}
+
+		// The next section actually calculates the payment amounts to update the payment channel state
+		// 1. (optional) sum already redeemed value of all merging lanes
+		redeemedFromOthers := big.Zero()
+		for _, merge := range sv.Merges {
+			if merge.Lane == sv.Lane {
+				rt.Abort(7, "voucher cannot merge lanes into its own lane")
+			}
+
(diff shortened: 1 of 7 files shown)


anorth:
There's nothing specifically wrong with this, but I expected to see the code for the payment channel actor! All the type definitions from https://github.com/filecoin-project/specs/pull/854 should be here.

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		if ls.Closed {
+			rt.Abort(5, "cannot redeem a voucher on a closed lane")
+		}
+
+		if ls.Nonce > sv.Nonce {
+			rt.Abort(6, "voucher has an outdated nonce, cannot redeem")
+		}
+
+		// The next section actually calculates the payment amounts to update the payment channel state
```
sternhenri: The below is a bit tricky and deserves review @anorth

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		ls, ok := st.LaneStates[sv.Lane]
+		// create voucher lane if it does not already exist
+		if !ok {
+			ls = new(LaneState)
+			ls.Redeemed = big.NewInt(0)
+			st.LaneStates[sv.Lane] = ls
+		}
+		if ls.Closed {
```
sternhenri: @whyrusleeping I don't think lanes ever get closed...?

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+	var st PaymentChannelActorState
+	rt.State().Transaction(&st, func() interface{} {
+
+		if st.ClosingAt == 0 {
+			rt.Abort(1, "payment channel not closing or closed")
+		}
+
+		if rt.CurrEpoch() < st.ClosingAt {
+			rt.Abort(2, "payment channel not closed yet")
+		}
```
sternhenri: why does the channel need to be closed in order to call `Collect`? The method resets `ToSend` at the end, so should be callable multiple times without issue... @whyrusleeping

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return st.From
+}
+
+func (pca *PaymentChannelActor) GetToSend(rt vmr.Runtime) abi.TokenAmount {
+	var st PaymentChannelActorState
+	rt.State().Readonly(&st)
+
+	return st.ToSend
+}
```
sternhenri: @anorth these aren't actually used in our code, but thought it useful to leave them as in Lotus. LMK if you want me to remove.

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+type PCAUpdateChannelStateParams struct {
+	Sv     SignedVoucher
+	Secret []byte
+	Proof  []byte
+}
+
+func hash(b []byte) []byte {
+	s := blake2b.Sum256(b)
+	return s[:]
+}
```
sternhenri: Worth having a convo @anorth @whyrusleeping to decide whether to keep SHA256 or blake2 throughout actors. Best to stick to a single fn

sternhenri:
Note that I am assuming `rt.transaction()` is atomic, hence my very wide cast in my methods. Lmk if I should scope it down @anorth

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+
+type PaymentChannelActor struct{}
+
+// type PaymentInfo struct {
+// 	PayChActor     address.Address
+// 	Payer          address.Address
+// 	ChannelMessage *cid.Cid
+
+// 	Vouchers []*types.SignedVoucher
+// }
```
anorth: Either we need this type or we don't. Please remove the disabled code.

actors/builtin/payment_channel/payment_channel_actor.go
anorth: I suggest we make hashing a syscall rather than import a concrete implementation directly into actor code. @whyrusleeping ?

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+func (pca *PaymentChannelActor) UpdateChannelState(rt vmr.Runtime, params *PCAUpdateChannelStateParams) *vmr.EmptyReturn {
+
+	var st PaymentChannelActorState
+	rt.State().Transaction(&st, func() interface{} {
+
+		sv := params.Sv
+
+		vb, nerr := sv.SigningBytes()
+		if nerr != nil {
+			rt.Abort(1, "failed to serialize signedvoucher")
```
anorth: Invalid exit codes, must be greater than `exitcode.FirstActorErrorCode`. Consider using those already defined in common.go where they are a good fit, but also absolutely ok to define others locally here, with symbolic constants.

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+
+		if len(sv.SecretPreimage) > 0 {
+			if !bytes.Equal(hash(params.Secret), sv.SecretPreimage) {
+				rt.Abort(4, "incorrect secret!")
+			}
+		}
+
+		if sv.Extra != nil {
+
+			_, code := rt.Send(
```
anorth: Illegal to Send() while inside a transaction, the transaction must finish first. Nothing prior to here mutates state, so I suggest using Readonly instead

actors/builtin/payment_channel/payment_channel_actor.go
anorth: I can't follow this giant method, but it's not your job to refactor it. I trust Lotus have tests we'll be able to port into here.

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -0,0 +1,247 @@
...
+			rt.Abort(1, "payment channel not closing or closed")
+		}
+
+		if rt.CurrEpoch() < st.ClosingAt {
+			rt.Abort(2, "payment channel not closed yet")
+		}
+
+		// send remaining balance to "From"
+
+		_, code := rt.Send(
```
anorth: Illegal Send inside transaction

actors/builtin/payment_channel/payment_channel_actor_state.go
```diff
@@ -0,0 +1,75 @@
...
+	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	big "github.com/filecoin-project/specs-actors/actors/abi/big"
+	acrypto "github.com/filecoin-project/specs-actors/actors/crypto"
+)
+
+type PaymentChannelActorState struct {
+	From addr.Address
+	To   addr.Address
+
+	ToSend big.Int
```
anorth: Use TokenAmount, at least until we decide to remove it.

actors/builtin/payment_channel/payment_channel_actor.go
anorth: I would prefer to remove them, but could be added to the state object instead.

actors/builtin/payment_channel/payment_channel_actor_state.go
```diff
@@ -0,0 +1,75 @@
...
+	"io"
+
+	addr "github.com/filecoin-project/go-address"
+
+	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	big "github.com/filecoin-project/specs-actors/actors/abi/big"
+	acrypto "github.com/filecoin-project/specs-actors/actors/crypto"
+)
+
+type PaymentChannelActorState struct {
```
anorth: Any documentation about any and all structures here would be much appreciated.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/ipfs/go-cid v0.0.4
 	github.com/ipfs/go-hamt-ipld v0.0.14
 	github.com/libp2p/go-libp2p-core v0.3.0
+	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
```
anorth: Remove this if we move hashing to a syscall.

actors/builtin/payment_channel/payment_channel_actor.go
sternhenri: Understood. Nonetheless I worry about atomicity here. Has this been worked out elsewhere yet?

sternhenri:
Ready for review @whyrusleeping 

Left some Qs for you. I also think we should get rid of `Extra` if we can. Doesn't seem all that necessary at this point.

actors/builtin/payment_channel/payment_channel_actor.go
anorth: Not sure I understand the question. When updating all the other actors to use transactions, I think I avoided illegal side effects, but unsure until we write tests.

sternhenri: As in what if the balance gets sent out, and then the transaction fails below and payment channel stays funded even though it's already paid out (in this case)

anorth: I can no longer link this comment to the code so I've lost context.

Note that if a method aborts, all state changes are rolled back, desipite the name transaction. Does that answer it?

sternhenri: yes and I guess this is the crux of the issue. You should be able to somehow have the transaction rollback on send failures as well in that both states should be in sync

actors/builtin/payment_channel/payment_channel_actor.go
sternhenri: per convo no need for this right now, this is already resolved by nonce logic.

PR #44, opened 2020-01-23
refactor: convert some storage power actor maps to hamts

frrist:


actors/abi/sector.go
```diff
@@ -28,7 +30,11 @@ type SectorID struct {
 type SectorWeight int64 // TODO big
 
 // The unit of storage power (measured in bytes)
-type StoragePower int64 // TODO big
+type StoragePower = big.Int
+
+func NewStoragePower(n int64) StoragePower {
+	return StoragePower(big.NewInt(n))
+}
```
frrist: Made this a BigInt now so that I would get the marshaling code for free. An alternative was to implement the marshaling interface on it with panics and then replace it later. I thought the former was more reasonable.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -235,6 +256,54 @@ func (st *StoragePowerActorState) _getPledgeSlashForConsensusFault(currPledge ab
...
+
+type addrKeyWrapper struct {
+	addr.Address
+}
+
+func (kw addrKeyWrapper) Key() string {
+	return string(kw.Bytes())
+}
+
+func getStoragePower(rt vmr.Runtime, root cid.Cid, a addr.Address) (abi.StoragePower, bool) {
```
frrist: returns a bool as some calls to this method make a decision if the address is not present in the hamt.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -91,28 +102,35 @@ func addrInArray(a addr.Address, list []addr.Address) bool {
...
-	for address := range st.PowerTable {
-		allMiners[index] = address
+	var index int64
+	var allMiners []addr.Address
+	if err := adt.NewMap(vmr.AsStore(rt), st.PowerTable).ForEach(func(k string, v interface{}) error {
+		maddr, err := addr.NewFromBytes([]byte(k))
+		if err != nil {
+			return err
+		}
+		allMiners[index] = maddr
```
anorth: Index out of bounds. But also the index seems completely unused, remove it and append to slice instead.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -235,6 +256,54 @@ func (st *StoragePowerActorState) _getPledgeSlashForConsensusFault(currPledge ab
 	}
 }
 
+func asKeyer(a addr.Address) adt.Keyer {
+	return addrKeyWrapper{a}
+}
+
+type addrKeyWrapper struct {
```
anorth: Just call it `addrKey`, and rename 259 to `asKey`.

actors/util/adt/map.go
```diff
@@ -90,3 +90,15 @@ func (h *Map) Delete(k Keyer) error {
 	h.root = newRoot
 	return nil
 }
+
+// ForEach applies fn to each key value in hamt.
+func (h *Map) ForEach(fn func(key string, v interface{}) error) error {
```
anorth: Use the CBORUnmarshalable interface from runtime (and replace the cbg references above with similar).

actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -201,7 +202,7 @@ func (a *StoragePowerActor) OnMinerSurprisePoStSuccess(rt Runtime) *vmr.EmptyRet
 	var st StoragePowerActorState
 	rt.State().Transaction(&st, func() interface{} {
 		delete(st.PoStDetectedFaultMiners, minerAddr)
-		st._updatePowerEntriesFromClaimedPower(minerAddr)
+		st._updatePowerEntriesFromClaimedPower(rt, minerAddr)
```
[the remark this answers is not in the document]
anorth: Limiting it to the Store interface would probably do it.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -235,6 +256,54 @@ func (st *StoragePowerActorState) _getPledgeSlashForConsensusFault(currPledge ab
...
+
+func getStoragePower(rt vmr.Runtime, root cid.Cid, a addr.Address) (abi.StoragePower, bool) {
+	hm := adt.NewMap(vmr.AsStore(rt), root)
+
+	var out abi.StoragePower
+	err := hm.Get(asKeyer(a), &out)
+	if err == hamt.ErrNotFound {
+		return abi.NewStoragePower(0), false
+	}
+	requireNoStateErr(rt, err, "failed to get storage power for address %v from claimed power HAMT", a)
```
anorth: After discussion with @whyrusleeping I think he'd rather we propagate the errors here up to the top level.

actors/util/adt/map.go
frrist: If I we do that here the hamt `ForEach` code won't accept it (unless there is some type-foo that can help here?). WIll do it for all other methods in this file.

actors/builtin/storage_power/storage_power_actor_state.go
frrist: Sounds good, will do that in a follow on as that on its own is a fairly large change.

actors/builtin/storage_power/storage_power_actor.go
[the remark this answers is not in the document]
frrist: Agreed, limiting the store and propagating errors will be a tightly coupled change, so will do in follow on.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -91,28 +102,35 @@ func addrInArray(a addr.Address, list []addr.Address) bool {
...
+func (st *StoragePowerActorState) _selectMinersToSurprise(rt vmr.Runtime, challengeCount int, randomness abi.Randomness) []addr.Address {
 	// this wont quite work -- a.PowerTable is a HAMT by actor address, doesn't
 	// support enumerating by int index. maybe we need that as an interface too,
 	// or something similar to an iterator (or iterator over the keys)
 	// or even a seeded random call directly in the HAMT: myhamt.GetRandomElement(seed []byte, idx int) using the ticket as a seed
 
-	ptSize := len(st.PowerTable)
-	allMiners := make([]addr.Address, len(st.PowerTable))
-	index := 0
+	// Reply to above comment: for now do the easy thing and use the hamt ForEach method as lotus does
```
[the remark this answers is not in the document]
sternhenri: cc @nikkolasg re the bias we have been discussing. I believe things should be ok as long as the order is stable.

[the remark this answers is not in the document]
nikkolasg: Specifically as long as the `RandomInt` that takes a bigint to a float64 takes the first 8 bytes and NOT a modulo operation, things are good for me !

PR #46, opened 2020-01-24
changes made during multisig integration push with lotus

whyrusleeping:
To be used in conjunction with: https://github.com/filecoin-project/lotus/pull/1148


actors/runtime/runtime.go
```diff
@@ -170,3 +172,8 @@ type CBORMarshalable interface {
 type CBORUnmarshalable interface {
 	UnmarshalCBOR(r io.Reader) error
 }
+
+type CBORAble interface {
+	CBORMarshalable
+	CBORUnmarshalable
+}
```
frrist: noting that here and elsewhere will conflict with https://github.com/filecoin-project/specs-actors/pull/45/files#diff-e2d8254b9ce8e65885afe1cf6a808a1dR173-R177

actors/builtin/multisig/multisig_actor.go
```diff
@@ -16,9 +18,9 @@ type TxnID int64
 
 func (t TxnID) Key() string {
 	// convert a TxnID to a HAMT key.
-	txnKey := make([]byte, 0, binary.MaxVarintLen64)
-	binary.PutVarint(txnKey, int64(t))
-	return string(txnKey)
+	txnKey := make([]byte, binary.MaxVarintLen64)
+	n := binary.PutVarint(txnKey, int64(t))
+	return string(txnKey[:n])
```
frrist: :pray:

actors/builtin/multisig/multisig_actor.go
```diff
@@ -69,12 +72,7 @@ type ProposeParams struct {
 	Params abi.MethodParams
 }
 
-type ProposeReturn struct {
-	TxnID TxnID
-}
-
-func (a *MultiSigActor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
-	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
+func (a *MultiSigActor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
```
anorth: Can we discuss this change?
1.  I added return structures symmetry with params objects, and consistent with service-method code generators everywhere (protobuf etc).
2. There's a CBOR-able big.Int type here already, we don't need to reference cbg directly here.

frrist: Why is this necessary? `ProposeReturn` was already CBORable.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -99,15 +97,15 @@ func (a *MultiSigActor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeR
...
+	out := cbg.CborInt(txnID)
+	return &out
 }
 
 type TxnIDParams struct {
 	ID TxnID
 }
 
 func (a *MultiSigActor) Approve(rt vmr.Runtime, params *TxnIDParams) *vmr.EmptyReturn {
-	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
```
anorth: This is a significant item coming from the spec, the source already of much debate, and still here by fiat from jbenet. We can't just remove it, but your runtime shim can no-op it for now if you want.

actors/runtime/exitcode/reserved.go
```diff
@@ -2,7 +2,7 @@ package exitcode
 
 import "strconv"
 
-type ExitCode int64
+type ExitCode uint8
```
anorth: We discussed this at length in person in November and since:
- we only use two integer types, with no unsigned variants
- there's no reason to restrict actors error code ranges, and if we did we can do it at runtime

Please revert this for now and cast in your shim

actors/builtin/multisig/multisig_actor.go
whyrusleeping: It also serializes as a struct, causing extra wasted bytes, and deviates from existing expectations around return values

whyrusleeping: 1) Struct wrapping all return values feels wasteful to me, functions should be able to have simple return types, i'd rather not have to define a new type and run codegen for every new function that wants to return an int
2) This isnt a big int

actors/builtin/multisig/multisig_actor.go
whyrusleeping: It's not doing anything productive. Its checking that the call is coming from some specific types of actors, when theres no good reason for that. So what if some new type of actor wants to interact with a multisig? The multisig explicitly doesnt care

actors/runtime/exitcode/reserved.go
whyrusleeping: I don't recall these conversations, but we can make this change later, sure.

actors/builtin/multisig/multisig_actor.go
anorth: Ok, I see the cost of the struct-wrapping bytes on chain. So methods can return serializable value, done.

And then I'm guessing CborInt is just a wrapper to implement the marshal methods around an int. cbg probably has a bunch of utilities like this that are unrelated to code generation but useful. Ok to land this now, but I might attempt to minimise external deps of this code later.

actors/builtin/multisig/multisig_actor.go
anorth: I 100% agree with you but already lost this battle. Please restore it here for now so we can land these changes and then discuss elsewhere.

actors/runtime/exitcode/reserved.go
anorth: Thanks!

actors/runtime/runtime.go
anorth: Yeah, sorry @whyrusleeping . I changed the interface names to match the more go-idiomatic ones from cbg.

actors/builtin/multisig/multisig_actor.go
whyrusleeping: *sigh* fine

whyrusleeping:
Alright, rebased and addressed comments. Should be good to go now

actors/builtin/multisig/multisig_actor.go
```diff
@@ -10,15 +10,17 @@ import (
 	vmr "github.com/filecoin-project/specs-actors/actors/runtime"
 	exitcode "github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
 	adt "github.com/filecoin-project/specs-actors/actors/util/adt"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
```
anorth: nit: prefer to group this with the other external import, `addr`

PR #47, opened 2020-01-24
simplify sector state

zixuanzh:
Simplify sector state with the following changes:
- [x] Active, TemporaryFault, and PreCommit still exists as concept but removed SectorState field from MinerActorState
- [x] Introduce PreCommittedSectorsAMT to keep track of PreCommit sectors
- [x] SectorAMT contains all Active and TemporaryFault sectors
- [x] TemporaryFault sectors are stored in FaultSet
- [x] ProvingSet is a snapshot of the current SectorAMT subtracting sectors in the FaultSet. It is a pointer to the root of SectorAMT
- [x] ProvingSet is updated at ProveCommit (assume sector gaining power at ProveCommit)and OnSurprisePoStChallenge. Lotus does this slightly differently and likely needs some changes but this ties to the discussion of when a sector gains power and ElectionPoSt.
- [x] Remove deal verification at PreCommit and only do that at ProveCommit, better separation of concerns and they are doing the same deal checks
- [x] VerifyDealsOnSectorProveCommit returns DealWeight upon successful verification to reduce one inter actor round trip
- [x] Remove unnecessary fields in PreCommitInfo and ProveCommitInfo
- [x] Add SectorPreCommitOnChainInfo to reduce onchain footprint as PreCommit related info can be discarded after ProveCommit
- [x] A sector is only in TemporaryFault state (hence in the FaultSet) after EffectiveFaultBeginEpoch() and not at DeclareTemporaryFault

TODO:
- [ ] align on power and update ProvingSet when HAMT/AMT are in place

cc @jzimmerman @sternhenri

### DIFF
--- a/actors/builtin/storage_miner/storage_miner_actor.go
+++ b/actors/builtin/storage_miner/storage_miner_actor.go
@@ -3,8 +3,7 @@ package storage_miner
 import (
 	"bytes"
 
-	big "github.com/filecoin-project/specs-actors/actors/abi/big"
-
+	cid "github.com/ipfs/go-cid"
 	addr "github.com/filecoin-project/go-address"
 	peer "github.com/libp2p/go-libp2p-core/peer"
 
@@ -60,9 +59,10 @@ func (a *StorageMinerActor) OnSurprisePoStChallenge(rt Runtime) *vmr.EmptyReturn
 		randomnessK := rt.GetRandomness(rt.CurrEpoch() - indices.StorageMining_SpcLookbackPoSt())
 		challengedSectorsRandomness := crypto.DeriveRandWithMinerAddr(crypto.DomainSeparationTag_SurprisePoStSampleSectors, randomnessK, rt.CurrReceiver())
 
+		st.ProvingSet = st.GetCurrentProvingSet()
 		challengedSectors := _surprisePoStSampleChallengedSectors(
 			challengedSectorsRandomness,
-			SectorNumberSetHAMT_Items(st.ProvingSet),
+			st.ProvingSet,
 		)
 
 		st.PoStState = MinerPoStState{
@@ -167,27 +167,11 @@ func (a *StorageMinerActor) PreCommitSector(rt Runtime, info SectorPreCommitInfo
 	depositReq := cidx.StorageMining_PreCommitDeposit(st.Info.SectorSize, info.Expiration)
 	builtin.RT_ConfirmFundsReceiptOrAbort_RefundRemainder(rt, depositReq)
 
-	// Verify deals with StorageMarketActor; abort if this fails.
-	// (Note: committed-capacity sectors contain no deals, so in that case verification will pass trivially.)
-	_, code := rt.Send(
-		builtin.StorageMarketActorAddr,
-		builtin.Method_StorageMarketActor_OnMinerSectorPreCommit_VerifyDealsOrAbort,
-		serde.MustSerializeParams(
-			info.DealIDs,
-			info,
-		),
-		abi.NewTokenAmount(0),
-	)
-	builtin.RequireSuccess(rt, code, "failed to verify deals")
-
 	rt.State().Transaction(&st, func() interface{} {
-		st.Sectors[info.SectorNumber] = SectorOnChainInfo{
-			State:            PreCommit,
+		st.PreCommittedSectors[info.SectorNumber] = SectorPreCommitOnChainInfo{
 			Info:             info,
 			PreCommitDeposit: depositReq,
 			PreCommitEpoch:   rt.CurrEpoch(),
-			ActivationEpoch:  epochUndefined,
-			DealWeight:       big.NewInt(-1),
 		}
 		return nil
 	})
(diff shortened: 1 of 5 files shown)


actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -167,27 +167,19 @@ func (a *StorageMinerActor) PreCommitSector(rt Runtime, info SectorPreCommitInfo
...
-		builtin.StorageMarketActorAddr,
-		builtin.Method_StorageMarketActor_OnMinerSectorPreCommit_VerifyDealsOrAbort,
-		serde.MustSerializeParams(
-			info.DealIDs,
-			info,
-		),
-		abi.NewTokenAmount(0),
-	)
-	vmr.RequireSuccess(rt, code, "failed to verify deals")
+	if info.SealEpoch >= (rt.CurrEpoch() + indices.StorageMining_SpcLookbackSeal()) {
```
anorth: What is `SealEpoch`? The epoch of pre-commit, or the deadline for provecommit (it's undocumented)? If the former, this condition can never succeed so is probably a bug. If the latter, can we name it as such? The error message below is very opaque, especially since the numbers have no labels. Try something like "seal epoch (%d) must be before current epoch (%d) + lookback (%d) = %d"

"Seal" is also not a word that's really in the vocab of the actors. Use pre-commit or provecommit. Does seal == prove? seal == commit?

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -447,9 +424,7 @@ func (a *StorageMinerActor) _rtCheckTemporaryFaultEvents(rt Runtime, sectorNumbe
 
 	storageWeightDesc := a._rtGetStorageWeightDescForSector(rt, sectorNumber)
 
-	if checkSector.State == Active && rt.CurrEpoch() == checkSector.EffectiveFaultBeginEpoch() {
-		checkSector.State = TemporaryFault
-
+	if !st.IsSectorInTemporaryFault(sectorNumber) && rt.CurrEpoch() == checkSector.EffectiveFaultBeginEpoch() {
```
anorth: I would feel much better if the `==` was `>=`. Ditto below.

actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -14,9 +15,12 @@ import (
 // Balance of a StorageMinerActor should equal exactly the sum of PreCommit deposits
 // that are not yet returned or burned.
 type StorageMinerActorState struct {
+	PreCommittedSectors PreCommittedSectorsAMT 
 	Sectors    SectorsAMT
+	FaultSet SectorNumberSetHAMT // TODO bitfield
+	ProvingSet cid.Cid
```
anorth: Please document what it's a cid of, e.g. `HAMT[sectorNumber]sectorInfo`. Same for the HAMTs and AMTs above.

Until implementing the real structure, it's probably more useful to use a map or slice so that the desired programming semantics can be expressed in code.

actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -56,23 +60,18 @@ func (mps *MinerPoStState) Is_DetectedFault() bool {
...
+	PreCommitEpoch abi.ChainEpoch
+}
 
 type SectorOnChainInfo struct {
-	State                 SectorState
 	Info                  SectorPreCommitInfo // Also contains Expiration field.
-	PreCommitDeposit      abi.TokenAmount
-	PreCommitEpoch        abi.ChainEpoch
-	ActivationEpoch       abi.ChainEpoch // -1 if still in PreCommit state.
+	ActivationEpoch       abi.ChainEpoch
```
anorth: Please document this and all fields. E.g. `The epoch at which the sector commitment was proven`.

actors/builtin/storage_miner/storage_miner_actor.go
sternhenri: sorry we tried to document it with Porcu in https://github.com/filecoin-project/specs-actors/blob/master/actors/abi/sector.go#L80.

`SealEpoch` is used to tie the sealing to the chain as a whole, so it is drawn from finality back at pre-commit (more details in spec). It matters on verification since it accounts for how long sealing actually takes. I'm not quitee sure what is happening in this function per se, but sealEpoch should be `currEpoch - spcLookbackSeal` when miner starts a precommit, thereafter the actor should verify it per (see https://github.com/filecoin-project/specs-actors/pull/29/files#diff-6c14f5a2f4e3689416a73d535d06d2b3R735-R740).

Regarding the nomenclature I think dig's terminology is used here cc @porcuquine

actors/runtime/indices/indices.go
```diff
@@ -365,6 +365,12 @@ func StorageMining_SpcLookbackSeal() abi.ChainEpoch {
 	return StorageMining_Finality() // should be approximately the same as finality
 }
 
+func StorageMining_SpcLookbackSealLimit() abi.ChainEpoch {
+	PARAM_FINISH()
+	return 0
+}
+
+
```
sternhenri: I don't think this is necessary. The point is that number is derived from the maxSealTime below. Happy to chat through it @zixuanzh

actors/builtin/storage_miner/storage_miner_actor.go
zixuanzh: Removing checks on SealEpoch from this PR for now. @sternhenri could you include a PR here for the right checks next week? thanks

actors/builtin/storage_miner/storage_miner_actor.go
zixuanzh: setting it to `>=` but noting that this check is scheduled on the CronActor and should be checking at the exact epoch.

actors/builtin/storage_miner/storage_miner_actor_state.go
zixuanzh: documented in comments, AMTs above have been defined in types.

PR #49, opened 2020-01-24
updated cron to enable multiple event types, and added worker key change

sternhenri:
This PR replaces https://github.com/filecoin-project/specs/pull/776

### DIFF
--- a/actors/builtin/storage_miner/storage_miner_actor.go
+++ b/actors/builtin/storage_miner/storage_miner_actor.go
@@ -43,7 +43,7 @@ func (a *StorageMinerActor) StageWorkerKeyChange(rt Runtime, key abi.PrivateKey)
 	// must be BLS since the worker key will be used alongside a BLS-VRF
 	keyAddr, err := addr.NewBLSAddress(key)
 	if err != nil {
-		rt.Abort(1, "Cannot make new address from key.")
+		rt.Abort(exitcode.ErrIllegalArgument, "Cannot make new address from key.")
 	}
 
 	keyChange := WorkerKeyChange{
@@ -68,11 +68,11 @@ func (a *StorageMinerActor) _rtCommitWorkerKeyChange(rt Runtime) *vmr.EmptyRetur
 	var st StorageMinerActorState
 	rt.State().Transaction(&st, func() interface{} {
 		if (st.Info.PendingKeyChange == WorkerKeyChange{}) {
-			rt.Abort(1, "No pending key change.")
+			rt.Abort(exitcode.ErrIllegalState, "No pending key change.")
 		}
 
 		if st.Info.PendingKeyChange.EffectiveAt > rt.CurrEpoch() {
-			rt.Abort(2, "Too early for key change.")
+			rt.Abort(exitcode.ErrIllegalState, "Too early for key change.")
 		}
 
 		st.Info.Worker = st.Info.PendingKeyChange.NewWorker
--- a/actors/builtin/storage_miner/storage_miner_actor.go
+++ b/actors/builtin/storage_miner/storage_miner_actor.go
@@ -56,6 +56,8 @@ func (a *StorageMinerActor) StageWorkerKeyChange(rt Runtime, newKey addr.Address
 		rt.ValidateImmediateCallerIs(st.Info.Owner)
 
 		// must be BLS since the worker key will be used alongside a BLS-VRF
+		// Specifically, this check isn't quite right
+		// TODO: check that the account actor at the other end of this address has a BLS key.
 		if newKey.Protocol() != addr.BLS {
 			rt.Abort(exitcode.ErrIllegalArgument, "Worker Key must be BLS.")
 		}
@@ -67,7 +69,7 @@ func (a *StorageMinerActor) StageWorkerKeyChange(rt Runtime, newKey addr.Address
 
 		// note that this may replace another pending key change
 		st.Info.PendingWorkerKey = keyChange
-		cronPayload := serde.MustSerializeParams(CronEventType_Miner_WorkerKeyChange, []abi.SectorNumber{})
+		cronPayload := serde.MustSerializeParams(CronEventType_Miner_WorkerKeyChange)
 		a._rtEnrollCronEvent(rt, rt.CurrEpoch()+keyChange.EffectiveAt, cronPayload)
 
 		return nil
@@ -141,7 +143,7 @@ func (a *StorageMinerActor) OnSurprisePoStChallenge(rt Runtime) *vmr.EmptyReturn
 
 	if challenged {
 		// Request deferred Cron check for SurprisePoSt challenge expiry.
-		cronPayload := serde.MustSerializeParams(CronEventType_Miner_SurpriseExpiration, []abi.SectorNumber{})
+		cronPayload := serde.MustSerializeParams(CronEventType_Miner_SurpriseExpiration)
 		surpriseDuration := indices.StorageMining_SurprisePoStChallengeDuration()
 		a._rtEnrollCronEvent(rt, rt.CurrEpoch()+surpriseDuration, cronPayload)
 	}
@@ -480,6 +482,7 @@ func (a *StorageMinerActor) OnDeferredCronEvent(rt Runtime, callbackPayload []by
 func (a *StorageMinerActor) Constructor(rt Runtime, ownerAddr addr.Address, workerAddr addr.Address, sectorSize abi.SectorSize, peerId peer.ID) *vmr.EmptyReturn {
 	rt.ValidateImmediateCallerIs(builtin.StoragePowerActorAddr)
 
+	// TODO: fix this, check that the account actor at the other end of this address has a BLS key.
 	if workerAddr.Protocol() != addr.BLS {
 		rt.Abort(exitcode.ErrIllegalArgument, "Worker Key must be BLS.")
 	}
--- a/actors/builtin/storage_miner/storage_miner_actor.go
+++ b/actors/builtin/storage_miner/storage_miner_actor.go
@@ -31,34 +31,47 @@ const epochUndefined = abi.ChainEpoch(-1)
 
 type StorageMinerActor struct{}
 
+type CronEventType int64
+
+const (
+	CronEventType_Miner_SurpriseExpiration = CronEventType(1)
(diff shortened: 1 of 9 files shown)


actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -200,10 +207,11 @@ func MinerInfo_New(
...
 	ret := &MinerInfo{
-		Owner:      ownerAddr,
-		Worker:     workerAddr,
-		PeerId:     peerId,
-		SectorSize: sectorSize,
+		Owner:            ownerAddr,
+		Worker:           workerAddr,
+		PeerId:           peerId,
+		SectorSize:       sectorSize,
+		PendingKeyChange: WorkerKeyChange{},
```
anorth: PendingWorkerKey?

actors/util/actor_util.go
```diff
@@ -67,6 +67,7 @@ type MinerSetHAMT map[addr.Address]bool
 type ActorIDSetHAMT map[abi.ActorID]bool
 
 type MinerEvent struct {
+	EventType abi.CronEventType
```
anorth: This event type is a parochial actor-specific issue leaking into the high levels. As is the sectors array below.

I think a superior pattern would be to replace both of those fields with a `[]byte` and let individual actors serialize something into those bytes as they wish (or not). Then the cron actor won't be burdened with unwanted knowledge about actor internals.

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -32,6 +32,58 @@ const epochUndefined = abi.ChainEpoch(-1)
...
+// Worker Key Change //
+///////////////////////
+
+func (a *StorageMinerActor) StageWorkerKeyChange(rt Runtime, key abi.PrivateKey) *vmr.EmptyReturn {
+	var st StorageMinerActorState
+	rt.State().Readonly(&st)
+	rt.ValidateImmediateCallerIs(st.Info.Owner)
+
+	// must be BLS since the worker key will be used alongside a BLS-VRF
+	keyAddr, err := addr.NewBLSAddress(key)
```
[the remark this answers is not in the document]
sternhenri: Sounds good. Decided to do a std check rather than assertion in order to be able to fail gracefully (`rt.Abort()`) rather than with a panic.

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -32,6 +32,58 @@ const epochUndefined = abi.ChainEpoch(-1)
...
+	if err != nil {
+		rt.Abort(exitcode.ErrIllegalArgument, "Cannot make new address from key.")
+	}
+
+	keyChange := WorkerKeyChange{
+		NewWorker:   keyAddr,
+		EffectiveAt: rt.CurrEpoch() + indices.StorageMining_WorkerKeyChangeFreeze(),
+	}
+
+	rt.State().Transaction(&st, func() interface{} {
```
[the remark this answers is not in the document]
sternhenri: Sounds good. Had felt odd to me to put caller val inside a transaction.

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -32,6 +32,58 @@ const epochUndefined = abi.ChainEpoch(-1)
...
+	rt.ValidateImmediateCallerIs(builtin.CronActorAddr)
+
+	var st StorageMinerActorState
+	rt.State().Transaction(&st, func() interface{} {
+		if (st.Info.PendingKeyChange == WorkerKeyChange{}) {
+			rt.Abort(exitcode.ErrIllegalState, "No pending key change.")
+		}
+
+		if st.Info.PendingKeyChange.EffectiveAt > rt.CurrEpoch() {
+			rt.Abort(exitcode.ErrIllegalState, "Too early for key change.")
```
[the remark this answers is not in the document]
sternhenri: will also be swalled by cron fyi... so either I should just return as well or likewise track the invalid state in the above condition no?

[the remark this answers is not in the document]
anorth: This actor should concern itself with what the appropriate behaviour is regardless of caller. I think abort is more appropriate here. But no objection if you return to aborting in the above case too.

actors/util/actor_util.go
sternhenri: So it's not quite as leaky as you think.

Specifically, this only applies to StorageMinerActor-triggered cron events. But the SMA relies on the SPA for callbacks here and never directly communicates with the CronActor. The SPA has a callback at every tick, and then uses these callback requests (and provided info) from the SMA to callback the appropriate miner actor as needed.

Changed nonetheless, and refactored to avoid noop checks with eventTypes

sternhenri:
cc @zixuanzh @jzimmerman for cron refactor. LMK if any issues. Seems more efficient (at the cost of slightly more complexity).

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -31,6 +31,71 @@ const epochUndefined = abi.ChainEpoch(-1)
...
+
+func (a *StorageMinerActor) StageWorkerKeyChange(rt Runtime, newKey addr.Address) *vmr.EmptyReturn {
+	var st StorageMinerActorState
+	rt.State().Transaction(&st, func() interface{} {
+		rt.ValidateImmediateCallerIs(st.Info.Owner)
+
+		// must be BLS since the worker key will be used alongside a BLS-VRF
+		if newKey.Protocol() != addr.BLS {
+			rt.Abort(exitcode.ErrIllegalArgument, "Worker Key must be BLS.")
+		}
```
anorth: Following on from my earlier comment, please leave a comment noting that this check isn't directly right, but we want to check that the account actor at the other end of this address has a BLS key.

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -31,6 +31,71 @@ const epochUndefined = abi.ChainEpoch(-1)
...
+
+		keyChange := WorkerKeyChange{
+			NewWorker:   newKey,
+			EffectiveAt: rt.CurrEpoch() + indices.StorageMining_WorkerKeyChangeFreeze(),
+		}
+
+		// note that this may replace another pending key change
+		st.Info.PendingWorkerKey = keyChange
+		cronPayload := serde.MustSerializeParams(CronEventType_Miner_WorkerKeyChange, []abi.SectorNumber{})
```
anorth: This is just a placeholder, but remove the empty last argument, which won't be involved.

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -76,8 +141,9 @@ func (a *StorageMinerActor) OnSurprisePoStChallenge(rt Runtime) *vmr.EmptyReturn
 
 	if challenged {
 		// Request deferred Cron check for SurprisePoSt challenge expiry.
-		provingPeriod := indices.StorageMining_SurprisePoStProvingPeriod()
-		a._rtEnrollCronEvent(rt, rt.CurrEpoch()+provingPeriod, []abi.SectorNumber{})
+		cronPayload := serde.MustSerializeParams(CronEventType_Miner_SurpriseExpiration, []abi.SectorNumber{})
```
anorth: Remove empty final param

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -360,25 +426,50 @@ func (a *StorageMinerActor) DeclareTemporaryFaults(rt Runtime, sectorNumbers []a
...
-func (a *StorageMinerActor) OnDeferredCronEvent(rt Runtime, sectorNumbers []abi.SectorNumber) *vmr.EmptyReturn {
+func (a *StorageMinerActor) OnDeferredCronEvent(rt Runtime, callbackPayload []byte) *vmr.EmptyReturn {
 	rt.ValidateImmediateCallerIs(builtin.StoragePowerActorAddr)
 
-	for _, sectorNumber := range sectorNumbers {
-		a._rtCheckTemporaryFaultEvents(rt, sectorNumber)
-		a._rtCheckPreCommitExpiry(rt, sectorNumber)
-		a._rtCheckSectorExpiry(rt, sectorNumber)
+	var payload CronEventPayload
+	serde.MustDeserialize(callbackPayload, payload)
```
anorth: Deserialize only inside the branches that will use the value

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -388,6 +479,11 @@ func (a *StorageMinerActor) OnDeferredCronEvent(rt Runtime, sectorNumbers []abi.
 
 func (a *StorageMinerActor) Constructor(rt Runtime, ownerAddr addr.Address, workerAddr addr.Address, sectorSize abi.SectorSize, peerId peer.ID) *vmr.EmptyReturn {
 	rt.ValidateImmediateCallerIs(builtin.StoragePowerActorAddr)
+
+	if workerAddr.Protocol() != addr.BLS {
+		rt.Abort(exitcode.ErrIllegalArgument, "Worker Key must be BLS.")
+	}
```
anorth: Please leave a comment here about the address type verification too

actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -98,6 +98,11 @@ type PreCommittedSectorsAMT map[abi.SectorNumber]SectorPreCommitOnChainInfo
 // TODO HAMT
 type SectorNumberSetHAMT map[abi.SectorNumber]bool
 
+type WorkerKeyChange struct {
+	NewWorker   addr.Address
```
anorth: // Must be an ID address (because the reverse lookup facility is not availble.)

actors/builtin/storage_miner/storage_miner_actor.go
sternhenri: These are getting deserialized into a structure that expects the struct. I thought it might be necessary..?

anorth: You might be right, until we write the real serialization code here. Fine to leave it.

actors/builtin/storage_miner/storage_miner_actor.go
sternhenri: I need to deserialize to branch (the event type is inside the serialized bytes)

PR #50, opened 2020-01-24
First implementation of mock runtime and construction unit tests for the multisig actor

anorth:
The test exposed a bug in the constructor failing to initialize a bigint value.

This is based on #45, which must land first.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -47,17 +47,18 @@ func (a *MultiSigActor) Constructor(rt vmr.Runtime, params *ConstructorParams) *
 		signers = append(signers, a)
 	}
 
-	var st MultiSigActorState
-	rt.State().Transaction(&st, func() interface{} {
+	rt.State().Construct(func() vmr.CBORMarshaler {
+		var st MultiSigActorState
 		st.Signers = signers
 		st.NumApprovalsThreshold = params.NumApprovalsThreshold
 		st.PendingTxns = autil.EmptyHAMT
+		st.InitialBalance = abi.NewTokenAmount(0)
 		if params.UnlockDuration != 0 {
-			st.UnlockDuration = params.UnlockDuration
 			st.InitialBalance = rt.ValueReceived()
+			st.UnlockDuration = params.UnlockDuration
 			st.StartEpoch = rt.CurrEpoch()
 		}
-		return nil
+		return &st
 	})
 	return &vmr.EmptyReturn{}
 }
--- a/support/mock/mockrt.go
+++ b/support/mock/mockrt.go
@@ -1,0 +1,326 @@
+package mock
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+
+	addr "github.com/filecoin-project/go-address"
+	cid "github.com/ipfs/go-cid"
+	mh "github.com/multiformats/go-multihash"
+
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/runtime"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	"github.com/filecoin-project/specs-actors/actors/runtime/indices"
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+)
+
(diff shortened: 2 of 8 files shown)


support/mock/mockrt.go
```diff
@@ -0,0 +1,326 @@
...
+func (rt *Runtime) ValidateImmediateCallerType(types ...cid.Cid) {
+	// Check and clear expectations.
+	if len(rt.expectValidateCallerType) == 0 {
+		rt.t.Errorf("unexpected validate caller code")
+	}
+	if !reflect.DeepEqual(rt.expectValidateCallerType, types) {
+		rt.t.Errorf("unexpected validate caller code %v, expected %v", types, rt.expectValidateCallerType)
+	}
+	defer func() {
+		rt.expectValidateCallerType = []cid.Cid{}
```
frrist: ditto my above comment here too.

support/mock/mockrt.go
```diff
@@ -0,0 +1,326 @@
...
+	if !found {
+		rt.Abort(exitcode.SysErrInternal, "actor state not found: %v", rt.state)
+	}
+}
+
+func (rt *Runtime) Transaction(st runtime.CBORer, f func() interface{}) interface{} {
+	if rt.inTransaction {
+		rt.Abort(exitcode.SysErrorIllegalActor, "nested transaction")
+	}
+	rt.Readonly(st)
```
frrist: Not a problem with the code here, but after seeing how this method is used I don't really like the name: `Readonly`. It gives me the impression this method does more than deserialize bytes into a state object, or maybe it intends to have more behaviour later in a non-mocked implementation?

support/mock/mockrt.go
```diff
@@ -0,0 +1,326 @@
...
+
+///// Inspection facilities /////
+
+func (rt *Runtime) StateRoot() cid.Cid {
+	return rt.state
+}
+
+func (rt *Runtime) GetState(o runtime.CBORUnmarshaler) {
+	rt.IpldGet(rt.state, o)
+}
```
frrist: Any reason to have this in addition to `Readonly`?

support/mock/mockrt.go
```diff
@@ -0,0 +1,326 @@
...
+
+func (rt *Runtime) IpldPut(o runtime.CBORMarshaler) cid.Cid {
+	r := bytes.Buffer{}
+	err := o.MarshalCBOR(&r)
+	if err != nil {
+		rt.Abort(exitcode.SysErrSerialization, err.Error())
+	}
+	data := r.Bytes()
+	key, err := cidBuilder.Sum(data)
```
frrist: `if err != nil` stuffs

actors/builtin/multisig/multisig_actor.go
```diff
@@ -47,17 +47,18 @@ func (a *MultiSigActor) Constructor(rt vmr.Runtime, params *ConstructorParams) *
...
 		st.PendingTxns = autil.EmptyHAMT
+		st.InitialBalance = abi.NewTokenAmount(0)
 		if params.UnlockDuration != 0 {
-			st.UnlockDuration = params.UnlockDuration
 			st.InitialBalance = rt.ValueReceived()
+			st.UnlockDuration = params.UnlockDuration
 			st.StartEpoch = rt.CurrEpoch()
 		}
-		return nil
+		return &st
```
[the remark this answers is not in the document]
anorth: This is because `Transaction` changed to `Construct`.

support/mock/mockrt.go
```diff
@@ -0,0 +1,326 @@
...
+		rt.t.Errorf("unexpected validate caller addrs")
+		return
+	}
+	if !reflect.DeepEqual(rt.expectValidateCallerAddr, addrs) {
+		rt.t.Errorf("unexpected validate caller addrs %v, expected %v", addrs, rt.expectValidateCallerAddr)
+		return
+	}
+	defer func() {
+		rt.expectValidateCallerAddr = []addr.Address{}
+	}()
```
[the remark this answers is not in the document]
anorth: Will use `nil` for now.

support/mock/mockrt.go
anorth: No, you're right, and we can probably improve the name (but will defer for now)

support/mock/mockrt.go
anorth: Removed

PR #51, opened 2020-01-24
Add network name to runtime. Move shared builtin code to builtin.

anorth:
This breaks the dependency of runtime on builtin.

### DIFF
--- a/actors/runtime/runtime.go
+++ b/actors/runtime/runtime.go
@@ -16,6 +16,9 @@ import (
 // Runtime is the VM's internal runtime object.
 // this is everything that is accessible to actors, beyond parameters.
 type Runtime interface {
+	// The network name, e.g. "mainnet".
+	NetworkName() string
+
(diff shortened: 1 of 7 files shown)


actors/runtime/runtime.go
```diff
@@ -16,6 +16,9 @@ import (
 // Runtime is the VM's internal runtime object.
 // this is everything that is accessible to actors, beyond parameters.
 type Runtime interface {
+	// The network name, e.g. "mainnet".
+	NetworkName() string
```
frrist: Do you think it would be worth making `NetworkName` a type? Something like:
```go
type Network int64

func (n Network) String() string {
	return [...]string{"mainnet", "testnet", "fishnet", "saftynet"}[n]
}

const (
	MainNet Network = iota
	TestNet
	FishNet
	SaftyNet
)
```

whyrusleeping: @frrist that particular implementation might be a bit panicky, but seems useful to me as a concept.

PR #52, opened 2020-01-24
refactor: propagate hamt errors through actor code and use store instead of runtime interface

frrist:


actors/builtin/multisig/multisig_actor_state.go
```diff
@@ -55,38 +55,40 @@ func (st *MultiSigActorState) _hasAvailable(currBalance abi.TokenAmount, amountT
...
-func (as *MultiSigActorState) getPendingTransaction(rt vmr.Runtime, txnID TxnID) MultiSigTransaction {
+func (as *MultiSigActorState) getPendingTransaction(rt vmr.Runtime, txnID TxnID) (MultiSigTransaction, error) {
 	hm := adt.NewMap(adt.AsStore(rt), as.PendingTxns)
 
 	var out MultiSigTransaction
-	err := hm.Get(txnID, &out)
-	requireNoStateErr(rt, err, "failed to get transaction %v from pending transaction HAMT", txnID)
+	found, err := hm.Get(txnID, &out)
+	if err != nil {
+		return MultiSigTransaction{}, errors.Wrapf(err, "failed to read transaction %v from HAMT %s", txnID, as.PendingTxns)
```
anorth: Both the key and the root are available inside the Map.Get, so I would put them in the inner message, then simplify this one to `failed to read transaction: `. General principle of including information at the lowest level it's available

actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -353,15 +381,20 @@ func (a *StoragePowerActor) Constructor(rt Runtime) *vmr.EmptyReturn {
...
 		return nil
 	})
 }
 
 func (a *StoragePowerActor) _rtDeductClaimedPowerForSectorAssert(rt Runtime, minerAddr addr.Address, storageWeightDesc SectorStorageWeightDesc) {
 	var st StoragePowerActorState
 	rt.State().Transaction(&st, func() interface{} {
-		st._deductClaimedPowerForSectorAssert(rt, minerAddr, storageWeightDesc)
+		if err := st._deductClaimedPowerForSectorAssert(adt.AsStore(rt), minerAddr, storageWeightDesc); err != nil {
+			rt.Abort(exitcode.ErrIllegalState, "failed to deducted claimed power for sector")
```
anorth: This non-exported method should probably return the wrapped error rather than abort.

actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -283,7 +308,10 @@ func (a *StoragePowerActor) ReportConsensusFault(rt Runtime, blockHeader1, block
 	var amountToSlasher abi.TokenAmount
 	var st StoragePowerActorState
 	rt.State().Transaction(&st, func() interface{} {
-		claimedPower, powerOk := getStoragePower(rt, st.ClaimedPower, slashee)
+		claimedPower, powerOk, err := getStoragePower(adt.AsStore(rt), st.ClaimedPower, slashee)
+		if err != nil {
+			rt.Abort(exitcode.ErrIllegalState, "spa.ReportConsensusFault failed to get miner power from claimed power table: %v", err)
```
anorth: "failed to read claimed power for fault"

actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -371,13 +404,17 @@ func (a *StoragePowerActor) _rtInitiateNewSurprisePoStChallenges(rt Runtime) {
...
 		randomness := crypto.DeriveRandWithEpoch(crypto.DomainSeparationTag_SurprisePoStSelectMiners, minerSelectionSeed, int(rt.CurrEpoch()))
 
 		IMPL_FINISH() // BigInt arithmetic (not floating-point)
 		challengeCount := math.Ceil(float64(st.MinerCount) / float64(provingPeriod))
-		surprisedMiners = st._selectMinersToSurprise(rt, int(challengeCount), randomness)
+		surprisedMiners, err = st._selectMinersToSurprise(adt.AsStore(rt), int(challengeCount), randomness)
+		if err != nil {
+			rt.Abort(exitcode.ErrIllegalState, "failed to select miner to surprise: %v", err)
+		}
 		return nil
```
anorth: Return the error from the Transaction and this method altogether

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -189,47 +194,72 @@ func (st *StoragePowerActorState) _updatePowerEntriesFromClaimedPower(rt vmr.Run
...
 }
 
-func (st *StoragePowerActorState) _setNominalPowerEntryInternal(rt vmr.Runtime, minerAddr addr.Address, updatedMinerNominalPower abi.StoragePower) {
+func (st *StoragePowerActorState) _setNominalPowerEntryInternal(s adt.Store, minerAddr addr.Address, updatedMinerNominalPower abi.StoragePower) error {
 	Assert(updatedMinerNominalPower.GreaterThanEqual(big.Zero()))
 
-	prevMinerNominalPower, ok := getStoragePower(rt, st.NominalPower, minerAddr)
+	prevMinerNominalPower, ok, err := getStoragePower(s, st.NominalPower, minerAddr)
+	if err != nil {
+		return errors.Wrap(err, " failed to get previous nominal miner power while setting nominal power table entry")
```
anorth: Unwanted space leading this message and some below

actors/builtin/storage_power/storage_power_actor.go
frrist: The name of this method seems to imply it will make some assertions about the success of the operation it performs. We can always change the name, but given how it reads _now_, returning errors feels wrong and also a big ugly:
```go
func (a *StoragePowerActor) _rtDeductClaimedPowerForSectorAssert(rt Runtime, minerAddr addr.Address, storageWeightDesc SectorStorageWeightDesc) error {
	var st StoragePowerActorState
	var txErr error
		if err := st._deductClaimedPowerForSectorAssert(adt.AsStore(rt), minerAddr, storageWeightDesc); err != nil {
                        // unless I can just return the error directly here??
			txErr = err
			return nil
		}
		return nil
	})
	return errors.Wrap(txErr, "failed to deducted claimed power for sector")
}
```
How strongly do we feel about errors propagating up from here (also cc @whyrusleeping)

actors/builtin/storage_power/storage_power_actor_state.go
frrist: whoops, sorry copy pasta :spaghetti:

actors/builtin/storage_power/storage_power_actor.go
anorth: Yeah please change the name.

PR #54, opened 2020-01-25
Implement NewEmptyMap, Map.CollectKeys and fix ForEach.

anorth:
Multisig and power actors correctly initialize empty maps at construction.

### DIFF
--- a/actors/util/adt/map.go
+++ b/actors/util/adt/map.go
@@ -1,11 +1,13 @@
 package adt
 
 import (
+	"bytes"
 	"context"
 
 	cid "github.com/ipfs/go-cid"
 	hamt "github.com/ipfs/go-hamt-ipld"
 	errors "github.com/pkg/errors"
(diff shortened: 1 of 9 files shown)


actors/util/adt/map.go
```diff
@@ -27,48 +29,50 @@ type Map struct {
...
-	oldRoot, err := hamt.LoadNode(h.store.Context(), h.store, h.root)
+	root, err := hamt.LoadNode(h.store.Context(), h.store, h.root)
 	if err != nil {
 		return false, errors.Wrapf(err, "Map Get failed to load node %v", h.root)
 	}
-	if err := oldRoot.Find(h.store.Context(), k.Key(), out); err != nil {
+	if err := root.Find(h.store.Context(), k.Key(), out); err != nil {
 		if err == hamt.ErrNotFound {
 			return false, nil
 		}
```
frrist: Changes I made while rebasing, I think its useful to return a boolean here for the case that the key was not found in the map.

PR #55, opened 2020-01-25
refactor: PoSt Detected Faults uses a HAMT set

frrist:
Rebased on #54 to implement a set by wrapping a map, will need to merge it before this.


actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -191,7 +192,9 @@ func (st *StoragePowerActorState) _updatePowerEntriesFromClaimedPower(s adt.Stor
 	// Currently, the only reason for these to differ is if the miner is in DetectedFault state
 	// from a SurprisePoSt challenge.
 	nominalPower := claimedPower
-	if st.PoStDetectedFaultMiners[minerAddr] {
+	if found, err := st.hasFault(s, minerAddr); err != nil {
+		return err
+	} else if !found {
```
anorth: This looks like an inversion from the original.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -277,6 +280,36 @@ func (st *StoragePowerActorState) _getPledgeSlashForConsensusFault(currPledge ab
...
+func (st *StoragePowerActorState) hasFault(s adt.Store, a addr.Address) (bool, error) {
+	hm := adt.AsSet(s, st.PoStDetectedFaultMiners)
+	found, err := hm.Has(asKey(a))
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get detected faults for address %v from set %s", a, st.PoStDetectedFaultMiners)
+	}
+	if !found {
+		return false, nil
+	}
+	return true, nil
```
anorth: `return found, nil`.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -277,6 +280,36 @@ func (st *StoragePowerActorState) _getPledgeSlashForConsensusFault(currPledge ab
 	}
 }
 
+func (st *StoragePowerActorState) hasFault(s adt.Store, a addr.Address) (bool, error) {
+	hm := adt.AsSet(s, st.PoStDetectedFaultMiners)
```
anorth: The name `hm` isn't informative here. `faultyMiners` or similar?

actors/util/adt/set.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+type hamtSetValue uint64
+
+func (s hamtSetValue) MarshalCBOR(w io.Writer) error {
+	if _, err := w.Write(cbg.CborEncodeMajorType(cbg.MajUnsignedInt, uint64(s))); err != nil {
+		return err
+	}
+	return nil
+}
```
anorth: It took me a while to figure out what's going on here, but I get it now - a sentinel value because you can't use nil.
- I think you can use `struct{}` (regardless of how we CBOR it), which will be cheaper in Go and more obviously empty, just like when used as a map value
- since it always encodes to the same byte, hardcode it rather than compute it
- is there prior art for using the value 1? We could use any single-byte encoding, such as the same thing as `EmptyReturn`, again a better representation of "nothing". 

Basically I think you can move/rename `EmptyReturn` to `adt.EmptyValue` and use it here.

actors/util/adt/set.go
```diff
@@ -0,0 +1,74 @@
...
+		m: AsMap(s, r),
+	}
+}
+
+// NewSet creates a new HAMT with root `r` and store `s`.
+func MakeEmptySet(s Store) (*Set, error) {
+	nd := hamt.NewNode(s)
+	newSet := AsSet(s, cid.Undef)
+	err := newSet.write(nd)
+	return newSet, err
```
anorth: Use MakeEmptyMap and wrap the result rather than duplicate the internals. Remove the then-unused `write` implementation.

actors/util/adt/set.go
frrist: Prior art here: https://github.com/filecoin-project/lotus/blob/master/chain/actors/actor_storagepower.go#L751

> Basically I think you can move/rename EmptyReturn to adt.EmptyValue and use it here.

will give it a try, good idea!

actors/util/adt/set.go
```diff
@@ -0,0 +1,83 @@
...
+	buf := make([]byte, 1)
+	_, err := r.Read(buf)
+	if err != nil {
+		return err
+	}
+	if buf[0] != emptyListEncoded {
+		return fmt.Errorf("invalid empty return %x", buf[0])
+	}
+	return nil
+}
```
anorth: Move this to `empty.go`.

PR #56, opened 2020-01-27
tests: add multisig actor unit test

frrist:


actors/builtin/callers.go
```diff
@@ -1,6 +0,0 @@
-package builtin
```
frrist: deleted this file since `CallerTypesSignable` has been moved to `codes.go`

actors/builtin/codes.go
```diff
@@ -37,4 +38,8 @@ func init() {
 	PaymentChannelActorCodeID = makeBuiltin("fil/1/paymentchannel")
 	MultisigActorCodeID = makeBuiltin("fil/1/multisig")
 	RewardActorCodeID = makeBuiltin("fil/1/reward")
+
+	// FIXME(frrist) this needs to happen here else it is initalized with cid.Undef.
+	// Set of actor code types that can represent external signing parties.
+	CallerTypesSignable = []cid.Cid{AccountActorCodeID, MultisigActorCodeID}
```
frrist: need to remove or address before merging

actors/util/adt/map.go
```diff
@@ -117,10 +117,18 @@ func (h *Map) ForEach(out vmr.CBORUnmarshaler, fn func(key string) error) error
...
 }
 
+type keyWrapper struct {
+	string
+}
+
+func (kw *keyWrapper) Key() string {
+	return kw.string
+}
+
```
frrist: This change make usage of the `CollectKeys()` method more ergonomic as it allows for values to be immiadetly fetched by their keys without casting, e.g.:
```go
keys, err := txns.CollectKeys()
var txn multisig.MultiSigTransaction
found, err := txns.Get(keys[0], &txn)
```

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,179 @@ func TestConstruction(t *testing.T) {
...
+			Value:    abi.NewTokenAmount(10),
+			Method:   0,
+			Params:   abi.MethodParams{},
+			Approved: []addr.Address{anne},
+		}, txn)
+	})
+
+	t.Run("propose with threshold met", func(t *testing.T) {
+		// TODO implement mockrt message send
+		t.SkipNow()
```
frrist: the mock runtime does not implement `Send()`, yet. Skip this test until its implemented.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -16,9 +16,9 @@ type TxnID int64
 
 func (t TxnID) Key() string {
 	// convert a TxnID to a HAMT key.
-	txnKey := make([]byte, 0, binary.MaxVarintLen64)
-	binary.PutVarint(txnKey, int64(t))
-	return string(txnKey)
+	txnKey := make([]byte, binary.MaxVarintLen64)
+	n := binary.PutVarint(txnKey, int64(t))
+	return string(txnKey[:n])
```
frrist: fixing a bug I introduced.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/filecoin-project/go-address v0.0.0-20200107215422-da8eea2842b5
+	github.com/filecoin-project/go-crypto v0.0.0-20191218222705-effae4ea9f03
```
frrist: used here: https://github.com/filecoin-project/specs-actors/pull/56/files#diff-d5777d603b48d25386edace2468db7cfR244-R256

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,214 @@ func TestConstruction(t *testing.T) {
...
+		assert.True(t, found)
+
+		// containing these values
+		assert.Equal(t, multisig.MultiSigTransaction{
+			To:       chuck,
+			Value:    abi.NewTokenAmount(10),
+			Method:   0,
+			Params:   abi.MethodParams{},
+			Approved: []addr.Address{anne},
+		}, txn)
```
frrist: This doesn't look right, in the event that the multisig actor has insufficient funds the transaction is still kept in its state.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+		var signers = []addr.Address{anne, bob}
+		var nilParams = abi.MethodParams{}
+
+		rt := builder.WithBalance(abi.NewTokenAmount(0), abi.NewTokenAmount(0)).Build()
+
+		actor.mustConstructMultisigActor(rt, numApprovals, unlockDuration, signers...)
+
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.mustProposeWithExitCode(rt, chuck, value, method, nilParams, exitcode.ErrInsufficientFunds)
```
anorth: To change in a follow-up: I think this should succeed, but the return value should contain the exit code of the nested send failure. This would be consistent with behaviour when N>1. But discuss with Lotus team first.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+//
+// Helper methods for calling multisig actor methods
+//
+
+type msActorHarness struct {
+	multisig.MultiSigActor
+	t testing.TB
+}
+
+func (h *msActorHarness) mustConstructMultisigActor(rt *mock.Runtime, numApprovalsThresh, unlockDuration int64, signers ...addr.Address) {
```
anorth: More concise names possible here, like `mustConstruct()`.

Since the methods call `verify`, which has the side-effect of resetting expectations, I think these methods should be named `verifyX` to make that more clear. They don't match the usual "must" convention of panicing (nor do they "require" fail hard).

So `verifyConstruct.`

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+		To:     to,
+		Value:  abi.NewTokenAmount(value),
+		Method: abi.MethodNum(method),
+		Params: params,
+	}
+	h.MultiSigActor.Propose(rt, proposeParams)
+	rt.Verify()
+}
+
+func (h *msActorHarness) mustProposeWithExitCode(rt *mock.Runtime, to addr.Address, value, method int64, params abi.MethodParams, exitCode exitcode.ExitCode) {
```
anorth: I don't think we need to duplicate each helper with a failing version. Would it work to
```
rt.ExpectAbort(exitCode, func() { a.verifyPropose(...) })
```

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+	}
+
+	rt.ExpectAbort(exitCode, func() {
+		h.MultiSigActor.Propose(rt, proposeParams)
+	})
+
+	rt.Verify()
+}
+
+func (h *msActorHarness) mustContainTransactions(rt *mock.Runtime, expected ...multisig.MultiSigTransaction) {
```
anorth: `assertTransactions` is a bit more clear they must exactly match, not just contain as a subset.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+	rt.Readonly(&st)
+
+	txns := adt.AsMap(rt.Store(), st.PendingTxns)
+	keys, err := txns.CollectKeys()
+	require.NoError(h.t, err)
+
+	if len(expected) == 0 {
+		require.Empty(h.t, keys)
+		return
+	}
```
anorth: Unnecessary, the line below does it.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+func (h *msActorHarness) mustConstructMultisigActor(rt *mock.Runtime, numApprovalsThresh, unlockDuration int64, signers ...addr.Address) {
+	constructParams := multisig.ConstructorParams{
+		Signers:               signers,
+		NumApprovalsThreshold: numApprovalsThresh,
+		UnlockDuration:        abi.ChainEpoch(unlockDuration),
+	}
+
+	rt.ExpectValidateCallerAddr(builtin.InitActorAddr)
+	constructRet := h.MultiSigActor.Constructor(rt, &constructParams)
+	require.Equal(h.t, runtime.EmptyReturn{}, *constructRet)
```
anorth: Use `assert` except in cases where the rest of the test is invalidated

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+	}
+
+	require.Equal(h.t, len(expected), len(keys))
+
+	for i, k := range keys {
+		var actual multisig.MultiSigTransaction
+		found, err := txns.Get(k, &actual)
+		require.NoError(h.t, err)
+		require.True(h.t, found)
+		require.Equal(h.t, expected[i], actual)
```
anorth: Assert, perhaps except for the `NoError` cases

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,10 +78,208 @@ func TestConstruction(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	pk := crypto.PublicKey(prv)
+	newAddr, err := addr.NewSecp256k1Address(pk)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return newAddr
+}
```
anorth: Remove

support/mock/mockrt.go
```diff
@@ -297,6 +315,41 @@ func (rt *Runtime) Store() adt.Store {
 
 ///// Mocking facilities /////
 
+type message struct {
```
anorth: Name this something like expectedMessage

support/mock/mockrt.go
```diff
@@ -44,6 +44,8 @@ type Runtime struct {
 	expectValidateCallerAny  bool
 	expectValidateCallerAddr []addr.Address
 	expectValidateCallerType []cid.Cid
+
+	sendingMessage *message
```
anorth: We're going to need a `[]*message` here, so let's just do it now. `ExpectSend` should append and `Send` should verify against the first in the queue.

support/mock/mockrt.go
```diff
@@ -185,7 +187,22 @@ func (rt *Runtime) Send(toAddr addr.Address, methodNum abi.MethodNum, params abi
...
 	}
-	panic("implement me")
+	sentMsg := &message{
+		to:     toAddr,
+		method: methodNum,
+		params: params,
+		value:  value,
+	}
+	if rt.sendingMessage == nil {
+		rt.t.Fatalf("unexpected message send: %v", sentMsg.Key())
```
anorth: Using a half-constructed expectedMessage here is a bit yuk. Just format the parameters into this string directly. Also no need for `fatal`.

support/mock/mockrt.go
```diff
@@ -297,6 +315,41 @@ func (rt *Runtime) Store() adt.Store {
...
+	value  abi.TokenAmount
+
+	// returns from applying message
+	sendReturn runtime.SendReturn
+	exitCode   exitcode.ExitCode
+}
+
+func (m *message) Key() string {
+	return fmt.Sprintf("to: %v from: %v params: %v value: %v", m.to.String(), m.method, m.params, m.value)
+}
```
anorth: I don't like the way this is abused as an informative message and a comparison key with a half-constructed other message. We can clean this up.

For verifying the expectation, instead write a `verify(t, to, from, etc...)` method

support/mock/mockrt.go
```diff
@@ -297,6 +315,41 @@ func (rt *Runtime) Store() adt.Store {
...
+
+func (rt *Runtime) ExpectSend(toAddr addr.Address, methodNum abi.MethodNum, params abi.MethodParams, value abi.TokenAmount, sendReturn runtime.SendReturn, exitCode exitcode.ExitCode) {
+	if rt.sendingMessage != nil {
+		rt.t.Errorf("previous unsent message %v", rt.sendingMessage.Key())
+	}
```
anorth: With a slice, this is fine. An actor must be able to send more than one message.

support/mock/mockrt.go
```diff
@@ -297,6 +315,41 @@ func (rt *Runtime) Store() adt.Store {
...
+	rt.sendingMessage = &message{
+		to:         toAddr,
+		method:     methodNum,
+		params:     params,
+		value:      value,
+		sendReturn: sendReturn,
+		exitCode:   exitCode,
+	}
+}
```
anorth: Nit: Move the above down below the existing expectations so they are in roughly the same order as the methods they're setting expectations about.

go.mod
anorth: Revert after removing that use.

support/mock/mockrt.go
frrist: If I add a `String()` method to message can we use that? It will save a considerable amount of typing (e.g. the line below this will need to be triple line wrapped)

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,7 +76,166 @@ func TestConstruction(t *testing.T) {
...
+
+		rt := builder.Build()
+
+		actor.verifyConstruct(rt, numApprovals, noUnlockDuration, signers...)
+
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.verifyPropose(rt, chuck, sendValue, builtin.MethodSend, nilParams)
+
+		actor.assertTransactions(rt, multisig.MultiSigTransaction{
```
anorth: As a final touch-up, add a brief statement of intent toward the end of these tests. E.g.
- // the transaction remains awaiting second approval
- // the transaction has been sent and cleaned up
- // the final approval was ignored until the balance is sufficient
- // the transaction is not persisted

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,7 +76,166 @@ func TestConstruction(t *testing.T) {
...
+//
+// Helper methods for calling multisig actor methods
+//
+
+type msActorHarness struct {
+	multisig.MultiSigActor
+	t testing.TB
+}
+
+func (h *msActorHarness) verifyConstruct(rt *mock.Runtime, numApprovalsThresh, unlockDuration int64, signers ...addr.Address) {
```
anorth: Final nit (sorry), let's adjust these names to `constructAndVerify` to make the state change more clear.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,7 +76,166 @@ func TestConstruction(t *testing.T) {
...
+
+func (h *msActorHarness) verifyPropose(rt *mock.Runtime, to addr.Address, value abi.TokenAmount, method abi.MethodNum, params abi.MethodParams) {
+	proposeParams := &multisig.ProposeParams{
+		To:     to,
+		Value:  value,
+		Method: method,
+		Params: params,
+	}
+	h.MultiSigActor.Propose(rt, proposeParams)
+	rt.Verify()
```
anorth: This isn't verifying anything, so let's not (and rename the method to `propose`)

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,7 +76,166 @@ func TestConstruction(t *testing.T) {
...
+		found, err := txns.Get(asKey(k), &actual)
+		require.NoError(h.t, err)
+		assert.True(h.t, found)
+		assert.Equal(h.t, expected[i], actual)
+	}
+}
+
+type strKeyWrapper struct {
+	string
+}
```
anorth: This could be just `type key string`

support/mock/mockrt.go
```diff
@@ -44,6 +44,8 @@ type Runtime struct {
 	expectValidateCallerAny  bool
 	expectValidateCallerAddr []addr.Address
 	expectValidateCallerType []cid.Cid
+
+	sendingMessage []*expectedMessage
```
anorth: `expectSends` or `expectMessages`

PR #59, opened 2020-01-28
update penalty deduction

zixuanzh:
Interpreter transfers gasReward from BurntFundsActor to RewardActor before Reward is paid out,

- gasPenalty is first deducted from gasReward+BlockReward
- penalty is sent back to the BurtFundsActor
- the remaining reward is then dispensed/garnished for the miner
- if penalty exceeds total reward, all reward will be penalized and the remaining penalty is forgone

### DIFF
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -112,11 +112,12 @@ func (a *RewardActor) WithdrawReward(rt vmr.Runtime) *adt.EmptyValue {
 	return &adt.EmptyValue{}
 }
 
+// gasReward is expected to be transferred to this actor by the runtime before invocation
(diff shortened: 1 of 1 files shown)


actors/builtin/reward/reward_actor.go
```diff
@@ -115,6 +115,7 @@ func (a *RewardActor) AwardBlockReward(
 	rt vmr.Runtime,
 	miner addr.Address,
 	penalty abi.TokenAmount,
+	gasReward abi.TokenAmount,
 	minerNominalPower abi.StoragePower,
 	currPledge abi.TokenAmount,
```
anorth: Please document these parameters. For gas reward, include the note that the gas reward value is expected to have been already transferred to this actor by the runtime priori to invoking this method.

PR #58, opened 2020-01-27
tests: add multisig approve tests

frrist:
This should go in after `frrist/multisig-unit-test`

PR #63, opened 2020-01-28
tests: add cancel transaction tests

frrist:


actors/builtin/multisig/multisig_test.go
```diff
@@ -358,7 +490,7 @@ func (h *msActorHarness) assertTransactions(rt *mock.Runtime, expected ...multis
 	keys, err := txns.CollectKeys()
 	assert.NoError(h.t, err)
 
-	assert.Equal(h.t, len(expected), len(keys))
+	require.Equal(h.t, len(expected), len(keys))
```
frrist: needs to be a require else the below loop will panic.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,134 @@ func TestApprove(t *testing.T) {
...
+	builder := mock.NewBuilder(context.Background(), t, receiver).WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
+
+	t.Run("simple propose and cancel", func(t *testing.T) {
+		rt := builder.Build()
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, signers...)
+
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		serParams := serde.MustSerialize(&multisig.ChangeNumApprovalsThresholdParams{NewThreshold: 1000})
```
anorth: I intended the `serde` package as just stubs to satisfy the compiler until we have cbor encoding for everything. Please avoid it here and serialize directly.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,134 @@ func TestApprove(t *testing.T) {
...
+		actor.propose(rt, chuck, sendValue, builtin.Method_MultiSigActor_ChangeNumApprovalsThreshold, serParams)
```
anorth: This is testing changing the number of approvals, which is orthogonal to cancellation. Remove this from this test, put it somewhere else. Such a test should verify that the new approval threshold is respected, by approving a transaction.

What happens if the threshold is reduced to a number that would cause some transaction/s to meet it? Sounds like more undefined behaviour.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,134 @@ func TestApprove(t *testing.T) {
...
+		rt.Verify()
+
+		rt.SetBalance(sendValue)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		rt.ExpectSend(chuck, builtin.MethodSend, nilParams, sendValue, nil, 0)
```
anorth: I do not expect cancelling a transaction to result in it being executed! How does this test pass? Maybe `Verify` is broken?

actors/builtin/multisig/multisig_test.go
anorth: Oooooh, this is really confusing because the proposed method is also on the multisig. That makes it really hard to follow. Please just use a SomeMethod=42 constant, and an arbitrary parameter object just for testing, nothing to do with multisig.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,134 @@ func TestApprove(t *testing.T) {
...
+
+	t.Run("signer fails to cancel transaction from another signer", func(t *testing.T) {
+		rt := builder.Build()
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, signers...)
+
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		serParams := serde.MustSerialize(&multisig.ChangeNumApprovalsThresholdParams{NewThreshold: 1000})
+		actor.propose(rt, chuck, sendValue, builtin.Method_MultiSigActor_ChangeNumApprovalsThreshold, serParams)
```
anorth: Ditto don't confuse multisig operations with delegated operations, here and elsewhere

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,134 @@ func TestApprove(t *testing.T) {
...
+		actor.assertTransactions(rt, multisig.MultiSigTransaction{
+			To:       chuck,
+			Value:    sendValue,
+			Method:   builtin.Method_MultiSigActor_ChangeNumApprovalsThreshold,
+			Params:   serParams,
+			Approved: []addr.Address{anne},
+		})
+	})
+
+	t.Run("fail to cancel transaction when not signer", func(t *testing.T) {
```
anorth: I can't find the difference between this test and the preceding. Please write more comments.

actors/builtin/multisig/multisig_test.go
frrist: In a [previous PR](https://github.com/filecoin-project/specs-actors/pull/58#discussion_r371912000) the suggestion was made to use non-zero methods and parameters here. Since the Multisig actor is the only actor with params I just grabbed them. These same parameters are also used on [lines above](https://github.com/filecoin-project/specs-actors/pull/63/files#diff-d5777d603b48d25386edace2468db7cfR202) as well. I'm going to change this, and the line above back to the zero method and nilParams since this makes the test hard to follow.

actors/builtin/multisig/multisig_test.go
frrist: The test above [bob](https://github.com/filecoin-project/specs-actors/pull/63/files#diff-d5777d603b48d25386edace2468db7cfR365) calls cancel and fails since they didn't create the transaction.

The test below [richard](https://github.com/filecoin-project/specs-actors/pull/63/files#diff-d5777d603b48d25386edace2468db7cfR394) calls cancel and fails since they are not a signer

actors/builtin/multisig/multisig_test.go
frrist: Indeed, verify is not ensuring the queue is empty. Will address.

actors/builtin/multisig/multisig_test.go
```diff
@@ -186,6 +185,8 @@ func TestApprove(t *testing.T) {
 	const noUnlockDuration = int64(0)
 	const numApprovals = int64(2)
 	const txnID = int64(0)
+	const fakeMethod = abi.MethodNum(42)
+	var fakeParams = abi.MethodParams{1, 2, 3, 4, 5}
```
anorth: In a follow-up, please delete the `abi.MethodParams` type and just use []byte. It's legacy.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,135 @@ func TestApprove(t *testing.T) {
...
+	t.Run("simple propose and cancel", func(t *testing.T) {
+		rt := builder.Build()
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, signers...)
+
+		// anne proposes a transaction
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.propose(rt, chuck, sendValue, fakeMethod, fakeParams)
+		rt.Verify()
```
anorth: Very helpful comments, thank you.

Later, blocks of lines like this could be factored into a one-liner, `proposeAndVerify`.

PR #65, opened 2020-01-28
test: add multisig tests for add, remove, swap signers and change threshold

frrist:


actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+		// Transaction should remain after invalid cancel
+		actor.assertTransactions(rt, multisig.MultiSigTransaction{
+			To:       chuck,
+			Value:    sendValue,
+			Method:   builtin.Method_MultiSigActor_ChangeNumApprovalsThreshold,
+			Params:   serParams,
+			Approved: []addr.Address{anne},
+		})
+	})
+}
```
anorth: I think this is duplicated from #63

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+		rt.Verify()
+
+		var st multisig.MultiSigActorState
+		rt.Readonly(&st)
+		assert.Equal(t, initialSigner, st.Signers)
+		assert.Equal(t, numApprovals, st.NumApprovalsThreshold)
+		assert.Equal(t, abi.NewTokenAmount(0), st.InitialBalance)
+		assert.Equal(t, abi.ChainEpoch(0), st.UnlockDuration)
+		assert.Equal(t, abi.ChainEpoch(0), st.StartEpoch)
+	})
```
anorth: These are great by-the-book orthogonal tests, but pragmatically I think you could combine these into one sequence against the same set-up state.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+		})
+		rt.Verify()
+
+		var st multisig.MultiSigActorState
+		rt.Readonly(&st)
+		assert.Equal(t, initialSigner, st.Signers)
+		assert.Equal(t, numApprovals, st.NumApprovalsThreshold)
+		assert.Equal(t, abi.NewTokenAmount(0), st.InitialBalance)
+		assert.Equal(t, abi.ChainEpoch(0), st.UnlockDuration)
+		assert.Equal(t, abi.ChainEpoch(0), st.StartEpoch)
```
anorth: Testing that the state hasn't changed after an abort is testing the mock, not the real code. Remove this - it's sufficient that abort is called.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+	var addedSigner = chuck
+
+	builder := mock.NewBuilder(context.Background(), t, receiver).WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
+
+	t.Run("simple add signer", func(t *testing.T) {
+		rt := builder.Build()
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, initialSigner...)
+
+		rt.SetCaller(receiver, builtin.AccountActorCodeID)
```
anorth: Please test that if the receiver is not the caller, the call is rejected.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+	var initialSigner = []addr.Address{anne, bob, chuck}
+
+	builder := mock.NewBuilder(context.Background(), t, receiver).WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
+
+	t.Run("simple remove signer", func(t *testing.T) {
+		rt := builder.Build()
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, initialSigner...)
+
+		rt.SetCaller(receiver, builtin.AccountActorCodeID)
```
anorth: Test caller validation for RemoveSigner too

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+
+		// the signers remain unchanged after a failed attempt to remove.
+		var st multisig.MultiSigActorState
+		rt.Readonly(&st)
+		assert.Equal(t, initialSigner, st.Signers)
+		assert.Equal(t, numApprovals, st.NumApprovalsThreshold)
+		assert.Equal(t, abi.NewTokenAmount(0), st.InitialBalance)
+		assert.Equal(t, abi.ChainEpoch(0), st.UnlockDuration)
+		assert.Equal(t, abi.ChainEpoch(0), st.StartEpoch)
+	})
```
anorth: Test failure to remove last signer (or success, if that's what it does)

Similar to above, I would collapse these into less code with more re-use of intermediate state.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+	var newSigner = chuck
+
+	builder := mock.NewBuilder(context.Background(), t, receiver).WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
+
+	t.Run("simple swap signer", func(t *testing.T) {
+		rt := builder.Build()
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, initialSigner...)
+
+		rt.SetCaller(receiver, builtin.AccountActorCodeID)
```
anorth: Test caller validation here too

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+		rt.Verify()
+
+		// threshold decreases
+		var st multisig.MultiSigActorState
+		rt.Readonly(&st)
+		assert.Equal(t, initialSigner, st.Signers)
+		assert.Equal(t, numApprovals-1, st.NumApprovalsThreshold)
+		assert.Equal(t, abi.NewTokenAmount(0), st.InitialBalance)
+		assert.Equal(t, abi.ChainEpoch(0), st.UnlockDuration)
+		assert.Equal(t, abi.ChainEpoch(0), st.StartEpoch)
```
anorth: Verifying these unchanged values is getting pretty repetitive. I think it would be fine to skip checking of things that aren't close to the value being changed. Alternative, factor this out into a `assertState(st, signers, approvals, ...)` function so it only takes one line.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +311,519 @@ func TestApprove(t *testing.T) {
...
+		// state remains unchanged
+		var st multisig.MultiSigActorState
+		rt.Readonly(&st)
+		assert.Equal(t, initialSigner, st.Signers)
+		assert.Equal(t, numApprovals, st.NumApprovalsThreshold)
+		assert.Equal(t, abi.NewTokenAmount(0), st.InitialBalance)
+		assert.Equal(t, abi.ChainEpoch(0), st.UnlockDuration)
+		assert.Equal(t, abi.ChainEpoch(0), st.StartEpoch)
+
+	})
```
anorth: I'd combine all these "failure to xxx" into one method.

actors/builtin/multisig/multisig_test.go
frrist: Talked in person, going to skip this as that would only test the mockrt.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,359 @@ func TestApprove(t *testing.T) {
...
+			initialSigners:   []addr.Address{anne},
+			initialApprovals: int64(2),
+
+			removeSigner: anne,
+			decrease:     false,
+
+			expectSigners:   nil,
+			expectApprovals: int64(1),
+			code:            exitcode.Ok,
+		},
```
frrist: Want to draw attention to this test: This doesn't seem like the right behaviour -- the Multisig actor should probably not allow the only signer to be removed.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -218,7 +218,7 @@ func (a *MultiSigActor) SwapSigner(rt vmr.Runtime, params *SwapSignerParams) *ad
 			rt.AbortStateMsg("Party not found")
 		}
 
-		if !st.isSigner(params.To) {
```
anorth: 🎉

actors/builtin/multisig/multisig_test.go
anorth: Please file and link an issue.

actors/builtin/multisig/multisig_test.go
```diff
@@ -311,6 +310,359 @@ func TestApprove(t *testing.T) {
...
+			initialThreshold: 2,
+			setThreshold:     -1,
+			code:             exitcode.ErrPlaceholder,
+		},
+		{
+			desc:             "fail to set threshold above number of signers",
+			initialThreshold: 2,
+			setThreshold:     int64(len(initialSigner) + 1),
+			code:             exitcode.ErrPlaceholder,
+		},
```
anorth: Please file an issue for "what happens when threshold is reduced below the number of approvers an existing transaction already has"

actors/builtin/multisig/multisig_actor.go
frrist: :bug: :hammer:

PR #66, opened 2020-01-28
Add exports, add methods struct

Kubuxu:


actors/builtin/multisig/multisig_actor.go
```diff
@@ -34,13 +34,41 @@ type MultiSigTransaction struct {
...
+	Approve                     int64
+	Cancel                      int64
+	ClearCompleted              int64
+	AddSigner                   int64
+	RemoveSigner                int64
+	SwapSigner                  int64
+	ChangeNumApprovalsThreshold int64
+}
+
+var Methods = musigMethods{1, 2, 3, 4, 5, 6, 7, 8, 9}
```
anorth: Please reconcile this with builtin/methods.go so the concept is not duplicated (for the multisig actor).

Since in general we need these to be outside the individual actor packages, to support inter-dependent storage power/miner/market actors, I suggest putting this in builtin/methods.go and replacing the existing values there. Leave the Exports() here though.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -34,13 +34,41 @@ type MultiSigTransaction struct {
...
+
+func (msa MultiSigActor) Exports() []interface{} {
+	return []interface{}{
+		1: msa.Constructor,
```
anorth: Can we use symbolic constants for the keys so it's more obviously correct?

`Methods.Constructor: msa.Constructor,`

Kubuxu: I cannot, the  would have to be constants.

anorth:
Sorry about the codecov block. @frrist will remove the check for now

actors/builtin/methods.go
```diff
@@ -25,14 +25,19 @@ const (
...
-	Method_MultiSigActor_RemoveAuthorizedParty
-	Method_MultiSigActor_SwapAuthorizedParty
-	Method_MultiSigActor_ChangeNumApprovalsThreshold
-)
+type msMethods struct {
+	Constructor                 abi.MethodNum
+	Propose                     abi.MethodNum
+	Approve                     abi.MethodNum
+	Cancel                      abi.MethodNum
+	ClearCompleted              abi.MethodNum
```
frrist: Multisig does automatic cleanup, given that I think we can remove this method.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -34,13 +34,29 @@ type MultiSigTransaction struct {
 
 type MultiSigActor struct{}
 
+func (msa MultiSigActor) Exports() []interface{} {
+	return []interface{}{
+		1: msa.Constructor,
+		2: msa.Propose,
+		3: msa.Approve,
+		4: msa.Cancel,
+		//5: msa.ClearCompleted, //TODO: sync with spec
```
frrist: ditto

frrist:
Clarify that merging this **isn't** blocked since codecov is not a requirement. But I will remove the :x:'s relating to code coverage in the list of checks.

actors/builtin/methods.go
Kubuxu: I would keep it here, until it is removed from https://github.com/filecoin-project/specs/blob/e331e26c185b8685eb52bb1b9df4e5519b43f38a/src/systems/filecoin_token/multisig/multisig_actor.id

Kubuxu:
@frrist the :x: is fine. It is to be expected from PR like this and as the codecov is not required, it is fine.

actors/builtin/methods.go
anorth: Ugh, that file shouldn't exist.

PR #67, opened 2020-01-28
add method to fetch available multisig balance

whyrusleeping:
This is a quick PR to prompt the discussion:
This is a very useful piece of information that people will want to query often. It makes sense to me to include it as a method here. However, I also had a conversation with @anorth about similar things, where we said it also makes sense to make these be methods on the state object, and that queries could load actor state into that struct and call the methods on the state objects.

Happy with whatever outcome, let's use this PR as a discussion space to all come to agreement on the desired approach.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -7,6 +7,7 @@ import (
 	cbg "github.com/whyrusleeping/cbor-gen"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
 	vmr "github.com/filecoin-project/specs-actors/actors/runtime"
 	exitcode "github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Additional context here:
I have an ask for a way to get this information from a group that we are working with. In lotus we have a CLI interface for the multisig, but we don't have any dedicated API methods. You can get this information for yourself through the API, but you would then have to implement all the mathematical logic clientside, which is suboptimal. If we add this as a method on the actor, then you can use the StateCall api to get the data you need.

anorth:
cc @acruikshank for the discussion

anorth:
My preferred approach would be to have queries and arithmetic like this as methods on the state object, and implementations provide access to the state object without involving the VM, method dispatch, or any actor code. Most state object methods probably need to be exported for external access.

The exported methods on the actors themselves would probably be limited to VM-dispatchable methods.

whyrusleeping:
@anorth the one problem i have with that approach (which is not necessarily a deal breaker) is that that logic is not accessible for anyone not writing go. For example, people interacting with lotus over the rpc api

acruikshank:
@whyrusleeping it seems this approach would just involve adding these state objects to the RPC layer. That would probably mean a slightly larger API, but individual queries would have simpler, more targeted signatures.

whyrusleeping:
@acruikshank I don't see how this solves the problem, unless youre saying "make special api calls for each of these state object methods". Which feels rather unsustainable to me, but on the other side, the same thing applies to adding more actor methods.

acruikshank:
@whyrusleeping I was assuming that would would have an endpoint for each of these methods, and also assuming that wouldn't be more difficult than exposing an actor method. I can see how that second assumption may be incorrect. I only have a slight preference for actor state methods over actor methods because they'll be cleaner to use in gfc. I would like to avoid having actor methods that do nothing but wrap actor state methods, but even that isn't the end of the world.

anorth:
We have been operating without these methods for a while now and it seems ok. I'm going to close the PR/discussion now, but welcome re-opening it if more info arises.

PR #68, opened 2020-01-29
Implement BalanceTable map specialization and use in power actor

anorth:
This same structure will then be usable in the market actor.

### DIFF
--- a/actors/builtin/storage_power/storage_power_actor.go
+++ b/actors/builtin/storage_power/storage_power_actor.go
@@ -1,6 +1,7 @@
 package storage_power
 
 import (
+	"fmt"
 	"math"
 
 	addr "github.com/filecoin-project/go-address"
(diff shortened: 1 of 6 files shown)


actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -37,44 +38,49 @@ type StoragePowerActor struct{}
 // Actor methods
 ////////////////////////////////////////////////////////////////////////////////
 
-func (a *StoragePowerActor) AddBalance(rt Runtime, minerAddr addr.Address) *adt.EmptyValue {
-	builtin.RT_MinerEntry_ValidateCaller_DetermineFundsLocation(rt, minerAddr, builtin.MinerEntrySpec_MinerOnly)
+func (a *StoragePowerActor) Constructor(rt Runtime) *adt.EmptyValue {
```
anorth: I moved the ctor to the top, and pushed the state initialization into the state type.

actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -294,32 +295,30 @@ func (a *StoragePowerActor) OnMinerEnrollCronEvent(rt Runtime, eventEpoch abi.Ch
...
+		store := adt.AsStore(rt)
+		claimedPower, powerOk, err := getStoragePower(store, st.ClaimedPower, target)
 		if err != nil {
 			rt.Abort(exitcode.ErrIllegalState, "spa.ReportConsensusFault failed to read claimed power for fault: %v", err)
 		}
 		if !powerOk {
 			rt.Abort(exitcode.ErrIllegalArgument, "spa.ReportConsensusFault: miner already slashed")
 		}
-		Assert(claimedPower.GreaterThan(big.Zero()))
+		Assert(claimedPower.GreaterThanEqual(big.Zero()))
```
anorth: @zixuanzh I think this was a bug, a miner might have dropped to zero power subsequent to misbehaving.

actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -294,32 +295,30 @@ func (a *StoragePowerActor) OnMinerEnrollCronEvent(rt Runtime, eventEpoch abi.Ch
...
 
-		currPledge, pledgeOk := st.getMinerPledge(slashee)
-		if !pledgeOk {
-			rt.Abort(exitcode.ErrIllegalArgument, "spa.ReportConsensusFault: miner has no pledge")
-		}
-		Assert(currPledge.GreaterThan(big.Zero()))
+		currPledge, err := st.getMinerPledge(store, target)
+		abortIfError(rt, err, "failed to get miner pledge")
+		Assert(currPledge.GreaterThanEqual(big.Zero()))
```
anorth: This changed too, it assumed the pledge function is always non-zero, but doesn't need that assumption to be correct.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -72,23 +92,6 @@ func (st *StoragePowerActorState) minerNominalPowerMeetsConsensusMinimum(s adt.S
...
-	Assert(ok)
-	st.EscrowTable = newTable
-
-	autil.TODO()
-	// Decide whether we can take any additional action if there is not enough
-	// pledge collateral to be slashed.
-
-	return amountSlashed
-}
-
```
anorth: Replaced with use of subtractMinerPledge

actors/util/adt/balancetable.go
```diff
@@ -0,0 +1,97 @@
...
+		return big.Zero(), err
+	}
+	return prev, nil
+}
+
+// Error type returned when an expected key is absent.
+type ErrNotFound struct {
+	Root cid.Cid
+	Key addr.Address
+}
```
anorth: For this particular interpretation of a map, a failed lookup is always an error, so I opted to return error rather than a boolean. This type exists so that the caller can set an appropriate exitcode.

actors/util/adt/store.go
```diff
@@ -0,0 +1,55 @@
...
+
+	addr "github.com/filecoin-project/go-address"
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-hamt-ipld"
+
+	vmr "github.com/filecoin-project/specs-actors/actors/runtime"
+)
+
+// Store defines an interface required to back the ADTs in this package.
+type Store interface {
```
anorth: This all is just moved out from other files.

actors/util/adt/store.go
```diff
@@ -0,0 +1,55 @@
+package adt
+
+import (
+	"context"
+
+	addr "github.com/filecoin-project/go-address"
+	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-hamt-ipld"
```
frrist: nit:
```suggestion
	cid "github.com/ipfs/go-cid"
	hamt "github.com/ipfs/go-hamt-ipld"
```

actors/builtin/storage_power/storage_power_actor.go
zixuanzh: possible 👍

PR #69, opened 2020-01-29
test: vesting from multisig actor

frrist:


actors/builtin/multisig/multisig_test.go
```diff
@@ -76,6 +76,114 @@ func TestConstruction(t *testing.T) {
...
+		rt.Verify()
+
+		// move the epoch ahead such that all FIL is unlocked.
+		rt.SetEpoch(unlockDuration)
+		// bob approves the transaction proposed by anne to send darlene FIL.
+		rt.SetCaller(bob, builtin.AccountActorCodeID)
+		// we expect a message send to darlene containing the method and value proposed by anne.
+		rt.ExpectSend(darlene, builtin.MethodSend, nilParams, abi.NewTokenAmount(multisigInitialBalance), nil, exitcode.Ok)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.approve(rt, 0)
```
anorth: This exposes a flaw in our runtime mock. The aborted `propose` should not persist state.

actors/builtin/multisig/multisig_actor_state.go
```diff
@@ -41,17 +41,26 @@ func (st *MultiSigActorState) isSigner(party address.Address) bool {
...
 	}
 
-	if big.Sub(currBalance, amountToSpend).LessThan(st.AmountLocked(currEpoch - st.StartEpoch)) {
-		return false
+	remainingBalance := big.Sub(currBalance, amountToSpend)
+	amountLocked := st.AmountLocked(currEpoch - st.StartEpoch)
+	thing := amountLocked.Int64()
+	thong := remainingBalance.Int64()
+	_ = thing
+	_ = thong
```
anorth: What's going on here?

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,6 +76,111 @@ func TestConstruction(t *testing.T) {
...
+
+	const unlockDuration = 10
+	var multisigInitialBalance = abi.NewTokenAmount(100)
+	var nilParams = abi.MethodParams{}
+
+	builder := mock.NewBuilder(context.Background(), receiver).
+		WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID).
+		WithEpoch(0).
+		// balance 0: current balance of the actor. receive: 100 the amount the multisig actor will be initalized with -- InitialBalance
+		WithBalance(abi.NewTokenAmount(0), multisigInitialBalance)
```
anorth: This isn't how it will happen. The balance will be >= received, so for construction they will match.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,6 +76,111 @@ func TestConstruction(t *testing.T) {
...
+		actor.constructAndVerify(rt, 2, unlockDuration, []addr.Address{anne, bob, charlie}...)
+
+		// anne proposes that darlene receives `multisgiInitialBalance` FIL.
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.propose(rt, darlene, multisigInitialBalance, builtin.MethodSend, nilParams)
+		rt.Verify()
+
+		// set the current balance of the multisig actor to its InitialBalance amount
+		rt.SetBalance(multisigInitialBalance)
```
anorth: This will become unnecessary.

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,6 +76,111 @@ func TestConstruction(t *testing.T) {
...
+		WithBalance(abi.NewTokenAmount(0), multisigInitialBalance)
+
+	t.Run("happy path full vesting", func(t *testing.T) {
+		rt := builder.Build(t)
+
+		actor.constructAndVerify(rt, 2, unlockDuration, []addr.Address{anne, bob, charlie}...)
+
+		// anne proposes that darlene receives `multisgiInitialBalance` FIL.
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
```
anorth: Set the received amount to zero here (and same in the other tests)

actors/builtin/multisig/multisig_test.go
```diff
@@ -75,6 +76,111 @@ func TestConstruction(t *testing.T) {
...
+		rt.SetEpoch(0 + unlockDuration/2)
+		rt.SetCaller(bob, builtin.AccountActorCodeID)
+		rt.ExpectSend(darlene, builtin.MethodSend, nilParams, big.Div(multisigInitialBalance, big.NewInt(2)), nil, exitcode.Ok)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.approve(rt, 0)
+		rt.Verify()
+
+	})
+
+	t.Run("fail to propose a transaction greater than locked amount", func(t *testing.T) {
```
anorth: the wording here is a bit off. You should be able to _propose_ a transaction spending the locked amount, but not submit the final approval for it. We should exercise that: roll it into the partial vesting test above: first try and fail to spend the whole balance.

this one might say "propose and autoapprove transaction above locked amount" (and you should also test that below the locked amount is ok.

support/mock/mockrt.go
```diff
@@ -207,7 +207,7 @@ func (rt *Runtime) Send(toAddr addr.Address, methodNum abi.MethodNum, params abi
 		rt.Abort(exitcode.SysErrorIllegalActor, "side-effect within transaction")
 	}
 	if len(rt.expectSends) == 0 {
-		rt.t.Errorf("unexpected expectedMessage to: %v method: %v, value: %v, params: %v", toAddr, methodNum, value, params)
+		rt.t.Fatalf("unexpected expectedMessage to: %v method: %v, value: %v, params: %v", toAddr, methodNum, value, params)
```
anorth: No need to be fatal?

actors/builtin/multisig/multisig_actor_state.go
frrist: cat walked across my keyboard, fixing..

support/mock/mockrt.go
frrist: If it's not fatal all the code below it will panic, but if we want this as an error I can modify the below to accommodate it.

PR #70, opened 2020-01-29
ci: remove red X's codecov shows on pr

frrist:


anorth:
Does it still post a comment? This PR doesn't seem to have one.

frrist:
Yes it should still post a comment, but only when the code coverage changes: https://github.com/filecoin-project/specs-actors/pull/70/files#diff-1f7b4b85ed8525c5239f741431a72872L16

anorth:
Ok, please change so that it always posts the comment.

PR #75, opened 2020-01-29
Mock runtime wraps the top-level call and rolls back state after an abort.

anorth:


support/mock/mockrt.go
```diff
@@ -392,18 +431,53 @@ func (rt *Runtime) Reset() {
...
+		if r == nil {
+			rt.t.Errorf("expected abort with code %v but call succeeded", expected)
+			return
+		}
 		a, ok := r.(abort)
 		if !ok {
 			panic(r)
 		}
 		if a.code != expected {
 			rt.t.Errorf("abort expected code %v, got %v %s", expected, a.code, a.msg)
```
frrist: I don't see any reason to continue execution (`Errorf`) if either of these assertions fail. Can we use `Fatalf` here?

PR #76, opened 2020-01-30
dep: port rle code from lotus

frrist:
RLE code as taken from here: https://github.com/filecoin-project/lotus/tree/master/extern/rleplus
Lazy package was taken from here: https://github.com/filecoin-project/lotus/tree/master/lib/rlepluslazy
The lazy package requires rle for testing.

In a follow on this will get hooked up with actor fault reporting.
Flagging that the issues found in https://github.com/filecoin-project/lotus/pull/1187 have been addressed in this PR.


Kubuxu:
Throw the not-lazy RLE+ code into an internal package so nobody uses it.

anorth:
It looks like this is a copy of code originally copied from go-filecoin (and subsequently improved). @Kubuxu @whyrusleeping if we commit this here will you import it back into Lotus so we can resolve the duplication? cc @ZenGround0 @acruikshank we should replace go-filecoin's too.

PR #78, opened 2020-01-31
Implemented Array and Multimap datastructures and use them for power actor cron callback queue

anorth:
~This doesn't compile yet only because it's waiting on https://github.com/filecoin-project/go-amt-ipld/pull/8 to unify the HAMT/AMT IPLD store interfaces (cc @frrist)~

The array type only supports contiguous arrays right now, but should be extendible to sparser arrays for other needs.


actors/builtin/storage_power/storage_power_actor.go
```diff
@@ -336,10 +336,10 @@ func (a *StoragePowerActor) OnMinerEnrollCronEvent(rt Runtime, params *OnMinerEn
 
 	var st StoragePowerActorState
 	rt.State().Transaction(&st, func() interface{} {
-		if _, found := st.CronEventQueue[params.eventEpoch]; !found {
-			st.CronEventQueue[params.eventEpoch] = []CronEvent{}
+		err := st.appendCronEvent(adt.AsStore(rt), params.eventEpoch, &minerEvent)
+		if err != nil {
+			rt.Abort(exitcode.ErrIllegalState, "failed to enroll cron event")
```
frrist: Include the wrapped error message in the Abort here and below.

PR #80, opened 2020-01-31
Replace init actor address map with HAMT.

anorth:
Also separated state into a different file.


actors/builtin/init/init_actor.go
```diff
@@ -16,44 +14,18 @@ import (
...
-	idAddr, err := addr.NewIDAddress(uint64(actorID))
-	autil.Assert(err == nil)
-	return idAddr
-}
-
+// The init actor uniquely has the power to create new actors.
+// It maintains a table resolving pubkey and temporary actor addresses to the canonical ID-addresses.
 type InitActor struct{}
 
 func (a *InitActor) Constructor(rt Runtime) *adt.EmptyValue {
```
frrist: To stick with the pattern used in other actors, this should accept a second param `_ *adt.EmptyValue`. Unless the init actor is a special case?

actors/builtin/init/init_actor_state.go
```diff
@@ -0,0 +1,70 @@
+package init
+
+import (
+	addr "github.com/filecoin-project/go-address"
+	cid "github.com/ipfs/go-cid"
+	errors "github.com/pkg/errors"
+	typegen "github.com/whyrusleeping/cbor-gen"
```
frrist: nit: we call this `cbg` everywhere except here.

actors/builtin/init/init_actor_state.go
```diff
@@ -0,0 +1,70 @@
...
+		AddressMap:  emptyMap.Root(),
+		NextID:      abi.ActorID(builtin.FirstNonSingletonActorId),
+		NetworkName: networkName,
+	}, nil
+}
+
+// Resolves an address to an ID-address, if possible.
+// If the provided address is not in the table, it is returned as-is. This means that ID-addresses
+// (which should only appear as values, not keys) and singleton actor addresses pass through unchanged.
+func (s *InitActorState) ResolveAddress(store adt.Store, address addr.Address) (addr.Address, error) {
```
frrist: Would it be worth checking if this (`address`) is an `ID` address before proceeding or is it not possible to reach that state?

anorth: The pass-through behaviour is useful to a caller who just wants to end up with an ID-address one way or another.

PR #81, opened 2020-01-31
Params object and cbor-gen for storage market actor and cron event data.

anorth:
cc @magik6k


actors/builtin/storage_market/storage_market_actor.go
```diff
@@ -316,7 +323,7 @@ func (a *StorageMarketActor) OnEpochTickEnd(rt Runtime) *adt.EmptyValue {
 	return &adt.EmptyValue{}
 }
 
-func (a *StorageMarketActor) Constructor(rt Runtime) *adt.EmptyValue {
+func (a *StorageMarketActor) Constructor(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
```
frrist: nit: can this be moved to the top of the file so it matches the rest of the actors?

actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -511,7 +527,10 @@ func (a *StorageMinerActor) OnDeferredCronEvent(rt Runtime, params *OnDeferredCr
 	rt.ValidateImmediateCallerIs(builtin.StoragePowerActorAddr)
 
 	var payload CronEventPayload
-	serde.MustDeserialize(params.callbackPayload, payload)
+	err := payload.UnmarshalCBOR(bytes.NewReader(params.callbackPayload))
+	if err != nil {
+		rt.Abort(exitcode.ErrIllegalArgument, "failed to deserialize event payload")
```
frrist: This abort message would be more helpful if it included the error. Can also save a line condensing the if statement `if err := ...; err != nil`

actors/builtin/storage_market/storage_market_actor.go
anorth: I'm trying to reduce conflicts with @magik6k implementing the state proper

PR #60, opened 2020-01-28
label TODOs

sternhenri:
**In this PR:**
- Handled a few rote TODOs, namely: 
  - `abortstatemsg` -> `abort`
  - `bytes` -> `[]byte`
  - stale TODOs
- labeled TODOs with likely DRIs

@zixuanzh see if you can knock any others out or pick other DRIs if you think there is better?

Thereafter good to merge.

PR #82, opened 2020-01-31
TODO fixes

zixuanzh:
per #60, in this PR,

- BigInt operations have been implemented in several places, removing IMPL_FINISH
- Implemented  BigInt in Reward vesting
- Reward vesting may still exist and linear should suffice unless research shows otherwise
- Tentatively, miners dont get consensus power for nominal power that they dont have pledge collateral for

### DIFF
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -2,7 +2,6 @@ package reward
 
 import (
 	"io"
-	"math"
 	"sort"
 
 	addr "github.com/filecoin-project/go-address"
@@ -24,10 +23,6 @@ type VestingFunction int64
 const (
 	None VestingFunction = iota
 	Linear
-	// TODO zx: potential options
-	// PieceWise
-	// Quadratic
-	// Exponential
 )
 
 type Reward struct {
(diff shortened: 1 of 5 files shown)


actors/builtin/reward/reward_actor.go
```diff
@@ -43,9 +38,9 @@ func (r *Reward) AmountVested(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 	case None:
 		return r.Value
 	case Linear:
-		TODO() // zx BigInt
-		vestedProportion := math.Max(1.0, float64(elapsedEpoch)/float64(r.StartEpoch-r.EndEpoch))
-		return big.Mul(r.Value, big.NewInt(int64(vestedProportion)))
+		vestDuration := big.Sub(big.NewInt(int64(r.EndEpoch)), big.NewInt(int64(r.StartEpoch)))
+		unitVest := big.Div(r.Value, vestDuration)
```
anorth: It's a nit, but when value (in attofil) is less than duration (in epochs), the vesting rate of < 1aF/epoch rounds to zero. This will happen when block rewards approach zero, (at which point we don't really need vesting).

The most correct solution is to rearrange arithmetic so the division happens _last_ (after the mul below).
Alternatively, add a guard before this to vest fully when the rate is <1 (generous) or vest fully when elapsed > duration (conservative).

PR #85, opened 2020-01-31
refactor: FaultSet is a RLE+ Encoded Bitset.

frrist:
This PR:
- ports the bit field type from lotus: 0fc7be7ae9abdab4f57eee64e8514c3f667281be
- Modifies it to fit the needs of the sotrage_miner actor: 01c81720096312fdec8cbf1679f6d7d965c81eac
- Instruments the storage miner actor with it: 2b2e4c84f89eaced9b50fd98709a596752e882fb


actors/abi/bitfield.go
```diff
@@ -168,11 +178,6 @@ func (bf BitField) AllMap(max uint64) (map[uint64]bool, error) {
 }
 
 func (bf BitField) MarshalCBOR(w io.Writer) error {
-	ints := make([]uint64, 0, len(bf.bits))
-	for i := range bf.bits {
-		ints = append(ints, i)
-	}
-
```
frrist: Not sure what this loop was here for, removing it since it does nothing.

actors/abi/bitfield.go
```diff
@@ -110,6 +110,16 @@ func (bf BitField) Set(bit uint64) {
...
+// Unset removes bit from the BitField
+func (bf BitField) Unset(bit uint64) {
+	delete(bf.bits, bit)
+}
+
+// Has returns true iff bif is set in the BitField.
+func (bf BitField) Has(bit uint64) bool {
+	return bf.bits[bit] == struct{}{}
+}
+
```
frrist: These didn't exist in the lotus implementation, added them here to fit our usage.

actors/abi/bitfield.go
```diff
@@ -0,0 +1,232 @@
...
+// Unset removes bit from the BitField
+func (bf BitField) Unset(bit uint64) {
+	delete(bf.bits, bit)
+}
+
+// Has returns true iff bif is set in the BitField.
+func (bf BitField) Has(bit uint64) bool {
+	return bf.bits[bit] == struct{}{}
```
anorth: Use the `_, ok := bits[bit]` style, which will stay correct if the value type changes.

actors/abi/bitfield.go
```diff
@@ -0,0 +1,232 @@
...
+
+	rlepluslazy "github.com/filecoin-project/specs-actors/actors/util/rleplus"
+)
+
+var ErrBitFieldTooMany = errors.New("to many items in RLE")
+
+type BitField struct {
+	rle rlepluslazy.RLE
+
+	bits map[uint64]struct{}
```
anorth: I trust this is a direct port from Lotus, but I must admit great confusion from reading the code what the rules or expectations are about when one or the other (or both?) of these fields is actually initialized.

I wish there were tests to come with it demonstrating expected use.

actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -17,7 +17,7 @@ import (
 type StorageMinerActorState struct {
 	PreCommittedSectors PreCommittedSectorsAMT // PreCommitted Sectors
 	Sectors             SectorsAMT             // Proven Sectors can be Active or in TemporaryFault
-	FaultSet            SectorNumberSetHAMT    // TODO zx bitfield of Sectors in TemporaryFault
+	FaultSet            abi.BitField           // TODO zx bitfield of Sectors in TemporaryFault
```
anorth: Remove TODO

actors/abi/bitfield.go
```diff
@@ -0,0 +1,232 @@
...
+
+}
+
+func BitFieldFromSet(setBits []uint64) BitField {
+	res := BitField{bits: make(map[uint64]struct{})}
+	for _, b := range setBits {
+		res.bits[b] = struct{}{}
+	}
+	return res
+}
```
anorth: This method is unused in Lotus and here. @magik6k can we drop it?

frrist:
@anorth I've filed an issue in lotus to track replacement of packages https://github.com/filecoin-project/lotus/issues/1194

actors/abi/bitfield.go
magik6k: Sounds good

actors/abi/bitfield.go
magik6k: I'm pretty sure these won't just work a implemented here, cc @Kubuxu

actors/abi/bitfield.go
Kubuxu: In essence, as implemented in lotus. BitField can either be used for listing all bits (`All`, `AllMap`) or adding bits.
When bits are added, they are set in the `bits` map and latter folded into RLE.

If there is a need of it, I can add `Has` function, adding `Delete` function would be harder.

frrist: It looks like the the contents of `bits` is not preserved after marshaling and unmarshaling, is that correct? 

If that is the case `Has()` logic could (naively) be implemented by calling `AllMap` and checking the map for existence? And `Unset` could be implemented by calling `All`, removing the value from the slice and then passing it to `BitFieldFromSet`?

actors/abi/bitfield.go
frrist: Now using this method here: https://github.com/filecoin-project/specs-actors/pull/85/files#diff-3718ef74dd0a657e370fc4ee95eb337fR127

actors/abi/bitfield.go
```diff
@@ -0,0 +1,257 @@
...
+		return err
+	}
+	bitslice, err := bf.All(count)
+	if err != nil {
+		return err
+	}
+	for i, b := range bitslice {
+		if b == bit {
+			bitslice[i] = bitslice[len(bitslice)-1]
+			*bf = BitFieldFromSet(bitslice[:len(bitslice)-1])
```
frrist: Would returning a new BitField without `bit` be prefer over this?

actors/abi/bitfield.go
Kubuxu: Yes, it could be implemented as such. For removing you would have to change how the marshaling is handled currently it does `or` operation between existing bitset and `bits`.

actors/abi/bitfield.go
Kubuxu: Why are we crating a new bitfield on every iteration of the loop?

frrist: We create a new bitfield when we find the bit we are unsetting (`bit`) is found in the current bitfield, then we remove it (without preserving order) from `bf` and return. This method should create at most 1 new BitField. 

I assume the slice returned from `All` doesn't contain any duplicates?

PR #86, opened 2020-01-31
Change Send() to take parameters as CBORMarshaler rather than []byte.

anorth:
Fix serialization for a bunch of param/return objects.
Remove placeholder serde methods in favour of real CBOR.


actors/abi/primitives.go
```diff
@@ -32,9 +32,6 @@ type ActorID int64
 // associated with methods in Filecoin VM Actors.
 type MethodNum int64
 
-// Method params are the CBOR-serialization of a heterogenous array of values.
-type MethodParams []byte
```
frrist: :tada:

actors/builtin/multisig/multisig_test.go
```diff
@@ -979,8 +980,8 @@ func (h *msActorHarness) assertTransactions(rt *mock.Runtime, expected ...multis
 	require.Equal(h.t, len(expected), len(keys))
 	for i, k := range keys {
 		var actual multisig.MultiSigTransaction
-		found, err := txns.Get(asKey(k), &actual)
-		require.NoError(h.t, err)
+		found, err_ := txns.Get(asKey(k), &actual)
+		require.NoError(h.t, err_)
```
frrist: Avoiding the shadow here feels a bit extreme, is this necessary?

anorth: I would like to reach the point of enabling it as a linter rule eventually

PR #89, opened 2020-02-01
Finish reward actor with HAMT/AMT, params objects, cbor, exports

anorth:


actors/builtin/reward/reward_actor_state.go
```diff
@@ -0,0 +1,130 @@
...
+	StartEpoch      abi.ChainEpoch
+	EndEpoch        abi.ChainEpoch
+	Value           abi.TokenAmount
+	AmountWithdrawn abi.TokenAmount
+}
+
+type VestingFunction int64
+const (
+	None VestingFunction = iota
+	Linear
```
zixuanzh: making a note that we might have one more option, more details to come next week cc @jzimmerman

actors/builtin/reward/reward_actor_state.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+type InitialReward struct {
```
zixuanzh: What are these rewards? Investor tokens should already be in multisig.

gen/gen.go
```diff
@@ -38,6 +39,19 @@ func main() {
...
 
+	if err := gen.WriteTupleEncodersToFile("./actors/builtin/reward/cbor_gen.go", "reward",
+		// actor state
+		reward.RewardActorState{},
+		// method params
+		reward.ConstructorParams{},
+		reward.AwardBlockRewardParams{},
+		// other types
+		reward.Reward{},
+		reward.InitialReward{},
```
zixuanzh: ditto

actors/builtin/reward/reward_actor_state.go
anorth: Great! I'll remove this.

PR #90, opened 2020-02-03
DO NOT MERGE - Incorrect test for vesting more funds than locked at the moment.

Alexey-N-Chernyshov:
MultisigActorState doesn't save initial amount spent. Here is an example how to spend 150% of initial amount in 3 transactions after 50% of locked period has past.

IMO MultisigActor logic is overloaded with vesting logic, i'd split into MultisigActor and VestingActor if locked amount is needed, so MultisigActor could send msgs to VestingActor to unlock amount with multisig.


anorth:
Resolved by #91

PR #91, opened 2020-02-03
test: mockrt updates balance on sends

frrist:
The mock runtime does not automatically track balance changes via message send. Currently this means test writers need to call `rt.SetBalance(/* expected balance after send */)` after a method performs a send operation to update actor balance. #90 has brought to our attention that this is not ergonomic for test writer. This PR updates the mock runtimes `Send()` operation to modify the `balance` value after a message is sent.

### DIFF
--- a/support/mock/mockrt.go
+++ b/support/mock/mockrt.go
@@ -217,7 +217,7 @@ func (rt *Runtime) Send(toAddr addr.Address, methodNum abi.MethodNum, params run
 	}
 
 	if value.GreaterThan(rt.balance) {
-		rt.Abort(exitcode.ErrInsufficientFunds, "cannot send value: %v exceeds balance: %v", value, rt.balance)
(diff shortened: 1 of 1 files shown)


support/mock/mockrt.go
```diff
@@ -209,15 +210,20 @@ func (rt *Runtime) Send(toAddr addr.Address, methodNum abi.MethodNum, params run
...
-
 	expectedMsg := rt.expectSends[0]
 
 	if expectedMsg.Equal(toAddr, methodNum, params, value) {
 		rt.t.Errorf("expectedMessage being sent does not match expectation. Message: to %v method: %v value: %v params: %v, Expected: %v", toAddr, methodNum, value, params, rt.expectSends[0].String())
 	}
-	// pop the expectedMessage from the queue
+
+	if value.GreaterThan(rt.balance) {
+		rt.Abort(exitcode.ErrInsufficientFunds, "cannot send value: %v exceeds balance: %v", value, rt.balance)
```
anorth: Nit, but this should use `SysErrInsufficientFunds` (errors from within the runtime are "system errors")

PR #95, opened 2020-02-04
remove param finish

zixuanzh:
In this PR,

- Commented out PARAM_FINISH but did not remove to keep track of parameters and functions that need to be set so that it can be run
- Implemented big int operation for slashing reward
- Proportion related calculation, specifically related to power, is left as zero, they will be fixed in a separate PR when the power table is removed as discussed

### DIFF
--- a/actors/builtin/storage_power/policy.go
+++ b/actors/builtin/storage_power/policy.go
@@ -1,26 +1,27 @@
 package storage_power
 
 import (
+	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
 	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
-	autil "github.com/filecoin-project/specs-actors/actors/util"
+	indices "github.com/filecoin-project/specs-actors/actors/runtime/indices"
 )
 
 // TerminationFault
 func pledgePenaltyForSectorTermination(weight SectorStorageWeightDesc, termType builtin.SectorTermination) abi.TokenAmount {
-	autil.PARAM_FINISH()
+	// autil.PARAM_FINISH()
 	return abi.NewTokenAmount(0)
 }
 
 // DetectedFault
 func pledgePenaltyForSurprisePoStFailure(claimedPower abi.StoragePower, failures int64) abi.TokenAmount {
-	autil.PARAM_FINISH()
+	// autil.PARAM_FINISH()
 	return abi.NewTokenAmount(0)
 }
 
 func pledgePenaltyForConsensusFault(currPledge abi.TokenAmount, faultType ConsensusFaultType) abi.TokenAmount {
 	// default is to slash all pledge collateral for all consensus fault
-	autil.TODO()
+	// autil.PARAM_FINISH()
 	switch faultType {
 	case DoubleForkMiningFault:
 		return currPledge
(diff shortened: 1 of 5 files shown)


actors/builtin/storage_power/policy.go
```diff
@@ -34,16 +35,18 @@ func pledgePenaltyForConsensusFault(currPledge abi.TokenAmount, faultType Consen
...
 func rewardForConsensusSlashReport(elapsedEpoch abi.ChainEpoch, collateralToSlash abi.TokenAmount) abi.TokenAmount {
-	autil.TODO()
-	// BigInt Operation
-	// var growthRate = builtin.SLASHER_SHARE_GROWTH_RATE_NUM / builtin.SLASHER_SHARE_GROWTH_RATE_DENOM
-	// var multiplier = growthRate^elapsedEpoch
-	// var slasherProportion = min(INITIAL_SLASHER_SHARE * multiplier, 1.0)
-	// return collateralToSlash * slasherProportion
-	return abi.NewTokenAmount(0)
+	// autil.PARAM_FINISH()
+	growthRate := big.Div(indices.ConsensusFault_SlasherShareGrowthRateNum(), indices.ConsensusFault_SlasherShareGrowthRateDenom())
```
anorth: This evaluates to a constant `1`. The point of separating numerator and denominator is to re-arrange the arithmetic to do division last. Please document the canonical expression in a comment, and then code the arithmetic to work with integers.

actors/builtin/storage_power/policy.go
```diff
@@ -34,16 +35,18 @@ func pledgePenaltyForConsensusFault(currPledge abi.TokenAmount, faultType Consen
...
+	multiplier := big.NewInt(1)
+	for i := int64(0); i < int64(elapsedEpoch); i++ {
+		multiplier = big.Mul(multiplier, growthRate)
+	}
+	initial_slasher_share := big.Div(indices.ConsensusFault_SlasherInitialShareNum(), indices.ConsensusFault_SlasherInitialShareDenom())
```
anorth: This is a constant `0`.

actors/builtin/storage_power/policy.go
```diff
@@ -34,16 +35,18 @@ func pledgePenaltyForConsensusFault(currPledge abi.TokenAmount, faultType Consen
...
+	slasher_proportion := big.Min(big.Mul(initial_slasher_share, multiplier), big.NewInt(1))
+	return big.Mul(collateralToSlash, slasher_proportion)
 }
 
 func consensusPowerForWeight(weight SectorStorageWeightDesc) abi.StoragePower {
-	autil.PARAM_FINISH()
-	panic("")
+	// autil.PARAM_FINISH()
+	return big.Mul(big.NewInt(int64(weight.SectorSize)), big.NewInt(int64(weight.Duration)))
```
anorth: This looks wrong. Power is measured in bytes, not byte-epochs.

actors/builtin/storage_power/policy.go
```diff
@@ -1,26 +1,27 @@
...
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
 	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
-	autil "github.com/filecoin-project/specs-actors/actors/util"
+	indices "github.com/filecoin-project/specs-actors/actors/runtime/indices"
 )
 
 // TerminationFault
 func pledgePenaltyForSectorTermination(weight SectorStorageWeightDesc, termType builtin.SectorTermination) abi.TokenAmount {
-	autil.PARAM_FINISH()
+	// autil.PARAM_FINISH()
```
anorth: Remove the `autil` and `()` from there everywhere. We can just use the token `PARAM_FINISH` in a comment. to mark these places

actors/runtime/indices/indices.go
```diff
@@ -146,25 +133,26 @@ func (inds *IndicesImpl) SectorWeight(
 ) abi.SectorWeight {
 	// for every sector, given its size, start, end, and deals within the sector
 	// assign sector power for the duration of its lifetime
-	PARAM_FINISH()
-	panic("")
+	// PARAM_FINISH()
+	sectorDuration := endEpoch - startEpoch
+	return abi.SectorWeight(big.NewInt(int64(sectorDuration) * int64(sectorSize)))
```
anorth: This is the wrong units. Just return sectorSize if you want to ignore deal weight for now.

actors/runtime/indices/indices.go
```diff
@@ -181,22 +169,23 @@ func (inds *IndicesImpl) StoragePower(
...
 }
 
 func (inds *IndicesImpl) CurrEpochBlockReward() abi.TokenAmount {
 	// total block reward allocated for CurrEpoch
 	// each expected winner get an equal share of this reward
 	// computed as a function of NetworkKPI, NetworkTarget, LastEpochReward, TotalUnmminedFIL, etc
-	PARAM_FINISH()
-	panic("")
+	// PARAM_FINISH()
+	return abi.NewTokenAmount(0)
```
anorth: At least make this non-zero so a network could progress.

Also ⚠️ here is a function that does not have the right type, and fails to declare the parameters it needs.

actors/runtime/indices/indices.go
```diff
@@ -214,68 +203,67 @@ func (inds *IndicesImpl) NetworkTransactionFee(
...
 }
 
 func (inds *IndicesImpl) GetCurrBlockRewardForMiner(
 	minerStoragePower abi.StoragePower,
 	minerPledgeCollateral abi.TokenAmount,
 ) abi.TokenAmount {
-	PARAM_FINISH()
-	panic("")
+	// PARAM_FINISH()
+	return abi.NewTokenAmount(0)
```
anorth: Non-zero please

actors/runtime/indices/indices.go
zixuanzh: Understood, it is unresolved how SectorWeight should be computed and integral of space over time can be an option. Will use sectorSize for now.

actors/builtin/storage_power/policy.go
zixuanzh: Similar to comment at SectorWeight, but will revert to just SectorSize for now

actors/runtime/indices/indices.go
zixuanzh: ack, per input type, we were assuming this is computed from the state tree before but I think we can update the input type here now.

zixuanzh: will leave out the details for a future PR

actors/runtime/indices/indices.go
```diff
@@ -82,6 +83,8 @@ type IndicesImpl struct {
 	// they are treated as globally available states
 	Epoch                      abi.ChainEpoch
 	NetworkKPI                 big.Int
+	NetworkIndicator           big.Int
+	NetworkTime                big.Int
```
anorth: Well those are opaque! Please document.

zixuanzh: will document this in a separate PR, thanks for flagging

PR #97, opened 2020-02-04
Test multisig actor removal

Alexey-N-Chernyshov:
Undefined behavior after tx creator removal - now the 2-nd signer is tx creator.

The order described [here](
https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go#L31) is violated.

### DIFF
--- a/actors/builtin/multisig/multisig_test.go
+++ b/actors/builtin/multisig/multisig_test.go
@@ -558,20 +558,20 @@ func TestCancel(t *testing.T) {
 		rt.Verify()
 
 		// anne fails to cancel a transaction - she is not a signer
-		rt.SetCaller(anne, builtin.AccountActorCodeID)
-		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
-		rt.ExpectAbort(exitcode.ErrForbidden, func() {
-			actor.cancel(rt, txnID)
-		})
-		rt.Verify()
+		//for some reason I cannot ExpectAbort again
+		//rt.SetCaller(anne, builtin.AccountActorCodeID)
+		//rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		//rt.ExpectAbort(exitcode.ErrForbidden, func() {
+		//	actor.cancel(rt, txnID)
+		//})
+		//rt.Verify()
 
 		// bob either fails to cancel a transaction - he is not a proposer
 		rt.SetCaller(bob, builtin.AccountActorCodeID)
 		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
-		// for some reason I cannot ExpectAbort again
-		//rt.ExpectAbort(exitcode.ErrForbidden, func() {
-		//	actor.cancel(rt, txnID)
-		//})
+		rt.ExpectAbort(exitcode.ErrForbidden, func() {
+			actor.cancel(rt, txnID)
+		})
 		rt.Verify()
 
 		// Transaction should remain after invalid cancel
--- a/actors/builtin/multisig/multisig_test.go
+++ b/actors/builtin/multisig/multisig_test.go
@@ -539,6 +539,50 @@ func TestCancel(t *testing.T) {
 			Approved: []addr.Address{anne},
 		})
 	})
+
+	t.Run("fail to cancel transaction from any account after removal", func(t *testing.T) {
+		rt := builder.Build(t)
+
+		actor.constructAndVerify(rt, numApprovals, noUnlockDuration, signers...)
+
+		// anne proposes a transaction ID: 0
+		rt.SetCaller(anne, builtin.AccountActorCodeID)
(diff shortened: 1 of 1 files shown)


anorth:
This behaviour seems fine to me, well defined. Could you update the test to remove the word "problem" and instead be a demonstration of what happens if the proposer is removed? And document the behaviour on the remove/swap signer methods too. 

Or propose some alternative behaviour.

anorth:
cc @whyrusleeping in case you'd desire different behaviour

actors/builtin/multisig/multisig_test.go
```diff
@@ -723,6 +767,45 @@ func TestRemoveSigner(t *testing.T) {
...
+			desc: "remove signer and decrease threshold to 0",
+
+			initialSigners:   []addr.Address{anne, bob, chuck},
+			initialApprovals: int64(1),
+
+			removeSigner: chuck,
+			decrease:     true,
+
+			expectSigners:   []addr.Address{anne, bob},
+			expectApprovals: int64(0),
```
Alexey-N-Chernyshov: seems not ok

actors/builtin/multisig/multisig_test.go
```diff
@@ -723,6 +767,45 @@ func TestRemoveSigner(t *testing.T) {
...
+			desc: "remove one more signer and decrease threshold to -1",
+
+			initialSigners:   []addr.Address{anne, bob},
+			initialApprovals: int64(0),
+
+			removeSigner: bob,
+			decrease:     true,
+
+			expectSigners:   []addr.Address{anne},
+			expectApprovals: int64(-1),
```
Alexey-N-Chernyshov: is it correct state?

actors/builtin/multisig/multisig_test.go
```diff
@@ -723,6 +767,45 @@ func TestRemoveSigner(t *testing.T) {
...
+		{
+			desc: "remove one more signer and decrease threshold to -2 with no signers at all",
+
+			initialSigners:   []addr.Address{anne},
+			initialApprovals: int64(-1),
+
+			removeSigner: anne,
+			decrease:     true,
+
+			expectSigners:   []addr.Address(nil),
```
Alexey-N-Chernyshov: pay attention to the actor state

actors/builtin/multisig/multisig_test.go
```diff
@@ -539,6 +539,50 @@ func TestCancel(t *testing.T) {
...
+		})
+		rt.Verify()
+
+		// bob either fails to cancel a transaction - he is not a proposer
+		rt.SetCaller(bob, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		// for some reason I cannot ExpectAbort again
+		//rt.ExpectAbort(exitcode.ErrForbidden, func() {
+		//	actor.cancel(rt, txnID)
+		//})
```
anorth: After anne is removed, I think bob is the new "proposer" (indeed, the only signer) so this call should succeed.

actors/builtin/multisig/multisig_test.go
anorth: Good catch, `decrease=true` should be rejected when the approvals is `1`.

anorth: Or, for consistency with `increase`, `decrease` should be ignored when it would lead to an illegal state. See #114

actors/builtin/multisig/multisig_test.go
```diff
@@ -539,6 +539,50 @@ func TestCancel(t *testing.T) {
...
+		actor.propose(rt, chuck, sendValue, fakeMethod, fakeParams)
+		rt.Verify()
+
+		// remove anne - tx creator
+		rt.SetCaller(receiver, builtin.MultisigActorCodeID)
+		rt.ExpectValidateCallerAddr(receiver)
+		actor.removeSigner(rt, anne, true)
+		rt.Verify()
+
+		// anne fails to cancel a transaction - she is not a signer
```
Alexey-N-Chernyshov: I don't understand why I cannot catch 2 aborts in one test.

actors/builtin/multisig/multisig_test.go
```diff
@@ -539,6 +539,50 @@ func TestCancel(t *testing.T) {
...
+		//for some reason I cannot ExpectAbort again
+		//rt.SetCaller(anne, builtin.AccountActorCodeID)
+		//rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		//rt.ExpectAbort(exitcode.ErrForbidden, func() {
+		//	actor.cancel(rt, txnID)
+		//})
+		//rt.Verify()
+
+		// bob either fails to cancel a transaction - he is not a proposer
```
Alexey-N-Chernyshov: Anne is removed from signer in MultisigActor state, but is not removed from approvers of created transactions. So, now nobody can cancel the tx since proposer is not a signer anymore. Do you think that removed signer should be removed from all transaction approvers?

actors/builtin/multisig/multisig_test.go
Alexey-N-Chernyshov: I uncommented this part, so you can run it.

PR #102, opened 2020-02-04
Add initial tests to StoragePowerActor

zixuanzh:


actors/builtin/storage_power/storage_power_test.go
```diff
@@ -1 +1,128 @@
...
+	"github.com/ipfs/go-cid"
+	"github.com/stretchr/testify/require"
+	"testing"
+
+	addr "github.com/filecoin-project/go-address"
+	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
+	storage_power "github.com/filecoin-project/specs-actors/actors/builtin/storage_power"
+	adt "github.com/filecoin-project/specs-actors/actors/util/adt"
+	mock "github.com/filecoin-project/specs-actors/support/mock"
+	assert "github.com/stretchr/testify/assert"
```
anorth: Please use goimports to sort and group arrays. GoLand has a facility to do this too (in settings)

actors/builtin/storage_power/storage_power_test.go
```diff
@@ -1 +1,128 @@
...
+			SectorSize: abi.SectorSize(int64(32)),
+			Peer: "miner1",
+		}
+
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		// owner1 send CreateMiner to StoragePowerActor
+		rt.SetCaller(owner1, builtin.AccountActorCodeID)
+		rt.SetReceived(big.Zero())
```
anorth: Set a non-zero amount here and verify it gets accounted as pledge collateral  for the miner

actors/builtin/storage_power/storage_power_test.go
```diff
@@ -1 +1,128 @@
...
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+
+		createMinerRet := &storage_power.CreateMinerReturn{
+			IDAddress:  miner1, // miner actor id address
+			RobustAddress: miner1, // should be long miner actor address
+		}
+		rt.ExpectSend(owner1, builtin.MethodSend, createMinerParams, abi.NewTokenAmount(10), &mock.ReturnWrapper{createMinerRet},0)
```
anorth: Thanks, I can probably make this return wrapper more ergonomic later.

actors/builtin/storage_power/storage_power_test.go
```diff
@@ -1 +1,128 @@
...
+		rt.Call(actor.StoragePowerActor.CreateMiner, createMinerParams)
+		rt.Verify()
+
+		var st storage_power.StoragePowerActorState
+		rt.GetState(&st)
+		assert.Equal(t, int64(1), st.MinerCount)
+		assert.Equal(t, abi.NewStoragePower(0), st.TotalNetworkPower)
+		assert.Equal(t, int64(0), st.NumMinersMeetingMinPower)
+		escrowTable := adt.AsMap(rt.Store(), st.ClaimedPower)
+		keys, err := escrowTable.CollectKeys()
```
anorth: assert length of keys == 1. Then don't loop below, just directly do a `Get` for the miner1 address.

actors/builtin/storage_power/storage_power_test.go
```diff
@@ -1 +1,128 @@
...
+		actor.constructAndVerify(rt)
+
+		// owner1 send CreateMiner to StoragePowerActor
+		rt.SetCaller(owner1, builtin.AccountActorCodeID)
+		rt.SetReceived(big.Zero())
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+
+		createMinerRet := &storage_power.CreateMinerReturn{
+			IDAddress:  miner1, // miner actor id address
+			RobustAddress: miner1, // should be long miner actor address
```
anorth: Use addr.Undef for this second one - it should not be used by the power actor at all

actors/builtin/storage_power/storage_power_test.go
```diff
@@ -1 +1,128 @@
...
+
+//
+// Misc. Utility Functions
+//
+
+type key string
+
+func asKey(in string) adt.Keyer {
+	return key(in)
+}
```
anorth: You shouldn't need this key wrapper, the tables are keyed by address.

actors/builtin/storage_power/storage_power_test.go
zixuanzh: thanks 🙏

actors/builtin/storage_power/storage_power_test.go
zixuanzh: setting that to Undef will throw a cbor error

actors/builtin/storage_power/storage_power_test.go
zixuanzh: looks like we still need it tho

PR #103, opened 2020-02-04
Move most constants and functions from indices into their respective actors

anorth:


actors/runtime/indices/indices.go
```diff
@@ -12,66 +10,22 @@ import (
...
-	NetworkKPI() big.Int
-	TotalNetworkSectorWeight() abi.SectorWeight
-	TotalPledgeCollateral() abi.TokenAmount
-	TotalNetworkEffectivePower() abi.StoragePower // power above minimum miner size
-	TotalNetworkPower() abi.StoragePower          // total network power irrespective of meeting minimum miner size
-
-	TotalMinedFIL() abi.TokenAmount
-	TotalUnminedFIL() abi.TokenAmount
-	TotalBurnedFIL() abi.TokenAmount
-	LastEpochReward() abi.TokenAmount
```
zixuanzh: Blocking: lets keep these accessors for now, at least NetworkKPI, NetworkIndicator, and NetworkTime (changes landed in a separate PR). TotalMined and TotalSupply can be used to compute block reward iteratively. For the other totals, is the idea to compute them in other actors?

actors/runtime/indices/indices.go
```diff
@@ -12,66 +10,22 @@ import (
...
 		networkTime big.Int,
 	) abi.TokenAmount
-	StorageMining_PreCommitDeposit(
-		sectorSize abi.SectorSize,
-		expirationEpoch abi.ChainEpoch,
-	) abi.TokenAmount
-	NetworkTransactionFee(
-		toActorCodeID cid.Cid,
-		methodNum abi.MethodNum,
-	) abi.TokenAmount
```
zixuanzh: Blocking: we will still need this as a mechanism

actors/builtin/storage_power/policy.go
```diff
@@ -1,66 +1,89 @@
...
-	for i := int64(0); i < int64(elapsedEpoch); i++ {
-		slasherShareDenomMultiplier = big.Mul(slasherShareDenomMultiplier, indices.ConsensusFault_SlasherShareGrowthRateDenom())
-	}
-
-	num := big.Mul(big.Mul(slasherShareNumMultiplier, indices.ConsensusFault_SlasherInitialShareNum()), collateralToSlash)
-	denom := big.Mul(slasherShareDenomMultiplier, indices.ConsensusFault_SlasherInitialShareDenom())
-	return big.Min(big.Div(num, denom), collateralToSlash)
+	num := big.Mul(big.Mul(slasherShareNumerator, consensusFaultReporterInitialShare.numerator), collateral)
+	denom := big.Mul(slasherShareDenominator, consensusFaultReporterInitialShare.denominator)
+	return big.Min(big.Div(num, denom), collateral)
```
jzimmerman: I'm not quite sure an exponential is the right choice for the reporter's share - either this will be vanishingly small at some points, or else will saturate to 1 at some points. Could we leave the entire function parametric?

actors/runtime/indices/indices.go
```diff
@@ -12,66 +10,22 @@ import (
...
-	StorageDeal_DurationBounds(
-		pieceSize abi.PieceSize,
-		startEpoch abi.ChainEpoch,
-	) (minDuration abi.ChainEpoch, maxDuration abi.ChainEpoch)
-	StorageDeal_StoragePricePerEpochBounds(
-		pieceSize abi.PieceSize,
-		startEpoch abi.ChainEpoch,
-		endEpoch abi.ChainEpoch,
-	) (minPrice abi.TokenAmount, maxPrice abi.TokenAmount)
-	StorageDeal_ProviderCollateralBounds(
```
jzimmerman: Blocking: This function (and others) was previously a method on Indices, which enabled it to be time/KPI-dependent. Making them non-time/KPI-dependent is a cryptoecon policy change, which we would need to evaluate separately.

actors/runtime/indices/indices.go
anorth: I was able to remove these because there are *zero callers* of them. When cryptoecon research provides a concrete function that requires something like this, I'll add it back in whatever way makes sense for that value when I know what it is.

actors/runtime/indices/indices.go
anorth: It doesn't happen in the actors, it happens in the interpreter.

actors/builtin/storage_power/policy.go
anorth: No, a concrete function is required. This PR is not about changing it. When there's a new mechanism proposal, we can implement it.

actors/runtime/indices/indices.go
anorth: When there is a concrete function that depends on some external quantity, we can restore it appropriately.

actors/builtin/storage_power/policy.go
Kubuxu: As a note, in lotus we implemented this as a lookup table.

PR #105, opened 2020-02-05
Clean up payment channel, including checking creator signing ability.

anorth:
Also generated CBOR serialization code.


actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -16,33 +17,34 @@ import (
...
-	// TODO anorth: ensure parties are valid
-	// rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
-	// toID, err := addr.IDFromAddress(params.To)
-	// if err != nil {
-	// 	rt.Abort(exitcode.ErrIllegalArgument, "receiver must be ID address")
-	// }
+func (pca *PaymentChannelActor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.EmptyValue {
+	rt.ValidateImmediateCallerType(builtin.AccountActorCodeID)
+
+	// Check that the channel creator is capable of signing vouchers.
```
anorth: This checks that the channel creator is an account address with associated crypto key (which all accounts have).

The destination address is *not* checked in the same way, which matches my understanding and most of the comments existing in this code. But @whyrusleeping this is contradicted by a comment and code in `UpdateChannelState` that "both parties must sign voucher" and then allowing either one party to be the signer. This didn't make sense to me, but if that code is correct,this construction should be updated to check that the destination address is also an account with a signing key.

Note that multisig actors cannot produce signatures and can't be a sender here. That restriction will extend to destinations if I make the changes I describe.

sternhenri: Sorry the signing comment there is a bit ambiguous. Specifically, one party must explicitly sign the voucher using the `Signature` field in the `signedVoucher`. The other party "implicitly" signs it by submitting it on chain, so if you look at the code there, we just ensure that the party who signed the voucher is not the caller...

Both accounts should be able to sign the voucher yes (ie be accounts with signing keys). Thereafter it seems odd to me that multisig can't produce sigs... but so be it.

actors/builtin/payment_channel/payment_channel_actor.go
```diff
@@ -16,33 +17,34 @@ import (
...
+	}
+	if signingAddress.Protocol() != addr.SECP256K1 && signingAddress.Protocol() != addr.BLS {
+		rt.Abort(exitcode.ErrIllegalArgument, "creator's signing address must use SECP or BLS protocol, was %v", signingAddress.Protocol())
+	}
 
-	// if !toID.PubKeyAddress.Protocol() == addr.SECP256K1 && !toID.PubKeyAddress.Protocol() == addr.BLS {
-	// 	rt.Abort(exitcode.ErrIllegalArgument, "address must use BLS or SECP protocol, got %v", toID.PubKeyAddress.Protocol())
-	// }
+	// Check that target is a canonical ID address.
+	// This is required for consistent caller validation.
```
sternhenri: must also check for from then since either party can call methods on this (this may implictly be done above already)

actors/builtin/payment_channel/payment_channel_actor.go
anorth: I talked with @whyrusleeping, agreed that both parties must be account actors (multisigs can't be parties here).

I'm also going to simplify the check here just to check for Account Actor code ID, which implies the necessary signing key.

actors/builtin/payment_channel/payment_channel_actor.go
anorth: Yes, implicit, `rt.ImmediateCaller()` is always an ID address.

PR #108, opened 2020-02-05
PoSt updates

sternhenri:
- removed unnecessary surprise challenge code. Only a constant number are challenged, and checked for.
- change electionPoSt eligibility to eligible unless surprise is missed (rather than unless is challenged)
- fix a couple mistakes
- remove no_surprise_challenge_period: fallback always occurs regardless of ePoSt

left TODO:
- power only after first post
- deal with ChallengedSectors stuff

### DIFF
--- a/actors/builtin/storage_miner/storage_miner_actor_state.go
+++ b/actors/builtin/storage_miner/storage_miner_actor_state.go
@@ -59,9 +59,6 @@ type MinerPoStState struct {
 	//  `epochUndefined` if not currently challeneged.
 	SurpriseChallengeEpoch abi.ChainEpoch
 
-	// Not empty iff the miner is challenged.
-	ChallengedSectors []abi.SectorNumber
-
 	// Number of surprised post challenges that have been failed since last successful PoSt.
 	// Indicates that the claimed storage power may not actually be proven. Recovery can proceed by
 	// submitting a correct response to a subsequent SurprisePoSt challenge, up until
@@ -261,7 +258,7 @@ func (mps *MinerPoStState) isChallenged() bool {
 }
 
 func (mps *MinerPoStState) isPoStOk() bool {
-	return !mps.isChallenged() && !mps.hasFailedPost()
+	return !mps.hasFailedPost()
 }
 
 func (mps *MinerPoStState) hasFailedPost() bool {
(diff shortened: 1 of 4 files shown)


actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -35,9 +35,6 @@ type MinerPoStState struct {
 	//  `epochUndefined` if not currently challeneged.
 	SurpriseChallengeEpoch abi.ChainEpoch
 
-	// Not empty iff the miner is challenged.
-	ChallengedSectors []abi.SectorNumber
```
zixuanzh: Blocking: unless I am missing something, dont think we should remove this even though a constant number is challenged. This is a snapshot of the sectors that are challenged. Let's discuss in sync

actors/runtime/indices/indices.go
```diff
@@ -255,20 +255,20 @@ func StorageMining_SurprisePoStProvingPeriod() abi.ChainEpoch {
...
 	const MAX_CONSECUTIVE_FAULTS = 3 // placeholder
 	return MAX_CONSECUTIVE_FAULTS
 }
 
+// number of sectors to be proven as part of the surprise PoSt
+func StoragePower_NumSurprisePoStSectors() int64 {
+	//PARAM_FINISH
+	const NUM_SURPRISE_SECTORS = 200 //placeholder
+	return NUM_SURPRISE_SECTORS
+}
```
anorth: This will move to storage_power/policy in #103

anorth:
cc @whyrusleeping

actors/builtin/storage_miner/storage_miner_actor_state.go
sternhenri: I don't understand what this does and assumed that this was based on a misunderstanding of how the proofs lib works. All sectors are submitted to the proofs lib, which randomly samples a wanted number cc @porcuquine. What is this meant to do @zixuanzh ?

sternhenri: synced w ZX. This is meant to be a cached version of the proving set used to know what got challenged (in case new sectors are added). Must be worked with concurrently with whole power accounting concept. But was wrong in actor code right now (or incomplete in any case). Handled in follow up PR.

anorth: That makes sense, should be simple as a CID.

PR #110, opened 2020-02-05
Fixup compute data commitment (#109) and cleaner cid types

### LINKED ISSUE #109: storage market actor should not send piece infos across method call boundaries
I had this discussion with the specs team a while ago, but it seems it was missed. 'GetPieceInfosForDealIDs' should be changed to 'ComputeDataCommitment' (https://github.com/filecoin-project/lotus/blob/master/chain/actors/actor_storagemarket.go#L614) Otherwise we end up having to serialize large numbers of piece infos across method call boundaries, leading to unnecessarily expensive actor calls. 

cc @magik6k

whyrusleeping:
Should resolve #109, also moves us closer to integration success.


actors/abi/sector.go
```diff
@@ -1,9 +1,8 @@
 package abi
 
 import (
-	cid "github.com/ipfs/go-cid"
-
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	cid "github.com/ipfs/go-cid"
```
anorth: Nit: Please follow the import ordering/grouping of `goimports -local` (or at least don't regress it).

actors/abi/sector.go
```diff
@@ -35,9 +34,6 @@ func NewStoragePower(n int64) StoragePower {
 	return StoragePower(big.NewInt(n))
 }
 
-type UnsealedSectorCID cid.Cid // CommD
-type SealedSectorCID cid.Cid   // CommR
-
```
anorth: Thanks, I have the same thing in an in-flight PR

actors/abi/sector.go
```diff
@@ -94,10 +90,10 @@ type PartialTicket []byte // 32 bytes
...
 type PoStVerifyInfo struct {
 	Randomness      PoStRandomness
-	CommR           SealedSectorCID
-	Candidates      []PoStCandidate              // From OnChain*PoStVerifyInfo
-	Proofs          []PoStProof                  // From OnChain*PoStVerifyInfo
-	EligibleSectors map[SectorID]SealedSectorCID // TODO: HAMT?
+	CommR           cid.Cid
+	Candidates      []PoStCandidate      // From OnChain*PoStVerifyInfo
+	Proofs          []PoStProof          // From OnChain*PoStVerifyInfo
+	EligibleSectors map[SectorID]cid.Cid // TODO: HAMT?
```
anorth: While we're here, what do you propose for this data structure. HAMT makes no sense here. How about a list of tuples?

actors/builtin/storage_market/storage_market_actor.go
```diff
@@ -228,31 +231,43 @@ func (a StorageMarketActor) VerifyDealsOnSectorProveCommit(rt Runtime, params *V
...
 }
 
-type GetPieceInfosForDealIDsParams struct {
-	DealIDs []abi.DealID
+type ComputeDataCommitmentParams struct {
+	DealIDs    []abi.DealID
+	SectorSize abi.SectorSize
+}
+
+// cbor gen and delete me
```
anorth: Don't forget!

whyrusleeping:
@anorth thank you for fixing up those things for me!

PR #111, opened 2020-02-05
storagemarket: Multiset for DealIDsByParty

magik6k:


actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	cid "github.com/ipfs/go-cid"
+	errors "github.com/pkg/errors"
+	cbg "github.com/whyrusleeping/cbor-gen"
+
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+)
+
+type Multiset struct {
```
anorth: This isn't quite a multiset (which would be a map of keys to ints), but a Set-Multimap: a multimap with set semantics for the values. And specifically with integer values. This contrasts with the List-Multimap that the existing multimap uses. Please rename to SetMultimap (and feel free to rename the other to ListMultimap). (cf [Guava](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimap.html))

For its use as a deal-by-client index, I think a HAMT may be more appropriate for the  value type: deal ids for a party will be sparse and large magnitude. You can use the existing `Set` ADT.

actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,123 @@
...
+	return &Multiset{m, s}, err
+}
+
+// Returns the root cid of the underlying HAMT.
+func (mm *Multiset) Root() cid.Cid {
+	return mm.mp.Root()
+}
+
+// Adds a value for a key.
+func (mm *Multiset) Add(key adt.Keyer, idx uint64) error {
```
anorth: I think it would be better to name this `Set` or `Put` so the set-semantics are clear. It does a different thing to the ListMultimap (which would retain duplicates and insertion order).

actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to delete multiset key %v root %v", key, mm.mp.Root())
+	}
+	return nil
+}
+
+// Iterates all entries for a key in the order they were inserted, deserializing each value in turn into `out` and then
+// calling a function.
+// Iteration halts if the function returns an error.
+// If the output parameter is nil, deserialization is skipped.
```
anorth: "in sorted order" rather than insertion, and delete the "deserializing ...." clause

actors/builtin/storage_market/storage_market_actor_state.go
```diff
@@ -43,7 +40,7 @@ type StorageMarketActorState struct {
 	NextID abi.DealID
 
 	// Metadata cached for efficient iteration over deals.
-	DealIDsByParty map[string]PartyDeals // TODO: figure out a way to drop this
+	DealIDsByParty cid.Cid // hamt[addr]Set
```
anorth:
```suggestion
	DealIDsByParty cid.Cid // SetMultimap, HAMT[addr]Set
```

actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/binary"
+
+	cid "github.com/ipfs/go-cid"
+	errors "github.com/pkg/errors"
+	cbg "github.com/whyrusleeping/cbor-gen"
+
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+)
+
+type SetMultimap struct {
```
anorth: Rename the file too

actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,127 @@
...
+	set, found, err := mm.get(key)
+	if err != nil {
+		return err
+	}
+	if !found {
+		return adt.ErrNotFound{
+			Root: set.Root(),
+			Key:  []interface{}{key, idx},
+		}
+	}
```
anorth: Is this worth erroring? We could just change the semantics to ensure it's not present afterwards? Else return a boolean?

actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Append to the array.
+	if err = set.Delete(adt.IntKey(idx)); err != nil {
+		return errors.Wrapf(err, "failed to remove multiset key %v", key)
+	}
+
+	// Store the new array root under key.
+	newArrayRoot := cbg.CborCid(set.Root())
```
anorth: array -> set

actors/builtin/storage_market/multiset.go
```diff
@@ -0,0 +1,127 @@
...
+	set, found, err := mm.get(key)
+	if err != nil {
+		return err
+	}
+	if found {
+		return set.ForEach(func(k string) error {
+			i, n := binary.Varint([]byte(k))
+			if n != len(k) {
+				return errors.New("failed to decode varint key")
+			}
```
anorth: Move this next to or into IntKey?

actors/util/adt/set.go
```diff
@@ -45,6 +45,11 @@ func (h *Set) Delete(k Keyer) error {
 	return h.m.Delete(k)
 }
 
+// Collects all the keys from the set into a slice of strings.
```
anorth: Mismatched comment.

PR #112, opened 2020-02-05
Remove power table and nominal power states

zixuanzh:


zixuanzh:
fixing cbor is the remaining todo and will do a separate PR to go through the flow of logic in SPA

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -77,8 +73,16 @@ func (st *StoragePowerActorState) minerNominalPowerMeetsConsensusMinimum(s adt.S
...
 	var minerSizes []abi.StoragePower
 	var pwr abi.StoragePower
-	if err := adt.AsMap(s, st.PowerTable).ForEach(&pwr, func(k string) error {
-		minerSizes = append(minerSizes, pwr.Copy())
+	if err := adt.AsMap(s, st.ClaimedPower).ForEach(&pwr, func(k string) error {
+		maddr, err := addr.NewFromBytes([]byte(k))
+		if err != nil {
+			return err
+		}
+		nominalPower, err := st.computeNominalPower(s, maddr)
```
anorth: Don't pass the address here only to redundantly look up the claimed power. `pwr` is the value just iterated: pass it as a parameter.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -92,15 +96,22 @@ func (st *StoragePowerActorState) minerNominalPowerMeetsConsensusMinimum(s adt.S
...
 	var allMiners []addr.Address
-	if err := adt.AsMap(s, st.PowerTable).ForEach(nil, func(k string) error {
+	var pwr abi.StoragePower
+	if err := adt.AsMap(s, st.ClaimedPower).ForEach(&pwr, func(k string) error {
 		maddr, err := addr.NewFromBytes([]byte(k))
 		if err != nil {
 			return err
 		}
-		allMiners = append(allMiners, maddr)
+		nominalPower, err := st.computeNominalPower(s, maddr)
```
anorth: Ditto pass `pwr` as claimed power.

actors/builtin/storage_power/storage_power_actor_state.go
```diff
@@ -193,16 +199,16 @@ func (st *StoragePowerActorState) deductClaimedPowerForSector(s adt.Store, miner
 	if err = st.setClaimedPower(s, minerAddr, big.Sub(currentPower, sectorPower)); err != nil {
 		return err
 	}
-	return st.updatePowerEntriesFromClaimed(s, minerAddr)
+	return nil
 }
 
-func (st *StoragePowerActorState) updatePowerEntriesFromClaimed(s adt.Store, minerAddr addr.Address) error {
+func (st *StoragePowerActorState) computeNominalPower(s adt.Store, minerAddr addr.Address) (abi.StoragePower, error) {
```
anorth: Aligned with the comments above, change the address parameter to be claimed power (and any other callers must look it up first).

actors/builtin/storage_power/storage_power_actor_state.go
zixuanzh: Ack, computeNominalPower still needs to take in a minerAddr to check if the miner is at fault. I have made the change and will land this PR to unblock other SPA related work. Let me know if we should do this differently.

PR #118, opened 2020-02-05
fix(abi/big): match CBOR Marshalling to impls

hannahhoward:
# Goals

Many data structs in Lotus/go-fil-markets have big int values that are treated as essentially optional -- they can be set or not. However, currently, the zero value for big.Int (i.e. nothing set, internal Int data member == nil) throws an error when marshalled to CBOR. This will cause several errors in existing lotus & go-fil-markets code that don't show until run (or tested). It also might require extensive code changes.

# Implementation

This PR changes the MarshalJSON & MarshalCBOR methods of big.Int to match big.Int from Lotus & tokenamount.Tokenamount from go-fil-markets, which both do not error when marshalling CBOR for an empty big.Int.

# For DIscussion

I am not sure this is the right solution -- rather it's a strawman to start a converation. Because the difference blocks, or at least makes difficult, moving much of the spec actor types into go-fil-markets.

I believe the implementations are treating big.Int{} as essentially an optional type -- i.e. value undefined or value empty. I realize the cbor encode/decode is not 1-to-1 I think, and maybe that is worth looking at. Or maybe defining an explicit big.OptionalInt? It's possible the lotus folks have already come up with a solution and are trying something different for their code.

cc: @magik6k @whyrusleeping @anorth

### DIFF
--- a/actors/abi/big/int.go
+++ b/actors/abi/big/int.go
@@ -247,3 +247,7 @@ func (bi *Int) UnmarshalCBOR(br io.Reader) error {
 func (bi *Int) IsZero() bool {
 	return bi.Int.Sign() == 0
 }
+
+func (bi *Int) Nil() bool {
+	return bi.Int == nil
+}
--- a/actors/abi/big/int.go
+++ b/actors/abi/big/int.go
@@ -127,6 +127,10 @@ func (bi Int) Equals(o Int) bool {
 }
 
 func (bi *Int) MarshalJSON() ([]byte, error) {
+	if bi.Int == nil {
+		zero := NewInt(0)
+		return json.Marshal(zero)
+	}
 	return json.Marshal(bi.String())
 }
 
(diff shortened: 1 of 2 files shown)


anorth:
I consider an uninitialized big.Int to be a bug and think we should always set an explicit `0` value. We dropped the nil handling consciously to surface such bugs.

I am willing to be overruled by @whyrusleeping if he has strong feelings about this (we expect Lotus to drop your copy of the Int type to import this common one).

I feel more strongly that the Int, which is used for many different types, should not have atto-fil assumptions built into the stringification or other methods. If explicit attofil behaviour is wanted, I think implementations should locally extend the type, without requiring all other implementations to have the same semantic mess.

I would also be ok with an AttoFil type extending the Int to be in this repo.

hannahhoward:
@anorth I support your desire to be principled! That said, we are looking at some complicated changes.

Here's my analysis:
I think what we're really talking about here is an optional type that can be unset and should be checked. 

In my ideal world, I would have: BigInt and OptionalBigInt, and OptionalBigInt would have the ability to serialize w/o error, even when BigInt.Int was not set (because it would write itself as a Nil type, while the other would be written BigInt) but to get to the value, you'd have to verify it was present first. But that's me, person who loves very strict types.

And, in this context, I believe the best initial approach to getting everyone using the same thing is to just use the behavior that already exists. Otherwise, we're looking at a good while to get all this inline.

PR #121, opened 2020-02-06
storagemarket: Separate proposal and deal arrays

magik6k:


actors/builtin/storage_market/storage_market_actor_state.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
-const epochUndefined = abi.ChainEpoch(-1)
+const epochUndefined = abi.ChainEpoch(0)
```
anorth: Please don't do this here. I see https://github.com/filecoin-project/specs/issues/874 but it needs broader acceptance. The genesis epoch isn't something we get to define here. The genesis may well have deals and this will cause us problems.

actors/builtin/storage_market/storage_market_actor_state.go
```diff
@@ -137,34 +138,44 @@ func (st *StorageMarketActorState) updatePendingDealState(rt Runtime, dealID abi
...
-	deals := AsDealArray(adt.AsStore(rt), st.Deals)
-	ocd, err := deals.Get(dealID)
-	if err != nil {
-		rt.Abort(exitcode.ErrIllegalState, "failed to get deal: %v", err)
-	}
-	ocd.LastUpdatedEpoch = epoch
-	if err := deals.Set(dealID, ocd); err != nil {
+	meta.LastUpdatedEpoch = epoch
+
+	deals := AsDealMetaArray(adt.AsStore(rt), st.Meta)
```
anorth:
```suggestion
	metas := AsDealMetaArray(adt.AsStore(rt), st.Meta)
```

actors/builtin/storage_market/storage_market_actor.go
```diff
@@ -166,10 +166,10 @@ func (a StorageMarketActor) PublishStorageDeals(rt Runtime, params *PublishStora
 			st.lockBalanceOrAbort(rt, deal.Provider, deal.ProviderBalanceRequirement())
 
 			id := st.generateStorageDealID()
-			onchainDeal := &OnChainDeal{
-				ID:               id,
-				Proposal:         deal,
+			onchainDeal := &DealMeta{
```
anorth:
```suggestion
			meta := &DealMeta{
```

actors/builtin/storage_market/storage_market_actor.go
```diff
@@ -166,10 +166,10 @@ func (a StorageMarketActor) PublishStorageDeals(rt Runtime, params *PublishStora
...
 				SectorStartEpoch: epochUndefined,
+				LastUpdatedEpoch: epochUndefined,
+				SlashEpoch:       epochUndefined,
 			}
 
 			err := deals.Set(id, onchainDeal)
```
anorth: Blocking: this is setting the meta in the wrong array. At 189 should update `st.Meta` instead.

This would all be less error-prone if the ADT manipulations were inside methods on the state, instead of directly here.

actors/builtin/storage_market/storage_market_actor.go
```diff
@@ -216,28 +216,35 @@ func (a StorageMarketActor) VerifyDealsOnSectorProveCommit(rt Runtime, params *V
 	rt.State().Transaction(&st, func() interface{} {
 		// if there are no dealIDs, it is a CommittedCapacity sector
 		// and the totalWeight should be zero
+		deals := AsDealMetaArray(adt.AsStore(rt), st.Meta)
```
anorth:
```suggestion
		metas := AsDealMetaArray(adt.AsStore(rt), st.Meta)
```

actors/builtin/storage_market/storage_market_actor_state.go
```diff
@@ -26,7 +26,8 @@ type PartyDeals struct {
 }
 
 type StorageMarketActorState struct {
-	Deals cid.Cid // AMT[]OnChainDeal
+	Deals cid.Cid // AMT[DealID]DealProposal
```
anorth: Since renaming the value type to DealProposal, call this `Proposals`, which will also make a bunch of the `proposals := AsDealArray()` variables above make more sense.

Alternatively, keep this as `Deals` but rename those variables to match.

actors/builtin/storage_market/storage_market_actor_state.go
```diff
@@ -26,7 +26,8 @@ type PartyDeals struct {
...
+	Meta  cid.Cid // AMT[DealID]DealMeta
```
anorth: Metas?

magik6k: Sounds a bit weird, 'states' sounds better I guess?

PR #125, opened 2020-02-06
Partial cleanup of the miner actor.

anorth:
Simpler control flow, merged some repeated state reads/transactions into one.


anorth:
cc @zixuanzh

PR #127, opened 2020-02-07
Cleanup some names

magik6k:
(There shouldn't by any logic changes in this PR)


actors/builtin/codes.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 // The built-in actor code IDs
-var SystemActorCodeID cid.Cid
+var SystemActorCodeID cid.Cid // todo: drop id?
```
anorth: Let's leave it for now. It's definitely not the code itself, but an identifier of the code.

actors/builtin/miner/miner_state.go
```diff
@@ -7,21 +7,21 @@ import (
 	errors "github.com/pkg/errors"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
-	storage_power "github.com/filecoin-project/specs-actors/actors/builtin/storage_power"
+	storage_power "github.com/filecoin-project/specs-actors/actors/builtin/power"
```
anorth: change the alias too, here and in other files.

PR #130, opened 2020-02-07
Implement up-front sector pledge collateral calculation and tracking.

anorth:
Pledge is computed when a sector is proven and fixed for the sector's lifetime. The computation depends on the total network power at the time, so must be stored with the sector metadata, and a per-miner total maintained by the power actor. Removing the placeholder for dynamic pledge collateral allows a complete implementation.


actors/builtin/miner/miner_actor.go
```diff
@@ -541,8 +539,9 @@ func (a *Actor) OnDeferredCronEvent(rt Runtime, params *OnDeferredCronEventParam
 	}
 
 	if payload.EventType == CronEventType_Miner_TempFault {
+		// should roll the huge number of state transactions in this loop into ~1
```
whyrusleeping:
```suggestion
		// TODO: should roll the huge number of state transactions in this loop into ~1
```

whyrusleeping: (makes this easier to grep for)

actors/builtin/power/power_state.go
```diff
@@ -71,13 +87,13 @@ func (st *State) minerNominalPowerMeetsConsensusMinimum(s adt.Store, minerPower
 	}
 
 	var minerSizes []abi.StoragePower
-	var pwr abi.StoragePower
-	if err := adt.AsMap(s, st.ClaimedPower).ForEach(&pwr, func(k string) error {
+	var claimed Claim
```
whyrusleeping: this method isnt used at all, and if it were, i would be throwing red flags at it from 50 feet away. Iterating the miner set during any onchain operation is asking for pain.

Can we just delete this one now?

anorth: As discussed, the conditions higher up in this function mean this iteration happens only very early on in the network life, when no miner yet meets the size threshold.

This method *should* be called somewhere, as part of validating a miner is eligible for EPost. Maybe that call should come from off-chain.

actors/builtin/miner/miner_actor.go
anorth: I've been reserving TODO temporarily for things that we absoluty must do, but this one is fixed by #135 anyway.

PR #132, opened 2020-02-07
Tests for payment channel actor: constructor, adding lane/voucher

shannonwells:
### Add tests 
. . . for construction of payment channel actor and a single test (more to come) for adding a lane.  Includes supporting functions:
* add Syscalls support to mock runtime
* add builder config functions to specify receiver type and add receiver to actorCodeCIDs
* ExpectAbortWithMsg function to check that something is failing for the right reason

### Fix other called-out issues with `paych.Actor`
* only `InitActor` may call the Constructor.
* Add `From` to the `ConstructorParams`. This is the sender of FIL, aka payer.
* Run same checks on `From` as are run on `To` param, the receiver of payment, aka payee

### Snake-killing
* fix a typo
* fix a bug in adt/store.Key


actors/util/adt/store.go
```diff
@@ -61,7 +61,7 @@ func (k AddrKey) Key() string {
 type IntKey int64
 
 func (k IntKey) Key() string {
-	var buf []byte
+	buf := make([]byte, 8)
 	n := binary.PutVarint(buf, int64(k))
```
shannonwells: can't PutVarint to a nil buf

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,179 @@
...
+func TestPaymentChannelActor_Constructor(t *testing.T) {
+	ctx := context.Background()
+	actor := pcActorHarness{Actor{}, t}
+
+	pcaAddr := tutil.NewIDAddr(t, 100)
+	callerAddr := tutil.NewIDAddr(t, 101)
+
+	t.Run("can create a payment channel actor", func(t *testing.T) {
+		builder := mock.NewBuilder(ctx, pcaAddr).
+			WithCaller(callerAddr, builtin.AccountActorCodeID).
```
anorth: This immediately reveals problems with the actor. The actor's constructor must be changed to only be callable by the InitActor (like all other actors). I think that means that the constructor params must be expanded to explicitly include the `From` address, which will no longer be the immediate caller.

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,179 @@
...
+			WithReceiverType(builtin.AccountActorCodeID)
+
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt, pcaAddr, callerAddr)
+	})
+
+	t.Run("fails if caller is not account actor", func(t *testing.T) {
```
anorth: The `ExpectValidateCallerType` is sufficient that we don't need to test this. This test is really testing that the runtime correctly rejects the ValidateCallerType with the wrong caller, but that's not actually testing the actor itself.

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,179 @@
...
+		})
+	})
+
+	t.Run("fails if target is not account actor", func(t *testing.T) {
+		builder := mock.NewBuilder(ctx, pcaAddr).
+			WithCaller(callerAddr, builtin.AccountActorCodeID).
+			WithReceiverType(builtin.CronActorCodeID)
+		rt := builder.Build(t)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID)
+		expMsg := "target actor [0 - 64] must be an account (bafkqadlgnfwc6mjpmfrwg33vnz2a), was bafkqactgnfwc6mjpmnzg63q"
```
anorth: Please don't add assertions about abort messages. I want maximum freedom for us to improve these strictly-diagnostic messages later. It's sufficient to check the exit code, even if that doesn't currently uniquely identify the cause.

actors/builtin/paych/paych_test.go
anorth: You can then factor this builder setup out of each t.Run, because it will always be the same.

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt, pcaAddr, callerAddr)
+	})
+
+	t.Run("fails if caller is not account actor", func(t *testing.T) {
+		builder := mock.NewBuilder(ctx, pcaAddr).WithCaller(callerAddr, builtin.CronActorCodeID)
+		rt := builder.Build(t)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID)
+		rt.ExpectAbort(18, func() {
```
frrist: In the `ExpectAbort` calls here and elsewhere use an exitcode from here instead: https://github.com/filecoin-project/specs-actors/blob/master/actors/runtime/exitcode/common.go

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,179 @@
...
+		builder := mock.NewBuilder(ctx, pcaAddr).
+			WithBalance(balance, received).
+			WithEpoch(abi.ChainEpoch(2)).
+			WithCaller(callerAddr, builtin.AccountActorCodeID).
+			WithReceiverType(builtin.AccountActorCodeID).
+			WithSysCalls(&syscalls)
+
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt, pcaAddr, callerAddr)
+		rt.ExpectValidateCallerAddr(callerAddr, pcaAddr)
```
anorth: Nit: move the expectations as close to the matching `Call` as possible. This location makes it seem related to the previous construction call.

support/mock/builder.go
```diff
@@ -80,3 +81,13 @@ func (b *RuntimeBuilder) WithBalance(balance, received abi.TokenAmount) *Runtime
 	b.rt.valueReceived = received
 	return b
 }
+
+func (b *RuntimeBuilder) WithReceiverType(code cid.Cid) *RuntimeBuilder {
```
anorth: This doesn't make sense to me, the receiving actor should never be inspecting its own type. But if you make a `WithActorType(addr, cid)` it might make more sense to me. And maybe this will drop out after the constructor is fixed.

support/mock/mockrt.go
```diff
@@ -69,7 +71,7 @@ var cidBuilder = cid.V1Builder{
 	MhLength: 0, // default
 }
 
-///// Implementation of the runtime API /////
+// /// Implementation of the runtime API /////
```
anorth: Unnecessary reformatting.

support/mock/mockrt.go
```diff
@@ -509,6 +511,29 @@ func (rt *Runtime) ExpectAbort(expected exitcode.ExitCode, f func()) {
 	f()
 }
 
+func (rt *Runtime) ExpectAbortWithMsg(expCode exitcode.ExitCode, expMsg string, f func()) {
```
anorth: Please remove

support/mock/builder.go
```diff
@@ -80,3 +81,13 @@ func (b *RuntimeBuilder) WithBalance(balance, received abi.TokenAmount) *Runtime
...
+	b.rt.actorCodeCIDs[b.rt.receiver] = code
+	return b
+}
+
+func (b *RuntimeBuilder) WithSysCalls(calls runtime.Syscalls) *RuntimeBuilder {
+	b.rt.syscalls = calls
+	return b
```
anorth: Rather than this, I'd rather add `WithVerifiesSig` here, or `SetVerifiesSig` on the mock that update individual syscall stubs. The mock syscalls object is great, but hide it from the user.

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	. "github.com/filecoin-project/specs-actors/actors/builtin/paych"
+	"github.com/filecoin-project/specs-actors/actors/crypto"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+	"github.com/filecoin-project/specs-actors/support/mock"
+	tutil "github.com/filecoin-project/specs-actors/support/testing"
+)
+
+func TestPaymentChannelActor_Constructor(t *testing.T) {
```
frrist: commentary: this would make a good table driven test ([example](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/account/account_test.go#L30))

shannonwells: yes... this form hasn't quite become habitual, but I have been trying to use it more.

support/mock/builder.go
shannonwells: @anorth  The thing I ended up doing (the simpler thing) was to make a syscaller struct but it's not terribly elegant and there is duplication.  I could do some reflection magic and reduce some duplication.  Let me know what your preference is, or if you have another suggestion.

actors/builtin/paych/paych_actor.go
```diff
@@ -34,33 +35,49 @@ func (a Actor) Exports() []interface{} {
...
+	}
+
+	rt.State().Construct(func() vmr.CBORMarshaler {
+			return ConstructState(params.From, params.To)
+	})
+	return &adt.EmptyValue{}
+}
+
+// validateActor requires an actor to be an account actor and to have a canonical ID address.
+// It checks that the embedded address is associated with an appropriate key.
```
anorth:
```suggestion
// The account actor constructor checks that the embedded address is associated with an appropriate key.
```

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,200 @@
...
+		callerCode cid.Cid
+		newActorCode cid.Cid
+		payerCode cid.Cid
+		expExitCode exitcode.ExitCode
+
+	} {
+		{ "fails if target (to) is not account actor",
+			newPaychAddr,
+			builtin.InitActorCodeID,
+			builtin.CronActorCodeID,
```
anorth: Nit: the multisig actor would be a more enlightening candidate here.

actors/builtin/paych/paych_test.go
```diff
@@ -0,0 +1,200 @@
...
+			Type: crypto.SigTypeBLS,
+			Data: []byte("doesn't matter"),
+		}
+		tl := abi.ChainEpoch(1)
+		sv := SignedVoucher{
+			TimeLock:  tl,
+			Lane:      lane,
+			Nonce:     nonce,
+			Amount:    amt,
+			Signature: sig,
```
anorth: Nit: I suggest inlining some of these vars/fields for brevity since they can still be referenced by good names like `sv.Amount` further down.

support/mock/builder.go
```diff
@@ -5,9 +5,9 @@ import (
 	"testing"
 
 	addr "github.com/filecoin-project/go-address"
-	cid "github.com/ipfs/go-cid"
+	"github.com/ipfs/go-cid"
```
anorth: Please retain these

support/mock/mock_syscaller.go
```diff
@@ -0,0 +1,54 @@
...
+
+type syscaller struct {
+	Verifier VerifyFunc
+	Hasher   HasherFunc
+}
+
+// Interface methods
+func (s *syscaller) VerifySignature(sig crypto.Signature, signer addr.Address, plaintext []byte) bool {
+	if s.Verifier == nil {
+		panic("set me")
```
anorth: I like the general way you've set this up, but please replace these panics with `s.t.Fatalf("no verifier set")` etc.

support/mock/mockrt.go
```diff
@@ -8,14 +8,14 @@ import (
 	"testing"
 
 	addr "github.com/filecoin-project/go-address"
-	cid "github.com/ipfs/go-cid"
+	"github.com/ipfs/go-cid"
```
anorth: Please retain the aliases, at least where they don't match the path. See https://github.com/golang/go/issues/28428

support/mock/mock_syscaller.go
shannonwells: This has a side effect the linter pointed out, which is, the builder does a copy of the runtime, and since syscalls would take testing.T in them, now those objects have RWMutexes, so it flags it as a copylock.  I’ll leave it as a panic but helpfully emit the function name in the panic message.

PR #133, opened 2020-02-07
Define method for signing a deal proposal

magik6k:


actors/crypto/signature.go
```diff
@@ -13,3 +17,9 @@ type Signature struct {
 	Type SigType
 	Data []byte
 }
+
+type SignFunc = func(context.Context, []byte) (*Signature, error)
+
+type Signable interface {
+	Sign(context.Context, SignFunc) error
```
magik6k: Note: lotus needs this context to be able to call wallet over RPC

actors/builtin/market/deal.go
```diff
@@ -1,8 +1,12 @@
 package market
 
 import (
+	"bytes"
+	"context"
+
 	addr "github.com/filecoin-project/go-address"
 	cid "github.com/ipfs/go-cid"
+	"golang.org/x/xerrors"
```
anorth: use `pkg/errors` for consistency with the rest of the project

actors/builtin/market/deal.go
```diff
@@ -22,7 +26,7 @@ type DealProposal struct {
 	PieceSize       abi.PaddedPieceSize
 	Client          addr.Address
 	Provider        addr.Address
-	ClientSignature acrypto.Signature
+	ClientSignature *acrypto.Signature
```
anorth: I literally just confirmed with @whyrusleeping that non-pointers is appropriate for all the nested structures. What motivates this change? Why not just use the zero value?

actors/crypto/signature.go
```diff
@@ -1,5 +1,9 @@
 package crypto
 
+import (
+	"context"
+)
+
 type SigType int64
```
anorth: Suggestion: use `0` for SigTypeUnknown so it's the ZeroValue?

PR #135, opened 2020-02-07
Batchify a bunch of per-sector operations to reduce messages and state transactions.

anorth:
This also paves the way for a fault representation that doesn't need to unset individual fault bits.

@Kubuxu tagging you for review as a good intro to the places the update the miner actor with a sequence of batch fault declarations.

cc @zixuanzh

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -37,7 +37,8 @@ const (
 
 type CronEventPayload struct {
 	EventType CronEventType
-	Sectors   []abi.SectorNumber // Empty for global events, such as SurprisePoSt expiration.
+	// TODO: replace sectors with a bitfield
+	Sectors []abi.SectorNumber // Empty for global events, such as SurprisePoSt expiration.
 }
 
 /////////////////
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -573,151 +568,145 @@ func (a *Actor) OnDeferredCronEvent(rt Runtime, params *OnDeferredCronEventParam
...
+			weight := asStorageWeightDesc(st.Info.SectorSize, sector)
+			allWeights = append(allWeights, weight)
 
-		err = st.deleteSector(store, sectorNumber)
-		if err != nil {
-			rt.Abort(exitcode.ErrIllegalState, "failed to delete sector: %v", err)
+			fault, err := st.FaultSet.Has(uint64(sectorNo))
+			AssertNoError(err)
+			Assert(fault == (sector.DeclaredFaultEpoch != epochUndefined))
+			Assert(false == (sector.DeclaredFaultDuration != epochUndefined))
```
Kubuxu: this should be `fault`.

PR #136, opened 2020-02-08
Big Int binary (un)marshal

ZenGround0:
For gfc interop

### DIFF
--- a/actors/abi/big/int.go
+++ b/actors/abi/big/int.go
@@ -24,7 +24,9 @@ func Zero() Int {
 	return NewInt(0)
 }
 
-func FromBytes(b []byte) Int {
+// PositiveFromUnsignedBytes interprets b as the bytes of a big-endian unsigned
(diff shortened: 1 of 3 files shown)


anorth:
It's my understanding that MarshalBinary should have nothing intrinsicly to do with CBOR. Is this just a convenient co-incidence, or do you intend to rely on the fact that MarshalBinary produces a CBOR representation. The latter would be a red flag, and also must definitely be documented.

ZenGround0:
> Is this just a convenient co-incidence?

Yes!  `FromCborBytes` and `CborBytes` don't write cbor, they write out the raw bytes that a cbor encoder will then pack into major type 2 (and then write the actual cbor header bytes with major type and length).  Really those functions should be renamed to `FromBytes` and `Bytes` as they are agnostic to cbor.  Let me know if you want me to do that rename.

anorth:
Yes, please do

PR #137, opened 2020-02-08
Seal TODOs

sternhenri:
from https://github.com/filecoin-project/specs/issues/867

In this PR:
- Made `MaxSealTime` a variable dependent on proof type everywhere (including cron callbacks for precommitExpiry)
- Added `PreCommitChallengeDelay` as Lotus uses to determine time between `PreCommit` appears on chain and challenge is drawn for interactive PoRep.
- Collapsing of redundant parameters `MaxSealTime` and `PoRepMaxDelay`. (@jzimmerman to verify no missing of purpose)

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -127,16 +127,11 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 			NewWorker:   worker,
 			EffectiveAt: effectiveEpoch,
 		}
-
-		cronPayload := CronEventPayload{
-			EventType: CronEventType_Miner_WorkerKeyChange,
-		}
 		return nil
 	})
 
 	cronPayload := CronEventPayload{
 		EventType: cronEventWorkerKeyChange,
-		Sectors:   nil,
 	}
 	a.enrollCronEvent(rt, effectiveEpoch, &cronPayload)
 	return &adt.EmptyValue{}
@@ -937,7 +932,7 @@ func (a Actor) verifySeal(rt Runtime, sectorSize abi.SectorSize, onChainInfo *ab
 	// Check randomness.
 	sealRandEarliest := rt.CurrEpoch() - ChainFinalityish - MaxSealDuration[onChainInfo.RegisteredProof]
 	if onChainInfo.SealRandEpoch < sealRandEarliest {
-		rt.Abort(exitcode.ErrIllegalArgument, "seal epoch %v too old, expected >= %v", onChainInfo.SealRandEpoch, sealRandEarliest)
+		rt.Abortf(exitcode.ErrIllegalArgument, "seal epoch %v too old, expected >= %v", onChainInfo.SealRandEpoch, sealRandEarliest)
 	}
 
 	commD := a.requestUnsealedSectorCID(rt, sectorSize, onChainInfo.DealIDs)
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -210,6 +210,11 @@ func (a Actor) OnDeleteMiner(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 // Sector Commitment //
 ///////////////////////
 
+type PreCommitSectorParams struct {
+	Info SectorPreCommitInfo
+	RegisteredProof abi.RegisteredProof
+}
+
 // Proposals must be posted on chain via sma.PublishStorageDeals before PreCommitSector.
 // Optimization: PreCommitSector could contain a list of deals that are not published yet.
 func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.EmptyValue {
@@ -248,11 +253,17 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 
 	// Request deferred Cron check for PreCommit expiry check.
 	cronPayload := CronEventPayload{
+<<<<<<< HEAD
 		EventType:       cronEventPreCommitExpiry,
 		Sectors:         &bf,
 		RegisteredProof: params.info.RegisteredProof,
+=======
+		EventType:       CronEventType_Miner_PreCommitExpiry,
+		Sectors:         []abi.SectorNumber{params.info.SectorNumber},
+		RegisteredProof: params.RegisteredProof,
+>>>>>>> address anorth comments
 	}
-	expiryBound := rt.CurrEpoch() + MaxSealDuration[params.info.RegisteredProof] + 1
+	expiryBound := rt.CurrEpoch() + MaxSealDuration[params.RegisteredProof] + 1
 	a.enrollCronEvent(rt, expiryBound, &cronPayload)
 
 	return &adt.EmptyValue{}
@@ -636,6 +647,7 @@ func (a Actor) checkPrecommitExpiry(rt Runtime, sectors *abi.BitField, regProof 
 		rt.Abortf(exitcode.ErrIllegalState, "failed to load sector %v: %v", sectorNo, err)
 	}
 
+<<<<<<< HEAD
 	sectorNos := bitfieldToSectorNos(rt, sectors)
 
 	depositToBurn := abi.NewTokenAmount(0)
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -264,8 +267,6 @@ func (a *Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *adt.
 	depositReq := precommitDeposit(st.getSectorSize(), params.info.Expiration-rt.CurrEpoch())
 	confirmPaymentAndRefundChange(rt, depositReq)
 
-	// TODO HS Check on valid SealEpoch
```
sternhenri: Remove from the Precommit: this is done in `verifySeal` after the commit is done. Functionally the same

actors/builtin/miner/miner_state.go
```diff
@@ -72,11 +71,12 @@ type WorkerKeyChange struct {
 }
 
 type SectorPreCommitInfo struct {
-	SectorNumber abi.SectorNumber
-	SealedCID    cid.Cid // CommR
-	SealEpoch    abi.ChainEpoch
-	DealIDs      []abi.DealID
-	Expiration   abi.ChainEpoch // Sector Expiration
+	RegisteredProof abi.RegisteredProof
```
sternhenri: allowed to add here @porcuquine ? I think we want it included with precommit so we can know when the PreCommit should expire if the SNARK doesn't appear on chain...?

actors/builtin/miner/policy.go
```diff
@@ -6,23 +6,19 @@ import (
 	power "github.com/filecoin-project/specs-actors/actors/builtin/power"
 )
 
-// An approximation to chain state finality.
+// An approximation to chain state finality (should include message propagation time as well).
 const ChainFinalityish = abi.ChainEpoch(500) // PARAM_FINISH
 
-// Lookback from current epoch from which to obtain a PoRep challenge.
-// TODO: HS why is this value unused?
-const PoRepLookback = ChainFinalityish // Should be approximately chain ~finality. PARAM_FINISH
```
sternhenri: This is to be used in subsystem code for drawing the sealRandEpoch, not on actors.

actors/builtin/miner/policy.go
```diff
@@ -6,23 +6,19 @@ import (
...
-
-// Minimum and maximum delay (inclusive) between a sector pre-commitment and corresponding proof of commitment.
-const PoRepMinDelay = abi.ChainEpoch(5)  // PARAM_FINISH
-const PoRepMaxDelay = abi.ChainEpoch(10) // PARAM_FINISH
```
sternhenri: I assume you added this @jzimmerman ?

I want to remove in favor of the `MaxSealDuration` param, which is more general (PoRep could be non-interactive)...

Specifically while `MaxSealDuration` is larger than `PoRepMaxDelay` (it is time for whole Seal so both for the pre-commit and proof, not just between the two (ie just the proof)), I think the time to PreCommit is comparatively negligible so we can use the single param... Initially we used MaxSealDuration to ensure the seal is tied to the chain using `SealRandEpoch` appropriately even though SRE is unverifiable.

Likewise removed the MinDelay since I couldn't fully see its purpose security wise... Lmk

actors/builtin/miner/miner_actor.go
```diff
@@ -652,17 +649,20 @@ func (a *Actor) checkPrecommitExpiry(rt Runtime, sectorNo abi.SectorNumber) {
...
-			err = st.deletePrecommitttedSector(store, sectorNo)
-			if err != nil {
-				rt.Abort(exitcode.ErrIllegalState, "failed to delete precommit %v: %v", sectorNo, err)
-			}
-			return nil
-		})
-		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, sector.PreCommitDeposit)
-		builtin.RequireSuccess(rt, code, "failed to burn funds")
+	if rt.CurrEpoch()-sector.PreCommitEpoch <= MaxSealDuration[regProof] {
+		rt.Abort(exitcode.ErrIllegalState, "Precommit not expired %v: %v", sectorNo, err)
```
anorth: Aborting here isn't particularly helpful. It will merely abort the call from cron and prevent further cleanup progress.

actors/builtin/miner/miner_state.go
anorth: Is it not sufficient to put it in the cron callback payload? Even there I think it's redundant because it's implied by the epoch that the event is scheduled for.

Putting it here is expensive - it must live for a long time on chain, way past its period of utility.

actors/builtin/miner/miner_actor.go
sternhenri: Gotcha. Reverted.

actors/builtin/miner/miner_state.go
sternhenri: point well taken. Fixed

anorth: Thank you. I should have noted is that an alternative is to restructure the on-chain structures so that the committed sector doesn't include the whole precommitted sector into, but extracts only the fields necessary. Arguably we should do that anyway. cc @magik6k

actors/builtin/miner/miner_actor.go
```diff
@@ -96,8 +97,9 @@ func (a *Actor) StageWorkerKeyChange(rt Runtime, params *StageWorkerKeyChangePar
 		st.Info.PendingWorkerKey = keyChange
 
 		cronPayload := CronEventPayload{
-			EventType: CronEventType_Miner_WorkerKeyChange,
-			Sectors:   nil,
+			EventType:       CronEventType_Miner_WorkerKeyChange,
+			Sectors:         nil,
+			RegisteredProof: abi.RegisteredProof_Undefined,
```
anorth: Nit: I think it's ok to leave the zero-valued fields here uninitialized.

actors/builtin/miner/miner_state.go
sternhenri: will leave for future PR. If you want to link me to area this is done in Lotus @magik6k I can PR

actors/builtin/miner/miner_actor.go
```diff
@@ -290,10 +288,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *adt.E
...
 	// Request deferred Cron check for PreCommit expiry check.
 	cronPayload := CronEventPayload{
-		EventType: CronEventType_Miner_PreCommitExpiry,
-		Sectors:   []abi.SectorNumber{params.Info.SectorNumber},
+		EventType:       CronEventType_Miner_PreCommitExpiry,
+		Sectors:         []abi.SectorNumber{params.Info.SectorNumber},
+		RegisteredProof: params.RegisteredProof,
 	}
-	expiryBound := rt.CurrEpoch() + PoRepMaxDelay + 1
+	expiryBound := rt.CurrEpoch() + MaxSealDuration[params.RegisteredProof] + 1
```
sternhenri: can pull the `regProof` from `MinerInfo` in actor state + invalid if not verified as is

actors/builtin/miner/miner_actor.go
```diff
@@ -319,20 +318,13 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 		rt.Abortf(exitcode.ErrNotFound, "no precommitted sector %v", sectorNo)
 	}
 
-	if rt.CurrEpoch() > precommit.PreCommitEpoch+PoRepMaxDelay || rt.CurrEpoch() < precommit.PreCommitEpoch+PoRepMinDelay {
```
sternhenri: leave this here. PreComm is a constant network wide

actors/builtin/miner/policy.go
sternhenri: Don't remove MinDelay, needed to prevent potential grinding issue

sternhenri: nvm that is the `PreCommitChallengeDelay`

actors/builtin/miner/miner_actor.go
sternhenri: It needs to be per sector actually, so submitted with `PreCommitInfo`

actors/builtin/miner/miner_state.go
sternhenri: Per convo with @whyrusleeping I think it does have to be on chain since it is part of a per sector configuration... Not sure what truth is here.

dignifiedquire: - it is not sufficient to store this in the callback, as the sector must be tied to the instance
- this should be very cheap as it is a single number
- all other places MUST look this up here and refer to it when interacting with this sector

actors/builtin/miner/miner_actor.go
```diff
@@ -319,20 +317,13 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
-	// Presumably they cannot be derived from the SectorProveCommitInfo provided by an untrusted party.
-
 	a.verifySeal(rt, st.Info.SectorSize, &abi.OnChainSealVerifyInfo{
-		SealedCID:    precommit.Info.SealedCID,
-		SealEpoch:    precommit.Info.SealEpoch,
-		Proof:        params.Proof,
-		DealIDs:      precommit.Info.DealIDs,
-		SectorNumber: precommit.Info.SectorNumber,
+		SealedCID:        precommit.Info.SealedCID,
+		InteractiveEpoch: precommit.PreCommitEpoch + PreCommitChallengeDelay,
```
dignifiedquire: This is a stricter check than before, as this has now to be drawn from a specific block, but it seems like a good change from my side.

PR #143, opened 2020-02-10
implement computeProvingSet()

sternhenri:
Updating PoSt to compute provingSet at appropriate time. 

The same `computeProvingSet` call should be used as part of computing electionPoSts, but without updating the `minerState.ProvingSet`after the call.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -25,8 +25,9 @@ type Runtime = vmr.Runtime
 const epochUndefined = abi.ChainEpoch(-1)
 
 type CronEventType int64
+
 const (
-	cronEventPoStExpiration CronEventType = iota
+	cronEventWindowedPoStExpiration CronEventType = iota
 	cronEventWorkerKeyChange
 	cronEventPreCommitExpiry
 	cronEventSectorExpiry
@@ -180,6 +181,7 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *abi.OnChainPoStVerifyInfo)
 			NumConsecutiveFailures: 0,
 		}
 
+		// reset provingSet to include all sectors (were not included during challenge period)
 		st.ProvingSet = st.Sectors
 
 		return nil
@@ -350,10 +352,7 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 			rt.Abortf(exitcode.ErrIllegalState, "failed to check miner sectors sizes: %v", err)
 		}
 		if len == 1 {
-			// enroll expiration check
-			a.enrollCronEvent(rt, st.PoStState.ProvingPeriodStart+power.WindowedPostChallengeDuration, &CronEventPayload{
-				EventType: cronEventPoStExpiration,
-			})
+			st.PoStState.ProvingPeriodStart = rt.CurrEpoch() + ProvingPeriod
 		}
 
 		// Do not update proving set during challenge window
@@ -381,7 +380,7 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 	if len == 1 {
 		// enroll expiration check
 		a.enrollCronEvent(rt, st.PoStState.ProvingPeriodStart+power.WindowedPostChallengeDuration, &CronEventPayload{
-			EventType: cronEventPoStExpiration,
+			EventType: cronEventWindowedPoStExpiration,
 		})
 	}
 
@@ -563,7 +562,7 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, params *OnDeferredCronEventParams
 		a.checkSectorExpiry(rt, payload.Sectors)
 	}
 
-	if payload.EventType == cronEventPoStExpiration {
+	if payload.EventType == cronEventWindowedPoStExpiration {
 		a.checkPoStProvingPeriodExpiration(rt)
 	}
 
@@ -904,7 +903,7 @@ func (a Actor) verifyWindowedPost(rt Runtime, st *State, onChainInfo *abi.OnChai
 	// the same was not used to generate the proof
 
 	store := adt.AsStore(rt)
-	provingSet := adt.AsArray(store, st.ProvingSet)
+	provingSet, err := st.ComputeProvingSet(store)
 
 	var sectorInfos []abi.SectorInfo
 	var ssinfo SectorOnChainInfo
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -230,11 +230,36 @@ func (st *State) IsSectorInTemporaryFault(store adt.Store, sectorNo abi.SectorNu
...
+		Assert(sector.DeclaredFaultEpoch != epochUndefined)
+		Assert(sector.DeclaredFaultDuration != epochUndefined)
+
+		if !fault {
+			err2 := provingSet.Set(uint64(sector.Info.SectorNumber), sector)
+			if err2 != nil {
+				return
+			}
+		}
+	})
```
anorth: Thanks for this. I think it's complicated™️ . 

When we talked, I think I'd indicated that there was no need to subtract the fault set from this, just copy the sectors root CID. Thanks for realising that we can't lazily subtract the faultset during evaluation, because the fault set might have changed. But:
- If we're just trying to record the IDs of the sectors in the proving set, we should use the much more compact bitfield representation rather than construct a new AMT. The lightweight clone approach of copying the root CID is only lightweight if the collection is identical. This will write a whole new AMT each proving period, which is quite expensive.
- But but, the values in this AMT are SectorInfos, which are mutable objects. So by cloning the AMT in this way we're actually baking in the present fault information in the form of the DeclaredFaultEpoch/Duration fields. So actually the faults are frozen after all so it *is* safe to just do the cheap copy (but have to traverse it to figure out the sector id set).. But only as fallout from the otherwise unfortunate design of having mutable sector state on chain.
- But but but, `st.ProvingSet` appears to be unused in the miner actor! So I can't tell what the more appropriate representation is to make a recommendation here. I can see a thread of spec debt where `PoStVerifyInfo.EligibleSectors` is unimplemented, a Go map that should almost certainly be something else (it can't be a H/AMT). Probably that's supposed to be a bitfield, which would then suggest option 1 above.

@magik6k can you confirm that approach?

sternhenri: Thanks for the catch. Just updated to call `st.ProvingSet` appropriately, @anorth. Sorry about that.

Thanks also for spelling out the above.

Your note does make me wonder why we even include a `FaultSet` in the first place? Is the hope to do easy-lookup of faults? Given that `IsSectorInTemporaryFault()` gets the sector as a whole, why not simply read its `DeclaredFaultEpoch` and `DeclaredFaultDuration` there and save ourselves a bitfield?

To repeat what you said above though (and make sure I get it), you are saying that given sectors include the fault fields, doing the AMT clone (simply copying the root CID of `sectors` as the`provingSet`) and then later traverse it to get only those IDs we want (you could imagine a `SetProvingSet()` method which does the snapshot while `ComputeProvingSet()` pares it down) should work given it will be a snapshot of those sectors as they were at that time (given further mutation will lead to a new root). Is that right?

actors/abi/sector.go
```diff
@@ -92,9 +92,9 @@ type PartialTicket []byte // 32 bytes
 type PoStVerifyInfo struct {
 	Randomness      PoStRandomness
 	CommR           cid.Cid
-	Candidates      []PoStCandidate      // From OnChain*PoStVerifyInfo
-	Proofs          []PoStProof          // From OnChain*PoStVerifyInfo
-	EligibleSectors map[SectorID]cid.Cid // TODO: HAMT?
+	Candidates      []PoStCandidate // From OnChain*PoStVerifyInfo
+	Proofs          []PoStProof     // From OnChain*PoStVerifyInfo
+	EligibleSectors cid.Cid
```
anorth: You can't do thAT in this struct. Sorry, I should have been more direct that a HAMT cannot be used here. This structure is passed to the PoSt verification syscall and ultimately the proofs library, and must be fully self-contained.

I think it should be either a `[]SectorID` or, probably, a bitfield. You could do the former for expediency now, but `TODO bitfield`.

sternhenri:
Ack. Makes sense. Not sure why I assumed the proofs lib could use the CID 😛

anorth:
> wonder why we even include a FaultSet in the first place?
Because that is the implementation-efficient approach that we've been using all along, until the spec departed from reality a while back

> Given that IsSectorInTemporaryFault() gets the sector as a whole...
Not any more, that was horribly inefficient and the method no longer exists. It is my hope that @Kubuxu 's upcoming improvements to the fault representation will also let us remove those fields from the sector info.


> you are saying that given sectors include the fault fields, doing the AMT clone ... and then later traverse it to get only those IDs we want ... should work given it will be a snapshot of those sectors as they were at that time?

(1) yes I think that will "work" given the current state of the code (but note that you can't just pass that CID to the proofs library, you have to actually traverse the snapshot), but (2) it's probably better to do the traversal and store the proving set sector ids in a bitfield, since that's probably what we need to pass to the proof verification anyway.

Please consult with @magik6k or @whyrusleeping to resolve the "probably"s here.

anorth:
See #148

sternhenri:
Should be ready for re-review, also fixed an error i introduced in #148 .

The one hesitation I have @anorth is wrt to the way @whyrusleeping sets `st.ProvingSet := sectors`on `ProveCommit`, and other changes to the sector set outside the challengeWindow. It's a bit unclean but more importantly I worry about the following: if the provingSet check happens too late (outside the challengeWindow), there's a chance the `provingSet` will have been changed before `computeProvingSet` is called by the verifier.

While it shouldn't happen, it's a subtle point introduced by the fact that so many parts of the actor touch the proving set.

A solution I could see would be to make a new `enrollCronEvent` on `windowedPoStChallenge` that "freezes" the ProvingSet in a single place at the beginning of the proving period. More inter-actor calls, most clean.

We can also leave as is since it should be fine.

actors/builtin/miner/miner_actor.go
```diff
@@ -894,7 +897,10 @@ func (a Actor) verifyWindowedPost(rt Runtime, st *State, onChainInfo *abi.OnChai
 	postRandomness := crypto.DeriveRandWithMinerAddr(crypto.DomainSeparationTag_WindowedPoStChallengeSeed, randomnessK, rt.Message().Receiver())
 
 	store := adt.AsStore(rt)
-	provingSet := adt.AsArray(store, st.ProvingSet)
+	provingSet, err := st.ComputeProvingSet(store)
```
anorth: Now that windowed post is in, we don't need this method (it used to persist the proving set).

Just check for faults at line 909 in the existing iteration over the proving set. Delete `ComputeProvingSet`.

sternhenri: Noted. derp.

I decided to move this into there nonetheless to keep verifySmaller and given that the compute function (substracting faults from the set) will be used outside of actors, so could be useful.

Happy to revert. What do you think of my above comment?

PR #144, opened 2020-02-11
feat: new message abstraction

icorderi:


icorderi:
cc: @anorth

actors/runtime/runtime.go
```diff
@@ -15,34 +15,24 @@ import (
...
 	// Validates the caller against some predicate.
 	// Exported actor methods must invoke at least one caller validation before returning.
 	ValidateImmediateCallerAcceptAny()
 	ValidateImmediateCallerIs(addrs ...addr.Address)
 	ValidateImmediateCallerType(types ...cid.Cid)
 
-	// The actor who mined the block in which the current invocation is being evaluated.
-	// Always an ID-address.
-	ToplevelBlockWinner() addr.Address
-
```
anorth: If ToplevelBlockWinner goes should CurrEpoch go with it?

Issue #140, opened 2020-02-08
Runtime IpldGet/Put should require context parameter


anorth:
The adt.Store runtime wrapper drops the context from the method signature when adapting to the cbor store interface. While it's possible for it to be hooked up behind the runtime, it would be more straightfoward to simply use the context already provided.

anorth:
For a counterpoint, @icorderi argues that the cbor.IpldStore interface the HAMT/AMT interfaces need should not require the Go context as a parameter (but it should be connected deeper down). In that case, we would not need to expose the Go context in the runtime at all.

This would certainly be nice, but it's not [canonical use of the Go context](https://golang.org/pkg/context/).
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

@whyrusleeping as our HAMT/AMT contact, do you have an opinion here? (cc @frrist )

PR #146, opened 2020-02-11
feat: move ipldput/get into a separate abstraction

icorderi:


icorderi:
cc: @anorth

actors/builtin/init/init_test.go
```diff
@@ -76,7 +76,7 @@ func TestExec(t *testing.T) {
 
 		var st init_.State
 		rt.GetState(&st)
-		actualIdAddr, err := st.ResolveAddress(rt.Store(), uniqueAddr1)
+		actualIdAddr, err := st.ResolveAddress(adt.AsStore(rt), uniqueAddr1)
```
magik6k: De we actually need to keep `adt.AsStore`?

anorth: This is annoying. The mock was already constructed not to require this explicit wrapping. Please restore a method to the mock runtime that can be reached directly by test code.

@magik6k if we do https://github.com/filecoin-project/specs-actors/issues/140 then we won't need any wrapping anywhere.

Issue #83, opened 2020-01-31
Add method ID to storage power cron-event


anorth:
The storage power actor's cron event stores the miner address and a serialized payload. The method ID on the storage miner to call back to is hardcoded to a single value. Inside that callback method, the miner actor reads an "event type" out of the serialized payload and then internally dispatches to a different method.

It would be more flexible and direct to simply register a callback method number with the cron event, removing the dispatch mechanism from the miner and generalizing the mechanism.

anorth:
This would also let different callbacks receive different payloads without a union type.

PR #148, opened 2020-02-11
switch post challenges from surprise to windowed

whyrusleeping:
Note: This isnt done yet, but it does build and the 'tests' 'pass'. I know i still need to do cbor gen, and the word surprise is still sticking around, but i'm going to sleep soon so i figured i'd put this up for people to look at.


actors/builtin/miner/miner_actor.go
```diff
@@ -142,66 +140,37 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-		if !st.PoStState.isChallenged() {
-			rt.Abort(exitcode.ErrIllegalState, "Not currently challenged")
+
+		if rt.CurrEpoch() <= st.PoStState.ProvingPeriodStart+ProvingPeriod {
+			rt.Abort(exitcode.ErrIllegalState, "Not currently in submission window for PoSt")
 		}
-		a.verifySurprisePost(rt, &st, &params.onChainInfo)
+		a.verifyWindowedPost(rt, &st, params)
+
+		// TODO: post is valid, do we do anything with faults now?
```
zixuanzh: There are two kinds of faults, Miner-Wide fault (failing fallback) and Sector-Specific fault (temporary fault). If a miner is not previously in Miner-Wide Fault (known as DetectedFault previously), then submitting FallbackPoSt should have no impact on sector state wrt fault. If a miner is in Miner-Wide Fault, then submitting FallbackPoSt successfully will get the miner out of the Miner-Wide Fault and put its sectors back in the sectorset. Miners may only gain power after the next PoSt.

zixuanzh: TemporaryFault recovery on the other hand is independent of Fallback as it has its own EffectiveEnd where the sector is added back to the sector set. To be consistent, it should only regain its power after the first PoSt.

actors/builtin/miner/miner_actor.go
```diff
@@ -142,66 +140,37 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-			LastSuccessfulPoSt:     rt.CurrEpoch(),
-			SurpriseChallengeEpoch: epochUndefined,
+			ProvingPeriodStart:     st.PoStState.ProvingPeriodStart + ProvingPeriod,
 			NumConsecutiveFailures: 0,
 		}
 		return nil
 	})
 
-	_, code := rt.Send(
+	_, code := rt.Send( // TODO: what does this actually do?
```
zixuanzh: this is for Power accounting in the Power actor.

actors/builtin/miner/miner_actor.go
```diff
@@ -734,20 +674,19 @@ func (a Actor) checkPoStProvingPeriodExpiration(rt Runtime) {
...
 	expired := rt.State().Transaction(&st, func() interface{} {
-		if !st.PoStState.isChallenged() {
-			return false // Already exited challenged state successfully prior to expiry.
-		}
 
-		window := power.SurprisePostChallengeDuration
-		if rt.CurrEpoch() < st.PoStState.SurpriseChallengeEpoch+window {
+		window := power.FallbackPostChallengeDuration
+		if rt.CurrEpoch() < st.PoStState.ProvingPeriodStart+ProvingPeriod+window {
+			// TODO: do we ever expect this to happen? seems like it might be a
```
zixuanzh: it depends on when and how `_rtCheckFallbackPoStExpiry` is triggered

actors/builtin/miner/miner_actor.go
```diff
@@ -874,10 +813,14 @@ func (a Actor) requestTerminatePower(rt Runtime, terminationType power.SectorTer
...
-func (a Actor) verifySurprisePost(rt Runtime, st *State, onChainInfo *abi.OnChainSurprisePoStVerifyInfo) {
-	Assert(st.PoStState.isChallenged())
+func (a Actor) verifyWindowedPost(rt Runtime, st *State, onChainInfo *abi.OnChainPoStVerifyInfo) {
 	sectorSize := st.Info.SectorSize
-	challengeEpoch := st.PoStState.SurpriseChallengeEpoch
+
+	// challenge issued at the end (beginning of next?) proving period
+	challengeEpoch := st.PoStState.ProvingPeriodStart + ProvingPeriod
+
+	// TODO: verifying no duplicates here seems wrong, we should be verifying
```
zixuanzh: 👍

actors/builtin/miner/miner_actor.go
```diff
@@ -889,22 +832,38 @@ func (a Actor) verifySurprisePost(rt Runtime, st *State, onChainInfo *abi.OnChai
...
+	postRandomness := crypto.DeriveRandWithMinerAddr(crypto.DomainSeparationTag_FallbackPoStChallengeSeed, randomnessK, rt.CurrReceiver())
+
+	store := adt.AsStore(rt)
+	provingSet := adt.AsArray(store, st.ProvingSet)
+
+	var sectorInfos []abi.SectorInfo
+	var ssinfo SectorOnChainInfo
+	provingSet.ForEach(&ssinfo, func(i int64) error {
+
+		// TODO: faults!!!!
```
zixuanzh: if any one of the sector here fails, miner enters Miner-Wide fault, power goes to zero.

actors/builtin/miner/miner_state.go
```diff
@@ -117,8 +110,7 @@ func ConstructState(store adt.Store, ownerAddr, workerAddr addr.Address, peerId
 			SectorSize:       sectorSize,
 		},
 		PoStState: PoStState{
-			LastSuccessfulPoSt:     epochUndefined,
-			SurpriseChallengeEpoch: epochUndefined,
+			ProvingPeriodStart:     epochUndefined,
 			NumConsecutiveFailures: 0,
```
zixuanzh: NumConsecutiveFailures is only relevant for Miner-Wide Fault

actors/builtin/miner/miner_actor.go
anorth: @zixuanzh that doesn't make sense. 

I think @whyrusleeping 's comment is noting that declared faulty sectors need to be excluded from the proving set.

However, if this proof verification fails, _nothing should happen_. The miner has until the end of the window to submit a good proof.

actors/builtin/miner/miner_actor.go
```diff
@@ -889,22 +832,38 @@ func (a Actor) verifySurprisePost(rt Runtime, st *State, onChainInfo *abi.OnChai
...
+		sectorInfos = append(sectorInfos, abi.SectorInfo{
+			CommR:    ssinfo.Info.SealedCID,
+			SectorID: ssinfo.Info.SectorNumber,
+		})
```
anorth: This potentially large accumulation is unfortunate. Since the proof will actually only sample some sectors, and which sectors is known by this time to the proof-submitter, can we avoid accumulating this somehow?

FYI note the the "TODO: potentially unbounded computation" in `requestTerminateAllDeals`. We could (should) bound the number of sectors, but we probably want to set that number very high.

actors/builtin/methods.go
```diff
@@ -75,25 +75,23 @@ var MethodsPower = struct {
 	OnSectorTemporaryFaultEffectiveBegin abi.MethodNum
 	OnSectorTemporaryFaultEffectiveEnd   abi.MethodNum
 	OnSectorModifyWeightDesc             abi.MethodNum
-	OnMinerSurprisePoStSuccess           abi.MethodNum
-	OnMinerSurprisePoStFailure           abi.MethodNum
+	OnMinerFallbackPoStSuccess           abi.MethodNum
+	OnMinerFallbackPoStFailure           abi.MethodNum
```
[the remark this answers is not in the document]
sternhenri: 👍 note that my current assumption is that it will be required by all miners even those regularly winning elections, since direction right now is to separate consensus considerations from these.

actors/builtin/miner/miner_actor.go
```diff
@@ -142,66 +140,37 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-		surpriseDuration := power.SurprisePostChallengeDuration
-		a.enrollCronEvent(rt, rt.CurrEpoch()+surpriseDuration, &cronPayload)
-	}
-	return &adt.EmptyValue{}
-}
-
-type SubmitSurprisePoStResponseParams struct {
-	onChainInfo abi.OnChainSurprisePoStVerifyInfo
-}
+// TODO: where should this live?
```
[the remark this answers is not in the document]
sternhenri: +1 to adding to power/policy. The `ProvingPeriod` is how often the miner will need to submit one such PoSt.

actors/builtin/miner/miner_actor.go
```diff
@@ -142,66 +140,37 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-func (a Actor) SubmitSurprisePoStResponse(rt Runtime, params *SubmitSurprisePoStResponseParams) *adt.EmptyValue {
+// Invoked by miner's worker address to submit their fallback post
+func (a Actor) SubmitFallbackPoSt(rt Runtime, params *abi.OnChainPoStVerifyInfo) *adt.EmptyValue {
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
 		rt.ValidateImmediateCallerIs(st.Info.Worker)
-		if !st.PoStState.isChallenged() {
-			rt.Abort(exitcode.ErrIllegalState, "Not currently challenged")
+
+		if rt.CurrEpoch() <= st.PoStState.ProvingPeriodStart+ProvingPeriod {
```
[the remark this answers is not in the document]
sternhenri: @whyrusleeping note that we have 

```
// The average period (i.e. 1/frequency) of surprise PoSt challenges to each miner.
const SurprisePoStPeriod = abi.ChainEpoch(5760) // ~2 days @ 30 second epochs. PARAM_FINISH

// The time a miner has to respond to a surprise PoSt challenge.
const SurprisePostChallengeDuration = abi.ChainEpoch(240) // ~2 hours @ 30 second epochs. PARAM_FINISH

// The minimum period after a PoSt challenge before a miner can be challenged again.
const SurprisePoStNoChallengePeriod = abi.ChainEpoch(240) // PARAM_FINISH
```

the `NoChallengePeriod` should be removed, but the `period` and `challengeDuration` can be adapted to our ends here. To @anorth 's point, expliciting the duration rather than doing so implicitly by offsetting starts off one another would be v helpful.

My assumption as to how this should work is:

```
| PP1    > CP1 | PP2     > CP2| ...
```
where PP is proving period, and CP is challenge period, so if PP = 500, and CP = 10, you would get challenged at epoch 490 of the PP... Feel free to correct my understanding

actors/builtin/miner/miner_actor.go
```diff
@@ -217,35 +186,6 @@ func (a Actor) OnDeleteMiner(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	return &adt.EmptyValue{}
 }
 
-//////////////////
-// ElectionPoSt //
-//////////////////
-
-// Called by the VM interpreter once an ElectionPoSt has been verified.
-func (a Actor) OnVerifiedElectionPoSt(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
```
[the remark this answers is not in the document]
sternhenri: per current red team direction, seems right.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
sternhenri: I see below that you have the CP part of the next PP, why.... Either works. Seems more confusing :b (means we have to say "you will be challenged once per PP", rather than "you will post once per PP")

actors/builtin/miner/miner_actor.go
sternhenri: Some in https://github.com/filecoin-project/specs-actors/pull/143

actors/builtin/miner/miner_actor.go
sternhenri: Alas, this means one of the following:
- 1 reimplement the sampling algo from proofs land here in order to only accumulate sampled sectors (pretty inefficient)
2- remove sampling from proofs and include here instead. not a bad idea -- this may have been a poorly set abstraction barrier, however this may be a large rework... Unfortunately the API right now is "pass all sectors to proofs" -- would have been nice that this be visible within actors

sternhenri: So probs something we must live with

actors/builtin/miner/miner_actor.go
whyrusleeping: choosing to abort for now, leaving a note to say that it needs testing and further consideration

actors/builtin/methods.go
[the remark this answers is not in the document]
whyrusleeping: Yeah, I think fallback is no longer a good name. I've been using `WindowedPoSt` a bunch, but i'm not picky about the names. I figure the word `Fallback` isnt used anywhere else, and a single global sed command can fix it all once we decide

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
whyrusleeping: Take another look at this code, I went through it with @zixuanzh today

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
whyrusleeping: correct

actors/builtin/miner/miner_actor.go
whyrusleeping: @anorth Yeah, I think it should be technically possible to just pass to proofs the set of sectors it needs to care about, but i havent been able to get that out of them yet. Definitely agree this is a DoS vector

actors/builtin/methods.go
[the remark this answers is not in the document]
anorth: I like "windowed"

actors/builtin/miner/miner_actor.go
```diff
@@ -36,7 +36,7 @@ const (
 type CronEventPayload struct {
 	EventType CronEventType
 	// TODO: replace sectors with a bitfield
-	Sectors []abi.SectorNumber // Empty for global events, such as SurprisePoSt expiration.
+	Sectors *abi.BitField
```
anorth: Remove the TODO.

Everywhere else, we avoid pointers in chain state. Is this different somehow? Am I right assuming it serializes the same either way?

actors/builtin/miner/miner_actor.go
```diff
@@ -142,70 +140,40 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-		}
+	rt.State().Transaction(&st, func() interface{} {
+		rt.ValidateImmediateCallerIs(st.Info.Worker)
 
-		st.PoStState = PoStState{
-			LastSuccessfulPoSt:     st.PoStState.LastSuccessfulPoSt,
-			SurpriseChallengeEpoch: rt.CurrEpoch(),
-			NumConsecutiveFailures: st.PoStState.NumConsecutiveFailures,
+		if rt.CurrEpoch() > st.PoStState.ProvingPeriodStart+power.FallbackPostChallengeDuration {
+			rt.Abort(exitcode.ErrIllegalState, "PoSt submission too late")
```
anorth: Nit: add as much detail as possible, here and other messages.

`rt.Abort(exitcode.ErrIllegalState, "PoSt submission too late, period start %v, duration %v, now %v", ...)`

actors/builtin/miner/miner_actor.go
```diff
@@ -142,70 +140,40 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
+		// implemented, with faults, we have potential power adjustments
 
 		st.PoStState = PoStState{
-			LastSuccessfulPoSt:     rt.CurrEpoch(),
-			SurpriseChallengeEpoch: epochUndefined,
+			ProvingPeriodStart:     st.PoStState.ProvingPeriodStart + ProvingPeriod,
 			NumConsecutiveFailures: 0,
 		}
+
+		st.ProvingSet = st.Sectors
```
anorth: This isn't wrong, but it should *also* happen when the challenge window opens. Deferring it until then so we only do it once would make that clear.

actors/builtin/miner/miner_actor.go
```diff
@@ -142,70 +140,40 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
 	})
 
-	_, code := rt.Send(
-		builtin.StoragePowerActorAddr,
-		builtin.MethodsPower.OnMinerSurprisePoStSuccess,
-		&adt.EmptyValue{},
-		abi.NewTokenAmount(0),
-	)
-	builtin.RequireSuccess(rt, code, "failed to notify storage power actor")
+	// TODO: register a callback with cron to check that we submitted our proof at the right time
```
anorth: Don't forget!

actors/builtin/miner/miner_actor.go
```diff
@@ -217,41 +185,12 @@ func (a Actor) OnDeleteMiner(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
-}
-
 ///////////////////////
 // Sector Commitment //
 ///////////////////////
 
 type PreCommitSectorParams struct {
-	info SectorPreCommitInfo
+	Info SectorPreCommitInfo
 }
```
anorth: Wanna unwrap this struct?

actors/builtin/miner/miner_actor.go
```diff
@@ -383,14 +326,20 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 			rt.Abort(exitcode.ErrIllegalState, "failed to delete precommit for sector %v: %v", sectorNo, err)
 		}
 
-		st.ProvingSet = st.ComputeProvingSet()
+		// Do not update proving set during challenge window
+		if !st.inChallengeWindow(rt) {
+			st.ProvingSet = st.Sectors
+		}
```
anorth: Rather than do this every time, how about copying sectors to proving set only/exactly when the challenge window opens?

I don't think the proving set is relevant for EPost - that just uses the committed set (except faults).

actors/builtin/miner/miner_actor.go
```diff
@@ -460,17 +409,19 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
...
-	sectorNumbers []abi.SectorNumber
+	Sectors *abi.BitField
 }
 
 func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *adt.EmptyValue {
 	var st State
 	rt.State().Readonly(&st)
 	rt.ValidateImmediateCallerIs(st.Info.Worker)
 
+	sectorNos := bitfieldToSectorNos(rt, params.Sectors)
```
anorth: We should set a max count here, probably independent of any bitfield limitatinos.

actors/builtin/miner/miner_actor.go
```diff
@@ -526,9 +482,10 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params DeclareTemporaryFaultsP
 	builtin.RequireSuccess(rt, code, "failed to burn fee")
 
 	// Request deferred Cron invocation to update temporary fault state.
+	// TODO: cant we just lazily clean this up?
```
anorth: When we change to the list of faultdecls stored in state, I think we'll be in much better shape to iterate that lazily.

actors/builtin/miner/miner_actor.go
```diff
@@ -734,20 +703,16 @@ func (a Actor) checkPoStProvingPeriodExpiration(rt Runtime) {
...
-		if !st.PoStState.isChallenged() {
-			return false // Already exited challenged state successfully prior to expiry.
-		}
 
-		window := power.SurprisePostChallengeDuration
-		if rt.CurrEpoch() < st.PoStState.SurpriseChallengeEpoch+window {
-			// Challenge not yet expired.
-			return false
+		window := power.FallbackPostChallengeDuration
+		if rt.CurrEpoch() < st.PoStState.ProvingPeriodStart+ProvingPeriod+window {
```
anorth: This method would be better named `checkPoStWindowExpiration`.

This should not add ProvingPeriod as well as the window, just the start + window. I think the expected cases are:
- `now == start + window` (the window has expired, return true)
- `now < start` (a PoSt was submitted and the period start moved into the future)

So error case would be:
- `now > start && now != start+window` (cron mis-timed)

At present, this closure never returns `false`, but it should.

actors/builtin/miner/miner_actor.go
```diff
@@ -142,70 +140,40 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-			Sectors:   nil,
+		if rt.CurrEpoch() <= st.PoStState.ProvingPeriodStart {
+			rt.Abort(exitcode.ErrIllegalState, "Not currently in submission window for PoSt")
 		}
-		surpriseDuration := power.SurprisePostChallengeDuration
-		a.enrollCronEvent(rt, rt.CurrEpoch()+surpriseDuration, &cronPayload)
-	}
-	return &adt.EmptyValue{}
-}
+		a.verifyWindowedPost(rt, &st, params)
```
anorth: Maybe worth dropping a comment that a failed verification doesn't immediately cause a penalty - the miner has until the end of the window to submit a good proof.

actors/builtin/miner/miner_actor.go
anorth: We don't need to re-implement any sampling here. The caller already sampled and should be able to tell us (assuming proof verification would fail if the sample didn't match what the proofs lib recomputes during verification)

actors/builtin/miner/miner_state.go
```diff
@@ -50,18 +50,12 @@ type MinerInfo struct {
...
 type PoStState struct {
-	// Epoch of the last succesful PoSt, either election post or surprise post.
-	LastSuccessfulPoSt abi.ChainEpoch
-
-	// If >= 0 miner has been challenged and not yet responded successfully.
-	// SurprisePoSt challenge state: The miner has not submitted timely ElectionPoSts,
-	// and as a result, the system has fallen back to proving storage via SurprisePoSt.
-	//  `epochUndefined` if not currently challeneged.
-	SurpriseChallengeEpoch abi.ChainEpoch
+	// Epoch that starts the current proving period
```
anorth: Maybe note also that this is the start of the challenge window (if I'm interpreting correctly that the window is at the start of the period).

actors/builtin/miner/miner_state.go
```diff
@@ -216,23 +209,17 @@ func (st *State) GetStorageWeightDescForSector(store adt.Store, sectorNo abi.Sec
 	return asStorageWeightDesc(st.Info.SectorSize, sectorInfo), nil
 }
 
+func (st *State) inChallengeWindow(rt Runtime) bool {
```
anorth: Help us clean up debt and move this function (with rt dependency) to the actor itself.

Also this should check now < start + window

actors/builtin/miner/policy.go
```diff
@@ -47,3 +47,9 @@ func precommitDeposit(sectorSize abi.SectorSize, duration abi.ChainEpoch) abi.To
 func temporaryFaultFee(weights []*power.SectorStorageWeightDesc, duration abi.ChainEpoch) abi.TokenAmount {
 	return big.Zero() // PARAM_FINISH
 }
+
+// MaxFaultsCount is the maximum number of faults that can be declared
+const MaxFaultsCount = 32 << 20
+
+// ProvingPeriod defines the frequency of PoSt challenges that a miner will have to respond to
+const ProvingPeriod = 300
```
anorth: That's really quick!

It occurs to me this might be one of those things where we want to inject values here at build time for testing networks.

actors/builtin/miner/miner_actor.go
```diff
@@ -165,16 +166,14 @@ func (a Actor) SubmitFallbackPoSt(rt Runtime, params *abi.OnChainPoStVerifyInfo)
...
 	})
 
-	_, code := rt.Send( // TODO: what does this actually do?
-		builtin.StoragePowerActorAddr,
-		builtin.MethodsPower.OnMinerFallbackPoStSuccess,
-		&adt.EmptyValue{},
-		abi.NewTokenAmount(0),
-	)
-	builtin.RequireSuccess(rt, code, "failed to notify storage power actor")
+	// TODO: register a callback with cron to check that we submitted our proof at the right time
```
sternhenri: I can take in a follow up also addressing https://github.com/filecoin-project/specs-actors/issues/83

actors/builtin/miner/miner_actor.go
```diff
@@ -323,14 +326,20 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 			rt.Abort(exitcode.ErrIllegalState, "failed to delete precommit for sector %v: %v", sectorNo, err)
 		}
 
-		st.ProvingSet = st.ComputeProvingSet()
+		// Do not update proving set during challenge window
+		if !st.inChallengeWindow(rt) {
+			st.ProvingSet = st.Sectors
```
sternhenri: haaaa this is why we do this here: note that with surprise this was unnecessary given the surprise was triggered by the powerActor, and so we could just update the provingSet there and nowhere else.

actors/builtin/miner/miner_actor.go
```diff
@@ -466,9 +482,10 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params DeclareTemporaryFaultsP
 	builtin.RequireSuccess(rt, code, "failed to burn fee")
 
 	// Request deferred Cron invocation to update temporary fault state.
+	// TODO: cant we just lazily clean this up?
```
sternhenri: I guess you could do that when computing the proving set, that is: on submitting the fallback post, you need to subtract faults from the provingSet. You could check whether thee fault is still active then... Assuming there is no deeper incidence on collateral and such cc @zixuanzh

actors/builtin/miner/miner_actor.go
sternhenri: Fwiw i am not sure this works, specifically this method is called on FallbackPoSt submission by the miner. This is arguably not when the cron callback should be registered. It should be registered on a miner's first Sector Commitment and registered with the PowerActor, that way the PowerActor has a map from miner to epoch by which a recent fallback Post must have been submitteed, and if not, it penalizes the miner.

On FallbackPoSt submission, that epoch number is incremented by `provingPeriod`.

What do you think?

actors/builtin/miner/miner_state.go
```diff
@@ -208,6 +209,10 @@ func (st *State) GetStorageWeightDescForSector(store adt.Store, sectorNo abi.Sec
...
+	return rt.CurrEpoch() > st.PoStState.ProvingPeriodStart // TODO: maybe also a few blocks beforehand?
```
sternhenri: nah... just lengthen the challengeePeriod

actors/builtin/miner/policy.go
sternhenri: please remove `SurprisePoStPeriod` from the `power/policy.go` file accordingly. Fwiw I think `fallbackPoStPeriod` or something is better. Proving is too confusing/overloaded

actors/builtin/miner/miner_actor.go
```diff
@@ -656,6 +681,10 @@ func (a Actor) terminateSectors(rt Runtime, sectorNos []abi.SectorNumber, termin
 			if err != nil {
 				rt.Abort(exitcode.ErrIllegalState, "failed to delete sector: %v", err)
 			}
+
+			if !st.inChallengeWindow(rt) {
+				st.ProvingSet = st.Sectors
+			}
```
sternhenri: In general I don't love that we have to do this on all sectors updates... But way more efficient than system that existed for surprise (which was necessary given the surprising nature of the thing).

actors/builtin/miner/policy.go
whyrusleeping: Yeah, thats why i was wondering where it should live. In lotus we have this defined in a global 'build' package that has different values if we're in tests vs debug mode vs testnet mode

actors/builtin/miner/miner_state.go
whyrusleeping: oh, i mean that you could hypothetically know the randomness at the challenge issuance a block or two in advance if youre a miner, and you could then submit faults just before that to cheat your post

actors/builtin/miner/miner_actor.go
Kubuxu: It is.

actors/builtin/miner/miner_actor.go
sternhenri: I agree with this @anorth (and it would be nicer to only do it on challenge window open rather than every time the sector set changes), but the challenge window opening is no longer an actual event, so unless we add that back (ie -- have it triggered by power actor -- i wrote something to that effect below), this is the way to do it I think.

whyrusleeping: @anorth it doesnt need to happen again at that point, once we submit the post, it bumps our window forward in time, so that we're no longer inside it

actors/builtin/miner/miner_state.go
whyrusleeping: > Help us clean up debt and move this function (with rt dependency) to the actor itself.

Ack

> Also this should check now < start + window

Nope, doesnt need to. Once you either submit your proof, or are slashed for being late, ProvingPeriodStart moves forward, meaning you'll never actually be in a situation where rt.CurrEpoch > ProvingPeriodStart+window

actors/builtin/miner/miner_actor.go
whyrusleeping: we don't have anything that actually happens when the window opens, and this assignment is free (we have to rewrite the entire struct anyways)

actors/builtin/miner/miner_actor.go
whyrusleeping: Yeah, its set in the bitfieldToSectorNos method. Do you mean other than that one?

actors/builtin/miner/miner_state.go
anorth: Ok. Maybe comment that so future devs aren't confused like me.

actors/builtin/miner/policy.go
anorth: As a first step, I would be ok with moving these out to a structure and then having two versions of that structure, and a static to chose between them (not this PR)? It's a bit inflexible. There must be better ideas.

actors/builtin/miner/miner_actor.go
anorth: Ok, I understand now that because there is no cron callback at the start of the window we can't snapshot then, and must maintain proving set throughout the dead period between challenge windows. In that case, we must be careful to keep them in sync (I see you've correctly done the same thing at terminateSector).

actors/abi/sector.go
```diff
@@ -92,9 +92,14 @@ type PartialTicket []byte // 32 bytes
 type PoStVerifyInfo struct {
 	Randomness      PoStRandomness
 	CommR           cid.Cid
-	Candidates      []PoStCandidate      // From OnChain*PoStVerifyInfo
-	Proofs          []PoStProof          // From OnChain*PoStVerifyInfo
-	EligibleSectors map[SectorID]cid.Cid // TODO: HAMT?
+	Candidates      []PoStCandidate // From OnChain*PoStVerifyInfo
+	Proof           []byte
+	EligibleSectors []SectorInfo
```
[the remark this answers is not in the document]
laser: I observe that this PR replaces `Proofs []PoStProof` (effectively `[][]byte`) with `Proof []byte`. Is this an intentional change?

actors/abi/sector.go
```diff
@@ -92,9 +92,14 @@ type PartialTicket []byte // 32 bytes
...
+}
+
+type SectorInfo struct {
+	SectorID SectorNumber
+	CommR    cid.Cid
```
laser: @whyrusleeping - Suggestion:

- change `CommR` field name to `SealedCID` (to match elsewhere in spec)
- change `SectorID` field name to `SectorNumber` (to match elsewhere in spec)

actors/builtin/miner/miner_actor.go
zixuanzh: ProvingSet will be updated again (aka by removing the faulty sectors) when a Challenge window begins right?

anorth: @zixuanzh no I am not expecting that change. We're taking the cheap copy, will subtract faults on the fly when evaluating. 

But, I think the fault set needs to be frozen when the window opens. so maybe we'll need that cron callback after all. probably for a follow-up

actors/abi/sector.go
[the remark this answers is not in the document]
sternhenri: I'm gonna revert this for now since I agree that there is no reason for making this a single proof... There should be one per candidate

actors/builtin/miner/miner_actor.go
sternhenri: Added a call to power actor below

actors/builtin/miner/miner_actor.go
sternhenri: I'm not so sure @whyrusleeping. I agree we should check exactly what we passed in (but proofs code and randomness check below do that I think). That said, given check for appropriate number of tix, checking there are no duplicates is key. 

Duplicates here does not mean same sector, but same actual challenge. :)

sternhenri:
- Added back `Proofs` (plural) to `PoStVerifyInfo`
- Change names per laser rec
- Add call to power actor on verified `WindowedPoSt` to remove detected fault
- set `ProvingPeriodStart` on first sector proven, and add cronEvent for verifying expiration

TBD
- the todos, but will also need to remove cron on removal of last sector (silent fault if not so nbd -- same as a cron callback if miner has ben erased)

actors/builtin/miner/miner_actor.go
```diff
@@ -319,14 +319,26 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
+		// if first sector, set proving period start at next period
+		len, err := adt.AsArray(store, st.Sectors).Length()
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "failed to check miner sectors sizes: %v", err)
+		}
+		if len == 1 {
+			// enroll expiration check
+			a.enrollCronEvent(rt, st.PoStState.ProvingPeriodStart+power.WindowedPostChallengeDuration, &CronEventPayload{
+				EventType: CronEventType_Miner_PoStExpiration,
+			})
```
sternhenri: mistake. Fixed in #143

actors/builtin/miner/miner_actor.go
sternhenri: see #143

PR #149, opened 2020-02-11
add test for address key encoding

Alexey-N-Chernyshov:
There is a lack of encoding in the specification. Please, consider adding tests for actor state serialisation.

Please, see https://github.com/filecoin-project/specs-actors/pull/150 as corresponding improvement.

PR #150, opened 2020-02-11
Address serialisation as string

Alexey-N-Chernyshov:
Encode Address for HAMT-keys as string but not a byte-array representation. See https://github.com/filecoin-project/specs-actors/pull/149 for current implementation.

### DIFF
--- a/actors/util/adt/adt_test.go
+++ b/actors/util/adt/adt_test.go
@@ -1,1 +1,22 @@
 package adt_test
+
+import (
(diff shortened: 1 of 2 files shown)


anorth:
@whyrusleeping could you please make a call here? This will make the address keys larger, which may be a no-go.

whyrusleeping:
Yeah, no. We need to use the raw bytes as the keys. Theres no reason for this to be a string.

PR #152, opened 2020-02-11
Initialize Market State fields

ingar:
## Problem
`market.State` was initializing `EscrowTable`, `LockedTable` fields as arrays, even though they are indexed by `address.Address`.  Also, the `States` field was left uninitialized.

## Solution
Initialize EscrowTable, LockedTable as maps, States as array.

### DIFF
--- a/actors/builtin/market/market_state_test.go
+++ b/actors/builtin/market/market_state_test.go
@@ -14,17 +14,29 @@ import (
 )
 
 func TestConstruction(t *testing.T) {
-	rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+	builder := mock.NewBuilder(context.Background(), address.Undef)
 
 	t.Run("initializes all fields", func(t *testing.T) {
-		state, err := market.ConstructState(adt.AsStore(rt))
+		rt := builder.Build(t)
+		store := adt.AsStore(rt)
 
+		emptyMap, err := adt.MakeEmptyMap(store)
 		assert.NoError(t, err)
-		assert.True(t, state.Proposals.Defined())
-		assert.True(t, state.States.Defined())
-		assert.True(t, state.EscrowTable.Defined())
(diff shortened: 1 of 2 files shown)


actors/builtin/market/market_state_test.go
```diff
@@ -0,0 +1,30 @@
...
+	rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+
+	t.Run("initializes all fields", func(t *testing.T) {
+		state, err := market.ConstructState(adt.AsStore(rt))
+
+		assert.NoError(t, err)
+		assert.True(t, state.Proposals.Defined())
+		assert.True(t, state.States.Defined())
+		assert.True(t, state.EscrowTable.Defined())
+		assert.NotNil(t, state.LockedTable.Defined())
```
anorth: `assert.True`?

actors/builtin/market/market_state_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/builtin/market"
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+	"github.com/filecoin-project/specs-actors/support/mock"
+)
+
+func TestConstruction(t *testing.T) {
+	rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+
+	t.Run("initializes all fields", func(t *testing.T) {
```
frrist: This test would be more robust if it verified fields containing CID's were the expected value, e.g.
```go
emptyMap, err := adt.MakeEmptyMap(rt.Store())
if err != nil {
    return nil, err
}
assert.Equal(r, emptyMap, state.EscrowTable)
/* ... similar assertions for EmptyArrarys ...  */
```

PR #159, opened 2020-02-13
Fix BalanceTable AddCreate typo

ingar:
## Problem
`BalanceTable.AddCreate()` was loading the table value into the incorrect value.

## Solution
Write value into the correct place.

### DIFF
--- a/actors/util/adt/adt_test.go
+++ b/actors/util/adt/adt_test.go
@@ -1,14 +1,12 @@
 package adt_test
 
 import (
-	"context"
-	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/specs-actors/actors/abi"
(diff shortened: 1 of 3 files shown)


actors/util/adt/adt_test.go
```diff
@@ -20,3 +24,33 @@ func TestAddrKey(t *testing.T) {
 		assert.Equal(t, "\x02\xaa\xd0\xb2\x98\xa9\xde\xab\xbb\xb6\u007f\x80\x5f\x66\xaa\x68\x8c\xdd\x89\xad\xf5", adt.AddrKey(actor_address_2).Key())
 	})
 }
+
+func TestBalanceTable(t *testing.T) {
```
anorth: Please move to a new balancetable_test.go.

Issue #115, opened 2020-02-05
Replace use of SHA with blake2b


anorth:
See https://github.com/filecoin-project/specs/issues/862 for detail. I don't think actor code should be invoking SHA directly.

```
[~/p/g/s/g/f/specs-actors](anorth/dealreturn)$ rg -i '\bSHA'
...
actors/runtime/runtime.go
105:	// Hashes input data using SHA256.

support/mock/mockrt.go
60:	MhType:   mh.SHA2_256,

actors/crypto/randomness.go
44:	return abi.Randomness(SHA256(buffer))
51:	ranHash := SHA256(input)
64:func SHA256(data []byte) []byte {
```

sternhenri:
Closing in favor of #161

PR #161, opened 2020-02-13
Updates to randomness function

sternhenri:
In This PR:

- remove randomness from specs-actors altogether, include as runtime calls, move to specs
- removing spec artifacts for cleaner code
- switch from SHA to blake2b (per https://github.com/filecoin-project/specs-actors/issues/115, https://github.com/filecoin-project/specs/issues/862)

Implicitly, this also means changes to impl code as follows (will be expressed in specs PR):
- changing some interfaces to simplify randomness, namely
  - epoch number and miner address are added as CBOR-serialized entropy to the tickets
- ticket drawn from chain is VRF digest, not proof (so hash of BLS sig) per common VRF semantics
- concat of round number to ticket only happens as needed, not each time (so no need to do it for drawing)
- removal of ticket drawing DST (since that is simply the VRF digest operation)
- big endian reps everywhere (in keeping with CBOR)

Lmk if you want me to PR lotus accordingly @whyrusleeping once this is approved

### DIFF
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -161,7 +161,7 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 	}
 
 	if len(sv.SecretPreimage) > 0 {
-		if !bytes.Equal(rt.Syscalls().Hash_Blake2b(params.Secret), sv.SecretPreimage) {
+		if !bytes.Equal(rt.Syscalls().HashBlake2b(params.Secret), sv.SecretPreimage) {
 			rt.Abortf(exitcode.ErrIllegalArgument, "incorrect secret!")
 		}
 	}
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -161,7 +161,8 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 	}
 
 	if len(sv.SecretPreimage) > 0 {
-		if !bytes.Equal(rt.Syscalls().HashBlake2b(params.Secret), sv.SecretPreimage) {
+		hashedSecret := rt.Syscalls().HashBlake2b(params.Secret)
+		if !bytes.Equal(hashedSecret[:], sv.SecretPreimage) {
 			rt.Abortf(exitcode.ErrIllegalArgument, "incorrect secret!")
 		}
 	}
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -161,7 +161,7 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 	}
 
 	if len(sv.SecretPreimage) > 0 {
-		if !bytes.Equal(rt.Syscalls().Hash_SHA256(params.Secret), sv.SecretPreimage) {
+		if !bytes.Equal(rt.Syscalls().Hash_Blake2b(params.Secret), sv.SecretPreimage) {
 			rt.Abortf(exitcode.ErrIllegalArgument, "incorrect secret!")
 		}
 	}
--- a/actors/crypto/cbor_gen.go
+++ b/actors/crypto/cbor_gen.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 
-	abi "github.com/filecoin-project/specs-actors/actors/abi"
 	cbg "github.com/whyrusleeping/cbor-gen"
 	xerrors "golang.org/x/xerrors"
 )
@@ -103,85 +102,6 @@ func (t *Signature) UnmarshalCBOR(r io.Reader) error {
 	t.Data = make([]byte, extra)
 	if _, err := io.ReadFull(br, t.Data); err != nil {
 		return err
-	}
-	return nil
-}
-
-func (t *AddressEpochEntropy) MarshalCBOR(w io.Writer) error {
-	if t == nil {
-		_, err := w.Write(cbg.CborNull)
-		return err
-	}
-	if _, err := w.Write([]byte{130}); err != nil {
-		return err
-	}
-
-	// t.MinerAddress (address.Address) (struct)
-	if err := t.MinerAddress.MarshalCBOR(w); err != nil {
-		return err
-	}
-
-	// t.Epoch (abi.ChainEpoch) (int64)
-	if t.Epoch >= 0 {
-		if _, err := w.Write(cbg.CborEncodeMajorType(cbg.MajUnsignedInt, uint64(t.Epoch))); err != nil {
-			return err
-		}
-	} else {
-		if _, err := w.Write(cbg.CborEncodeMajorType(cbg.MajNegativeInt, uint64(-t.Epoch)-1)); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-func (t *AddressEpochEntropy) UnmarshalCBOR(r io.Reader) error {
-	br := cbg.GetPeeker(r)
-
-	maj, extra, err := cbg.CborReadHeader(br)
-	if err != nil {
-		return err
-	}
-	if maj != cbg.MajArray {
-		return fmt.Errorf("cbor input should be of type array")
-	}
-
-	if extra != 2 {
-		return fmt.Errorf("cbor input had wrong number of fields")
-	}
-
-	// t.MinerAddress (address.Address) (struct)
-
-	{
-
-		if err := t.MinerAddress.UnmarshalCBOR(br); err != nil {
-			return err
-		}
-
-	}
-	// t.Epoch (abi.ChainEpoch) (int64)
-	{
-		maj, extra, err := cbg.CborReadHeader(br)
-		var extraI int64
-		if err != nil {
-			return err
-		}
-		switch maj {
-		case cbg.MajUnsignedInt:
-			extraI = int64(extra)
-			if extraI < 0 {
-				return fmt.Errorf("int64 positive overflow")
-			}
-		case cbg.MajNegativeInt:
-			extraI = int64(extra)
-			if extraI < 0 {
-				return fmt.Errorf("int64 negative oveflow")
-			}
-			extraI = -1 - extraI
-		default:
-			return fmt.Errorf("wrong type for int64 field: %d", maj)
-		}
-
-		t.Epoch = abi.ChainEpoch(extraI)
 	}
 	return nil
 }
--- a/actors/crypto/cbor_gen.go
+++ b/actors/crypto/cbor_gen.go
@@ -105,3 +105,32 @@ func (t *Signature) UnmarshalCBOR(r io.Reader) error {
 	}
 	return nil
 }
+
+func (t *AddressEpochEntropy) MarshalCBOR(w io.Writer) error {
+	if t == nil {
+		_, err := w.Write(cbg.CborNull)
+		return err
+	}
+	if _, err := w.Write([]byte{128}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (t *AddressEpochEntropy) UnmarshalCBOR(r io.Reader) error {
+	br := cbg.GetPeeker(r)
+
+	maj, extra, err := cbg.CborReadHeader(br)
+	if err != nil {
+		return err
+	}
+	if maj != cbg.MajArray {
+		return fmt.Errorf("cbor input should be of type array")
+	}
+
+	if extra != 0 {
+		return fmt.Errorf("cbor input had wrong number of fields")
+	}
+
+	return nil
+}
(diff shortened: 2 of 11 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -533,9 +532,10 @@ func (a Actor) initiateNewSurprisePoStChallenges(rt Runtime) error {
 	rt.State().Transaction(&st, func() interface{} {
 		var err error
 		// sample the actor addresses
-		minerSelectionSeed := rt.GetRandomness(rt.CurrEpoch())
-		randomness := crypto.DeriveRandWithEpoch(crypto.DomainSeparationTag_SurprisePoStSelectMiners, minerSelectionSeed, int64(rt.CurrEpoch()))
+		// minerSelectionSeed := rt.DrawRandomnessFromChain(rt.CurrEpoch())
+		// randomness := crypto.DeriveRandWithMinerAddrAndEpoch(crypto.DomainSeparationTag_SurprisePoStSelectMiners, minerSelectionSeed, int64(rt.CurrEpoch()))
 
+		var randomness []byte
```
nikkolasg: So what's the goal here ?

actors/crypto/randomness.go
```diff
@@ -15,42 +16,56 @@ import (
...
 
-func _deriveRandInternal(tag DomainSeparationTag, randSeed abi.RandomnessSeed, index int, s []byte) abi.Randomness {
+func DeriveRandWithMinerAddrAndEpoch(tag DomainSeparationTag, randSeed []byte, minerAddr addr.Address, epoch abi.ChainEpoch) []byte {
+	entropy := &AddressEpochEntropy{
+		minerAddress: minerAddr,
+		epoch:        epoch,
+	}
+	_ = entropy
+	var entrBuf bytes.Buffer
+	err := entropy.MarshalCBOR(&entrBuf)
```
nikkolasg: ? Not sure what's happening here ?

actors/crypto/randomness.go
```diff
@@ -15,42 +16,56 @@ import (
...
+// TODO hs: remove once 148 lands
 // Computes an unpredictable integer less than limit from inputs seed and nonce.
-func RandomInt(seed abi.Randomness, nonce int64, limit int64) int64 {
+func RandomInt(seed []byte, nonce int64, limit int64) int64 {
 	nonceBytes := BigEndianBytesFromInt(nonce)
 	input := append(seed, nonceBytes...)
-	ranHash := SHA256(input)
-	hashInt := big.PositiveFromUnsignedBytes(ranHash)
+	ranHash := blake2b.Sum256(input)
+	hashInt := big.PositiveFromUnsignedBytes(ranHash[:])
```
nikkolasg: I can't find a definition of this method. Can you point it to me somewhere ?

actors/builtin/power/power_actor.go
sternhenri: Sorry this whole section is getting ripped out in https://github.com/filecoin-project/specs-actors/pull/148, so you can ignore all of this

actors/crypto/randomness.go
sternhenri: fixed. ty

actors/crypto/randomness.go
sternhenri: https://github.com/filecoin-project/specs-actors/blob/master/actors/abi/big/int.go#L29

actors/crypto/cbor_gen.go
```diff
@@ -105,3 +105,32 @@ func (t *Signature) UnmarshalCBOR(r io.Reader) error {
 	}
 	return nil
 }
+
+func (t *AddressEpochEntropy) MarshalCBOR(w io.Writer) error {
+	if t == nil {
+		_, err := w.Write(cbg.CborNull)
+		return err
+	}
+	if _, err := w.Write([]byte{128}); err != nil {
```
ZenGround0: Im a bit new to cbg but it looks like this just writes 0x80 and calls it a day without writing fields.

actors/crypto/cbor_gen.go
```diff
@@ -105,3 +105,32 @@ func (t *Signature) UnmarshalCBOR(r io.Reader) error {
...
+
+	maj, extra, err := cbg.CborReadHeader(br)
+	if err != nil {
+		return err
+	}
+	if maj != cbg.MajArray {
+		return fmt.Errorf("cbor input should be of type array")
+	}
+
+	if extra != 0 {
```
ZenGround0: Yeah similarly it looks like this is not reading and unmarshaling anything.

actors/crypto/randomness.go
```diff
@@ -15,42 +16,56 @@ import (
...
 	autil.AssertNoError(err)
 
-	return _deriveRandInternal(tag, tix, -1, addrBuf.Bytes())
+	return _deriveRandInternal(tag, randSeed, addrBuf.Bytes())
 }
 
-func DeriveRandWithEpoch(tag DomainSeparationTag, tix abi.RandomnessSeed, epoch int64) abi.Randomness {
-	return _deriveRandInternal(tag, tix, -1, BigEndianBytesFromInt(epoch))
+type AddressEpochEntropy struct {
+	minerAddress addr.Address // Must be an ID-addr
```
ZenGround0: If I was a gambling man I would wager that these fields should be capitalized so that cbg will actually read and write them

actors/runtime/runtime.go
```diff
@@ -33,8 +33,8 @@ type Runtime interface {
 	// Look up the code ID at an actor address.
 	GetActorCodeCID(addr addr.Address) (ret cid.Cid, ok bool)
 
-	// Randomness returns a (pseudo)random string for the given epoch and tag.
-	GetRandomness(epoch abi.ChainEpoch) abi.RandomnessSeed
+	// Randomness returns a (pseudo)random byte array from the chain at a given epoch
+	DrawRandomnessFromChain(epoch abi.ChainEpoch) []byte
```
ZenGround0: As discussed offline, it makes sense from my perspective for runtime the runtime endpoint to contain all the necessary context for creating the randomness (i.e. DrawRandomness with miner addr and epoch, DrawRandomness with just epoch, DrawRandomness).  Then all the utilities for creating that randomness can sit outside within the node not the vm code.

Doesn't need to happen in this PR.

ZenGround0: If you agree with the above it would make sense to specify how to do randomness drawing in the needed cases only in the text part of the spec, removing the randomness utility here.

Not sure if that blocks existing intermediate assumptions /testing capabilities in sp-act (i'm new here).

actors/builtin/paych/paych_actor.go
```diff
@@ -161,7 +161,7 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 	}
 
 	if len(sv.SecretPreimage) > 0 {
-		if !bytes.Equal(rt.Syscalls().Hash_SHA256(params.Secret), sv.SecretPreimage) {
+		if !bytes.Equal(rt.Syscalls().Hash_Blake2b(params.Secret), sv.SecretPreimage) {
```
anorth: While renaming it, please use canonical Go style (the old one caused linter problems for us!).

`HashBlake2b(b)`, or, if you like, `Hash(enum, b)` defining a const for blake2b hash algo.

actors/crypto/randomness.go
```diff
@@ -15,42 +16,55 @@ import (
...
 	autil.AssertNoError(err)
 
-	return _deriveRandInternal(tag, tix, -1, addrBuf.Bytes())
+	return _deriveRandInternal(tag, randSeed, addrBuf.Bytes())
 }
 
-func DeriveRandWithEpoch(tag DomainSeparationTag, tix abi.RandomnessSeed, epoch int64) abi.Randomness {
-	return _deriveRandInternal(tag, tix, -1, BigEndianBytesFromInt(epoch))
+type AddressEpochEntropy struct {
+	MinerAddress addr.Address // Must be an ID-addr
```
anorth: We could extract the ID from the address prior to here, preventing the possibility of error. Other "must be an ID-addr" comments in the code are documenting preconditions for external callers.

actors/crypto/randomness.go
```diff
@@ -15,42 +16,55 @@ import (
...
+
+func _deriveRandInternal(tag DomainSeparationTag, randSeed []byte, serializedEntropy []byte) []byte {
 	buffer := []byte{}
 	buffer = append(buffer, BigEndianBytesFromInt(int64(tag))...)
-	buffer = append(buffer, BigEndianBytesFromInt(int64(index))...)
 	buffer = append(buffer, randSeed...)
-	buffer = append(buffer, s...)
-	return SHA256(buffer)
+	buffer = append(buffer, serializedEntropy...)
+	bufHash := blake2b.Sum256(buffer)
```
anorth: Blocking: the whole point of the hash method on the runtime is to avoid importing a specific hashing implementation here. You need to plumb the runtime's hash method through to here.

actors/runtime/runtime.go
anorth: I have previously expressed the opposite point of view, which is that the runtime should be the thinnest interface we can define, and pulling code that can be defined in terms of those primitives into the actor spec is a desirable outcome. For one, it's a lot easier to test the actors than the VMs (and will remain true when there are many impls of each).

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/ipfs/go-hamt-ipld v0.0.15-0.20200131012125-dd88a59d3f2e
 	github.com/ipfs/go-ipld-cbor v0.0.4
 	github.com/libp2p/go-libp2p-core v0.3.0
+	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
```
anorth: Please revert.

actors/runtime/runtime.go
anorth: I am confused by the removal of the `RandomnessSeed` type but still having a `Randomness` type, and the `DrawRandomness` method not returning a value of that type. Also, reference to the blockchain is intentionally minimised inside the VM (everything inside should roughly make sense if there was some other consensus distributed database mechanism).

So how about `DrawRandomSeed(epoch) abi.RandomSeed` (unless I've misinterpreted)?

actors/crypto/randomness.go
```diff
@@ -15,42 +16,55 @@ import (
...
+	DomainSeparationTag_TicketProduction DomainSeparationTag = 1 + iota
 	DomainSeparationTag_ElectionPoStChallengeSeed
 	DomainSeparationTag_SurprisePoStChallengeSeed
 	DomainSeparationTag_SurprisePoStSelectMiners
 	DomainSeparationTag_SurprisePoStSampleSectors
 )
 
 // Derive a random byte string from a domain separation tag and the appropriate values
-func DeriveRandWithMinerAddr(tag DomainSeparationTag, tix abi.RandomnessSeed, minerAddr addr.Address) abi.Randomness {
+func DeriveRandWithMinerAddr(tag DomainSeparationTag, randSeed []byte, minerAddr addr.Address) []byte {
```
anorth: `abi.Randomness` still exists, why doesn't this return one? I think there is some legit type safety on offer here with those types.

actors/crypto/randomness.go
sternhenri: It's a point well taken. Makes me think @ZenGround0's approach above of just abstracting this to sysCall and having this be node code makes sense, given these functions will also need to be called outside the VM (as part of generating a new block, ticket etc)... It needs to run the same way inside or outside the VM...

What do you think?

actors/runtime/runtime.go
sternhenri: Yeah I think the `randomnessSeed` vs `Randomness` distinction is confusing so I'm going to move to using Randomness everywhere.

actors/crypto/randomness.go
sternhenri: Agree, but I'd rather do that outside the randomness lib (ie with the external caller) rather than pass in the `initState` to the lib to do it here. That seems like a separate concern.

actors/crypto/randomness.go
```diff
@@ -1,70 +1,24 @@
...
-	autil.AssertNoError(err)
-	return buf.Bytes()
-}
-
-func SHA256(data []byte) []byte {
-	autil.TODO()
-	return []byte{}
+type AddressEpochEntropy struct {
+	MinerAddress addr.Address // Must be an ID-addr
+	Epoch        abi.ChainEpoch
```
anorth: Remove this now-unused structure.

actors/runtime/runtime.go
```diff
@@ -105,8 +106,8 @@ type Message interface {
 type Syscalls interface {
 	// Verifies that a signature is valid for an address and plaintext.
 	VerifySignature(signature crypto.Signature, signer addr.Address, plaintext []byte) bool
-	// Hashes input data using SHA256.
-	Hash_SHA256(data []byte) []byte
+	// Hashes input data using blake2b with 256 bit output.
+	HashBlake2b(data []byte) []byte
```
anorth: Make this return `[8]byte` since we can.

PR #164, opened 2020-02-14
feat: add methods to check if an actor is builtin or singleton

icorderi:


actors/builtin/singletons.go
```diff
@@ -25,3 +26,13 @@ func mustMakeAddress(id uint64) addr.Address {
 	autil.AssertNoError(err)
 	return address
 }
+
+// IsSingletonActor returns true if the code belongs to a singleton actor.
+func IsSingletonActor(code cid.Cid) bool {
+	return code.Equals(CronActorCodeID) ||
+		code.Equals(InitActorCodeID) ||
+		code.Equals(RewardActorCodeID) ||
+		code.Equals(CronActorCodeID) ||
```
anorth: Cron is here twice.

PR #166, opened 2020-02-14
Check challengeCount < MinerCount in PowerActor. selectMinersToSurprise()

Alexey-N-Chernyshov:
It is an infinite loop when requested count larger than miner count.


anorth:
Thanks, but this code is being deleted in #148

PR #168, opened 2020-02-14
rework minerNominalPowerMeetsConsensusMinimum

Alexey-N-Chernyshov:
Unfortunately, there is no test to prove it works correctly.

### DIFF
--- a/actors/builtin/power/power_state.go
+++ b/actors/builtin/power/power_state.go
@@ -2,7 +2,6 @@ package power
 
 import (
 	"reflect"
-	"sort"
 
 	addr "github.com/filecoin-project/go-address"
 	cid "github.com/ipfs/go-cid"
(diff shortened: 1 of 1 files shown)


actors/builtin/power/power_state.go
```diff
@@ -98,15 +97,15 @@ func (st *State) minerNominalPowerMeetsConsensusMinimum(s adt.Store, minerPower
...
+		}
 		return nil
 	}); err != nil {
 		return false, errors.Wrap(err, "failed to iterate power table")
 	}
 
-	// get size of MIN_MINER_SIZE_TARGth largest miner
-	sort.Slice(minerSizes, func(i, j int) bool { return i > j })
-	return minerPower.GreaterThanEqual(minerSizes[ConsensusMinerMinMiners-1]), nil
+	return minerPower.GreaterThanEqual(maxPower), nil
```
anorth: This appears to change the semantics. The (poorly documented) goal is to return true for the top `ConsensusMinerMinMiners` miners.

PR #171, opened 2020-02-15
More payment channel Actor tests

shannonwells:
Remaining tests for payment channel Actor:
* More tests for UpdateChannelState, including lane Merges and Extra
* Collect
* Settle

Fix a copy-paste bug in Collect.

### DIFF
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -284,7 +284,6 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	}
 
 	// send remaining balance to "From"
-
 	_, codeFrom := rt.Send(
 		st.From,
 		builtin.MethodSend,
@@ -294,9 +293,8 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	builtin.RequireSuccess(rt, codeFrom, "Failed to send balance to `From`")
 
 	// send ToSend to "To"
-
 	_, codeTo := rt.Send(
-		st.From,
+		st.To,
 		builtin.MethodSend,
 		nil,
 		abi.NewTokenAmount(st.ToSend.Int64()),
--- a/actors/builtin/paych/paych_test.go
+++ b/actors/builtin/paych/paych_test.go
@@ -115,6 +115,8 @@ func TestPaymentChannelActor_CreateLaneSuccess(t *testing.T) {
 		rt := builder.Build(t)
 		actor.constructAndVerify(rt, payerAddr, newPaychAddr)
 		amt := big.NewInt(10)
+		lane := uint64(999)
+		nonce := uint64(1)
 		sig := &crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte("doesn't matter")}
 		tl := abi.ChainEpoch(1)
 		sv := SignedVoucher{TimeLock: tl, Lane: lane, Nonce: nonce, Amount: amt, Signature: sig}
--- a/actors/builtin/paych/paych_test.go
+++ b/actors/builtin/paych/paych_test.go
@@ -165,7 +165,7 @@ func TestPaymentChannelActor_CreateLaneFailure(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
 			versig := func(sig crypto.Signature, signer addr.Address, plaintext []byte) bool { return tc.verifySig }
-			hasher := func(data []byte) []byte { return data }
+			hasher := func(data []byte) [8]byte { return [8]byte{} }
 
 			builder := mock.NewBuilder(ctx, paychAddr).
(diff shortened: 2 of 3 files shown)


actors/builtin/paych/paych_actor.go
```diff
@@ -295,9 +294,8 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	builtin.RequireSuccess(rt, codeFrom, "Failed to send balance to `From`")
 
 	// send ToSend to "To"
-
 	_, codeTo := rt.Send(
-		st.From,
+		st.To,
```
shannonwells: copy-paste bug

actors/builtin/paych/paych_test.go
```diff
@@ -23,156 +25,705 @@ func TestPaymentChannelActor_Constructor(t *testing.T) {
...
 		}, {"fails if sender (from) is not account actor",
-			newPaychAddr,
+			paychAddr,
 			builtin.InitActorCodeID,
 			builtin.MultisigActorCodeID,
 			builtin.AccountActorCodeID,
 			exitcode.ErrIllegalArgument,
 		}, {"fails if addr is not ID type",
-			tutil.NewActorAddr(t, "beach blanket babylon"),
+			tutil.NewSECP256K1Addr(t, "beach blanket babylon"),
```
shannonwells: some corrections to tests that were producing false positives.

actors/builtin/paych/paych_test.go
```diff
@@ -23,156 +25,705 @@ func TestPaymentChannelActor_Constructor(t *testing.T) {
...
+
+			rt.SetCaller(st1.From, builtin.AccountActorCodeID)
+			rt.ExpectValidateCallerAddr(st1.From, st1.To)
+			rt.ExpectAbort(tc.expExitCode, func() {
+				rt.Call(actor.UpdateChannelState, ucp)
+			})
+
+		})
+	}
+	t.Run("When lane doesn't exist, fails with: voucher specifies invalid merge lane 999", func(t *testing.T) {
```
shannonwells: I've changed some of the descriptions to reflect the expected reason for failure.  This helped me identify some false positives.

actors/builtin/paych/paych_test.go
```diff
@@ -23,158 +25,707 @@ func TestPaymentChannelActor_Constructor(t *testing.T) {
...
-	versig := func(sig crypto.Signature, signer addr.Address, plaintext []byte) bool { return true }
-	hasher := func(data []byte) [8]byte {
-		return [8]byte{}
+	sig := &crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte("doesn't matter")}
+
+	testCases := []struct {
+		desc       string
+		targetCode cid.Cid
+
+		balance  int64
```
anorth: `balance` looks unused.

Please factor out the `abi.NewTokenAmount(9)` below so it's encountered before the test cases (or use it here).

actors/builtin/paych/paych_test.go
```diff
@@ -23,158 +25,707 @@ func TestPaymentChannelActor_Constructor(t *testing.T) {
...
+			amt: 1, epoch: 1, tl: 10,
+			sig: sig, verifySig: true,
+			expExitCode: exitcode.ErrIllegalArgument},
+		{desc: "fails if signature not verified", targetCode: builtin.AccountActorCodeID,
+			amt: 1, epoch: 1, tl: 1, sig: sig, verifySig: false,
+			expExitCode: exitcode.ErrIllegalArgument},
+		{desc: "fails if SigningBytes fails", targetCode: builtin.AccountActorCodeID,
+			amt: 1, epoch: 1, tl: 1, sig: sig, verifySig: true,
+			secretPreimage: make([]byte, 2<<21),
+			expExitCode:    exitcode.ErrIllegalArgument},
```
anorth: These are great, but without a success case it's hard to be confident that there isn't something that causes all these to fail that's not the thing actually being exercised (tho I have no doubt you verified in a debugger). Can you add a success case so it's clear that each of these other cases differs from that by only one factor?

Can possibly replace TestPaymentChannelActor_CreateLaneSuccess, tho not necessary.

support/mock/mockrt.go
```diff
@@ -400,6 +400,11 @@ func (rt *Runtime) SetBalance(amt abi.TokenAmount) {
 	rt.balance = amt
 }
 
+// Get balance outside of a runtime call for test purposes
+func (rt *Runtime) GetBalance() abi.TokenAmount {
+	return rt.balance
+}
```
anorth: I don't think you need this: just call CurrentBalance()

shannonwells: This was to avoid using a constant or scattering bare ints around, to know the balance we created down in requireCreateWithLanes. `CurrentBalance` fails in such a context, b/c it calls `requireInCall`. I could remove that requirement but all of the mocked API calls have it, so I hesitated to change it.  I could simply return the balance used I suppose.... I'll do that instead.

anorth: Ah, ok. GetBalance is fine.

Issue #126, opened 2020-02-06
Syscalls should return an error instead of bool


hannahhoward:
I noticed SysCalls.VerifySignature returns a boolean.

Existing code we've been working with for signatures returns an error.
Generally an error can communicate more information about reasons for failure than a boolean.

Maybe this is worth exploring?

anorth:
What kind of error information is possible to propagate?

hannahhoward:
Truthfully, I'm not totally sure but here is some verification code I'm aware of -- it certainly would appear a lot of information is communicated. Alternatively, this may be more relevant in the markets context but I was hoping to match the call signatures.

```golang
func (s *Signature) Verify(addr address.Address, msg []byte) error {
       if addr.Protocol() == address.ID {
               return fmt.Errorf("must resolve ID addresses before using them to verify a signature")
       }
       b2sum := blake2b.Sum256(msg)

       switch s.Type {
       case KTSecp256k1:
               pubk, err := crypto.EcRecover(b2sum[:], s.Data)
               if err != nil {
                       return err
               }

               maybeaddr, err := address.NewSecp256k1Address(pubk)
               if err != nil {
                       return err
               }

               if addr != maybeaddr {
                       return fmt.Errorf("signature did not match")
               }

               return nil
       case KTBLS:
               digests := []bls.Digest{bls.Hash(bls.Message(msg))}

               var pubk bls.PublicKey
               copy(pubk[:], addr.Payload())
               pubkeys := []bls.PublicKey{pubk}

               var sig bls.Signature
               copy(sig[:], s.Data)

               if !bls.Verify(&sig, digests, pubkeys) {
                       return fmt.Errorf("bls signature failed to verify")
               }

               return nil
       default:
               return fmt.Errorf("cannot verify signature of unsupported type: %s", s.Type)
       }
}
```

anorth:
Ok, I can see some value there. Thanks for the suggestion.

Kubuxu:
This is the current code in lotus: https://github.com/filecoin-project/lotus/blob/master/lib/sigs/secp/init.go#L37-L58

anorth:
This should go for all the syscalls that return bools

anorth:
Done

PR #174, opened 2020-02-18
Working on lotus integration

whyrusleeping:
Fixing a number of bugs, and a few interface changes (syscalls don't have the ability to call abort, so must return errors).

Not entirely done yet, but I guess its one step closer, so maybe we just merge it anyways?


actors/builtin/market/market_actor.go
```diff
@@ -92,16 +93,16 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 }
 
 // Deposits the received value into the balance held in escrow.
-func (a Actor) AddBalance(rt Runtime, providerOrClientAdrress *addr.Address) *adt.EmptyValue {
-	escrowAddress(rt, *providerOrClientAdrress)
+func (a Actor) AddBalance(rt Runtime, providerOrClientAddress *addr.Address) *adt.EmptyValue {
+	btabAddr := escrowAddress(rt, *providerOrClientAddress)
```
anorth: This changes behaviour in a way you probably don't want (see https://filecoinproject.slack.com/archives/CHMNDCK9P/p1581987982474800)

This logic is obscure. It may be better to duplicate it between the power and miner actors than attempt the partial code-sharing.

actors/builtin/market/market_actor.go
```diff
@@ -160,7 +172,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			amountSlashedTotal = big.Add(amountSlashedTotal, st.updatePendingDealStatesForParty(rt, deal.Proposal.Provider))
 
 			st.lockBalanceOrAbort(rt, deal.Proposal.Client, deal.Proposal.ClientBalanceRequirement())
-			st.lockBalanceOrAbort(rt, deal.Proposal.Provider, deal.Proposal.ProviderBalanceRequirement())
+			st.lockBalanceOrAbort(rt, owner, deal.Proposal.ProviderBalanceRequirement())
```
anorth: This looks wrong: the collateral should be on balance for for the miner actor, not owner. It _pays out_ to the owner, which is computed at withdrawal time.

actors/builtin/market/market_actor.go
```diff
@@ -138,14 +139,25 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 	// This allows us to retain and verify only the client's signature in each deal proposal itself.
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 
+	// resolve worker keys for providers in deals
+	provkeys := resolveControlKeys(rt, params.Deals)
```
anorth: Are you explicitly trying to support the batch publication of deals for different miners but the same worker? I think that's more complicated than necessary. Without that, the deals must all have the same provider and hence same worker, so we can do a simpler check that the first deal provider has caller == worker, and that all the others match. Only one `RequestMinerControlAddrs` required.

actors/builtin/market/market_state.go
```diff
@@ -3,6 +3,7 @@ package market
 import (
 	addr "github.com/filecoin-project/go-address"
 	"github.com/ipfs/go-cid"
+	xerrors "golang.org/x/xerrors"
```
anorth: Please use pkg/errors for consistency with the rest of this module.

actors/builtin/market/market_state.go
```diff
@@ -383,10 +380,10 @@ func dealProposalIsInternallyValid(rt Runtime, proposal ClientDealProposal) bool
 	// authenticity of the on-chain message publishing the deal.
 	sigVerified := rt.Syscalls().VerifySignature(proposal.ClientSignature, proposal.Proposal.Client, m)
 	if !sigVerified {
-		return false
+		return xerrors.Errorf("signature proposal invalid")
```
anorth: FYI see https://github.com/filecoin-project/specs-actors/issues/126. I concur syscalls should return errors.

actors/builtin/miner/miner_state.go
```diff
@@ -4,9 +4,11 @@ import (
 	"reflect"
 
 	addr "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-amt-ipld/v2"
 	cid "github.com/ipfs/go-cid"
 	peer "github.com/libp2p/go-libp2p-core/peer"
 	errors "github.com/pkg/errors"
+	xerrors "golang.org/x/xerrors"
```
anorth: pkg/errors

actors/builtin/miner/miner_state.go
```diff
@@ -156,7 +159,12 @@ func (st *State) HasSectorNo(store adt.Store, sectorNo abi.SectorNumber) (bool,
 	var info SectorOnChainInfo
 	found, err := sectors.Get(uint64(sectorNo), &info)
 	if err != nil {
-		return false, errors.Wrapf(err, "failed to get sector %v", sectorNo)
+		var enf *amt.ErrNotFound
+		if xerrors.As(err, &enf) {
+			return false, nil
+		}
```
anorth: This looks messed up. `Array.Get` already returns a boolean indicating found, checking for the same error type. If there's something wrong with that check, can we resolve it lower down?

actors/runtime/runtime.go
```diff
@@ -107,13 +107,13 @@ type Syscalls interface {
...
-	HashBlake2b(data []byte) [8]byte
+	HashBlake2b(data []byte) [32]byte
 	// Computes an unsealed sector CID (CommD) from its constituent piece CIDs (CommPs) and sizes.
 	ComputeUnsealedSectorCID(sectorSize abi.SectorSize, pieces []abi.PieceInfo) (cid.Cid, error)
 	// Verifies a sector seal proof.
-	VerifySeal(sectorSize abi.SectorSize, vi abi.SealVerifyInfo) bool
+	VerifySeal(sectorSize abi.SectorSize, vi abi.SealVerifyInfo) (bool, error)
 	// Verifies a proof of spacetime.
-	VerifyPoSt(sectorSize abi.SectorSize, vi abi.PoStVerifyInfo) bool
+	VerifyPoSt(sectorSize abi.SectorSize, vi abi.PoStVerifyInfo) (bool, error)
```
anorth: Suggest we drop the boolean return value here and use only the error.

actors/util/adt/array.go
```diff
@@ -116,14 +117,14 @@ func (a *Array) write(root *amt.Root) error {
 func (a *Array) Get(k uint64, out runtime.CBORUnmarshaler) (bool, error) {
 	root, err := amt.LoadAMT(a.store.Context(), a.store, a.root)
 	if err != nil {
-		return false, errors.Wrapf(err, "array get failed to load root %v", a.root)
+		return false, xerrors.Errorf("array get failed to load root %v: %w", a.root, err)
```
anorth: Maybe we need to talk about errors. We talked about it a few weeks ago but you keep (inadvertently?) making it inconsistent. What are you trying to achieve here that the prior state doesn't?

go.mod
```diff
@@ -3,7 +3,8 @@ module github.com/filecoin-project/specs-actors
 go 1.13
 
 require (
-	github.com/filecoin-project/go-address v0.0.0-20200107215422-da8eea2842b5
+	github.com/filecoin-project/go-address v0.0.2-0.20200218010043-eb9bb40ed5be
+	github.com/filecoin-project/go-amt-ipld v1.0.0
```
anorth: This is probably an error, please revert.

actors/builtin/market/market_actor.go
anorth: Edit: it looks like I already split the power actor's use of similar concepts, so these are specific to the market.

actors/builtin/market/market_actor.go
whyrusleeping: Yeah, I think I had it in my head that we allowed clients to submit deals, which would imply you might have a bunch of different miners that the client was submitting a batch of deals for. We don't allow that right now it seems (though the restriction is hidden deep down a few methods away from here). I can definitely simplify this

actors/builtin/market/market_state.go
whyrusleeping: pkg/errors loses a lot of information, and doesnt conform to go's new error handling mechanics. Strongly prefer xerrors (or simply fmt.Errorf, which has the same api, but doesnt preserve stack information)

go.mod
whyrusleeping: Yep, the wrong amt package snuck in through goimports trying to be smart.

actors/util/adt/array.go
whyrusleeping: Take a look at https://blog.golang.org/go1.13-errors

Pkg/errors is not canonical go, and the package is being deprecated: https://github.com/pkg/errors#roadmap

We can either use fmt.Errorf with the %w verb, or use xerrors.Errorf with the %w verb. We tend to prefer xerrors as it has the exact api, but includes stack information that is super helpful for debugging.

actors/builtin/miner/miner_state.go
whyrusleeping: Yeah, it wasnt happy. Let's see if I can sink it deeper, could have been the wrong solution to the problem

actors/util/adt/array.go
anorth: Thanks for the links. I wish we could have had this information when we first talked about it before @frrist and I standardised this repo on pkg/errors!

FWIW I have only heard negative things from other Go devs about the new errors. We don't *have* to follow. But so long as we do something consistent so that new developers have a simple pattern to follow, I don't mind too much.

whyrusleeping:
@anorth Cleaned up the parts you mentioned.

@magik6k we're probably good to merge this now, yeah?

actors/builtin/miner/miner_actor.go
```diff
@@ -131,7 +131,8 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 	})
```
ZenGround0: ControlAddresses method needs a call to `rt.ValidateImmediateCallerAcceptAny()` on line 103 to satisfy the caller verification invariant required by spec.  Determined while using this branch to integrate go-filecoin.

actors/abi/bitfield.go
```diff
@@ -202,7 +202,12 @@ func (bf BitField) AllMap(max uint64) (map[uint64]bool, error) {
 	return out, nil
 }
 
-func (bf BitField) MarshalCBOR(w io.Writer) error {
+func (bf *BitField) MarshalCBOR(w io.Writer) error {
+	if bf == nil {
+		_, err := w.Write(cbg.CborNull)
+		return err
+	}
+
```
anorth: @Kubuxu please propagate this fix to https://github.com/filecoin-project/go-bitfield

actors/builtin/init/init_actor_state.go
```diff
@@ -55,7 +57,7 @@ func (s *State) ResolveAddress(store adt.Store, address addr.Address) (addr.Addr
 	}
 
 	// Not found.
-	return addr.Undef, errors.New("address not found")
+	return addr.Undef, ErrAddressNotFound
```
anorth: If anything needs to actually inspect this I would much prefer we return a boolean indicating found or not.

actors/builtin/market/market_actor.go
anorth: This still needs cleanup (esp for consistency with the comment at 150).

actors/builtin/market/market_actor.go
anorth: Still needs fixing.

actors/builtin/miner/miner_actor.go
```diff
@@ -131,7 +131,8 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
 
 	cronPayload := CronEventPayload{
-		EventType: cronEventWorkerKeyChange,
+		EventType: CronEventWorkerKeyChange,
+		Sectors:   nil,
```
anorth: Unnecessary?

actors/builtin/market/market_actor.go
whyrusleeping: I have updated this to use the code on master

actors/builtin/miner/miner_actor.go
ZenGround0: orthogonal so put it up here: https://github.com/filecoin-project/specs-actors/pull/202

whyrusleeping: thats.... dumb

actors/builtin/market/market_actor.go
```diff
@@ -156,22 +157,30 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 		proposals := AsDealProposalArray(adt.AsStore(rt), st.Proposals)
 		dbp := AsSetMultimap(adt.AsStore(rt), st.DealIDsByParty)
 		// All storage proposals will be added in an atomic transaction; this operation will be unrolled if any of them fails.
+
+		_, worker := builtin.RequestMinerControlAddrs(rt, params.Deals[0].Proposal.Provider)
```
anorth: Check that deals is non-empty before here (your call whether to succeed vacuously or abort).

actors/builtin/market/market_actor.go
```diff
@@ -156,22 +157,30 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+
 		for _, deal := range params.Deals {
-			if deal.Proposal.Provider != rt.Message().Caller() {
+
+			if worker != rt.Message().Caller() {
```
anorth: Pull this check out of the loop over deals, but add a check that the deal provider is the same as the first deal's provider.

Note: we don't resolve the provider address yet, but maybe should? It's the actor's address, so could conceivably be the reorg-stable actor address instead of ID address. We should either resolve it or assert it's an ID address (the former is probably more helpful).

If you resolve it `s/provider/resolved provider/g` in the first sentence of this comment.

actors/builtin/market/market_actor.go
```diff
@@ -156,22 +157,30 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
 			//
 			// Note: as an optimization, implementations may cache efficient data structures indicating
 			// which of the following set of updates are redundant and can be skipped.
-			amountSlashedTotal = big.Add(amountSlashedTotal, st.updatePendingDealStatesForParty(rt, deal.Proposal.Client))
+			amountSlashedTotal = big.Add(amountSlashedTotal, st.updatePendingDealStatesForParty(rt, client))
 			amountSlashedTotal = big.Add(amountSlashedTotal, st.updatePendingDealStatesForParty(rt, deal.Proposal.Provider))
 
-			st.lockBalanceOrAbort(rt, deal.Proposal.Client, deal.Proposal.ClientBalanceRequirement())
+			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientBalanceRequirement())
 			st.lockBalanceOrAbort(rt, deal.Proposal.Provider, deal.Proposal.ProviderBalanceRequirement())
```
anorth: Use the resolved provider address here too.

actors/builtin/market/market_actor.go
```diff
@@ -181,7 +190,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 				rt.Abortf(exitcode.ErrIllegalState, "set deal: %v", err)
 			}
 
-			if err = dbp.Put(deal.Proposal.Client, id); err != nil {
+			if err = dbp.Put(client, id); err != nil {
 				rt.Abortf(exitcode.ErrIllegalState, "set client deal id: %v", err)
 			}
 			if err = dbp.Put(deal.Proposal.Provider, id); err != nil {
```
anorth: Resolved provider here too

actors/builtin/market/market_actor.go
```diff
@@ -200,6 +209,22 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+		}
+
+		owner, worker := builtin.RequestMinerControlAddrs(rt, d.Proposal.Provider)
+
+		out[d.Proposal.Provider] = []address.Address{owner, worker}
+	}
+
+	return out
+}
+
```
anorth: Delete this, now unused.

actors/builtin/miner/miner_actor.go
```diff
@@ -282,13 +282,23 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 		rt.Abortf(exitcode.ErrNotFound, "no precommitted sector %v", sectorNo)
 	}
 
+	if rt.CurrEpoch() > precommit.PreCommitEpoch+MaxSealDuration[precommit.Info.RegisteredProof] || rt.CurrEpoch() < precommit.PreCommitEpoch+PreCommitChallengeDelay {
+		rt.Abortf(exitcode.ErrIllegalArgument, "Invalid ProveCommitSector epoch (cur=%d, pcepoch=%d)", rt.CurrEpoch(), precommit.PreCommitEpoch)
+	}
```
anorth: `verifySeal`, called immediately below here, has another epoch check. Let's move them next to each other so the constraints are clear.

actors/builtin/miner/miner_actor.go
```diff
@@ -282,13 +282,23 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
+
+	TODO()
+	// TODO HS: How are SealEpoch, InteractiveEpoch determined (and intended to be used)?
+	// Presumably they cannot be derived from the SectorProveCommitInfo provided by an untrusted party.
+
+	// will abort if seal invalid
```
anorth: I think there's a rebase artefact here, HS completed and removed this TODO on master.

actors/builtin/miner/policy.go
```diff
@@ -25,7 +25,7 @@ const PoStLookback = abi.ChainEpoch(1) // PARAM_FINISH
 // Lookback from the current epoch for state view for elections; for Election PoSt, same as the PoSt lookback.
 const ElectionLookback = PoStLookback // PARAM_FINISH
 
-// Number of sectors to be sampled as part of surprise PoSt
+// Number of sectors to be sampled as part of fallback PoSt
```
magik6k:
```suggestion
// Number of sectors to be sampled as part of windowed PoSt
```

actors/builtin/market/market_actor.go
whyrusleeping: Eh, I guess. Since we're using panics for everything we would just catch this, but yeah. cleaner to explicitly check.

anorth: Actor code itself should *never* panic.

PR #180, opened 2020-02-19
Add address resolution call to runtime; resolve miner worker and power/market account-keeping addresses.

anorth:
This makes the miner worker address much more ergonomic to set, avoiding the need for either yanking the account's ID address out of the state, or waiting for finality for it to be safe to use.

The miner and power escrow keys were very easy to mis-use beforehand. This makes it clear that the address under which all balances are held is the canonical ID address of the miner or client.

An alternative to this approach would be to send a message to the Init actor. This seems like a reasonable facility to special-case, with less actual execution cost as well as less gas to be paid, since we can expect this table to be cached.

FYI @icorderi

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -71,10 +71,10 @@ type ConstructorParams = power.MinerConstructorParams
 func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.InitActorAddr)
 
-	verifyAddressAsOwner(rt, params.OwnerAddr)
-	verifyAddressAsWorker(rt, params.WorkerAddr)
+	owner := resolveOwnerAddress(rt, params.OwnerAddr)
+	worker := resolveWorkerAddress(rt, params.WorkerAddr)
 
-	state, err := ConstructState(adt.AsStore(rt), params.OwnerAddr, params.WorkerAddr, params.PeerId, params.SectorSize)
+	state, err := ConstructState(adt.AsStore(rt), owner, worker, params.PeerId, params.SectorSize)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


anorth:
Note: I did *not* resolve addresses in the multisig actor. I think there would be benefit to doing so, but also a tradeoff: resolving addresses requires the accounts to actually exist. It may be an intentional or desired behaviour that the signers of a multisig don't need to exist on chain at the time they are declared.

@whyrusleeping lmk if you would prefer that we do resolve them, and we can do it in a follow-up.

actors/builtin/miner/miner_actor.go
```diff
@@ -991,49 +991,54 @@ func (a Actor) commitWorkerKeyChange(rt Runtime) *adt.EmptyValue {
...
-	ret, code := rt.Send(address, builtin.MethodsAccount.PubkeyAddress, &adt.EmptyValue{}, big.Zero())
-	builtin.RequireSuccess(rt, code, "failed to fetch account pubkey from %v", address)
-	var pubkey addr.Address
-	err := ret.Into(&pubkey)
-	if err != nil {
-		rt.Abortf(exitcode.ErrSerialization, "failed to deserialize address result: %v", ret)
-	}
-	if pubkey.Protocol() != addr.BLS {
-		rt.Abortf(exitcode.ErrIllegalArgument, "worker account %v must have BLS pubkey, was %v", address, pubkey.Protocol())
+	if raw.Protocol() != addr.BLS {
```
icorderi: Leave the BLS check out until go-filecoin has BLS fixed. 
We can't build the genesis block with this check in.

anorth: It looks like the reported non-working of BLS in go-filecoin may be based on a misunderstanding. I'm going to land this in the desired form, and if more BLS problems arise that'll be my top priority.

PR #185, opened 2020-02-19
fix: ResolveAddress does not always return an ID address

icorderi:


actors/builtin/init/init_actor_state.go
```diff
@@ -33,22 +33,35 @@ func ConstructState(store adt.Store, networkName string) (*State, error) {
...
-// (which should only appear as values, not keys) and singleton actor addresses pass through unchanged.
+// ResolveAddress resolves an address to an ID-address, if possible.
+// If the provided address is an ID address, it is returned as-is.
+// This means that ID-addresses (which should only appear as values, not keys) and singleton actor addresses
+// pass through unchanged.
+//
+// Post-condition: all addresses succesfully returned by this method satisfy `addr.Protocol() == addr.ID`.
 func (s *State) ResolveAddress(store adt.Store, address addr.Address) (addr.Address, error) {
+	// short-circuit ID address resolution
+	// Note: ID addresses are not stored in the init actor
```
anorth: I think this is more accurately captured in the comment above `appear as values, not keys`

PR #187, opened 2020-02-20
Unwrap ProofBytes structure in proof verification info structs

anorth:
These structures used to have other fields, but now don't.

FYI @dignifiedquire, please alert if we'll need to revert this because there will be more fields again.


actors/abi/sector.go
```diff
@@ -32,25 +32,26 @@ type SectorWeight = big.Int
...
-	RegisteredProof_StackedDRG32GiBPoSt    = RegisteredProof(4)
-	RegisteredProof_StackedDRG1KiBSeal     = RegisteredProof(5)
-	RegisteredProof_StackedDRG1KiBPoSt     = RegisteredProof(6)
-	RegisteredProof_StackedDRG16MiBSeal    = RegisteredProof(7)
-	RegisteredProof_StackedDRG16MiBPoSt    = RegisteredProof(8)
-	RegisteredProof_StackedDRG256MiBSeal   = RegisteredProof(9)
-	RegisteredProof_StackedDRG256MiBPoSt   = RegisteredProof(10)
-	RegisteredProof_StackedDRG1GiBSeal     = RegisteredProof(11)
-	RegisteredProof_StackedDRG1GiBPoSt     = RegisteredProof(12)
+	RegisteredProofWinStackedDRG32GiBSeal = RegisteredProof(1)
```
dignifiedquire: winstacked is not a valid algorithm anymore, why is this still here?

actors/abi/sector.go
```diff
@@ -32,25 +32,26 @@ type SectorWeight = big.Int
...
+	RegisteredProofWinStackedDRG32GiBPoSt = RegisteredProof(2)
+	RegisteredProofStackedDRG32GiBSeal    = RegisteredProof(3)
+	RegisteredProofStackedDRG32GiBPoSt    = RegisteredProof(4)
+	RegisteredProofStackedDRG1KiBSeal     = RegisteredProof(5)
```
dignifiedquire: if this is spec code, the non spec compliant sectors should be probably marked as such,

actors/abi/sector.go
anorth: Because no-one responsible who knows that updated the spec.

actors/abi/sector.go
anorth: Indeed. Please tell us what they are (preferably by PR).

anorth:
We can't unwrap them after all.

PR #192, opened 2020-02-21
StoragePowerActor - deleteMiner ExitCodes

Alexey-N-Chernyshov:
Not found is an IllegalArgument Error, not IllegalState.
If claim power > 0 return  exitcode.ErrForbidden.


anorth:
The reason for the IllegalState codes is that either of these cases indicate an internal inconsistency in the power actor's state. There should be no sequence of calls from the outside that can lead to it.

The checks above already produce an ErrNotFound if the miner isn't registered, or ErrForbidden if balance is >0 (which should be the case for claim >0).

That said, I'm ok with leaving the ErrForbidden for the nonzero claim.

PR #194, opened 2020-02-22
fix bug preventing creation of multisig actor

frrist:
- add test to init actor for multisig creation

### DIFF
--- a/actors/builtin/init/init_test.go
+++ b/actors/builtin/init/init_test.go
@@ -139,6 +139,29 @@ func TestExec(t *testing.T) {
 		assert.Equal(t, addr.Undef, actualUnknownAddr)
 	})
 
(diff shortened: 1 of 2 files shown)


actors/builtin/init/init_test.go
```diff
@@ -139,6 +139,29 @@ func TestExec(t *testing.T) {
...
 	})
 
+	t.Run("happy path create multisig actor", func(t *testing.T) {
+		rt := builder.Build(t)
+
+		actor.constructAndVerify(rt)
+
+		// actor creating the multisig actor
+		someAccountActor := tutil.NewSECP256K1Addr(t, "pubbub")
+		rt.SetCaller(someAccountActor, builtin.AccountActorCodeID)
```
anorth: Callers are always ID addresses (possibly of account actors)

Issue #184, opened 2020-02-19
Consider resolving multisig signer addresses to ID-addresses in state


anorth:
Where addresses are provided by users, most actors resolve them to canonical ID addresses for use in state. Not doing so would make their use very error-prone.

The multisig actor does not (yet). Tesolving addresses requires the accounts to actually exist. It may be an intentional or desired behaviour that the signers of a multisig don't need to exist on chain at the time they are declared

See https://github.com/filecoin-project/specs-actors/pull/180#issuecomment-587971542

anorth:
https://github.com/filecoin-project/specs-actors/pull/195#issuecomment-594206413

PR #195, opened 2020-02-22
resolve address to public key for multisig verification

frrist:
Since the caller address is not guaranteed to be a public key address (pretty sure its always an ID address) it need to be resolved to a pubkey address for verifying if its a signer in the multisig actor state.

Will hold off on further implementation until #184 is resolved.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -322,7 +322,8 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID) {
 }
 
 func (a Actor) validateSigner(rt vmr.Runtime, st *State, address addr.Address) {
-	if address.Protocol() != addr.ID {
+	// resolve address to a public key before verifying.
+	if address.Protocol() == addr.ID {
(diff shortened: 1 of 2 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -93,7 +94,7 @@ type ProposeParams struct {
 
 func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
-	callerAddr := rt.Message().Caller()
+	callerAddr := resolveToPubKey(rt, rt.Message().Caller())
```
frrist: Almost every method on the Multisig actor is going to need to make this call. Holding off on completing the fix here until I am told this is the right solution.

anorth:
It would be much simpler to store canonical ID addresses in state. However that may not be possible if we want to support creating multisig actors that reference accounts that don't exist on-chain yet (#195).

I think we need to discuss with @whyrusleeping

whyrusleeping:
@anorth I think the only 'special' case here would be that we want to allow proposing sends to account actors that don't exist yet. All other addresses should be resolved to IDs

frrist:
closing in favor of #325

PR #196, opened 2020-02-24
Set deal state

ZenGround0:
The deal state is currently unset after publish storage deals finishes.  Maybe this is working as intended, but it prevents me from following lotus's lead [here](https://github.com/filecoin-project/lotus/blob/feat/specs-actors/chain/gen/genesis/miners.go#L164) by using the market methods to calculate deal weight.  Feel free to close if this doesn't line up with what you want.  In that case we'll just do direct setting of market stat in gen construction (cc @magik6k)


ZenGround0:
A little more context: an unset deal state causes VerifyDealsOnStorageCommit to fail on `states.Get`. I think that method may be the site of the unintended behavior because instead of returning the undefined deal state that DealMetaArray:Get tries to return it just errors.  I'll happily adjust this PR to fix states.Get to correctly recognize amt not found calls if that's preferred.

ZenGround0:
This is fixed in lotus integration PR

PR #198, opened 2020-02-24
update NumMinersMeetingMinPower

sternhenri:
Note that this explicitly bases all `minMinerSize` considerations on nominal power (cc @zixuanzh to confirm this is right -- seems right).

Also added actor endpoint for checking this (when validating ePost)

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -46,6 +46,7 @@ func (a Actor) Exports() []interface{} {
 		13:                        a.EnrollCronEvent,
 		14:                        a.ReportConsensusFault,
 		15:                        a.OnEpochTickEnd,
+		16:                        a.MinerNominalPowerMeetsConsensusMinimum,
 	}
 }
 
@@ -528,6 +529,19 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 		rt.Abortf(exitcode.ErrIllegalState, "Failed to process deferred cron events: %v", err)
 	}
 	return nil
+}
+
+func (a Actor) MinerNominalPowerMeetsConsensusMinimum(rt Runtime, _ *adt.EmptyValue) bool {
+	var st State
+	rt.State().Readonly(&st)
+	store := adt.AsStore(rt)
+	minerAddr := rt.Message().Receiver()
+
+	gtMinSize, err := st.minerNominalPowerMeetsConsensusMinimum(store, minerAddr)
+	if err != nil {
+		rt.Abortf(exitcode.ErrIllegalState, "Failed to check miner %v meets min required size: %v", minerAddr, err)
+	}
+	return gtMinSize
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -46,7 +46,6 @@ func (a Actor) Exports() []interface{} {
 		13:                        a.EnrollCronEvent,
 		14:                        a.ReportConsensusFault,
 		15:                        a.OnEpochTickEnd,
-		16:                        a.MinerNominalPowerMeetsConsensusMinimum,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -544,6 +545,19 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
+	store := adt.AsStore(rt)
+	minerAddr := rt.Message().Receiver()
+
+	gtMinSize, err := st.minerNominalPowerMeetsConsensusMinimum(store, minerAddr)
+	if err != nil {
+		rt.Abortf(exitcode.ErrIllegalState, "Failed to check miner %v meets min required size: %v", minerAddr, err)
+	}
+	return gtMinSize
+}
+
```
anorth: Please do not add an actor method unless it needs to be called by some other actor. Extra-actor code (including EPost validation) will access this state directly.

actors/builtin/power/power_state.go
```diff
@@ -146,10 +159,29 @@ func (st *State) addToClaim(s adt.Store, miner addr.Address, power abi.StoragePo
...
+	oldNominalPower, err := st.computeNominalPower(s, miner, claim.Power)
+	if err != nil {
+		return err
+	}
 	claim.Power = big.Add(claim.Power, power)
+	newNominalPower, err := st.computeNominalPower(s, miner, claim.Power)
+	if err != nil {
+		return err
+	}
+
```
anorth: Prefer to move this after the following line so the claim pledge/power updates remain grouped.

anorth:
@sternhenri to be clear, I believe `put/deleteDetecteFault` must update the count.

actors/builtin/power/power_state.go
```diff
@@ -146,23 +159,49 @@ func (st *State) addToClaim(s adt.Store, miner addr.Address, power abi.StoragePo
...
 	return st.setClaim(s, miner, claim)
 }
 
 func (st *State) computeNominalPower(s adt.Store, minerAddr addr.Address, claimedPower abi.StoragePower) (abi.StoragePower, error) {
 	// Compute nominal power: i.e., the power we infer the miner to have (based on the network's
 	// PoSt queries), which may not be the same as the claimed power.
-	// Currently, the only reason for these to differ is if the miner is in DetectedFault state
-	// from a SurprisePoSt challenge. TODO: hs update this
+	// Currently, the nominal power may differ from claimed power because of
+	// collateral and faults (declared or detected)
```
anorth: It doesn't differ due to declared faults. Declared faults are already reflected in the claimed power - a miner is not claiming a faulty sector has power. The claim will already include any "trickiness".

This means the `else` case below is unnecessary (but maybe add a comment so someone else doesn't have the same misconception).

sternhenri: ahhhh I had misunderstood your above comment, as you saw. Will update accordingly. With that said, could you tell me where `DeclareTempFaults` updates the "power `claim`s"? I see where it updates the sector itself in the miner actor, but not where it updates the `Claim` from the power actor. Is that missing (should be handled in subseq PR)?

sternhenri:
Do we not want to account for collateral in this calculation?

actors/builtin/power/power_state.go
anorth: DeclareTempFaults registers a cron handler that does the real work. See `requestBeginFaults` which invokes the power actor.

actors/builtin/power/power_state.go
```diff
@@ -183,6 +219,23 @@ func (st *State) putDetectedFault(s adt.Store, a addr.Address) error {
...
+		return err
+	}
+	if !ok {
+		return errors.Errorf("no claim for actor %v", a)
+	}
+	nominalPower, err := st.computeNominalPower(s, a, claim.Power)
+	if err != nil {
+		return err
+	}
+	if nominalPower.GreaterThanEqual(ConsensusMinerMinPower) {
```
anorth: Never true. This will be zero precisely because the miner now has a fault. We need to check the prior state: the post state is known.

actors/builtin/power/power_state.go
sternhenri: I was missing it: it all happens thanks to the cron actor in `checkTemporaryFaultEvents()`

sternhenri:
We must ensure that a miner can *never* be undercollaterallized (as is intended).

Specifically, check whether `withdrawFund()` ever causes pledge collateral to go down (in case where temp faults have been declared -> claimed power goes down -> nominal power goes down -> associated collateral requirement ? (it should not decrease)).

actors/builtin/power/power_state.go
sternhenri: good catch!

Issue #189, opened 2020-02-20
Add 'CheckSectorProved' method to miner actor 


whyrusleeping:
In order to allow for a trustless PoRep outsourcing market, we only need a single method on the miner actor that returns exit code 0 if the referenced sector is proven, and returns a nonzero exit code if the sector has not been proven.

This will also require that 'ProveCommitSector' can be called by anyone (which is fine) so that the prover can submit the proof on their own.

The final choice in the design space is a timeout, we can add the timeout as a parameter to the method we're adding here (which is fine because the prover cannot change the method parameters since they are signed over by the payer) or we can generalize it and add a timeout field to the payment channel itself. I'm actually a fan of having the payment channel have a timeout field, as this is also needed for cross chain atomic swaps.

cc @magik6k @jbenet @arajasek

anorth:
Closed by #199

PR #199, opened 2020-02-25
Changes to support a PoRep outsourcing market

arajasek:
As in #189, this PR has 3 commits, one for each of the following changes:

- Add a new method to the miner actor, `CheckSectorProven`, which succeeds if a given sector has been proven, fails otherwise
- Allow miner actor's `ProveCommitSector` to be called by anyone
- Add a `TimeLockMax` to vouchers in payment channels. Vouchers cannot be redeemed beyond the `TimeLockMax`.

### DIFF
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -93,8 +93,11 @@ type UpdateChannelStateParams struct {
 // A voucher is sent by `From` to `To` off-chain in order to enable
 // `To` to redeem payments on-chain in the future
 type SignedVoucher struct {
-	// TimeLock sets a min epoch before which the voucher cannot be redeemed
-	TimeLock abi.ChainEpoch
+	// TimeLockMin sets a min epoch before which the voucher cannot be redeemed
+	TimeLockMin abi.ChainEpoch
+	// TimeLockMax sets a max epoch beyond which the voucher cannot be redeemed
+	// TimeLockMax set to 0 means no timeout
+	TimeLockMax abi.ChainEpoch
 	// (optional) The SecretPreImage is used by `To` to validate
 	SecretPreimage []byte
 	// (optional) Extra can be specified by `From` to add a verification method to the voucher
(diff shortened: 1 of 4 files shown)


actors/builtin/paych/paych_actor.go
```diff
@@ -154,10 +156,14 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
...
 	}
 
-	if rt.CurrEpoch() < sv.TimeLock {
+	if rt.CurrEpoch() < sv.TimeLockMin {
 		rt.Abortf(exitcode.ErrIllegalArgument, "cannot use this voucher yet!")
 	}
 
+	if rt.CurrEpoch() > sv.TimeLockMax {
+		rt.Abortf(exitcode.ErrIllegalArgument, "this voucher has expired!")
+	}
```
anorth: Please add a test exercising this check. Probably another case next to the one labelled "fails if too early for voucher"

anorth:
FYI @zixuanzh

actors/builtin/paych/paych_actor.go
```diff
@@ -154,10 +156,14 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 		rt.Abortf(exitcode.ErrIllegalArgument, "voucher signature invalid")
 	}
 
-	if rt.CurrEpoch() < sv.TimeLock {
+	if rt.CurrEpoch() < sv.TimeLockMin {
 		rt.Abortf(exitcode.ErrIllegalArgument, "cannot use this voucher yet!")
 	}
 
+	if rt.CurrEpoch() > sv.TimeLockMax {
```
whyrusleeping: We should check that the TimeLockMax is not zero, to allow people to set it to zero and mean 'no timeout'

arajasek: Good point, thanks!

actors/builtin/paych/paych_actor.go
arajasek: Done, thank you!

arajasek:
> Without a design rationale written somewhere (like an issue) it's a bit hard to judge whether this is appropriate for the goals. I know previous work has also lacked documented design, but I'm hoping to lift the bar here.
> 
> However I don't see anything wrong with this code, and willing to trust that the design ideas have in fact been reviewed in some other forum.
> 
> LGTM after adding a test.

Do you think the design rationale provided in #189 is lacking? I can flesh it out a bit if you'd like :)

anorth:
#189 leaves a lot to inference, but no I don't think it's necessary to flesh out now.

frrist:
Heads up to reviews here, this PR missed both of these:
- https://github.com/filecoin-project/specs-actors/pull/221
- https://github.com/filecoin-project/specs-actors/pull/224

This isn't the first time cases like this have been missed, there are a lot of details to keep track of here. Solutions to prevent this from happening again could be:
- Write some code that uses reflection to check that method params implement the required interface (high effort approach)
- Add a PR templated that states something like: "Did you add, change, or remove an actor method, if you did be sure to 1 add code gen, 2 run code gen 3 update method nums" (low effort approach)

PR #200, opened 2020-02-25
plumb required values into PoSt verification-input

laser:
## Why is this PR needed?

The existing PoSt verification types are lacking in a few ways:

1. The PoSt verifier needs some way to know the `RegisteredProof` that corresponds to a `PoStProof`
1. The PoSt verifier needs some value from which they can derive a 32-byte prover ID
1. Sector size is a function of `RegisteredProof`

## What's in this PR?

This changeset:

1. Enables a storage miner's proving set to contain sectors sealed and proven using more than one `RegisteredProof`; it adds `RegisteredProof` to the `PoStProof` struct.
1. Adds `Prover ActorID` to the `PoStVerifyInfo` struct so that verifiers can map to a 32-byte prover id (for libfilecoin's `verify_post` call).
1. Drops the `sectorSize SectorSize` parameter from `VerifySeal` and `VerifyPoSt` calls

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -178,17 +178,14 @@ type SectorInfo struct {
 }
 
 type OnChainElectionPoStVerifyInfo struct {
-	// There should be one RegisteredProof for each PoSt Candidate
-	RegisteredProofs []RegisteredProof
-	Candidates       []PoStCandidate
-	Proofs           []PoStProof
-	Randomness       PoStRandomness
+	Candidates []PoStCandidate // each PoStCandidate has its own RegisteredProof
+	Proofs     []PoStProof     // each PoStProof has its own RegisteredProof
+	Randomness PoStRandomness
 }
 
 type OnChainPoStVerifyInfo struct {
-	ProofType  RegisteredProof
-	Candidates []PoStCandidate
-	Proofs     []PoStProof
+	Candidates []PoStCandidate // each PoStCandidate has its own RegisteredProof
+	Proofs     []PoStProof     // each PoStProof has its own RegisteredProof
 }
 
 type PoStCandidate struct {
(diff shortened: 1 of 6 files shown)


laser:
@icorderi - This PR makes incremental progress towards [the thing which you asked for in Slack](https://filecoinproject.slack.com/archives/CEGB67XJ8/p1582157103459800). Without this PR, the `PoStVerifyInfo` doesn't provide all of the information that a verifier needs to verify a PoSt proof.

actors/abi/sector.go
```diff
@@ -166,10 +166,10 @@ type PartialTicket []byte // 32 bytes
 // TODO Porcu: refactor these types to get rid of the squishy optional fields.
```
anorth: Please expunge this TODO while you're here.

actors/abi/sector.go
```diff
@@ -200,6 +197,7 @@ type PoStCandidate struct {
 }
 
 type PoStProof struct { //<curve, system> {
+	RegisteredProof
```
anorth: Do you anticipate a similar thing being required for SealProof? 

That will conflict with #174 (but we should land this to signal/avoid that regression). Also invalidates #187. cc @magik6k

porcuquine: As far as I saw, seal was not broken suring the refactor performed when you moved the actors, so it will should be fine.

laser: @anorth 

> Do you anticipate a similar thing being required for SealProof?

I do not. It looks like `SealVerifyInfo.OnChain.RegisteredProof` represents the `SealProof`'s `RegisteredProof`, so we should be good to go.

laser: @porcuquine 

> As far as I saw, seal was not broken during the refactor performed when you moved the actors, so it should be fine.

@porcuquine roger that, thanks for doing that archaeology for me

actors/abi/sector.go
laser: donezo

PR #201, opened 2020-02-25
add challenge count computing-method and plumb value to PoSt verification method

laser:
See [this thread](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1582657646068400?thread_ts=1582648075.057600&cid=CHMNDCK9P) for more information.

## Why is this PR needed?

The libfilecoin `post_verify` function needs to know the prover's challenge count. Without that value, the verifier can't verify the provided proofs. 

## What's in this changeset?

This changeset adds a challenge count-computing method (which is a function of proving set) and plumbs its value to the `VerifyPoSt` method.

## What's not in this changeset?

This changeset does not add an operation which can be used to generate ePoSt challenge count.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -237,6 +237,8 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
 }
 
 func (st *State) ComputeWindowedPoStChallengeCount(store adt.Store) (uint64, error) {
+	// TODO: This is too slow. Subtract fault count from sector count when these
+	// are maintained consistently.
 	provingSet, err := st.ComputeProvingSet(store)
 	if err != nil {
 		return 0, errors.Wrapf(err, "failed to compute proving set: %v", err)
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -236,13 +236,13 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
 	return sectorInfos, nil
 }
 
-func (st *State) ComputePoStChallengeCount(store adt.Store) (uint64, error) {
+func (st *State) ComputeWindowedPoStChallengeCount(store adt.Store) (uint64, error) {
 	provingSet, err := st.ComputeProvingSet(store)
 	if err != nil {
 		return 0, errors.Wrapf(err, "failed to compute proving set: %v", err)
 	}
 
-	return uint64(len(provingSet)) * (PoStSampleRateNum / PoStSampleRateDenom), nil
+	return uint64(len(provingSet)) * (WindowedPoStSampleRateNum / WindowedPoStSampleRateDenom), nil
 }
 
 func (mps *PoStState) IsPoStOk() bool {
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -236,6 +236,18 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
 	return sectorInfos, nil
 }
 
+func (st *State) ComputeWindowedPoStChallengeCount(store adt.Store) (uint64, error) {
+	provingSet, err := st.ComputeProvingSet(store)
```
anorth: This is very slow. I think we can do something faster but it will depend on changes and discipline around the fault representation so that we can safely just subtract the fault count from the sector count. Please leave a comment here: `TODO: this is too slow, subtract fault count from sector count when these are maintained consistently`.

laser: done!

sternhenri:
I'm sorry to be a bit late to the party but this PR doesn't completely make sense to me.

Why are we adding a `ChallengedCount` field on-chain, and then computing one that is dependent on miner size.

The use of the sampling rate here is what is currently used for election PoSt, rather than windowed-Post, but all of this happens off-chain regardless.

As for the `ChallengedCount` for windowedPoSt it is likely that the heuristic for calculating it will be available in the actor code (if it is not just a network param), so the field here is likely unnecessary. (this is subject to change pending post soundness params analysis ending tm)

For now I might just pass in a constant or do whatever lotus is doing... Lmk if you disagree @anorth 

I'm really sorry again that I'm only getting to this now @laser

laser:
@sternhenri 

> I'm really sorry again that I'm only getting to this now @laser

Not a problem at all. Let's see if we can come to a shared understanding of how you want this thing should work.

> Why are we adding a ChallengedCount field on-chain, and then computing one that is dependent on miner size.

Where can I learn about the on-chain ChallengedCount field? I [don't see `ChallengedCount` appear in the specs](https://github.com/filecoin-project/specs/search?q=%22ChallengedCount%22&unscoped_q=%22ChallengedCount%22) repo, nor do I see it in [specs-actors](https://github.com/filecoin-project/specs-actors/search?q=%22ChallengedCount%22&type=Issues), nor in [lotus](https://github.com/filecoin-project/lotus/search?q=%22ChallengeCount%22&unscoped_q=%22ChallengeCount%22).

> As for the ChallengedCount for windowedPoSt it is likely that the heuristic for calculating it will be available in the actor code (if it is not just a network param), so the field here is likely unnecessary. (this is subject to change pending post soundness params analysis ending tm)

I don't understand what you mean by a "field" here. I've added a _method_ to the miner `State` struct - but no additional fields.

> For now I might just pass in a constant 

Just so I'm clear: Do you suggest that challenge count (for purposes of windowed PoSt) be a fixed number (a constant) for all miners - regardless of proving set-size?

> or do whatever lotus is doing or keep as is if it's no extra work for them

Lotus doesn't compute non-Election PoSts yet, so there isn't any prior art here.

sternhenri:
Sorry. I thought this was better defined everywhere. 

> Just so I'm clear: Do you suggest that challenge count (for purposes of windowed PoSt) be a fixed number (a constant) for all miners - regardless of proving set-size?

Yes exactly. Sorry: I'd misread your actor method as a field.

Lmk if that and out-of-band activity has answered your Qs (I believe they have).

Result: leave as is for now, likely to touch up as soon as final params are in.

anorth:
@laser given this information, would you follow up with a reversion of most of this change?

laser:
@anorth (cc @sternhenri)

> @laser given this information, would you follow up with a reversion of most of this change?

surely

PR #204, opened 2020-02-26
Storage Market Actor tests #1

ingar:
## Summary
Testing of Storage Market Actor and State, part 1..

This is a partial test coverage of the storage market actor; submitting it now because it's getting too large.  As far as changes to the actor/state, I've exposed a few State methods and moved some state manipulation into the state object (`AddEscrowBalance`, `AddLockedBalance`), as well as a couple of fixes.

This PR covers testing of the Actor's `Constructor`, `AddBalance`, `WithdrawBalance`

## Todo
Test the logic within `State` via the exported actor methods.


actors/builtin/market/market_state_test.go
```diff
@@ -38,4 +62,300 @@ func TestConstruction(t *testing.T) {
...
+			// There should be no balance changes, as payments are done
+			assertBalances(t, rt, st, client, 50, 0)
+			assertBalances(t, rt, st, provider, 507, 0)
+
+			// State stops tracking this deal
+			rt.ExpectAbort(exitcode.ErrIllegalState, func() {
+				st.mustGetDeal(rt, dealId)
+			})
+		})
+	})
```
anorth: Most of these direct-state tests are too low level. There is a bunch of technical debt in the market actor state. These tests will cement that debt and make the fixes painful. Two particular examples are:
- direct use of the runtime
- unscalable iteration over all deals in updatePendingDealState

The scenarios described in the test descriptions are great, but please exercise them via the actor's exported API, not the state object. On balance, I'd prefer to withdraw this file from the PR so the rest of it can land, then bring these scenarios back a different way.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+	t.Run("simple construction", func(t *testing.T) {
+		rt := builder.Build(t)
+
+		rt.ExpectValidateCallerAddr(builtin.SystemActorAddr)
+
+		params := adt.EmptyValue{}
+		ret := rt.Call(actor.Constructor, &params).(*adt.EmptyValue)
+
+		assert.Equal(t, adt.EmptyValue{}, *ret)
+		rt.Verify()
```
anorth: Move the constructor state checks out from the state test into this one (and delete the state test)

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+	marketActor := tutil.NewIDAddr(t, 100)
+	provider := tutil.NewIDAddr(t, 101)
+	worker := tutil.NewIDAddr(t, 102)
+	client := tutil.NewIDAddr(t, 103)
+
+	var st market.State
+
+	setup := func() (*mock.Runtime, *marketActorTestHarness) {
+		builder := mock.NewBuilder(context.Background(), marketActor).
+			WithBalance(abi.NewTokenAmount(100), abi.NewTokenAmount(0)).
```
anorth: I'd prefer not to set the balance inaccurately here. Set it only directly when it should known. For example, it would be pretty reasonable for the actor to have a sanity check that the sum of its balance tables matches the balance actually held  (I'd love you to add a facility like that to the state object, actually).

I suspect right now the actor never looks at it's balance so we could get away with zero always until we add such checks.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+			delta int64
+			total int64
+		}{
+			{10, 10},
+			{20, 30},
+			{40, 70},
+		}
+
+		for _, tc := range testCases {
+			rt.SetCaller(provider, builtin.StorageMinerActorCodeID)
```
anorth: AddBalance is not called by the miner actor, it's called by an account actor (the miner's worker account).

I am really surprised this worked at all, since the `escrowAddress` function contains `rt.ValidateImmediateCallerIs(ownerAddr, workerAddr)`. Please investigate and add a case that explicitly aborts because the caller is an account that's not the miner's worker or owner

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+			rt.SetReceived(abi.NewTokenAmount(tc.delta))
+			actor.expectProviderControlAddress(rt, provider, worker)
```
anorth: This might be the problem, the miner's owner address is _not the miner's address_. We need to carefully track all three miner/owner/worker addresses.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+			rt.ExpectValidateCallerType(builtin.CallerTypesSignable...)
+
+			rt.Call(actor.AddBalance, &client)
+
+			rt.Verify()
+
+			rt.GetState(&st)
+			assert.Equal(t, abi.NewTokenAmount(tc.total), st.GetEscrowBalance(rt, client))
+		}
+	})
```
anorth: Please add a case for the somewhat confusing possibility of a miner worker account functioning as the client of a deal (i.e. the worker account address as the parameter), ensuring the balance is credited to the worker account, not the miner account.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+	actor := marketActorTestHarness{t: t}
+	marketActor := tutil.NewIDAddr(t, 100)
+	provider := tutil.NewIDAddr(t, 101)
+	worker := tutil.NewIDAddr(t, 102)
+	client := tutil.NewIDAddr(t, 103)
+
+	var st market.State
+
+	builder := mock.NewBuilder(context.Background(), marketActor).
+		WithBalance(abi.NewTokenAmount(100), abi.NewTokenAmount(0)).
```
anorth: Ditto

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+
+	assert.Equal(h.t, adt.EmptyValue{}, *ret)
+	rt.Verify()
+}
+
+// addProviderFunds is a helper method to setup provider market funds
+func (h *marketActorTestHarness) addProviderFunds(rt *mock.Runtime, provider address.Address, worker address.Address, amount abi.TokenAmount) {
+	rt.SetReceived(amount)
+	rt.SetCaller(provider, builtin.StorageMinerActorCodeID)
+	h.expectProviderControlAddress(rt, provider, worker)
```
anorth: These are incorrect as described above.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+
+	t.Run("withdraws from provider escrow funds", func(t *testing.T) {
+		rt := builder.WithBalance(abi.NewTokenAmount(100), abi.NewTokenAmount(0)).Build(t)
+		actor.constructAndVerify(rt)
+		actor.addProviderFunds(rt, provider, worker, abi.NewTokenAmount(20))
+
+		rt.GetState(&st)
+		assert.Equal(t, abi.NewTokenAmount(20), st.GetEscrowBalance(rt, provider))
+
+		rt.SetCaller(provider, builtin.StorageMinerActorCodeID)
```
anorth: Caller should be miner worker. (It could also be the owner, but using worker is stronger since we later must check that the transfer actually goes to the owner, despite the owner not being the caller).

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+		actor.expectProviderControlAddress(rt, provider, worker)
+		rt.ExpectSend(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, abi.NewTokenAmount(0), nil, exitcode.Ok)
```
anorth: Wow, could you please fix the code so this doesn't happen?

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+
+		withdrawAmount := abi.NewTokenAmount(1)
+		rt.ExpectSend(provider, builtin.MethodSend, nil, withdrawAmount, nil, exitcode.Ok)
```
anorth: This must send to the owner address (please use distinct owner/worker addresses so this can be asserted).

actors/builtin/market/market_test.go
```diff
@@ -1 +1,252 @@
...
+		rt.ExpectSend(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, abi.NewTokenAmount(0), nil, exitcode.Ok)
+		rt.ExpectSend(client, builtin.MethodSend, nil, withdrawAmount, nil, exitcode.Ok)
+
+		rt.Call(actor.WithdrawBalance, &params)
+
+		rt.Verify()
+
+		rt.GetState(&st)
+		assert.Equal(t, abi.NewTokenAmount(19), st.GetEscrowBalance(rt, client))
+	})
```
anorth: Please add cases for:
- miner worker account used as deal client
- fail due to insufficient balance

Add a placeholder/todo for
- provider/client fail with sufficient balance, but locked (requires deals)

actors/util/adt/array.go
```diff
@@ -56,7 +56,7 @@ func (a *Array) Set(i uint64, value runtime.CBORMarshaler) error {
 	if err != nil {
 		return errors.Wrapf(err, "array set failed to load root %v", a.root)
 	}
-	if root.Set(a.store.Context(), i, value) != nil {
+	if err := root.Set(a.store.Context(), i, value); err != nil {
```
anorth:
```suggestion
	if err = root.Set(a.store.Context(), i, value); err != nil {
```

support/mock/mockrt.go
```diff
@@ -406,6 +406,10 @@ func (rt *Runtime) SetCaller(address addr.Address, actorType cid.Cid) {
 	rt.actorCodeCIDs[address] = actorType
 }
 
+func (rt *Runtime) SetInCall(inCall bool) {
+	rt.inCall = inCall
+}
+
```
anorth: Remove this after rolling back the state tests that required it.

actors/builtin/market/market_state_test.go
```diff
@@ -1,22 +1,28 @@
-package market_test
+package market
```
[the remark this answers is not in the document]
ingar: Yeah, I was unfortunately testing unexported state methods.  State test is gone now.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,347 @@
...
+			WithActorType(owner, builtin.AccountActorCodeID).
+			WithActorType(worker, builtin.AccountActorCodeID).
+			WithActorType(providerIdAddr, builtin.StorageMinerActorCodeID).
+			WithActorType(client, builtin.AccountActorCodeID)
+
+		rt := builder.Build(t)
+
+		rt.AddIDAddress(owner, owner)
+		rt.AddIDAddress(worker, worker)
+		rt.AddIDAddress(client, client)
```
anorth: It only makes sense to call this with a first argument that's not an ID address. The non-ID -> ID address resolution is a very important set of functionality to test, perhaps in the next round. But these muddy the waters.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,347 @@
...
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+	"github.com/filecoin-project/specs-actors/support/mock"
+	tutil "github.com/filecoin-project/specs-actors/support/testing"
+)
+
+func TestMarketActor(t *testing.T) {
+	marketActor := tutil.NewIDAddr(t, 100)
+	owner := tutil.NewIDAddr(t, 101)
+	provider := tutil.NewActorAddr(t, "provider")
```
anorth: I know it's confusing, but just use an ID address here for the canonical case (using the actor address is an important but specific case to exercise).

actors/builtin/market/market_test.go
```diff
@@ -1 +1,347 @@
...
+		assert.Equal(t, emptyMap.Root(), state.LockedTable)
+		assert.Equal(t, abi.DealID(0), state.NextID)
+		assert.Equal(t, emptyMultiMap.Root(), state.DealIDsByParty)
+	})
+
+	t.Run("AddBalance", func(t *testing.T) {
+		t.Run("adds to provider escrow funds", func(t *testing.T) {
+			testCases := []struct {
+				delta int64
+				total int64
```
anorth: Maybe for a follow-up, but a table-driven approach like this might work well for exercising all of the provider, owner, and worker as non-ID addresses as well as ID addresses.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,347 @@
...
+			}
+
+			// Test adding provider funds from both worker and owner address
+			for _, callerAddr := range []address.Address{owner, worker} {
+				rt, actor := setup()
+
+				for _, tc := range testCases {
+					rt.SetCaller(callerAddr, builtin.AccountActorCodeID)
+					rt.SetReceived(abi.NewTokenAmount(tc.delta))
+					actor.expectProviderControlAddress(rt, providerIdAddr, owner, worker)
```
anorth: ~This should also expect a `ValidateImmediateCallerIs(owner, worker)`. I'm confused as to why this test didn't fail due to the actor making that call, but there being no expectation for it.~

Edit: I see it's inside this method call. Please rename something like `expectValidateCallerIsProviderControlAddress` or `expectProviderControlAddressAndValidateCaller` or pull that caller validation out and make it explicit here.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,347 @@
...
+
+			rt.GetState(&st)
+			assert.Equal(t, abi.NewTokenAmount(20), st.GetEscrowBalance(rt, providerIdAddr))
+
+			// worker calls WithdrawBalance, balance is transferred to owner
+			rt.SetCaller(worker, builtin.AccountActorCodeID)
+			actor.expectProviderControlAddress(rt, providerIdAddr, owner, worker)
+
+			withdrawAmount := abi.NewTokenAmount(1)
+			rt.ExpectSend(owner, builtin.MethodSend, nil, withdrawAmount, nil, exitcode.Ok)
```
anorth: Use adt.EmptyValue for the third parameter. An in-flight change will enforce this.

actors/builtin/market/market_test.go
```diff
@@ -1 +1,347 @@
...
+			rt.GetState(&st)
+			assert.Equal(t, abi.NewTokenAmount(19), st.GetEscrowBalance(rt, client))
+		})
+
+		t.Run("client withdrawing more than escrow balance limits to available funds", func(t *testing.T) {
+			rt, actor := setup()
+			actor.addParticipantFunds(rt, client, abi.NewTokenAmount(20))
+
+			rt.GetState(&st)
+			assert.Equal(t, abi.NewTokenAmount(20), st.GetEscrowBalance(rt, client))
```
anorth: Unnecessary IMO

actors/builtin/market/market_test.go
ingar: Oh right, I didn't see that it early-outs with an id address.

actors/builtin/market/market_test.go
ingar: @anorth should all `Send` calls with `nil` params in the market actor be changed to `adt.EmptyValue{}`?

anorth: Yes please

PR #205, opened 2020-02-26
implement reward actor unit tests

frrist:
The [makeBlockRewardMessage](https://filecoin-project.github.io/specs/#vminterpreter-implementation) method defined in the spec indicates that messages being sent to the reward actor from the system actor for block reward should contain a value of zero. I believe this means the RewardActor should not assert the value received is equal to the gas reward since value received will always be zero.
```
// Builds a message for paying block reward to a miner's owner.
func _makeBlockRewardMessage(state st.StateTree, minerAddr addr.Address, penalty abi.TokenAmount, gasReward abi.TokenAmount) msg.UnsignedMessage {
	params := serde.MustSerializeParams(minerAddr, penalty)
	TODO() // serialize other inputs to BlockRewardMessage or get this from query in RewardActor

	sysActor, ok := state.GetActor(builtin.SystemActorAddr)
	Assert(ok)

	return &msg.UnsignedMessage_I{
		From_:       builtin.SystemActorAddr,
		To_:         builtin.RewardActorAddr,
		Method_:     builtin.Method_RewardActor_AwardBlockReward,
		Params_:     params,
		CallSeqNum_: sysActor.CallSeqNum(),
		Value_:      0,
		GasPrice_:   0,
		GasLimit_:   msg.GasAmount_SentinelUnlimited(),
	}
}

```

### DIFF
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -67,7 +67,10 @@ type AwardBlockRewardParams struct {
 // - a penalty amount, provided as a parameter, which is burnt,
 func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
-	AssertMsg(params.Miner.Protocol() == addr.ID, "miner address must be ID-address (%d), was %d", addr.ID, params.Miner.Protocol())
+	AssertMsg(params.Miner.Protocol() == addr.ID,
+		"miner address must be ID-address protocol (%d), was protocol %d", addr.ID, params.Miner.Protocol())
+	AssertMsg(rt.CurrentBalance().GreaterThanEqual(params.GasReward),
+		"actor current balance %v insufficient to pay gas reward %v", rt.CurrentBalance(), params.GasReward)
 	priorBalance := rt.CurrentBalance()
 
 	var penalty abi.TokenAmount
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -68,8 +68,6 @@ type AwardBlockRewardParams struct {
 func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
 	AssertMsg(params.Miner.Protocol() == addr.ID, "miner address must be ID-address (%d), was %d", addr.ID, params.Miner.Protocol())
-	AssertMsg(params.GasReward.Equals(rt.Message().ValueReceived()),
-		"expected value received %v to match gas reward %v", rt.Message().ValueReceived(), params.GasReward)
 	priorBalance := rt.CurrentBalance()
 
 	var penalty abi.TokenAmount
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -20,8 +20,8 @@ const TokenPrecision = int64(1_000_000_000_000_000_000)
 // Target reward released to each block winner.
 var BlockRewardTarget = big.Mul(big.NewInt(100), big.NewInt(TokenPrecision))
 
-const rewardVestingFunction = None            // PARAM_FINISH
-const rewardVestingPeriod = abi.ChainEpoch(0) // PARAM_FINISH
+const RewardVestingFunction = None            // PARAM_FINISH
+const RewardVestingPeriod = abi.ChainEpoch(0) // PARAM_FINISH
 
 type Actor struct{}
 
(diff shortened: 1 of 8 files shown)


actors/builtin/reward/reward_actor.go
```diff
@@ -68,8 +68,6 @@ type AwardBlockRewardParams struct {
 func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
 	AssertMsg(params.Miner.Protocol() == addr.ID, "miner address must be ID-address (%d), was %d", addr.ID, params.Miner.Protocol())
-	AssertMsg(params.GasReward.Equals(rt.Message().ValueReceived()),
```
anorth: The assertion that should be made instead is that the balance is greater than or equal to the params.GasReward.

support/mock/mockrt.go
```diff
@@ -496,6 +496,33 @@ func (rt *Runtime) Reset() {
...
+		a, ok := r.(abort)
+		if ok {
+			rt.t.Errorf("expected panic with message %v but got abort %v", expected, a)
+			return
+		}
+		p, ok := r.(string)
+		if !ok {
+			panic(r)
+		}
+		if p != expected {
```
frrist: I think being explicit on the expected reason for an assertion failure is appropriate here since calls to `AssertMsg` are infrequent.

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+	receiver := tutil.NewIDAddr(t, 1000)
+	builder := mock.NewBuilder(context.Background(), receiver).WithCaller(builtin.SystemActorAddr, builtin.SystemActorCodeID)
+	rt := builder.Build(t)
+	actor.constructAndVerify(rt)
+}
+
+func TestAwardBlockReward(t *testing.T) {
+	actor := rewardHarness{reward.Actor{}, t}
+
+	receiver := tutil.NewIDAddr(t, 1000)
```
anorth: We actually know the reward actor's address, use the singleton.

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+	builder := mock.NewBuilder(context.Background(), receiver).WithCaller(builtin.SystemActorAddr, builtin.SystemActorCodeID)
+
+	t.Run("block reward with no payable reward", func(t *testing.T) {
```
anorth: More specific "no reward when penalty equals reward and reward actor has no balance except gas".

We should also check for penalty exceeding reward.

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+		rt.GetState(&st)
+		rewardMap := adt.AsMultimap(adt.AsStore(rt), st.RewardMap)
+		assert.Equal(t, big.Zero(), st.RewardTotal)
+		var r []reward.Reward
+		var rwd reward.Reward
+		err := rewardMap.ForEach(reward.AddrKey(miner), &rwd, func(i int64) error {
+			r = append(r, rwd)
+			return nil
+		})
+		require.NoError(t, err)
```
anorth: Factor out these state inspection methods into helpers

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+		assert.Equal(t, 0, len(r))
+	})
+
+	t.Run("block reward with reward payable", func(t *testing.T) {
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		miner := tutil.NewIDAddr(t, 1000)
+		penalty := abi.NewTokenAmount(10)
+		gasreward := abi.NewTokenAmount(10)
```
anorth: If they're going to cancel out, just use zero for this test (so the reward total is jsut block reward)

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+		nominalpwr := abi.NewStoragePower(10)
+		rt.SetBalance(big.NewInt(20))
+		actor.blockRewardAndVerify(rt, miner, penalty, gasreward, nominalpwr)
+
+		var st reward.State
+		rt.GetState(&st)
+		rewardMap := adt.AsMultimap(adt.AsStore(rt), st.RewardMap)
+		assert.Equal(t, big.NewInt(10), st.RewardTotal)
```
anorth: I don't understand how this is 10. It should be 20 (being the reward actor's balance, the target reward would be 100 if it had sufficient).

The balance < target is a good case to exercise nevertheless.

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+
+		assert.Equal(t, abi.ChainEpoch(0), r[0].StartEpoch)
+		assert.Equal(t, 0+reward.RewardVestingPeriod, r[0].EndEpoch)
+		assert.Equal(t, big.NewInt(10), r[0].Value)
+		assert.Equal(t, big.Zero(), r[0].AmountWithdrawn)
+		assert.Equal(t, reward.RewardVestingFunction, r[0].VestingFunction)
+
+	})
+
+	t.Run("block reward send failure", func(t *testing.T) {
```
anorth: "aborts when burning penalty fails"

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+		})
+		rt.Verify()
+	})
+
+	t.Run("assertion failure when miner is not ID-address protocol", func(t *testing.T) {
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		miner := tutil.NewSECP256K1Addr(t, "bleepbloob")
+		penalty := abi.NewTokenAmount(10)
```
anorth: If not semantically meaningful, use zero. Even better, inline big.Zero where this used.

actors/builtin/reward/reward_test.go
```diff
@@ -1 +1,191 @@
...
+			rt.Call(actor.AwardBlockReward, &reward.AwardBlockRewardParams{
+				Miner:        miner,
+				Penalty:      penalty,
+				GasReward:    gasreward,
+				NominalPower: nominalpwr,
+			})
+		})
+		rt.Verify()
+	})
+	// TODO test reward payable + penalty exceeds balance assertion failure when blockreward function is complete
```
anorth: Probably impossible to exercise this, which is fine.

frrist:
reward actor has changed and this PR is stale, closing.

PR #207, opened 2020-02-26
need to provide PoSt proof type to verification function

laser:
## Why does this PR exist?

Miners's proving sets may contain a variety of sectors, each sealed using a different version of proofs. As such, we need to pass the per-sector `RegisteredProof` value (i.e. the proofs version) to the PoSt verification operation so that candidates and replicas, and proofs can be grouped together by version and dispatched to the correct (internal) verification-functions.

## What's in this PR?

This PR adds a new type `abi.PoStSectorInfo` which is a tuple of `abi.SectorInfo` and `abi.RegisteredProof`. This PR augments the provided `[]abi.SectorInfo` with a `RegisteredProof` using information from the sector pre-commit.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -227,8 +227,9 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
 		// if not a temp fault sector, add to computed proving set
 		if !fault {
 			sectorInfos = append(sectorInfos, abi.SectorInfo{
-				SealedCID:    ssinfo.Info.SealedCID,
-				SectorNumber: ssinfo.Info.SectorNumber,
+				SealedCID:       ssinfo.Info.SealedCID,
+				SectorNumber:    ssinfo.Info.SectorNumber,
+				RegisteredProof: ssinfo.Info.RegisteredProof,
 			})
 		}
 		return nil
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -227,8 +227,9 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
 		// if not a temp fault sector, add to computed proving set
 		if !fault {
 			sectorInfos = append(sectorInfos, abi.SectorInfo{
-				SealedCID:    ssinfo.Info.SealedCID,
-				SectorNumber: ssinfo.Info.SectorNumber,
+				SealedCID:       ssinfo.Info.SealedCID,
+				SectorNumber:    ssinfo.Info.SectorNumber,
+				RegisteredProof: ssinfo.Info.RegisteredProof,
```
anorth: Remember to map the seal proof to the post proof.

Perhaps rename SectorInfo to PoStSectorInfo to clarify.

laser: @anorth 

> Remember to map the seal proof to the post proof.

I was going to do the mapping in the implementation of `VerifyPoSt` so as to keep this diff small. Other structs (e.g. `PoStCandidate`) embed `RegisteredProof`, so for consistency's sake I did the same with `SectorInfo`. Thoughts?

anorth: Ok, great. Could you simply document that in the struct?

laser: done

PR #208, opened 2020-02-26
readdr miner owner param to createminer call

whyrusleeping:
Not sure why this got changed, but theres no reason not to allow someone to create a miner on someone elses behalf. We use this in the faucet to help people set up their miners.


anorth:
FYI @zixuanzh.

PR #214, opened 2020-02-27
Use an empty value for params instead of nil on send

ZenGround0:
This nil param on send requires non-ideal go-filecoin runtime workaround.  I think we want to enforce that all send calls use an adt empty val param when there are no params to send.


anorth:
We've reversed policy on this, nil is now the preferred empty value.

PR #220, opened 2020-03-02
for now, don't abort on poorly timed post check

whyrusleeping:


anorth:
No problem on the surface, but I don't understand where the mistimed check comes from. I see only a single place scheduling the event, when the sector count transitions from zero to one. Is it in the recovery path when a miner was once slashed?

whyrusleeping:
The mistimed call comes from the fact that we can't cancel cron events, and the pre-registered cron event continues to fire anyway

PR #223, opened 2020-03-02
Remove error when unmarshalling unknown signature

ZenGround0:
go-filecoin wants a bit more leniency in unmarshaling so that zero valued signatures can be encoded and decoded.


ZenGround0:
Since I see you got pulled in for review @whyrusleeping fyi this has zero impact on the cbor gen cbor marshalling function and so almost certainly has zero impact on lotus.

PR #226, opened 2020-03-03
Improve bitfield handling

Kubuxu:
The Unset function in go-bitfield is current a bit slow but I can improve it in future
with the same API


actors/builtin/miner/miner_actor.go
```diff
@@ -605,6 +614,18 @@ func (a Actor) checkTemporaryFaultEvents(rt Runtime, sectors *abi.BitField) {
...
+		maxAllowedFaults, err := st.GetMaxAllowedFaults(store)
+		if err != nil {
+			rt.Abortf(exitcode.SysErrInternal, "failed to get number of sectors")
+		}
+		faultsMap, err := st.FaultSet.AllMap(maxAllowedFaults)
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "too many faults")
+		}
+
+		unsets := make(map[uint64]struct{})
```
anorth: This looks to be unused.

Kubuxu: Yeah, it is. Remainder of earlier version.

PR #227, opened 2020-03-03
Accept nil as empty params for Send(); document interpretation of PoSt lookback.

anorth:
FYI @frrist @shannonwells @ingar


support/mock/mockrt.go
```diff
@@ -183,6 +184,9 @@ func (rt *Runtime) Send(toAddr addr.Address, methodNum abi.MethodNum, params run
 	if len(rt.expectSends) == 0 {
 		rt.t.Fatalf("unexpected expectedMessage to: %v method: %v, value: %v, params: %v", toAddr, methodNum, value, params)
 	}
+	if params == nil {
+		params = adt.EmptyValue{}
+	}
```
frrist: Unsure if this is necessary. Could we just use nil?

anorth: Right, on further thought I think we should normalize to nil.

PR #229, opened 2020-03-04
fix reward actor to use miner owner

whyrusleeping:


anorth:
This is explicitly not what I want it to do. Maybe the slack conversation wasn't conclusive.

anorth:
To be more clear: I think a better solution is that the reward actor holds balances on account for the miner actor (the same way balances are held in the market and power actors). Then, upon withdrawal only, the miner's owner address is looked up by calling the MinerControlAddrs method and the payment made to that address.

anorth:
Ref: https://github.com/filecoin-project/specs/issues/787

whyrusleeping:
That assumes that the reward actor is only ever used to pay out to miner actors, which is probably a fine assumption to make. but its definitely additional assumptions (at one weird point in time, multisig vesting also happened in here)

whyrusleeping:
@anorth Addressed that, should be good to go

actors/builtin/reward/reward_actor.go
```diff
@@ -49,10 +49,9 @@ func (a Actor) Constructor(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 }
 
 type AwardBlockRewardParams struct {
-	Miner        addr.Address
-	Penalty      abi.TokenAmount // penalty for including bad messages in a block
-	GasReward    abi.TokenAmount // gas reward from all gas fees in a block
-	NominalPower abi.StoragePower
+	Miner     address.Address
+	Penalty   abi.TokenAmount // penalty for including bad messages in a block
+	GasReward abi.TokenAmount // gas reward from all gas fees in a block
```
anorth: Just FYI, I believe we'll need to add a winning ticket count parameter here later.

actors/builtin/reward/reward_actor.go
```diff
@@ -67,10 +66,14 @@ type AwardBlockRewardParams struct {
...
 	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
-	AssertMsg(params.Miner.Protocol() == addr.ID,
-		"miner address must be ID-address protocol (%d), was protocol %d", addr.ID, params.Miner.Protocol())
 	AssertMsg(rt.CurrentBalance().GreaterThanEqual(params.GasReward),
 		"actor current balance %v insufficient to pay gas reward %v", rt.CurrentBalance(), params.GasReward)
+
+	miner, ok := rt.ResolveAddress(params.Miner)
+	if !ok {
+		rt.Abortf(exitcode.ErrIllegalState, "failed to resolve given owner address")
+	}
```
anorth: The sender here will be the VM/interpreter, which has no excuse not to provide an ID address (which it gets from the block header). So I prefer the assertion, but this is ok.

actors/builtin/reward/reward_actor.go
```diff
@@ -109,14 +112,20 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
 	return &adt.EmptyValue{}
 }
 
-func (a Actor) WithdrawReward(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
+func (a Actor) WithdrawReward(rt vmr.Runtime, minerin *address.Address) *adt.EmptyValue {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
```
anorth: I concur that the miner address needs to be a parameter, but in that case we probably want to lock down the caller to be the owner (or possibly the owner|worker). Wide open means _anyone_ could cause a miner's reward to be paid out to the owner account, which it's hard to see a legitimate reason for, and could be part of an attempted theft.

So I suggest `ValidateImmediateAddr(owner)` after resolving the owner.

actors/builtin/reward/reward_actor.go
whyrusleeping: Ah, yeah. we will need that.

actors/builtin/reward/reward_actor.go
whyrusleeping: I can see many reasons to not want to use your owner key here, I think restricting to owner or worker makes sense

PR #230, opened 2020-03-04
return nil instead of adt.EmptyValue

frrist:


actors/builtin/cron/cron_test.go
```diff
@@ -92,13 +92,13 @@ func (h *cronHarness) constructAndVerify(rt *mock.Runtime, entries ...cron.Entry
 	params := cron.ConstructorParams{Entries: entries}
 	rt.ExpectValidateCallerAddr(builtin.SystemActorAddr)
 	ret := rt.Call(h.Constructor, &params).(*adt.EmptyValue)
-	assert.Equal(h.t, &adt.EmptyValue{}, ret)
+	assert.Nil(h.t, ret)
```
anorth: Skip the cast on the line above (everywhere you made this change)

anorth:
FYI @magik6k

actors/builtin/cron/cron_test.go
frrist: whoops, missed this one. Will do

PR #231, opened 2020-03-04
Fix consensus fault reporting method params and syscall

anorth:
Add third parameter to consensus fault reporting for the parent grinding witness.
Remove parameters that are derived from the block headers, return them from the syscall instead.

### DIFF
--- a/actors/builtin/power/cbor_gen.go
+++ b/actors/builtin/power/cbor_gen.go
@@ -900,7 +900,7 @@ func (t *ReportConsensusFaultParams) MarshalCBOR(w io.Writer) error {
 		_, err := w.Write(cbg.CborNull)
 		return err
 	}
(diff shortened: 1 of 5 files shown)


whyrusleeping:
@anorth this whole method looks super broken. It seems like i can craft two blocks myself that are slashable, then put some other miners info in the parameters.

whyrusleeping:
For fixing the method in general, I recommend looking at the abitrate consensus fault method from lotus: https://github.com/filecoin-project/lotus/blob/master/chain/actors/actor_storagepower.go#L139

anorth:
The comment is far from the whole spec. The block signatures are checked, for one thing.

anorth:
I think you've misunderstood the intent here. The code for arbitrating the fault, that your linked code has inside the actor, goes instead inside the system call. The actor only contains the logic for what to do if a fault is indeed detected.

PR #236, opened 2020-03-06
only count power towards network total once miners break threshold

whyrusleeping:


actors/builtin/power/power_state.go
```diff
@@ -146,10 +146,19 @@ func (st *State) AddToClaim(s adt.Store, miner addr.Address, power abi.StoragePo
...
 	claim.Power = big.Add(claim.Power, power)
 	claim.Pledge = big.Add(claim.Pledge, pledge)
+
+	stillBelow := claim.Power.LessThan(ConsensusMinerMinPower)
+
+	if !prevBelow && !stillBelow {
+		st.TotalNetworkPower = big.Add(st.TotalNetworkPower, power)
+	} else if prevBelow && !stillBelow {
+		st.TotalNetworkPower = big.Add(st.TotalNetworkPower, claim.Power)
+	}
```
anorth: This doesn't appear to handle when a claim is reduced (`power` is negative). That's handled here rather than a duplicative `subtractFromClaim` method. I think you want:
```
if prevBelow != stillBelow { add power }
```

anorth:
Ah, I see #198 is in already. In that case I think this is incomplete unless it also adjusts total power in `put/deleteDetectedFaults`.

actors/builtin/power/power_state.go
sternhenri: argh good catch alex

PR #238, opened 2020-03-07
handle total network power in the presence of faults

whyrusleeping:
It's a bit late so i might not be thinking through this right, but we werent adjusting the total network power up or down on faults. This should roughly do the thing.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -414,6 +414,10 @@ func (a Actor) OnMinerWindowedPoStFailure(rt Runtime, params *OnMinerWindowedPoS
 			rt.Abortf(exitcode.ErrIllegalState, "Failed to check if miner was faulty already: %s", err)
 		}
 
+		if faulty {
+			return nil
+		}
+
 		if err := st.putDetectedFault(adt.AsStore(rt), minerAddr); err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "Failed to put miner fault: %v", err)
 		}
@@ -427,12 +431,9 @@ func (a Actor) OnMinerWindowedPoStFailure(rt Runtime, params *OnMinerWindowedPoS
 			rt.Abortf(exitcode.ErrIllegalState, "Failed to find miner power in claimed power table for surprise PoSt failure")
 		}
 
-		if claim.Power.GreaterThanEqual(ConsensusMinerMinPower) && !faulty {
+		if claim.Power.GreaterThanEqual(ConsensusMinerMinPower) {
 			// Ensure we only deduct this once...
 			st.TotalNetworkPower = big.Sub(st.TotalNetworkPower, claim.Power)
-			if err := st.setClaim(adt.AsStore(rt), minerAddr, claim); err != nil {
-				rt.Abortf(exitcode.ErrIllegalState, "failed to rewrite power claim after slashing miner: %s", err)
-			}
 		}
 		return nil
 	})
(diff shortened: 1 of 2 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -409,19 +409,31 @@ func (a Actor) OnMinerWindowedPoStFailure(rt Runtime, params *OnMinerWindowedPoS
...
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
+		faulty, err := st.hasDetectedFault(adt.AsStore(rt), minerAddr)
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "Failed to check if miner was faulty already: %s", err)
+		}
+
 		if err := st.putDetectedFault(adt.AsStore(rt), minerAddr); err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "Failed to put miner fault: %v", err)
 		}
```
sternhenri: We can avoid this if faulty (in fact we should otherwise we are substracting numMinersMeetingMinPower once too many in `putDetectedFault`). May be worth moving that check from #198 to your below check

actors/builtin/power/power_state.go
```diff
@@ -277,6 +285,7 @@ func (st *State) deleteDetectedFault(s adt.Store, a addr.Address) error {
 	if nominalPower.GreaterThanEqual(ConsensusMinerMinPower) {
 		// just regained a miner > min size
 		st.NumMinersMeetingMinPower++
+		st.TotalNetworkPower = big.Add(st.TotalNetworkPower, claim.Power)
```
anorth: If total network power is to be independent of the min miner sizes (i.e. option (b)) this should be outside the if on 285.

actors/builtin/power/power_actor.go
```diff
@@ -409,19 +409,31 @@ func (a Actor) OnMinerWindowedPoStFailure(rt Runtime, params *OnMinerWindowedPoS
...
 		if !found {
 			rt.Abortf(exitcode.ErrIllegalState, "Failed to find miner power in claimed power table for surprise PoSt failure")
 		}
+
+		if claim.Power.GreaterThanEqual(ConsensusMinerMinPower) && !faulty {
+			// Ensure we only deduct this once...
+			st.TotalNetworkPower = big.Sub(st.TotalNetworkPower, claim.Power)
+			if err := st.setClaim(adt.AsStore(rt), minerAddr, claim); err != nil {
+				rt.Abortf(exitcode.ErrIllegalState, "failed to rewrite power claim after slashing miner: %s", err)
+			}
```
anorth: I don't quite get what's going on with re-setting the claim. The claim object hasn't been modified. I don't think it _should_ be modified: we should hold the claim object as power-less-declared-faults, and use the detected fault set to include/exclude in total power etc.

whyrusleeping: yeah, that was leftover from trying something else

whyrusleeping:
TotalNetworkPower should be (c) from your list of choices. This is the number we use to perform the sortition, doing (b) would mean all the miners who have less than the minimum size would still contribute to total power, and thus making the effective value of 'e' lower (aka, power inflation)

sternhenri:
Per @anorth 's point though, this does mean we need to special case the sortition when no miner is larger than min size...

Kubuxu:
In lotus we used to not allow for removal of last miner.

sternhenri:
Ack. Well this is a lot simpler than the current min_miner_target approach we have... What do you prefer @anorth ? I would suggest we either move like lotus here (and potentially remove all the min miner stuff if no one reached min size per the assumption), or we add the special case to the sortition...

PR #242, opened 2020-03-11
fix sector creation, drop winpost candidate req

whyrusleeping:
These sector fields need to be initialized to -1 otherwise it fails when we try to enumerate the sector set for post.

I also lowered the number of sectors we need down to 1, so simplify implementation work. This is going to change anyways, so no big deal here.

### DIFF
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -30,7 +30,7 @@ const PoStLookback = abi.ChainEpoch(1) // PARAM_FINISH
 const ElectionLookback = PoStLookback // PARAM_FINISH
 
 // Number of sectors to be sampled as part of windowed PoSt
-const NumWindowedPoStSectors = 200 // PARAM_FINISH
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/policy.go
```diff
@@ -30,7 +30,7 @@ const PoStLookback = abi.ChainEpoch(1) // PARAM_FINISH
 const ElectionLookback = PoStLookback // PARAM_FINISH
 
 // Number of sectors to be sampled as part of windowed PoSt
-const NumWindowedPoStSectors = 200 // PARAM_FINISH
+const NumWindowedPoStSectors = 1 // PARAM_FINISH
```
anorth: I'm worried this will mask issues until we push it back up, maybe with little time to resolve whatever they are. What does this simplify?

whyrusleeping: The post we are running now cares about this, The post we will ship with testnet will not.

PR #244, opened 2020-03-11
power: add cron null round test

travisperson:
Test to ensure that cron-events from null rounds are executed.

Amendment to #243

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -563,9 +563,11 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
 
 			cronEvents = append(cronEvents, epochEvents...)
 
-			err = st.clearCronEvents(store, epoch)
-			if err != nil {
-				return errors.Wrapf(err, "failed to clear cron events at %v", epoch)
+			if len(epochEvents) > 0 {
+				err = st.clearCronEvents(store, epoch)
+				if err != nil {
+					return errors.Wrapf(err, "failed to clear cron events at %v", epoch)
+				}
 			}
 		}
 
--- a/actors/builtin/power/power_test.go
+++ b/actors/builtin/power/power_test.go
@@ -24,7 +24,6 @@ import (
 
 func TestConstruction(t *testing.T) {
 	actor := spActorHarness{power.Actor{}, t}
-	powerActor := tutil.NewIDAddr(t, 100)
 
 	owner1 := tutil.NewIDAddr(t, 101)
 	worker1 := tutil.NewIDAddr(t, 102)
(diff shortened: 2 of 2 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -563,9 +563,11 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
...
 			cronEvents = append(cronEvents, epochEvents...)
 
-			err = st.clearCronEvents(store, epoch)
-			if err != nil {
-				return errors.Wrapf(err, "failed to clear cron events at %v", epoch)
+			if len(epochEvents) > 0 {
+				err = st.clearCronEvents(store, epoch)
+				if err != nil {
+					return errors.Wrapf(err, "failed to clear cron events at %v", epoch)
+				}
```
frrist: From investigation we did together, `Delete` isn't idempotent and the error being returned: https://github.com/ipfs/go-hamt-ipld/blob/d4e402c05ab0aa9fa74d5d5f84533108b7f048da/hamt.go#L370

Not sure if we should be checking the returned error (:thinking:), checking for existence before deleting (:nauseated_face:), or mofiy go-hamt-ipld s.t. `Delete` is idempotent (:star:)

frrist: cc @whyrusleeping @anorth what do y'all think?

frrist: also, errors returned inside here are being ignored. Here is an example of how we have handled these in the past: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/power_actor.go#L603

actors/builtin/power/power_test.go
```diff
@@ -22,10 +25,16 @@ import (
 func TestConstruction(t *testing.T) {
 	actor := spActorHarness{power.Actor{}, t}
 	powerActor := tutil.NewIDAddr(t, 100)
```
frrist: Since you are already changing this file, can this be changed `builtin_spec.StoragePowerActorAddr`.

actors/builtin/power/power_actor.go
anorth: 🤢

actors/builtin/power/power_test.go
```diff
@@ -22,10 +25,16 @@ import (
...
 	owner1 := tutil.NewIDAddr(t, 101)
 	worker1 := tutil.NewIDAddr(t, 102)
 	miner1 := tutil.NewIDAddr(t, 103)
-	unused := tutil.NewIDAddr(t, 104)
+
+	owner2 := tutil.NewIDAddr(t, 104)
+	worker2 := tutil.NewIDAddr(t, 105)
+	miner2 := tutil.NewIDAddr(t, 106)
+
+	unused := tutil.NewIDAddr(t, 107)
```
anorth: Maybe use `9999` so the name stays accurate easily

actors/builtin/power/power_test.go
```diff
@@ -105,6 +114,91 @@ func TestConstruction(t *testing.T) {
...
+			rt.SetReceived(abi.NewTokenAmount(1))
+			rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+
+			createMinerRet := &power.CreateMinerReturn{
+				IDAddress:     miner,  // miner actor id address
+				RobustAddress: robust, // should be long miner actor address
+			}
+			createMinerBuf := new(bytes.Buffer)
+			err := createMinerParams.MarshalCBOR(createMinerBuf)
+			require.NoError(t, err)
```
anorth: `createMinerBuf` is unnecessary

actors/builtin/power/power_test.go
```diff
@@ -105,6 +114,91 @@ func TestConstruction(t *testing.T) {
...
+
+			initCreateMinerBuf := new(bytes.Buffer)
+			err = initCreateMinerParams.MarshalCBOR(initCreateMinerBuf)
+			require.NoError(t, err)
```
anorth: Break the creation and serializatio of initCreateMinerParams into a function, because it will be re-used a lot.

PR #248, opened 2020-03-21
Add a proposal hash to multisig "Approve" message params

droark:
As written, multisig has an issue. When approving a message, approvers have no cryptographic method for confirming that they are truly signing a message related to the original proposal. Add a proposal hash field to the "Approve" message params. The field contains a BLAKE2B-256 hash of the CBOR bytes of the multisig wallet address, the proposer's address, and the fields in the multisig "Propose" params. When an approval occurs, the actor code compares the hash it receives with the data from the original transaction it received. If there's a match, the approval proceeds. If not, the code errors out.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -161,6 +161,12 @@ func (a Actor) Cancel(rt vmr.Runtime, params *TxnIDParams) *adt.EmptyValue {
 			rt.Abortf(exitcode.ErrForbidden, "Cannot cancel another signers transaction")
 		}
 
+                // confirm the hashes match
+                calculatedHash := a.getApprovalHash(rt, txn)
+                if params.ProposalHash != nil && bytes.Compare(params.ProposalHash, calculatedHash[:]) != 0 {
+                        rt.Abortf(exitcode.ErrIllegalState, "hash does not match proposal params")
+                }
+
 		if err = st.deletePendingTransaction(adt.AsStore(rt), params.ID); err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "failed to delete transaction for cancel: %v", err)
 		}
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -43,11 +43,11 @@ type Transaction struct {
 // Subtle point: The "To" field must be the multisig wallet address, otherwise the hash
 // loses much of its value, if not all of it.
 type ProposalHashData struct {
-        Requester addr.Address
-        To        addr.Address
-        Value     abi.TokenAmount
-        Method    abi.MethodNum
-        Params    []byte
+	Requester addr.Address
+	To        addr.Address
+	Value     abi.TokenAmount
+	Method    abi.MethodNum
+	Params    []byte
 }
 
 type Actor struct{}
@@ -136,8 +136,10 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
 		return nil
 	})
 
-	// Proposal implicitly includes approval of a transaction.
-	a.approveTransaction(rt, txnID)
+	// Proposal implicitly includes approval of a transaction. Bypass hash check
+	// because PROPOSE is the reference point for other approvers.
+	var emptyArray []byte
+	a.approveTransaction(rt, txnID, emptyArray, false)
 
 	// Note: this ID may not be stable across chain re-orgs.
 	// https://github.com/filecoin-project/specs-actors/issues/7
@@ -147,7 +149,8 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
 }
 
 type TxnIDParams struct {
-	ID TxnID
+	ID           TxnID
+	ProposalHash []byte
 }
 
 func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *adt.EmptyValue {
@@ -158,7 +161,7 @@ func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *adt.EmptyValue {
 		a.validateSigner(rt, &st, callerAddr)
 		return nil
 	})
-	a.approveTransaction(rt, params.ID)
+	a.approveTransaction(rt, params.ID, params.ProposalHash, true)
 	return nil
 }
 
@@ -179,7 +182,7 @@ func (a Actor) Cancel(rt vmr.Runtime, params *TxnIDParams) *adt.EmptyValue {
 		}
 
 		// confirm the hashes match
-                calculatedHash := a.GetProposalHash(rt, txn)
+		calculatedHash := a.GetProposalHash(rt, txn)
 		if params.ProposalHash != nil && bytes.Compare(params.ProposalHash, calculatedHash[:]) != 0 {
 			rt.Abortf(exitcode.ErrIllegalState, "hash does not match proposal params")
 		}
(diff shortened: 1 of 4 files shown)


anorth:
Hey @droark, thanks for this proposal. We don't have the bandwidth to evaluate it in the next few days, but do intend to at some point soon. Hang tight.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -23,15 +25,24 @@ func (t TxnID) Key() string {
 }
 
 type Transaction struct {
-	To     addr.Address
-	Value  abi.TokenAmount
-	Method abi.MethodNum
-	Params []byte
+        To         addr.Address
```
whyrusleeping: theres a bunch of whitespace only changes here, did you run `go fmt` before pushing?

droark: Hi. The only commands I recall issuing are `make`, `make gen`, and `make test`. It's possible I ran something by accident and don't remember. It's also possible my IDE did something weird. Either way, if it's a problem, I can undo it one way or another, at least for the non-generated code.

arajasek: Yeah, your IDE might have decided to do some funny business...Running `go fmt` on the changed files should get you on the same formatting as us.

droark: Thanks. `gp fmt` did the trick.

droark:
Hello. Regarding the current state of the patch, there are a couple of things that need to be added or considered.

- Add `CANCEL` support.
- I'm considering switching out the hash with the CID of the original proposal. This is more comprehensive but it might be overkill, and it would make the patch a bit more intrusive. I hope to make a final decision by tomorrow.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -32,6 +34,15 @@ type Transaction struct {
 	Approved []addr.Address
 }
 
+type ApprovalHashData struct {
```
dwhjames, line 37: naming inconsistency:
this is called `ApprovalHashData` but its hash in `TxnIDParams` is call `ProposalHash`, and its contents is the `ProposeParams` (+ `MsigWallet` and `Requester`)

actors/builtin/multisig/multisig_actor.go
```diff
@@ -32,6 +34,15 @@ type Transaction struct {
...
+	MsigWallet addr.Address
```
dwhjames, line 38: unneeded as the address of the multisig actor is the To of the Approve/Cancel Message, and so is already protected by signing scheme

actors/builtin/multisig/multisig_actor.go
```diff
@@ -279,6 +297,32 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
...
 
+func (ahd *ApprovalHashData) Serialize() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	if err := ahd.MarshalCBOR(buf); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func (a Actor) getApprovalHash(rt vmr.Runtime, txn Transaction) [32]byte {
```
dwhjames, line 308: naming inconsistency following from `ApprovalHashData`
this is the hash of the proposal

actors/builtin/multisig/multisig_actor.go
```diff
@@ -279,6 +305,31 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
...
+		Method:    txn.Method,
+		Params:    txn.Params,
+	}
+
+	data, err := hashData.Serialize()
+	if err != nil {
+		rt.Abortf(exitcode.ErrIllegalState, "failed to construct multisig approval hash: %v", err)
+	}
+
+	return blake2b.Sum256(data)
```
whyrusleeping, line 330: use `rt.Syscalls().HashBlake2b`, trying to stick with the ethereum opcode approach for these sorts of things.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -328,6 +379,62 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID) {
 	}
 }
 
+func (a Actor) approveTransactionWithHash(rt vmr.Runtime, params *TxnIDParams) {
```
whyrusleeping, line 382: I think we can just delete the old `approveTransaction` method, right?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -279,6 +305,31 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
...
 
+func (ahd *ProposalHashData) Serialize() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	if err := ahd.MarshalCBOR(buf); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func (a Actor) GetProposalHash(rt vmr.Runtime, txn Transaction) [32]byte {
```
whyrusleeping, line 316: lets just return a `[]byte` so its cleaner for the user

actors/builtin/multisig/multisig_actor.go
droark: This call seems to be killing the unit tests I wrote/modded. I'm seeing the following when the unit tests are run.

```
--- FAIL: TestVesting (0.00s)
    --- FAIL: TestVesting/happy_path_full_vesting (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
    --- FAIL: TestVesting/partial_vesting_propose_to_send_half_the_actor_balance_when_the_epoch_is_hald_the_unlock_duration (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
    --- FAIL: TestVesting/fail_to_vest_more_than_locked_amount (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 19 but call succeeded
--- FAIL: TestApprove (0.00s)
    --- FAIL: TestApprove/simple_propose_and_approval (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
    --- FAIL: TestApprove/fail_approval_with_bad_proposal_hash (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 20 but call succeeded
    --- FAIL: TestApprove/fail_approve_transaction_more_than_once (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 20 but call succeeded
    --- FAIL: TestApprove/fail_approve_transaction_that_does_not_exist (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 17 but call succeeded
    --- FAIL: TestApprove/fail_to_approve_transaction_by_non-signer (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 18 but call succeeded
--- FAIL: TestCancel (0.00s)
    --- FAIL: TestCancel/simple_propose_and_cancel (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
    --- FAIL: TestCancel/fail_cancel_with_bad_proposal_hash (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 20 but call succeeded
    --- FAIL: TestCancel/signer_fails_to_cancel_transaction_from_another_signer (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 18 but call succeeded
    --- FAIL: TestCancel/fail_to_cancel_transaction_when_not_signer (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 18 but call succeeded
    --- FAIL: TestCancel/fail_to_cancel_a_transaction_that_does_not_exist (0.00s)
        mockrt.go:594: invalid runtime invocation outside of method call
        mockrt.go:507: expected abort with code 17 but call succeeded
```

actors/builtin/multisig/multisig_actor.go
droark: We can't as written. `approveTransaction` is used to institute the `PROPOSE` approval. There won't be a hash to check. That said, I can rewrite the code such that it uses a boolean to bypass the hash check in such a case. I'll do that.

droark:
Force pushed my fixes. The unit tests are still failing.

PR #253, opened 2020-03-24
Remove duplicate balance add in power actor state

austinabell:
amount is being added twice, which does not affect anything regarding consensus because the root is only updated after the first, but there are unnecessary values written to store (and of course unnecessary logic). 

With your implementation the call the second time is written to the underlying data storage, meaning you have a data leak of sorts having to store this updated state which may never be used.

### DIFF
--- a/actors/builtin/power/power_state.go
+++ b/actors/builtin/power/power_state.go
@@ -136,10 +136,11 @@ func (st *State) setMinerBalance(store adt.Store, miner addr.Address, amount abi
 func (st *State) addMinerBalance(store adt.Store, miner addr.Address, amount abi.TokenAmount) error {
 	Assert(amount.GreaterThanEqual(big.Zero()))
 	table := adt.AsBalanceTable(store, st.EscrowTable)
-	if table.Add(miner, amount) == nil {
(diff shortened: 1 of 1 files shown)


actors/builtin/power/power_state.go
```diff
@@ -136,10 +136,11 @@ func (st *State) setMinerBalance(store adt.Store, miner addr.Address, amount abi
 func (st *State) addMinerBalance(store adt.Store, miner addr.Address, amount abi.TokenAmount) error {
 	Assert(amount.GreaterThanEqual(big.Zero()))
 	table := adt.AsBalanceTable(store, st.EscrowTable)
-	if table.Add(miner, amount) == nil {
+	err := table.Add(miner, amount)
+	if err == nil {
```
anorth: Would you mind rewriting this as:

```
if err := table.Add(miner, amount); err != nil {
	return err
}
st.EscrowTable = table.Root()
return nil
```

... which is more canonical Go. Someone (probably me) trying to be tricky caused the bug in the first place.

(I know Go isn't your primary language, thanks for the fixes)

PR #254, opened 2020-03-24
Remove conversion to bigint through a i64 value

austinabell:
Edge case where the values are greater than the max i64 value, incorrect value will be used


anorth:
Thanks

PR #262, opened 2020-03-29
Update amount vested utility function

austinabell:
- Removes unnecessary big.Int allocations
- Only initialize elapsed on linear vesting (not used for non-vesting)
- Panic on invalid vesting function instead of just returning 0
  - Obviously doesn't matter now because `None` vesting function is hard coded, but just want to make sure the call will not succeed in future if an invalid function

The heap allocations and verbose checks were hard to look at. Let me know what you would want the doc comment or panic message to say, or if you want this to behave differently than this

### DIFF
--- a/actors/builtin/reward/reward_state.go
+++ b/actors/builtin/reward/reward_state.go
@@ -115,7 +115,8 @@ func (r *Reward) AmountVested(currEpoch abi.ChainEpoch) abi.TokenAmount {
 			return r.Value
 		}
 
-		// totalReward * elapsedEpoch / vestDuration
+		// (totalReward * elapsedEpoch) / vestDuration
+		// Division must be done last to avoid precision loss with integer values
 		return big.Div(big.Mul(r.Value, big.NewInt(int64(elapsed))), big.NewInt(int64(vestDuration)))
 	default:
 		return abi.NewTokenAmount(0)
--- a/actors/builtin/reward/reward_state.go
+++ b/actors/builtin/reward/reward_state.go
@@ -1,6 +1,8 @@
 package reward
 
 import (
+	"fmt"
+
 	addr "github.com/filecoin-project/go-address"
 	cid "github.com/ipfs/go-cid"
 	errors "github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


actors/builtin/reward/reward_state.go
```diff
@@ -103,20 +105,21 @@ func (st *State) withdrawReward(store adt.Store, owner addr.Address, currEpoch a
...
+		if elapsed >= vestDuration {
 			return r.Value
 		}
 
 		// totalReward * elapsedEpoch / vestDuration
-		return big.Div(big.Mul(r.Value, big.NewInt(int64(elapsed))), vestDuration)
+		return big.Mul(r.Value, big.NewInt(int64(elapsed/vestDuration)))
 	default:
-		return abi.NewTokenAmount(0)
+		panic(fmt.Sprintf("invalid vesting function %v", r.VestingFunction))
```
anorth: On balance I don't think we should panic here. A panic means "stop the blockchain" if such a message gets in a block somehow, and is quite difficult to deal with in validation. While I agree it certainly signals a programmer error somewhere if this happens, keeping funds locked while we arrange a protocol upgrade to fix the underlying issue is probably better than stopping.

actors/builtin/reward/reward_state.go
```diff
@@ -103,20 +105,21 @@ func (st *State) withdrawReward(store adt.Store, owner addr.Address, currEpoch a
...
-		if big.NewInt(int64(elapsed)).GreaterThanEqual(vestDuration) {
+		elapsed := currEpoch - r.StartEpoch
+		vestDuration := r.EndEpoch - r.StartEpoch
+		if elapsed >= vestDuration {
 			return r.Value
 		}
 
 		// totalReward * elapsedEpoch / vestDuration
-		return big.Div(big.Mul(r.Value, big.NewInt(int64(elapsed))), vestDuration)
+		return big.Mul(r.Value, big.NewInt(int64(elapsed/vestDuration)))
```
anorth: This changes the behaviour incorrectly. `elapsed/vestDuration` as a  real number is always < 1 here, which means this integer division will truncate to zero. The division must be performed as the _last_ operation in the expression. Deferring the bigint allocation until here is good though, thanks.

This is a general guideline for all integer arithmetic in the project. But since it tripped up at least one person here, would you add a comment explaining it to the next reader?

actors/builtin/reward/reward_state.go
austinabell: Oh, assumed since equivalent panics happen for similar checks that it would be okay, but I assume the difference is that you don't want this done on reward state for a call outside of the VM execution?

actors/builtin/reward/reward_state.go
austinabell: ah, yes, a lapse on my part. I'll fix that ordering now

actors/builtin/reward/reward_state.go
anorth: They're judgement calls, and I'm not yet confident in them all yet. Some things would be so broken if violated that halting seems reasonable.

PR #264, opened 2020-03-31
Baseline Minting 

zixuanzh:


actors/builtin/reward/reward_state.go
```diff
@@ -86,9 +146,6 @@ func (st *State) withdrawReward(store adt.Store, owner addr.Address, currEpoch a
 		return big.Zero(), err
 	}
 
-	AssertMsg(withdrawableSum.LessThan(st.RewardTotal), "withdrawable %v exceeds recorded prior total %v",
-		withdrawableSum, st.RewardTotal)
```
davidad, line 90: Does #269 resolve this?

actors/builtin/reward/reward_test.go
```diff
@@ -65,5 +65,4 @@ func (h *rewardHarness) constructAndVerify(rt *mock.Runtime) {
 	rt.GetState(&st)
 	emptyMap := adt.AsMultimap(adt.AsStore(rt), st.RewardMap)
 	assert.Equal(h.t, emptyMap.Root(), st.RewardMap)
-	assert.Equal(h.t, big.Zero(), st.RewardTotal)
```
davidad, line 68: Does anyone know why this needs to be removed?

zixuanzh: RewardTotal is the sum of unwithdrawn but minted funds, RewardActor.balance - RewardTotal is the amount of FIL that has not been minted in supply. We are moving away from the iterative computation for block reward and will send funds to miner actor right after a reward is minted and hence no need for keep tracking of RewardTotal.

actors/abi/big/int.go
```diff
@@ -75,6 +75,21 @@ func Exp(a Int, e Int) Int {
...
+func Lsh(a Int, n uint) Int {
+	return Int{big.NewInt(0).Lsh(a.Int, n)}
+}
+
+// Returns x >> n
+func Rsh(a Int, n uint) Int {
+	return Int{big.NewInt(0).Rsh(a.Int, n)}
+}
+
+// TODO: check this
```
whyrusleeping, line 88: seems fine

PR #268, opened 2020-04-02
finalize constants; add tests

davidad:
attn @zixuanzh @whyrusleeping


davidad:
The only meaningful change I made that we hadn't already discussed is removing `taylorSeriesExpansion`'s dependence on State. It is a pure function, it's used as such, and it was easier to test that way. Hoping that doesn't violate some convention. There are also some whitespace changes that were generated by `go fmt`.

PR #269, opened 2020-04-03
Reward state should allow withdrawableSum to equal RewardTotal

arajasek:
There's no conceptual reason to want withdrawableSum to be strictly less.

### DIFF
--- a/actors/builtin/reward/reward_test.go
+++ b/actors/builtin/reward/reward_test.go
@@ -2,6 +2,7 @@ package reward_test
 
 import (
 	"context"
+	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


actors/builtin/reward/reward_test.go
```diff
@@ -50,6 +51,40 @@ func TestAwardBlockReward(t *testing.T) {
...
+
+	t.Run("successfully withdraw reward total", func(t *testing.T) {
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		gasreward := abi.NewTokenAmount(10)
+		rt.SetBalance(abi.NewTokenAmount(100000000))
+
+		rt.ExpectValidateCallerAddr(builtin.SystemActorAddr)
+		rt.ExpectSend(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, big.Zero(), nil, 0)
```
anorth, line 71: Would you mind optimizing out the burning of zero funds in the actor?

actors/builtin/reward/reward_test.go
```diff
@@ -50,6 +51,40 @@ func TestAwardBlockReward(t *testing.T) {
...
+		WithCaller(builtin.SystemActorAddr, builtin.SystemActorCodeID).
+		WithActorType(m, builtin.StorageMinerActorCodeID).
+		WithActorType(owner, builtin.AccountActorCodeID)
+
+	t.Run("successfully withdraw reward total", func(t *testing.T) {
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		gasreward := abi.NewTokenAmount(10)
+		rt.SetBalance(abi.NewTokenAmount(100000000))
```
anorth, line 68: This is an artificially low balance that is clamping the total reward payable.
- If you intended this, please document it, and please also remove the non-zero gas reward since it is not payable
- If you didn't, set this value to be above the `BlockRewardTarget`, and expect 100*10^18 + 10 (gas reward) tokens to be withdrawn

arajasek:
With #264, the change to source code in this PR is redundant. Gonna close it for now, will reopen to add the test when reward logic is in its final state.

PR #271, opened 2020-04-03
Pledge structure

zixuanzh:


actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,184 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+	return nil
+}
+
+func (st *State) slashPledge(store adt.Store, currEpoch abi.ChainEpoch, amountToSlash abi.TokenAmount) (abi.TokenAmount, error) {
+
+	// var pc PledgeCollateral
+	// store.Get(st.PledgeCollateral, &pc)
+
+	// pledges := adt.AsMultimap(store, pc.LockedFunds)
+	// indices, err := pc.LockedIndices.All(32 << 20)
```
davidad, line 321: This number is exactly the number of sectors a 1EiB miner would have, so maybe a little small for an absolute upper limit here

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,184 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+	// 	lockedFunds, err = adt.MakeEmptyArray(store)
+	// 	if err != nil {
+	// 		return errors.Wrapf(err, "failed to initialize in add pledge: %v", err)
+	// 	}
+	// }
+
+	// lockedFunds.AppendContinuous(pledge)
+	// // update the bitfield
+	// pc.LockedIndices.Set(uint64(pledge.CliffStartEpoch)) // TODO: ChainEpoch should be uint64
+	// pc.LockedFunds = lockedFunds.Root()
```
davidad, line 307: I think `lockedFunds` is to be inserted or updated within `pledges`; it is `pledges.Root()` which then needs to be updated into `pc` here.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,184 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+
+	// amountSlashed := big.Zero()
+	// remainToSlash := amountToSlash
+
+	// for _, index := range indices {
+	// 	// AMT[LockedFund]
+	// 	lockedFunds, found, err := pledges.Get(index)
+	// 	Assert(found)
+	// 	if err != nil {
+	// 		return big.Zero(), errors.Wrapf(err, "failed to locate lockd fund: %w", err)
```
davidad, line 334: typo

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +272,129 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
 
+func (st *State) addLockedFund(store adt.Store, currEpoch abi.ChainEpoch, pledgeAmount abi.TokenAmount) error {
+	Assert(pledgeAmount.GreaterThanEqual(big.Zero()))
+
+	lockedFundsArray := adt.AsArray(store, st.LockedFundsArray)
+
+	cliffStartEpoch := currEpoch + PledgeCliff
+	fullyVestedEpoch := currEpoch + PledgeCliff + PledgeVestingPeriod
+
+	for i := cliffStartEpoch; i < fullyVestedEpoch; i += VestIncrement {
```
whyrusleeping, line 283: Quick order of magnitude for how many times this loop executes on average?

whyrusleeping: ah, from the constants above it looks like 365 times. Thats gonna be reallly expensive. Can we cut this down to maybe like 10?

whyrusleeping: Also, i think its extremely unlikely that you'll get any overlap here if youre keying this on a block height, and using the current epoch as your base. Maybe would be better if you rounded it to every ~10k blocks or something?

whyrusleeping: Basically, as written, youre gonna rewrite the entire AMT every time this method is called. Thats not acceptable

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +272,129 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+		var lockedFundEntry abi.TokenAmount
+		quantizedVestEpoch := getQuantizedEpoch(i)
+		found, err := lockedFundsArray.Get(epochKey(quantizedVestEpoch), &lockedFundEntry)
```
whyrusleeping, line 286: lets add a TODO note here, these fancy ADT types do a full flush on every operation. We should avoid doing that.

actors/builtin/miner/miner_state.go
zixuanzh: We can parametrize this to vest on a weekly basis and the quantization will also help.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+
+func (st *State) slashLockedFund(store adt.Store, currEpoch abi.ChainEpoch, amountToSlash abi.TokenAmount) (abi.TokenAmount, error) {
+
+	lockedFundsArray := adt.AsArray(store, st.LockedFundsArray)
+	amountSlashed := big.Zero()
+	remainToSlash := amountToSlash
+
+	var lockedFund abi.TokenAmount
+	var toDelete []uint64
+	// Assume ordering in lockedFundsArray by CliffStartEpoch
```
anorth, line 317: This comment doesn't make any sense. the entries are in order of release from vesting.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+	err := lockedFundsArray.ForEach(&lockedFund, func(k int64) error {
+		if remainToSlash.GreaterThan(big.Zero()) {
+			if k >= int64(currEpoch) {
+				// slashable
+				slashingAmount := big.Max(remainToSlash, lockedFund)
+				lockedFund = big.Sub(lockedFund, slashingAmount)
+
+				amountSlashed = big.Add(amountSlashed, slashingAmount)
+				remainToSlash = big.Sub(remainToSlash, slashingAmount)
```
anorth, line 326: You don't need to track both these in the loop. Just track one and compute the other at the end.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+				Assert(remainToSlash.GreaterThanEqual(big.Zero()))
+
+				if lockedFund.IsZero() {
+					// we can delete this entry
+					toDelete = append(toDelete, uint64(k))
+				}
+			}
+			return nil
+		} else {
+			return errorContinue
```
anorth, line 336: "finished" would make more sense than "continue", since it does the opposute to what the `continue` keyword in many languages does.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+		}
+	})
+
+	if err != nil {
+		if err != errorContinue {
```
anorth, line 341: Collapse to one line

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+			return big.Zero(), err
+		}
+	}
+
+	// TODO: expose batch delete on AMT
```
anorth, line 346: That's not a TODO in this repo. File an issue on AMT and change this to "If AMT exposed a batch delete, we could use that to save some writes here".

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+// iterate over LockedFunds and delete fully vested locked funds and return newly unlocked funds
+// update st.TotalLockedPledge
+func (st *State) vestNewFunds(store adt.Store, currEpoch abi.ChainEpoch) (abi.TokenAmount, error) {
+
+	lockedFundsArray := adt.AsArray(store, st.LockedFundsArray)
+	newlyUnlockedFund := big.Zero()
+
+	var lockedFund abi.TokenAmount
+	var toDelete []uint64
+	// Assume ordering in lockedFundsArray by CliffStartEpoch
```
anorth, line 367: Remove duplicated confusing comment

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+			toDelete = append(toDelete, uint64(k))
+		} else {
+			// stop iterating remaining funds
+			return errorContinue
+		}
+		return nil
+	})
+
+	if err != nil {
+		if err != errorContinue {
```
anorth, line 381: Collapse

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+			return big.Zero(), err
+		}
+	}
+
+	// TODO: expose batch delete on AMT
```
anorth, line 386: Ditto

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +273,138 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+	for _, i := range toDelete {
+		err = lockedFundsArray.Delete(i)
+		if err != nil {
+			return big.Zero(), errors.Wrapf(err, "failed to delete locked fund during slash: %v", err)
+		}
+	}
+
+	st.LockedFundsArray = lockedFundsArray.Root()
```
anorth, line 354: Total locked pledge is now inconsistent with the sum of the lock-ups

actors/builtin/miner/miner_state.go
```diff
@@ -279,6 +425,11 @@ func sectorKey(e abi.SectorNumber) adt.Keyer {
 	return adt.UIntKey(uint64(e))
 }
 
+func epochKey(e abi.ChainEpoch) uint64 {
+	// TODO: ChainEpoch is int64
```
anorth, line 429: This is probably fine

actors/builtin/power/power_actor.go
```diff
@@ -229,33 +151,32 @@ func (a Actor) DeleteMiner(rt Runtime, params *DeleteMinerParams) *adt.EmptyValu
...
+			rt.Abortf(exitcode.ErrIllegalState, "deletion requested for miner %v with raw byte power %v", nominal, claim.RawBytePower)
+		}
+		if claim.QualityAdjPower.GreaterThan(big.Zero()) {
+			rt.Abortf(exitcode.ErrIllegalState, "deletion requested for miner %v with quality adjusted power %v", nominal, claim.QualityAdjPower)
+		}
 
-	if balance.GreaterThan(abi.NewTokenAmount(0)) {
-		rt.Abortf(exitcode.ErrForbidden, "deletion requested for miner %v with pledge balance %v", nominal, balance)
-	}
+		ownerAddr, workerAddr := builtin.RequestMinerControlAddrs(rt, nominal)
```
anorth, line 170: Illegal send within a transaction

actors/builtin/power/power_actor.go
```diff
@@ -516,18 +401,18 @@ type ReportConsensusFaultParams struct {
 func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultParams) *adt.EmptyValue {
 	currEpoch := rt.CurrEpoch()
 	earliest := currEpoch - ConsensusFaultReportingWindow
-	fault, err := rt.Syscalls().VerifyConsensusFault(params.BlockHeader1, params.BlockHeader2, params.BlockHeaderExtra, earliest)
+	reporter := rt.Message().Caller()
+	fault, err := rt.Syscalls().VerifyConsensusFault(params.BlockHeader1, params.BlockHeader2, params.BlockHeaderExtra, earliest, reporter)
```
anorth, line 405: How is the reporter relevant to verifying the fault?

actors/builtin/power/power_actor.go
```diff
@@ -590,10 +470,44 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	return nil
 }
 
+func (a Actor) UpdatePledgeCollateralTally(rt Runtime, pledgeAmount abi.TokenAmount) *adt.EmptyValue {
+	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
+	msgValue := rt.Message().ValueReceived()
+	if !msgValue.IsZero() {
+		rt.Abortf(exitcode.ErrIllegalArgument, "fund is not required to update pledge collateral tally")
+	}
```
anorth, line 478: This is silly. We could put a similar check on all methods that don't expect a value, but we don't. Writing a test for the miner would be a great alternative.

actors/runtime/exitcode/reserved.go
```diff
@@ -102,20 +102,20 @@ const (
...
+	SysErrOutOfGas:           "SysErrOutOfGas",
+	SysErrForbidden:          "SysErrForbidden",
+	SysErrorIllegalActor:     "SysErrorIllegalActor",
+	SysErrorIllegalArgument:  "SysErrorIllegalArgument",
+	SysErrSerialization:      "SysErrSerialization",
+	SysErrorReserved3:        "SysErrorReserved3",
+	SysErrorReserved4:        "SysErrorReserved4",
+	SysErrorReserved5:        "SysErrorReserved5",
+	SysErrInternal:           "SysErrInternal",
+}
```
anorth, line 121: Unrelated cosmetic change

actors/runtime/runtime.go
```diff
@@ -129,7 +129,7 @@ type Syscalls interface {
 	// the "parent grinding fault", in which case it must be the sibling of h1 (same parent tipset) and one of the
 	// blocks in the parent of h2 (i.e. h2's grandparent).
 	// Returns nil and an error if the headers don't prove a fault.
-	VerifyConsensusFault(h1, h2, extra []byte, earliest abi.ChainEpoch) (*ConsensusFault, error)
+	VerifyConsensusFault(h1, h2, extra []byte, earliest abi.ChainEpoch, reporter addr.Address) (*abi.ConsensusFault, error)
```
anorth, line 132: ooooooh, nasty. Reporter has been added here just to get it into the ConsensusFault object so that object can then be used as a parameter to a method? Please don't do that.
- if you re-organize as I suggested, this won't be necessary at all
- even if you don't, the SlashConsensusFault method should take a parameters struct with exactly the info it needs, which is _not_ the full ConsensusFault object but would include the reporter.

gen/gen.go
```diff
@@ -81,7 +82,7 @@ func main() {
 		// method params
 		reward.AwardBlockRewardParams{},
 		// other types
-		reward.Reward{},
+		// reward.Reward{},
```
anorth, line 85: Delete

anorth:
Where are the tests?

actors/builtin/miner/miner_actor.go
```diff
@@ -336,18 +354,24 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
 	)
 	builtin.RequireSuccess(rt, code, "failed to notify power actor")
-	AssertNoError(ret.Into(&pledgeRequirement))
+	AssertNoError(ret.Into(&initialPledge))
 
-	// add sector to miner state
+	// add sector and pledge to miner state
 	rt.State().Transaction(&st, func() interface{} {
+
+		_, err = st.vestNewFunds(adt.AsStore(rt), rt.CurrEpoch())
```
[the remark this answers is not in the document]
zixuanzh: this was to save a transaction to power actor as the total pledge is updated in power.OnSectorProveCommit but we can do the tally here for more accuracy.

actors/builtin/miner/miner_actor.go
```diff
@@ -387,6 +420,7 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 	a.enrollCronEvent(rt, precommit.Info.Expiration, &cronPayload)
 
 	// If first sector
+	// TODO: move this into a transaction
```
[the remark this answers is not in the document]
zixuanzh: I see.

actors/builtin/miner/miner_actor.go
```diff
@@ -579,6 +618,125 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params *DeclareTemporaryFaults
...
+type WithdrawBalanceParams struct {
+	AmountRequested abi.TokenAmount
+}
+
+func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.EmptyValue {
+	var st State
+	currBalance := rt.CurrentBalance()
+
+	if params.AmountRequested.LessThanEqual(big.Zero()) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "negative fund requested for withdrawal: %w", params.AmountRequested)
```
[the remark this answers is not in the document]
zixuanzh: allowing for zero

actors/builtin/miner/miner_state.go
```diff
@@ -81,17 +87,25 @@ type SectorOnChainInfo struct {
 	Info                  SectorPreCommitInfo
 	ActivationEpoch       abi.ChainEpoch  // Epoch at which SectorProveCommit is accepted
 	DealWeight            abi.DealWeight  // Integral of active deals over sector lifetime, 0 if CommittedCapacity sector
-	PledgeRequirement     abi.TokenAmount // Fixed pledge collateral requirement determined at activation
+	InitialPledge         abi.TokenAmount // initial pledge collateral requirement determined at activation
```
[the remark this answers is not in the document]
zixuanzh: removing InitialPledge from SectorOnChainInfo

actors/builtin/miner/policy.go
```diff
@@ -56,3 +56,65 @@ const ProvingPeriod = 300
...
+	slasherShareNumerator := big.Exp(consensusFaultReporterShareGrowthRate.numerator, elapsed)
+	slasherShareDenominator := big.Exp(consensusFaultReporterShareGrowthRate.denominator, elapsed)
+
+	num := big.Mul(big.Mul(slasherShareNumerator, consensusFaultReporterInitialShare.numerator), collateral)
+	denom := big.Mul(slasherShareDenominator, consensusFaultReporterInitialShare.denominator)
+	return big.Min(big.Div(num, denom), big.Div(big.Mul(collateral, maxReporterShareNum), maxReporterShareDen))
+}
+
+func computeVestAmount(total abi.TokenAmount, cliffStartEpoch abi.ChainEpoch, fullyVestedEpoch abi.ChainEpoch, quantizedEpoch abi.ChainEpoch) abi.TokenAmount {
+	// TODO: placeholder linear vesting
```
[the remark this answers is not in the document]
zixuanzh: Changing this to PARAM_FINISH, we might want to support other function types.

actors/runtime/exitcode/reserved.go
zixuanzh: go fmt changes

actors/builtin/miner/miner_actor.go
```diff
@@ -873,10 +1048,11 @@ func (a Actor) requestBeginFaults(rt Runtime, weights []*power.SectorStorageWeig
 	builtin.RequireSuccess(rt, code, "failed to request faults %v", weights)
 }
 
-func (a Actor) requestEndFaults(rt Runtime, weights []*power.SectorStorageWeightDesc, pledge abi.TokenAmount) {
+// TODO: this should put sectors back in the proving set
+// power is only gained when sectors are proven at least once
```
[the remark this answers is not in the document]
zixuanzh: Removing the comment and think we are aligned on sectors gaining power after a PoSt when they recover from faults.

actors/builtin/miner/miner_actor.go
```diff
@@ -783,18 +949,22 @@ func (a Actor) terminateSectors(rt Runtime, sectorNos []abi.SectorNumber, termin
...
-		a.requestEndFaults(rt, faultedWeights, faultPledge)
+		a.requestEndFaults(rt, faultedWeights)
 	}
 
 	a.requestTerminateDeals(rt, dealIDs)
-	a.requestTerminatePower(rt, terminationType, allWeights, allPledge)
+	a.requestTerminatePower(rt, terminationType, allWeights)
 }
 
+// TODO: to be updated
```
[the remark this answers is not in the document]
zixuanzh: removing this TODO but this function should be updated with the WindowPoSt PR.

actors/builtin/miner/miner_actor.go
```diff
@@ -783,18 +949,22 @@ func (a Actor) terminateSectors(rt Runtime, sectorNos []abi.SectorNumber, termin
 				st.ProvingSet = st.Sectors
 			}
 		}
+
+		a.slashPledgeCollateral(rt, amountToSlash)
 		return nil
 	})
 
+	// TODO: check this
```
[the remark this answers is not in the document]
zixuanzh: removing the comment, expecting this funciton to be updated with WindowPoSt PR.

actors/runtime/exitcode/reserved.go
zixuanzh: go fmt changes

actors/builtin/miner/miner_state.go
zixuanzh: this method returns an amount slashed that is used to update totalPledge at call site.

actors/abi/primitives.go
```diff
@@ -60,3 +62,24 @@ func NewTokenAmount(t int64) TokenAmount {
 
 // Randomness is a string of random bytes
 type Randomness []byte
+
+// Result of checking two headers for a consensus fault.
+type ConsensusFault struct {
```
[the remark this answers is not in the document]
zixuanzh: moved back to runtime.go

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +269,151 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
 	return sectorInfos, nil
 }
 
+func computeVestAmount(total abi.TokenAmount, cliffStartEpoch abi.ChainEpoch, fullyVestedEpoch abi.ChainEpoch, quantizedEpoch abi.ChainEpoch) abi.TokenAmount {
+	// PARAM_FINISH linear vesting placeholder
+	numVestWindows := (fullyVestedEpoch - cliffStartEpoch) / VestIncrement
```
anorth, line 274: This integer division will truncate and result in too low a window count, and thus too low a vest amount to reach the total. Once again, with integers, we cannot assume that the exact same amount will vest at each point.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +269,151 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+	_ = (quantizedEpoch - cliffStartEpoch) / VestIncrement // number of VestIncrement elapsed for other vest functions
+	quantizedVestAmount := big.Div(total, big.NewInt(int64(numVestWindows)))
+	return quantizedVestAmount
```
anorth, line 277: So the result here now appears not to depend on the current epoch, this is just a "constant" return value of the total divided by the number of vesting stages. But the number of vesting stages depends on the total length of the vesting period. Is that right?

This integer division will also truncate and short-change the miner if the total is not an exact multiple of the windows. You can't use a function like this to compute a fixed value to vest at each point.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +269,151 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+}
+
+func getQuantizedEpoch(e abi.ChainEpoch) abi.ChainEpoch {
+	// PARAM_FINISH
+	// default is the nearest next epoch that is multiple of QuantizedGranularity after a VestIncrement offset
+	// computeVestAmount can assume at least one VestIncrement has elapsed
+	offsetEpoch := e + VestIncrement
```
anorth, line 284: Why is VestIncrement used here at all? I expect quantization to ... just quantize. 

If `e` is an exact multiple of the QuantizedGranularity, this should return `e`.

actors/builtin/miner/miner_state.go
```diff
@@ -259,6 +269,151 @@ func (st *State) ComputeProvingSet(store adt.Store) ([]abi.SectorInfo, error) {
...
+		_, err := vestingFunds.Get(epochKey(quantizedVestEpoch), &lockedFundEntry)
+
+		if err != nil {
+			return err
+		}
+
+		vestAmount := computeVestAmount(pledgeAmount, cliffStartEpoch, fullyVestedEpoch, quantizedVestEpoch)
+		newLockedFundEntry := big.Add(lockedFundEntry, vestAmount)
+		vestingFunds.Set(epochKey(quantizedVestEpoch), &newLockedFundEntry)
+	}
```
anorth, line 311: As from comments on computeVestAmount, this loop cannot be structured like this and still be correct in the face of integer division. It needs to be structured like:
```
while totalVested < total {
	x := next payment from (total, totalVested, remainingEpochs, vestIncrement)
    vest x
	totalVested += x
}
```

actors/builtin/power/power_actor.go
```diff
@@ -223,39 +145,39 @@ type DeleteMinerParams struct {
...
+
 	nominal, ok := rt.ResolveAddress(params.Miner)
 	if !ok {
 		rt.Abortf(exitcode.ErrIllegalArgument, "failed to resolve address %v", params.Miner)
 	}
 
-	var st State
-	rt.State().Readonly(&st)
+	ownerAddr, workerAddr := builtin.RequestMinerControlAddrs(rt, nominal)
+	rt.ValidateImmediateCallerIs(ownerAddr, workerAddr)
```
anorth, line 155: Nooo, I don't think so. The miner _actor_ will be the only allowed caller here.

actors/builtin/power/power_actor.go
```diff
@@ -590,10 +414,79 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
+		Assert(claim.RawBytePower.GreaterThanEqual(big.Zero()))
+		Assert(claim.QualityAdjPower.GreaterThanEqual(big.Zero()))
+
+		st.TotalQualityAdjPower = big.Sub(st.TotalQualityAdjPower, claim.QualityAdjPower)
+		st.TotalRawBytePower = big.Sub(st.TotalRawBytePower, claim.RawBytePower)
+
+		return nil
+	})
+
+	err := a.updatePledgeCollateralTally(rt, pledgeAmount)
```
anorth, line 460: This should be subtracting the former miner's pledge, not adding it.

Since you already have a transaction right above here, delete `updatePledgeCollateralTally` (inline it in `UpdatePledgeCollateralTally `) and do the update in the existing transaction.

actors/runtime/runtime.go
```diff
@@ -183,6 +183,8 @@ type StateHandle interface {
 
 // Result of checking two headers for a consensus fault.
 type ConsensusFault struct {
+	// Address of the slasher who reported the fault.
+	Reporter addr.Address
```
anorth, line 187: Remove this

actors/builtin/miner/miner_actor.go
```diff
@@ -336,18 +354,35 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
+	newlyVestedAmount := rt.State().Transaction(&st, func() interface{} {
+
+		newlyVestedFund, err := st.vestNewFunds(adt.AsStore(rt), rt.CurrEpoch())
+
+		// Unlock PreCommit deposit upon successful ProveCommit and PreCommitDeposit can be used for initial pledge.
+		Assert(st.PreCommitDeposits.GreaterThanEqual(precommit.PreCommitDeposit))
+		st.PreCommitDeposits = big.Sub(st.PreCommitDeposits, precommit.PreCommitDeposit)
+
+		availableBalance := st.getAvailableBalance(rt.CurrentBalance())
+		if big.Cmp(availableBalance, initialPledge) < 0 {
```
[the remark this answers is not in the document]
zixuanzh: ack, was following code that wast there, removed.

actors/builtin/miner/miner_actor.go
```diff
@@ -465,23 +498,39 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
...
+
+		newlyVestedFund, err := st.vestNewFunds(adt.AsStore(rt), rt.CurrEpoch())
+		availableBalance := st.getAvailableBalance(rt.CurrentBalance())
+
+		if big.Cmp(availableBalance, newInitialPledge) < 0 {
+			rt.Abortf(exitcode.ErrInsufficientFunds, "not enough funds for new initial pledge requirement")
+		}
+
+		// Lock funds immediately after new initial pledge is checked against available funds.
+		if err = st.addLockedFund(adt.AsStore(rt), rt.CurrEpoch(), newInitialPledge); err != nil {
```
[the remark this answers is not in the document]
zixuanzh: ah this is intentional, we will not know if the original initial pledge has vested so this is a new initial pledge that we are requiring.

actors/builtin/miner/miner_actor.go
```diff
@@ -579,6 +628,150 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params *DeclareTemporaryFaults
...
+}
+
+type ReportConsensusFaultParams struct {
+	BlockHeader1     []byte
+	BlockHeader2     []byte
+	BlockHeaderExtra []byte
+}
+
+func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultParams) *adt.EmptyValue {
+
```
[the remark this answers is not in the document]
zixuanzh: fixed in other functions too

actors/builtin/power/power_actor.go
zixuanzh: pledgeAmount has been negated before invoking this method, moving the negation into this file and adding change as suggested.

actors/builtin/miner/miner_actor.go
```diff
@@ -579,6 +628,150 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params *DeclareTemporaryFaults
...
+	var st State
+	newlyVestedAmount := rt.State().Transaction(&st, func() interface{} {
+		newlyVestedFund, err := st.vestNewFunds(adt.AsStore(rt), rt.CurrEpoch())
+		AssertNoError(err)
+
+		return newlyVestedFund
+	}).(abi.TokenAmount)
+
+	totalLockedFund := st.LockedFunds
+	Assert(totalLockedFund.GreaterThanEqual(big.Zero()))
```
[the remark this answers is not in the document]
zixuanzh: Yes, there was a change in policy here (cc @davidad). We previously only burn the unvested balance and hence slasher share is a fraction of unvested balance. We have since moved to burning everything but the slasher share remains as a fraction of the unvested balance. Given that the slasher share will be subject to the same terms, im inclined to make slasher share a fraction of the total balance to be slashed instead of the unvested share.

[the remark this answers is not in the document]
davidad: I am fine with slasher reward being a fraction of all burned tokens. In the event of premeditated hijinks, fully vested tokens would be already withdrawn by this point anyhow.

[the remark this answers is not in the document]
davidad: I am fine with slasher reward being a fraction of all burned tokens. In the event of premeditated hijinks, fully vested tokens would be already withdrawn by this point anyhow.

actors/builtin/miner/miner_state.go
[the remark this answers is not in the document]
anorth: Sounds like we might want it back later for penalty reasons, but great to remove it now while unused.

actors/builtin/miner/miner_state.go
anorth: It wasn't, but I fixed it by maintaining the invariant here.

actors/builtin/miner/miner_actor.go
```diff
@@ -579,6 +628,150 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params *DeclareTemporaryFaults
...
+type ReportConsensusFaultParams struct {
+	BlockHeader1     []byte
+	BlockHeader2     []byte
+	BlockHeaderExtra []byte
+}
+
+func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultParams) *adt.EmptyValue {
+
+	// only another miner actor can report consensus fault
+	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
```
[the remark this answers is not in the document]
anorth: I fixed this

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
anorth: Changing to reward a fraction of all burnt tokens.

actors/builtin/power/power_actor.go
anorth: I misunderstood this. Miner deletion is now a bit wierd: the caller has to call the power actor which then deletes the miner. I think it would be better to follow the same pattern as consensus fault reporting now does, where the caller invokes a method on the miner, and the miner notifies the power actor. But that's for later.

anorth:
@zixuanzh I have squashed the history and added a single additional commit with my changes from our pairing session and some follow-up work. It's good to land now, I think.

Implementation of penalties is incomplete now, but let's leave it until after window post changes.

actors/builtin/miner/miner_actor.go
```diff
@@ -638,37 +635,35 @@ func (a Actor) AwardReward(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
-	}
+	AssertMsg(amountToSlash.GreaterThanEqual(big.Zero()), "negative slash amount %s", amountToSlash)
 
 	var st State
-	amountSlashed := rt.State().Transaction(&st, func() interface{} {
-		slashedPledge, err := st.slashLockedFund(adt.AsStore(rt), rt.CurrEpoch(), amountToSlash)
+	slashable := rt.State().Transaction(&st, func() interface{} {
+		// Unlock *unvested* amounts in order to burn them.
+		slashable, err := st.UnlockUnvestedFunds(adt.AsStore(rt), rt.CurrEpoch(), amountToSlash)
 		if err != nil {
```
zixuanzh, line 655: this should never error out?

anorth: I don't quite understand what you mean, but it's never acceptable to ignore the error. Perhaps especially so if it should "never" happen!

PR #273, opened 2020-04-06
Increase max bound of dealPricePerEpoch

arajasek:
The current number seems too small (it's under the default ask of a Lotus miner)


arajasek:
Yeah, I just increased it to a number that would keep Lotus happy, it's still just a placeholder. I can make it something bigger, though.

PR #274, opened 2020-04-06
first pass at updating abi/sector to for window and winning post

### LINKED ISSUE #249: Remove unused proof fields
Apparently 
- `PoStCandidate.PrivateProof` is unused in proofs/ffi, no reason to exist.
- PoStCandidate.SectorID.Miner is unused (and looks to duplicate PoStCandidate.Prover anyway)

dignifiedquire:
Closes #249

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -221,7 +221,6 @@ type WinningPoStVerifyInfo struct {
 	Randomness        PoStRandomness
 	Proofs            []PoStProof
 	ChallengedSectors []SectorInfo
-	EligibleSectors   []SectorNumber
 	Prover            ActorID // used to derive 32-byte prover ID
 }
 
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -69,13 +69,25 @@ type RegisteredProof int64
 
 const (
 	RegisteredProof_StackedDRG32GiBSeal  = RegisteredProof(1)
-	RegisteredProof_StackedDRG32GiBPoSt  = RegisteredProof(2)
+	RegisteredProof_StackedDRG32GiBPoSt  = RegisteredProof(2) // TODO: remove
 	RegisteredProof_StackedDRG2KiBSeal   = RegisteredProof(3)
-	RegisteredProof_StackedDRG2KiBPoSt   = RegisteredProof(4)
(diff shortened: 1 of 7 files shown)


anorth:
I've pushed a commit that deletes all the deprecated structures. If looks right to you @dignifiedquire please mark ready for review and I'll land it for you.

dignifiedquire:
@anorth thank you, looks good from my side

actors/abi/sector.go
```diff
@@ -69,75 +69,98 @@ type RegisteredProof int64
...
+		return RegisteredProof_StackedDRG512MiBWindowPoSt, nil
 	default:
 		return 0, errors.Errorf("unsupported mapping from %+v to PoSt-specific RegisteredProof", p)
 	}
 }
 
+// RegisteredPoStProof produces the PoSt-specific RegisteredProof corresponding
+// to the receiving RegisteredProof.
+func (p RegisteredProof) RegisteredPoStProof() (RegisteredProof, error) {
+	panic("unsupported method")
```
whyrusleeping, line 150: let's not have a panic here. either delete the method if we don't want it called, or return an error

actors/abi/sector.go
```diff
@@ -179,50 +201,39 @@ type OnChainSealVerifyInfo struct {
...
-	PrivateProof   PrivatePoStCandidateProof // Optional — should be ommitted for verification.
-	SectorID       SectorID
-	ChallengeIndex int64
+// Information needed to verify a Winning PoSt attached to a block header.
+// Note: this is not used within the state machine, but by the consensus/election mechanisms.
+type WinningPoStVerifyInfo struct {
+	Randomness        PoStRandomness
+	Proofs            []PoStProof
+	ChallengedSectors []SectorInfo
+	EligibleSectors   []SectorNumber
```
anorth, line 224: As discussed, this will grow to be a very large list. Two options instead of this include:
- use a bitfield of EligibleSectors instead, to be indexed inside proof verification
- split proof verification into two: generate the challenge index from randomness, and then verify with only the challenged sector infos

whyrusleeping: it seems like this list will be at most ~2k sectors?

anorth: No, for _winning_ PoSt it's all the sectors. 10s of millions.

PR #275, opened 2020-04-06
move peer ID serialization to byte arrays

### LINKED ISSUE #263: Improve peer.ID serialization to byte-array
The peer ID in miner info serializes as a string, but is really just a byte array. This can cause problems for languages that are more opinionated about the difference between those.

It would be preferable to alias or wrap the peer ID so we can control the serialization, and use []byte. Or even drop the dependency on libp2p altogether and leave conversion to the extra-VM code. IIRC, use of peer ID is the only reason this repo depends on libp2p, and tighter deps would be very reassuring.

whyrusleeping:
This was requested by other implementers, apparently peer IDs are serializing as technically invalid cbor (random bytes are not valid utf-8, therefore not valid cbor).

Closes #263

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -1,12 +1,15 @@
 package miner
 
 import (
+	"fmt"
+	"io"
 	"reflect"
 
 	addr "github.com/filecoin-project/go-address"
 	cid "github.com/ipfs/go-cid"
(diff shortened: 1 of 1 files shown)


whyrusleeping:
> Do you think this is significantly better than just changing the peer ID to a []byte in the worker info and leaving the wrapping to the outside?

Yeah, I think its a little bit nicer to have the real types here. But not a huge amount

PR #281, opened 2020-04-10
avoid flushing merkle structures on every single operation

whyrusleeping:
This is the first phase, where we just change the apis a bit to only flush when root is called, and actually load the objects up front when we instantiate the structures.

I put in a few helpers in an attempt to write less code, but expect things to change more in the next step where I try to combine multiple operations on the same structure in the same transaction.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -41,22 +41,22 @@ var _ abi.Invokee = Actor{}
 func (a Actor) Constructor(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
 
-	emptyArray, err := adt.MakeEmptyArray(adt.AsStore(rt))
+	emptyArray, err := adt.MakeEmptyArray(adt.AsStore(rt)).Root()
 	if err != nil {
 		rt.Abortf(exitcode.ErrIllegalState, "failed to create storage market state: %v", err)
 	}
 
-	emptyMap, err := adt.MakeEmptyMap(adt.AsStore(rt))
+	emptyMap, err := adt.MakeEmptyMap(adt.AsStore(rt)).Root()
 	if err != nil {
 		rt.Abortf(exitcode.ErrIllegalState, "failed to create storage market state: %v", err)
 	}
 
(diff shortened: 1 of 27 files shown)


whyrusleeping:
One thing i'm considering here is making the adt 'constructor' methods all take a runtime instead of a store, and have them abort if the loading fails. Would result in the code being a bit cleaner, though makes those constructors a little less reuseable

actors/builtin/market/market_actor.go
```diff
@@ -209,8 +224,18 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 
 			newDealIds = append(newDealIds, id)
 		}
-		st.Proposals = proposals.Root()
-		st.DealIDsByParty = dbp.Root()
+		propc, err := proposals.Root()
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "failed to flush proposal set: %w", err)
+		}
+		st.Proposals = propc
```
anorth, line 231: I think you could generally do

```
st.Proposals, err = proposals.Root()
if err != nil { ... }
```
because if we're aborting anyway, the state will be rolled back.

support/testing/address.go
```diff
@@ -44,3 +45,16 @@ func NewActorAddr(t *testing.T, data string) addr.Address {
...
+}
+
+func MustRoot(t testing.TB, r rooter) cid.Cid {
+	t.Helper()
+	c, err := r.Root()
+	if err != nil {
+		t.Fatal(err)
+	}
+	return c
+}
```
anorth, line 60: This is a weird file to put something about CIDs and storage in.

whyrusleeping: mmm, yeah. That is a weird location...

whyrusleeping:
@anorth 

> it's leaning further into the exception-style error handling which you've advocated against in the past, and we've almost exclusively constrained to top-level actor methods

Yeah, agree. In general, I don't mind a bit of extra error checking. We can switch the constructor to not load the object up and lazily do it on each operation, but thats also a ton of typing (though yes, it does clean up the actor method code a little bit)

whyrusleeping:
Going to merge this now, and if we figure something out we like more in the future, we can always change it

PR #282, opened 2020-04-10
update blacklist of actors init cant exec

### LINKED ISSUE #219: Restrict init actor exec's ability to create singleton actors
#218 relaxed the init actor's constraints on what actors can be instantiated freely too much, e.g. allowing user-creation of reward actor, system actor etc. Switch to either an explicit capability list (rather than exclusion list), or more exhaustively enumerate the exclusions such that it won't rot in the future.

whyrusleeping:
closes #219

### DIFF
--- a/actors/builtin/init/init_actor.go
+++ b/actors/builtin/init/init_actor.go
@@ -87,7 +87,7 @@ func (a Actor) Exec(rt runtime.Runtime, params *ExecParams) *ExecReturn {
 
 func canExec(callerCodeID cid.Cid, execCodeID cid.Cid) bool {
 	switch execCodeID {
-	case builtin.AccountActorCodeID, builtin.InitActorCodeID, builtin.StoragePowerActorCodeID, builtin.StorageMarketActorCodeID, builtin.StorageMinerActorCodeID, cid.Undef:
+	case builtin.AccountActorCodeID, builtin.InitActorCodeID, builtin.StoragePowerActorCodeID, builtin.StorageMarketActorCodeID, builtin.StorageMinerActorCodeID, builtin.RewardActorCodeID, builtin.SystemActorCodeID, builtin.CronActorCodeID, cid.Undef:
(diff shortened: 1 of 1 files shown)


anorth:
I don't really like this solution because it fails unsafely. Future changes elsewhere could lead to exactly the same bug being re-introduced. 

A whitelist, rather than blacklist, would at least fail without introducing unknown behaviour. Or some technique where this code can be guaranteed to exhaustively include all code types, including if another is added in the future.

whyrusleeping:
A whitelist makes testing anything extremely hard, I much prefer a blacklist, as we already have to remember to type all of these somewhere anyways. And realistically, who cares if any of these actors (aside from the init actor) get execed? Its not really a big deal if they do (its unlikely they will work well, so idk why anyone would do it, but its not going to hurt anything)

actors/builtin/init/init_actor.go
```diff
@@ -87,7 +87,7 @@ func (a Actor) Exec(rt runtime.Runtime, params *ExecParams) *ExecReturn {
 
 func canExec(callerCodeID cid.Cid, execCodeID cid.Cid) bool {
 	switch execCodeID {
-	case builtin.AccountActorCodeID, builtin.InitActorCodeID, builtin.StoragePowerActorCodeID, builtin.StorageMarketActorCodeID, builtin.StorageMinerActorCodeID, cid.Undef:
+	case builtin.AccountActorCodeID, builtin.InitActorCodeID, builtin.StoragePowerActorCodeID, builtin.StorageMarketActorCodeID, builtin.StorageMinerActorCodeID, builtin.RewardActorCodeID, builtin.SystemActorCodeID, builtin.CronActorCodeID, cid.Undef:
```
ZenGround0, line 90: Because of rot this is missing `VerifiedRegistryActorCodeID`

If you rely on `IsSingletonActor` (minus accountactor) it would help keep this list in one place.

magik6k:
Closing as https://github.com/filecoin-project/specs-actors/pull/405/files does roughly the same thing, just a bit differently

PR #283, opened 2020-04-10
make seal randomness spec compliant

### LINKED ISSUE #260: Seal randomness must include miner address as entropy
The miner actor computes seal randomness with no entropy. It must use the miner address according to   https://github.com/filecoin-project/specs/pull/883

cc @sternhenri @magik6k

whyrusleeping:
closes #260 

FYI @magik6k


anorth:
FYI @acruikshank @ZenGround0 we'll need the parallel change in seal generation when integrating this.

PR #286, opened 2020-04-14
New Window PoSt construction over all sectors

anorth:
Implement the new shiny Window PoSt construction which requires a proof of all sectors every 24hr.

This has all the basics. A few things to come in follow-ups:
- Process skipped faults
- Load-balance while filling deadlines with new sectors
- Implement verifyPledgeMeetsInitialRequirements
- Terminate deals with market actor (existing debt)
- Compute non-zero fees and penalties (existing debt)

There's one outstanding issue before it's mergeable:
- [x] Fix incorrect `ComputeCurrentDeadline`

Thanks @magik6k @whyrusleeping and @frrist for contributing.


actors/builtin/miner/miner_state.go
```diff
@@ -309,46 +301,185 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
 
 // Adds some sector numbers to the new sectors bitfield.
 func (st *State) AddNewSectors(newSectors ...abi.SectorNumber) error {
-	panic("Implement me") // TODO
+	for _, sector := range newSectors {
+		st.NewSectors.Set(uint64(sector))
+	}
```
anorth, line 306: @magik6k I know the current use case is one at a time, but would it generally be better to compute a bitfield of many and then merge bitfields?

actors/builtin/miner/miner_state.go
```diff
@@ -309,46 +301,185 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+
+	bf := abi.NewBitField()
+	_, err = arr.Get(uint64(expiry), &bf)
+	if err != nil {
+		return err
+	}
+
+	for _, sector := range newSectors {
+		bf.Set(uint64(sector))
+	}
```
anorth, line 332: Ditto

actors/builtin/miner/miner_state.go
```diff
@@ -309,46 +301,185 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+
+	if err := arr.Set(uint64(expiry), &bf); err != nil {
+		return err
+	}
+
+	st.SectorExpirations, err = arr.Root()
+	return err
+}
+
+func checkBitFieldIntersect(expectAllIntersect bool, a, b abi.BitField) (bool, uint64, error) {
```
anorth, line 342: As implemented, this doesn't check the intersection but the subset/contains operation. If that's what you want, rename to `containsAny`/`containsAll`. But if you need intersection, must also iterate over `am` in an outer loop and check in `bm` (unless sizes are equal)

actors/builtin/miner/miner_state.go
```diff
@@ -309,46 +301,185 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
 
 // Adds sectors numbers to faults and fault epochs.
 // Precondition: sectors are not already in faults.
-func (st *State) AddFaults(sectorNos abi.BitField, faultEpoch abi.ChainEpoch) error {
-	panic("implement me") // TODO
+func (st *State) AddFaults(store adt.Store, sectorNos abi.BitField, faultEpoch abi.ChainEpoch) (err error) {
+	// TODO: check max count after adding
+
+	{
+		ok, bad, err := checkBitFieldIntersect(false, st.Faults, sectorNos)
```
anorth, line 369: Idea: rather than expand maps for this check, you could do the merge and check the expected size is the sum of sizes.

actors/builtin/miner/miner_state.go
```diff
@@ -374,7 +505,11 @@ func (d *Deadlines) CountDeadlines() [WPoStPeriodDeadlines]uint64 {
 
 // Adds sector numbers to a deadline.
 func (d *Deadlines) AddToDeadline(deadline int, newSectors ...abi.SectorNumber) error {
-	panic("implement me") // TODO
+	for _, sector := range newSectors {
+		d.Due[deadline].Set(uint64(sector))
+	}
```
anorth, line 510: This one will definitely be called with >1 new sector, so merge bitfields?

actors/builtin/miner/miner_state.go
magik6k: Here this doesn't matter as we're immedietally putting the bitfield into an amt

actors/builtin/miner/miner_state.go
```diff
@@ -525,16 +939,25 @@ func quantizeUp(e abi.ChainEpoch, unit abi.ChainEpoch) abi.ChainEpoch {
 	return e - remainder + unit
 }
 
-func sectorKey(e abi.SectorNumber) adt.Keyer {
+// Rounds e to the nearest exact multiple of the quantization unit, rounding down.
+func quantizeDown(e abi.ChainEpoch, unit abi.ChainEpoch) abi.ChainEpoch {
+	remainder := e % unit
```
frrist, line 944: This and `quantizeUp` will panic when `unit == 0` which is a possible (but unlikely) scenario given the defition here: https://github.com/filecoin-project/specs-actors/pull/286/files#diff-2580e4b82238370ee854a52e8d5428e7R37

actors/builtin/miner/miner_state.go
```diff
@@ -267,64 +306,430 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+		for _, n := range ns {
+			d.Due[i].Unset(n)
+		}
+	}
+	return nil
 }
 
+// Computes the first partition index and number of sectors for a deadline.
+// Partitions are numbered globally for the miner, not per-deadline.
+func (d *Deadlines) PartitionsForDeadline(deadlineIndex uint64) (firstPartition, sectorCount uint64, _ error) {
```
frrist, line 712: @anorth What is meant by "first partition index"?

anorth: It's the index of the first (lowest) partition number due at the requested deadline. The partitions are indexed from 0 by enumerating those in the first deadline, then the second, etc for all the deadlines.

actors/builtin/miner/miner_state.go
```diff
@@ -525,6 +597,15 @@ func quantizeUp(e abi.ChainEpoch, unit abi.ChainEpoch) abi.ChainEpoch {
 	return e - remainder + unit
 }
 
+// Rounds e to the nearest exact multiple of the quantization unit, rounding down.
+func quantizeDown(e abi.ChainEpoch, unit abi.ChainEpoch) abi.ChainEpoch {
+	remainder := e % unit
+	if remainder == 0 {
+		return e
+	}
+	return e - remainder
```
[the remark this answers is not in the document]
anorth: I don't think so. E.g. `quantizeDown(13, 10)`, remainder will be 3 and desired output is 10. Your suggestion would yield 6

actors/builtin/miner/miner_state.go
```diff
@@ -131,6 +163,13 @@ func (st *State) GetSectorSize() abi.SectorSize {
 	return st.Info.SectorSize
 }
 
+// Returns the epoch that starts the current proving period, and whether that period starts on or after epoch 0.
+// The period start can be negative during the first proving period of the chain.
```
frrist, line 167: for what case does this method return negative values? I don't see how that's possible given the impl of `quantizeDown`.

actors/builtin/miner/miner_actor.go
```diff
@@ -541,88 +652,172 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *adt
...
+
+		// Traverse earlier submissions and enact detected faults.
+		// This is necessary to prevent the miner "declaring" a fault for a PoSt already missed.
+		deadlineIndex, _ := ComputeCurrentDeadline(periodStart, currEpoch)
+		detectedFaultSectors, penalty = checkMissingPoStFaults(rt, st, store, deadlines, periodStart, deadlineIndex, currEpoch)
+
+		var decaredSectors []*abi.BitField
+		for _, decl := range params.Faults {
+			err := validateFRDeclaration(currEpoch, periodStart, deadlines, decl.Deadline, decl.Sectors)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "invalid fault declaration")
```
nicola, line 703: error should be more explicit: "Faults for deadline decl.Deadline are too late for submission"

actors/builtin/miner/miner_actor.go
```diff
@@ -541,88 +652,172 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *adt
...
+		}
+		return nil
+	})
 
-	// Request deferred Cron invocation to update temporary fault state.
-	// TODO: cant we just lazily clean this up?
-	cronPayload := CronEventPayload{
-		EventType: CronEventTempFault,
-		Sectors:   &params.SectorNumbers,
+	// Remove power for new faulty sectors.
```
nicola, line 764: power must be removed at the deadline and not at the time it was announced

actors/builtin/miner/miner_actor.go
```diff
@@ -541,88 +652,172 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *adt
...
-			if err = st.PutSector(store, sector); err != nil {
-				rt.Abortf(exitcode.ErrIllegalState, "failed to update sector %v: %v", sectorNumber, err)
-			}
+			// Load info for sectors.
+			declaredFaultSectors, err = st.LoadSectorInfos(store, newFaults)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load fault sectors")
+
+			// Unlock penalty for declared faults.
+			declaredPenalty, err := unlockPenalty(store, &st, currEpoch, declaredFaultSectors, pledgePenaltyForSectorDeclaredFault)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to charge fault fee")
```
nicola, line 748: shouldn't we check here if this is the 14th period for which it was declared faulty?

actors/builtin/miner/miner_actor.go
```diff
@@ -160,53 +173,166 @@ func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyVa
...
+		if contains {
+			rt.Abortf(exitcode.ErrIllegalArgument, "duplicate PoSt partition")
+		}
+		err = st.AddPoStSubmissions(postedPartitions)
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "failed to record submissions for partitions %s: %v", params.Partitions, err)
+		}
+
+		// If the PoSt was successful, the declared recoveries should be restored
+		err = st.RemoveFaults(store, declaredRecoveries)
```
nicola, line 300: What happens if a sector was not declared as recovered but gets proven in this windowPost?

It must not be accounted for power until the next windowpost

frrist:
@anorth branch with moar tests here: https://github.com/filecoin-project/specs-actors/pull/294, includes coverage for `ComputeCurrentDeadline` and `assignProvingPeriodBoundary`. Will add some for `ComputeFaultDeclarationCutoff` next.

actors/builtin/miner/miner_actor.go
anorth: The nested `err` will be included in the log and has those details.

actors/builtin/miner/miner_actor.go
anorth: Let's discuss that outside this PR as a potential enhancement.

actors/builtin/miner/miner_actor.go
anorth: Faults are declared only once, the transition from not fault to faulty. Then they stay faulty and the daily cron checks how old ongoing faults are. This is definitely the first period that these specific sectors have been faulty.

actors/builtin/miner/miner_actor.go
anorth: The PoSt will fail validation. The good sectors + declared recoveries must exactly match the set proven.

actors/builtin/miner/miner_actor.go
```diff
@@ -160,53 +173,164 @@ func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyVa
...
 	var st State
+	var detectedFaultSectors []*SectorOnChainInfo
+	var penalty abi.TokenAmount
+	var recoveredSectors []*SectorOnChainInfo
 	rt.State().Transaction(&st, func() interface{} {
 		rt.ValidateImmediateCallerIs(st.Info.Worker)
 
-		if rt.CurrEpoch() > st.PoStState.ProvingPeriodStart+power.WindowedPostChallengeDuration {
-			rt.Abortf(exitcode.ErrIllegalState, "PoSt submission too late")
+		// Every epoch is during some deadline's challenge window.
```
whyrusleeping, line 206: I would strongly recommend having the user submit the partition they are targeting. Its likely just one extra byte in the parameters, and will make diagnosing issues *much* easier

actors/builtin/miner/miner_actor.go
```diff
@@ -872,63 +1227,51 @@ func (a Actor) checkPrecommitExpiry(rt Runtime, sectors *abi.BitField, regProof
...
-		}
-		faultsMap, err := st.FaultSet.AllMap(maxAllowedFaults)
-		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "too many faults")
+			rt.Abortf(exitcode.ErrIllegalState, "failed to expand faults")
 		}
 
-		for _, sectorNo := range sectorNos {
-			sector, found, err := st.GetSector(store, sectorNo)
+		_ = sectorNos.ForEach(func(sectorNo uint64) error {
```
whyrusleeping, line 1273: ignoring errors makes me ever so slightly nervous

actors/builtin/miner/miner_state.go
```diff
@@ -267,64 +294,439 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
 	}
 
-	return asStorageWeightDesc(st.Info.SectorSize, sectorInfo), nil
+	count, err := st.NewSectors.Count()
+	if err != nil {
+		return err
+	}
+	if count > NewSectorsPerPeriodMax {
+		return fmt.Errorf("too many new sectors %d, max %d", count, NewSectorsPerPeriodMax)
+	}
```
frrist, line 314: `st.NewSectors`'s value is modified by this method (line 303) even if it errors, is that the intended effect?

actors/builtin/miner/miner_actor.go
anorth: The _partitions_ are explicitly submitted, but assuming you meant deadline: 👍  I'll do that.

actors/builtin/miner/miner_actor.go
anorth: This particular code aborts rather than return errors, but that nervousness is well-founded so I'll change all the non-trivial cases of this to check the error anyway.

actors/builtin/miner/miner_state.go
anorth: It was, but you're right to feel uneasy about it. https://github.com/filecoin-project/specs-actors/issues/300

PR #290, opened 2020-04-14
implement puppet actor

frrist:
### Motivation
Can be used for exercising inner sends without actor level validation getting in the way.

### Open Questions
Draft until the below are answered/addressed.
- Should TestActor's send return values or ignore them?
- Are there an additional methods that should be added? For example, a method that returns an encodable value?


actors/test/test_actor.go
```diff
@@ -0,0 +1,48 @@
+package test
```
anorth, line 1: I think we can do better than `test` for the name, it'll be confusing with other things. I chatted briefly with @whyrusleeping and others and `puppet` came out as short and to the point.

actors/test/test_actor.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (a Actor) Send(rt runtime.Runtime, params *SendParams) *adt.EmptyValue {
+	rt.ValidateImmediateCallerAcceptAny()
+	_, _ = rt.Send(
+		params.To,
+		params.Method,
+		runtime.CBORBytes(params.Params),
+		params.Value,
+	)
+	return nil
```
anorth, line 45: I'd like to get both the return value and exit code out, probably by returning a structure containing both.

For the return value, we just want the raw bytes (since we don't know the type). I suggest extending the `runtime.CBORBytes` to implement `CBORUnmarshaler` to get the raw bytes out.

actors/test/test_actor.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	addr "github.com/filecoin-project/go-address"
+
+	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
+	runtime "github.com/filecoin-project/specs-actors/actors/runtime"
+	adt "github.com/filecoin-project/specs-actors/actors/util/adt"
+)
+
+type Actor struct{}
```
anorth, line 12: Please add some documentation.

> This actor exists to aid testing the runtime and environment in which it's embedded. It provides direct access to the runtime methods, including sending arbitrary messages to other actors, without any preconditions or invariants to get in the way.

anorth, line 12: We'll need a code ID for this actor. Put it in this file rather than alongside the real builtin actors.

Similarly, we also need a table of method numbers so they can be referenced by number.

PR #295, opened 2020-04-17
Miner test harness

anorth:
Basic setup for miner actor tests.

- [x] Wait for #286 to land and rebase

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -1,12 +1,160 @@
 package miner_test
 
 import (
+	"bytes"
+	"context"
 	"testing"
 
+	addr "github.com/filecoin-project/go-address"
+	"github.com/minio/blake2b-simd"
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1,12 +1,160 @@
...
+	t.Run("simple construction", func(t *testing.T) {
+		rt := builder.Build(t)
+		params := miner.ConstructorParams{
+			OwnerAddr:  owner,
+			WorkerAddr: worker,
+			SectorSize: SectorSize,
+			PeerId:     "peer",
+		}
+
+		provingPeriodBoundary := abi.ChainEpoch(2386) // This is just set from running the code.
```
frrist, line 50: Since the runtime's `Hasher` is swappable we _could_ pass a mockHasher that returns a value we define, but this seems reasonable enough.

PR #296, opened 2020-04-17
Remove adding LockedFund in ExtendSectorExpiration

zixuanzh:
Remove adding new LockedFund in SectorExpiration.
The only operation in ExtendSectorExpiration is changing sector duration.


actors/abi/sector.go
```diff
@@ -228,5 +228,5 @@ type WindowPoStVerifyInfo struct {
 
 // Information submitted by a miner to provide a Window PoSt.
 type OnChainWindowPoStVerifyInfo struct {
-	Proofs            []PoStProof
+	Proofs []PoStProof
```
anorth, line 231: I know gofmt did this, but please try to avoid unrelated cosmetic changes. This just adds rebase pain.

PR #297, opened 2020-04-17
update AwardReward to AddLockedFund

zixuanzh:
- Rename AwardReward to AddLockedFund that can be called by any one to add locked fund as collateral.
- Caller can specify an amountToLock that needs to be less than or equal to the available balance.
- RewardActor calls this method for awarding block reward and adding locked fund.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -56,7 +56,7 @@ func (a Actor) Exports() []interface{} {
 		11:                        a.DeclareFaultsRecovered,
 		12:                        a.OnDeferredCronEvent,
 		13:                        a.CheckSectorProven,
-		14:                        a.AwardReward,
+		14:                        a.AddLockedFund,
 		15:                        a.ReportConsensusFault,
 		16:                        a.WithdrawBalance,
 	}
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -630,19 +630,27 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params *DeclareTemporaryFaults
 // Pledge Collateral //
 ///////////////////////
 
-func (a Actor) AwardReward(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
-	rt.ValidateImmediateCallerIs(builtin.RewardActorAddr)
-	pledgeAmount := rt.Message().ValueReceived()
+func (a Actor) AddLockedFund(rt Runtime, amountToLock *abi.TokenAmount) *adt.EmptyValue {
+	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
```
anorth, line 634: This is wrong because the reward actor is not a signable. The strictest thing that might work here is to allow the owner/worker and reward actor. I'll do that.

actors/builtin/miner/miner_actor.go
```diff
@@ -630,19 +630,27 @@ func (a Actor) DeclareTemporaryFaults(rt Runtime, params *DeclareTemporaryFaults
...
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		if err := st.AddLockedFunds(adt.AsStore(rt), rt.CurrEpoch(), pledgeAmount, &PledgeVestingSpec); err != nil {
+		_, err := st.UnlockVestedFunds(adt.AsStore(rt), rt.CurrEpoch())
```
anorth, line 638: Ignored return value breaks the delta calculations 😱

PR #298, opened 2020-04-17
Tests for miner state, fix shadowing in RemoveRecoveries

frrist:
### What's done:
- PreCommit sectors store tests
- Sectors store tests
- NewSectors bit field tests
- SectorExpirations store tests
### What's left:
- [x] ~decide if we want this 18 second test: https://github.com/filecoin-project/specs-actors/pull/298/files#diff-4b9e78dc63ba81fa9651a3f08d1973a9R362~ We don't want it.
- [x] Fault store tests
- [x] Recoveries bitfield tests
- [x] PoStSubmissions bitfield tests


actors/builtin/miner/miner_state_test.go
```diff
@@ -0,0 +1,587 @@
...
+}
+
+func (h *minerStateHarness) assertClearNewSectors() {
+	h.assertNewSectorsCount(0)
+}
+
+func (h *minerStateHarness) assertNewSectorsCount(count uint64) {
+	sectorCount, err := h.s.NewSectors.Count()
+	assert.NoError(h.t, err)
+	assert.Equal(h.t, count, sectorCount)
```
anorth, line 438: In general I think pushing the semantically relevant assertions into helpers like this detracts from test clarity. This would be better as a `countNewSectors()` method that returns the count (but no error), and then putting the `assert.Equal` in the calling test code directly.

So the harness generally would have a bunch of `require.NoError` but no/very few `assertEqual`, and thus ~none of the method names start with `assert`. There may be some exceptions for a complex but repeatedly-used assertion.

actors/builtin/miner/miner_state_test.go
```diff
@@ -0,0 +1,587 @@
...
+	return out
+}
+
+func (h *minerStateHarness) assertHasSectorNo(rt runtime.Runtime, sectorNo abi.SectorNumber) {
+	found, err := h.s.HasSectorNo(adt.AsStore(rt), sectorNo)
+	assert.NoError(h.t, err)
+	assert.True(h.t, found)
+}
+
+func (h *minerStateHarness) assertSectorNotFound(rt runtime.Runtime, sectorNo abi.SectorNumber) {
```
anorth, line 457: Remove in favour of the caller doing `assert.False(t, h.hasSector(n))`.

actors/builtin/miner/miner_state_test.go
```diff
@@ -0,0 +1,587 @@
...
+// Type Construction Methods
+//
+
+// returns a unique SectorPreCommitOnChainInfo with each invocation with SectorNumber set to `sectorNo`.
+func (h *minerStateHarness) newSectorPreCommitOnChainInfo(sectorNo abi.SectorNumber) *miner.SectorPreCommitOnChainInfo {
+	seed := h.getSeed()
+	info := h.newSectorPreCommitInfo(sectorNo)
+	return &miner.SectorPreCommitOnChainInfo{
+		Info:             *info,
+		PreCommitDeposit: abi.NewTokenAmount(int64(seed)),
```
anorth, line 521: This would be much better as an explicit arg.

actors/builtin/miner/miner_state_test.go
```diff
@@ -0,0 +1,587 @@
...
+
+	// assert NewSectors bitfield was constructed correctly (empty)
+	newSectorsCount, err := state.NewSectors.Count()
+	assert.NoError(t, err)
+	assert.Equal(t, uint64(0), newSectorsCount)
+
+	return &minerStateHarness{s: state, t: t, cidGetter: tutils.NewCidForTestGetter(), seed: 0}
+}
+
+func newBitFieldFromSectorNos(sectorNos ...abi.SectorNumber) *abi.BitField {
```
anorth, line 581: Use `bitfield.NewFromSet`. You'll notice that some of the state methods use `uint64` in place of abi.SectorNumber to ease this.

support/testing/cid.go
```diff
@@ -0,0 +1,22 @@
+package testing
+
+import (
+	"github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
```
anorth, line 5: We don't really want this dep in the repo at all. It's unfortunate that this package brings the IPLD store interface (good) along with a deep dependency on refmt and etc (bad). Please don't use WrapObject.

I much prefer explicit construction to this like this stateful "getter" pattern. Use something like:
```
var DefaultCidBuilder = cid.V1Builder{Codec: cid.DagCBOR, MhType: DefaultHashFunction}
// Or cid.V1Builder{Codec: cid.Raw, MhType: mh.IDENTITY} for more transparent CIDs
func CidFromString(t *testing.T, input string) cid.Cid {
	c, err := constants.DefaultCidBuilder.Sum([]byte(input))
	require.NoError(t, err)
	return c
```
And give the CIDs explicit values.

actors/builtin/miner/miner_state_test.go
```diff
@@ -0,0 +1,587 @@
...
+}
+
+// returns a unique SectorPreCommitInfo with each invocation with SectorNumber set to `sectorNo`.
+func (h *minerStateHarness) newSectorPreCommitInfo(sectorNo abi.SectorNumber) *miner.SectorPreCommitInfo {
+	seed := h.getSeed()
+	return &miner.SectorPreCommitInfo{
+		RegisteredProof: abi.RegisteredProof_StackedDRG32GiBPoSt,
+		SectorNumber:    sectorNo,
+		SealedCID:       h.cidGetter(),
+		SealRandEpoch:   abi.ChainEpoch(seed),
```
anorth, line 544: Use an explicit value. For the purposes of these tests there's probably no reason not to use a constant.

actors/builtin/miner/miner_state_test.go
```diff
@@ -0,0 +1,587 @@
...
+		assert.Empty(t, empty2)
+	})
+
+	t.Run("fail to add more than SectorsMax number of sectors", func(t *testing.T) {
+		rt := builder.Build(t)
+		harness := constructStateHarness(t, rt, abi.ChainEpoch(0))
+
+		tooManySectors := make([]uint64, miner.SectorsMax+1)
+
+		for i := uint64(0); i < miner.SectorsMax+1; i++ {
```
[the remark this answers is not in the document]
frrist: :snail: About 18 seconds which is less than ideal, should I omit this from the PR? Could also add test flags if we think this check is important.

frrist:
@anorth `TestFaultStore` is failing, I believe the test to be correct and have a hunch that the bug is in `RemoveFaults` change calculation.

go.mod
```diff
@@ -7,14 +7,16 @@ require (
 	github.com/filecoin-project/go-amt-ipld/v2 v2.0.1-0.20200131012142-05d80eeccc5e
 	github.com/filecoin-project/go-bitfield v0.0.0-20200416002808-b3ee67ec9060
 	github.com/gopherjs/gopherjs v0.0.0-20190812055157-5d271430af9f // indirect
-	github.com/ipfs/go-cid v0.0.4
+	github.com/ipfs/go-cid v0.0.5
+	github.com/ipfs/go-datastore v0.4.1
 	github.com/ipfs/go-hamt-ipld v0.0.15-0.20200131012125-dd88a59d3f2e
+	github.com/ipfs/go-ipfs-blockstore v1.0.0
```
anorth, line 13: We can avoid this dependency, I'll add a tiny in-memory  store (which really should be provided by the ipld-cbor package)

PR #304, opened 2020-04-20
Remove SysErrInternal definition, use of which is always a bug

anorth:
This came out of interop efforts last time round. Everywhere this was being used was incorrect (usually because the state transition should fail to be evaluated, rather than have this exist code persist on chain).


actors/runtime/exitcode/reserved.go
```diff
@@ -92,9 +92,7 @@ const (
 	SysErrorReserved3 = ExitCode(12)
 	SysErrorReserved4 = ExitCode(13)
 	SysErrorReserved5 = ExitCode(14)
-
-	// Indicates something broken within the VM.
-	SysErrInternal = ExitCode(15)
```
magik6k, line 97: I'd keep the definition, maybe rename to something indicating it's a VM error

anorth: The problem is that this can never safely be used. These exit codes go on chain, but every time there's a VM error of this form (idiosyncratic, not reproducible between impls) it should *not* go on chain; instead the state evaluation should just fail.

It should never have existed. If Lotus still uses it, they are all bugs that should either be a different exit code, or a "fatal" error.

PR #309, opened 2020-04-21
improve deal accounting performance

whyrusleeping:
This should accomplish the same thing, except now payments for storage deals only happen once every hundred blocks or so (not bucketed though, otherwise that would create hot spots)

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -284,9 +284,12 @@ func (a Actor) VerifyDealsOnSectorProveCommit(rt Runtime, params *VerifyDealsOnS
 		}
 
 		for _, dealID := range params.DealIDs {
-			deal, _, err := states.Get(dealID)
+			deal, found, err := states.Get(dealID)
 			if err != nil {
 				rt.Abortf(exitcode.ErrIllegalState, "get deal %v", err)
+			}
+			if !found {
+				rt.Abortf(exitcode.ErrIllegalArgument, "failed to find deal given in params: %d", dealID)
 			}
 			proposal, err := proposals.Get(dealID)
 			if err != nil {
@@ -385,9 +388,12 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
 			}
 			Assert(deal.Provider == minerAddr)
 
-			state, _, err := states.Get(dealID)
+			state, found, err := states.Get(dealID)
 			if err != nil {
 				rt.Abortf(exitcode.ErrIllegalState, "get deal: %v", err)
+			}
+			if !found {
+				rt.Abortf(exitcode.ErrIllegalState, "no state found for deal in sector being terminated")
 			}
 
 			// Note: we do not perform the balance transfers here, but rather simply record the flag
@@ -455,19 +461,16 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 
 				if state.SectorStartEpoch == epochUndefined {
 					// Not yet appeared in proven sector; check for timeout.
-					if rt.CurrEpoch() > deal.StartEpoch {
-						slashed := st.processDealInitTimedOut(rt, et, lt, dealID, deal, state)
-						if !slashed.IsZero() {
-							amountSlashed = big.Add(amountSlashed, slashed)
-						}
-						if deal.VerifiedDeal {
-							timedOutVerifiedDeals = append(timedOutVerifiedDeals, deal)
-						}
-						return nil
-					}
+					AssertMsg(rt.CurrEpoch() >= deal.StartEpoch, "if sector start is not set, we must be in a timed out state")
 
-					// This should not be able to happen
-					rt.Abortf(exitcode.ErrIllegalState, "invalid deal state, unstarted, not timed out")
+					slashed := st.processDealInitTimedOut(rt, et, lt, dealID, deal, state)
+					if !slashed.IsZero() {
+						amountSlashed = big.Add(amountSlashed, slashed)
+					}
+					if deal.VerifiedDeal {
+						timedOutVerifiedDeals = append(timedOutVerifiedDeals, deal)
+					}
+					return nil
 				}
 
 				slashAmount, nextEpoch := st.updatePendingDealState(rt, state, deal, dealID, et, lt, rt.CurrEpoch())
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -395,8 +395,8 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
 			// is performed. // TODO: Do that here
 
 			state.SlashEpoch = rt.CurrEpoch()
-			err = states.Set(dealID, state)
-			if err != nil {
+
+			if err := states.Set(dealID, state); err != nil {
 				rt.Abortf(exitcode.ErrIllegalState, "set deal: %v", err)
 			}
 		}
@@ -414,6 +414,8 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.CronActorAddr)
 	amountSlashed := big.Zero()
 
+	var timedOutVerifiedDeals []*DealProposal
+
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
 		dbe, err := AsSetMultimap(adt.AsStore(rt), st.DealOpsByEpoch)
@@ -423,9 +425,42 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 
 		updatesNeeded := make(map[abi.ChainEpoch][]abi.DealID)
 
+		states, err := AsDealStateArray(adt.AsStore(rt), st.States)
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "get state state: %v", err)
+		}
+
 		for i := st.LastCron; i <= rt.CurrEpoch(); i++ {
-			if err := dbe.ForEach(i, func(deal abi.DealID) error {
-				slashAmount, nextEpoch := st.updatePendingDealState(rt, deal, rt.CurrEpoch())
+			if err := dbe.ForEach(i, func(dealID abi.DealID) error {
+				state, found, err := states.Get(dealID)
+				if err != nil {
+					rt.Abortf(exitcode.ErrIllegalState, "failed to get deal: %d", dealID)
+				}
+
+				if !found {
+					return nil
+				}
+
+				deal := st.mustGetDeal(rt, dealID)
+
+				if state.SectorStartEpoch == epochUndefined {
+					// Not yet appeared in proven sector; check for timeout.
+					if rt.CurrEpoch() > deal.StartEpoch {
+						slashed := st.processDealInitTimedOut(rt, dealID, deal, state)
+						if !slashed.IsZero() {
+							amountSlashed = big.Add(amountSlashed, slashed)
+						}
+						if deal.VerifiedDeal {
+							timedOutVerifiedDeals = append(timedOutVerifiedDeals, deal)
+						}
+						return nil
+					}
+
+					// This should not be able to happen
+					rt.Abortf(exitcode.ErrIllegalState, "invalid deal state, unstarted, not timed out")
+				}
+
+				slashAmount, nextEpoch := st.updatePendingDealState(rt, state, deal, dealID, rt.CurrEpoch())
 				if !slashAmount.IsZero() {
 					amountSlashed = big.Add(amountSlashed, slashAmount)
 				}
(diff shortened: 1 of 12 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -427,78 +410,62 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
...
 			}
 		}
 
-		return &slashed
-	}).(abi.TokenAmount)
-
-	// TODO: award some small portion of slashed to caller as incentive
-
-	// Restore verified dataset allowance for verified clients.
-	for _, deal := range verifiedDeals {
```
zixuanzh, line 487: we will need to put this logic in `processDealInitTimedOut`

actors/builtin/market/market_actor.go
```diff
@@ -82,7 +81,6 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 	rt.State().Transaction(&st, func() interface{} {
 		// Before any operations that check the balance tables for funds, execute all deferred
 		// deal state updates.
-		amountSlashedTotal = big.Add(amountSlashedTotal, st.updatePendingDealStatesForParty(rt, nominal))
```
zixuanzh, line 85: What if someone calls this method after its deal collateral is marked slashed but not yet processed by Cron? and Cron is the last message in a tipset?

actors/builtin/market/market_actor.go
```diff
@@ -216,29 +213,17 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			deal.Proposal.Provider = provider
 			deal.Proposal.Client = client
 
-			// Before any operations that check the balance tables for funds, execute all deferred
-			// deal state updates.
-			//
-			// Note: as an optimization, implementations may cache efficient data structures indicating
-			// which of the following set of updates are redundant and can be skipped.
-			amountSlashedTotal = big.Add(amountSlashedTotal, st.updatePendingDealStatesForParty(rt, client))
```
zixuanzh, line 224: similar issue to WithdrawBalance

actors/builtin/market/market_actor.go
anorth: Good thought. Penalties are probably worth processing immediately, and we know exactly what the deal IDs are in that case anyway.

actors/builtin/market/market_state.go
```diff
@@ -151,22 +118,29 @@ func (st *State) updatePendingDealState(rt Runtime, dealID abi.DealID, epoch abi
...
-		return
+		return amountSlashed, 0
 	}
 
+	next := epoch + DealUpdatesInterval
+	if next > deal.EndEpoch {
+		next = deal.EndEpoch
+	}
+
+	// TODO: can we avoid having this field?
```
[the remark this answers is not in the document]
zixuanzh: i believe this is for lazy deal payment iirc

actors/builtin/market/market_actor.go
whyrusleeping: @zixuanzh this is processed in cron as needed, unless i'm missing your concern

zixuanzh: assuming penalties are not processed immediately (which @anorth suggested we probably should), we are processing slashing at the end of every tipset, the following edge case might happen, in 3 consecutive messages, a miner first gets slashed but its not processed, then it PublishStorageDeals which might lock some available funds that should already be slashed or Withdraw in the withdraw case, and then cron processes the slashed funds which may no longer exist. I think this has to do with the underlying BalanceTable accounting but i have been away from the specifics for a while now.

actors/builtin/market/market_actor.go
```diff
@@ -427,78 +410,62 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
...
+		}
 
-	// TODO: award some small portion of slashed to caller as incentive
+		updatesNeeded := make(map[abi.ChainEpoch][]abi.DealID)
 
-	_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, slashed)
-	builtin.RequireSuccess(rt, code, "failed to burn funds")
-	return nil
-}
+		for i := st.LastCron; i <= rt.CurrEpoch(); i++ {
```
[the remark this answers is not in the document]
whyrusleeping: But its not an AMT, so thats really doable

actors/builtin/market/market_state.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
+const DealUpdatesInterval = 100
```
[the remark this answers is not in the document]
whyrusleeping: 100 is just a completely arbitrary number, its just cutting down the number of deal updates we have to do. Probably doesnt need to be defined in terms of clock time as it doesnt really care about clock time

actors/builtin/market/market_state.go
```diff
@@ -47,72 +50,36 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyMSetCid cid.Cid) *State {
...
+	deal := st.mustGetDeal(rt, dealID)
 
 	if state.SectorStartEpoch == epochUndefined {
 		// Not yet appeared in proven sector; check for timeout.
 		if epoch > deal.StartEpoch {
-			return st.processDealInitTimedOut(rt, dealID)
+			return st.processDealInitTimedOut(rt, dealID, deal, state), 0
 		}
-		return
+		return amountSlashed, 0
```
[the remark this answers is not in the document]
whyrusleeping: I actually think the case youre commenting on will never be hit, it would mean that we called this method for a deal that hasnt been proven, before its starting deadline. Nothing registers a callback for deals at that point, so we really shouldnt ever hit that. I could call abort here, but i'm not 100% sure

actors/builtin/market/market_state.go
```diff
@@ -47,72 +50,36 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyMSetCid cid.Cid) *State {
...
 	}
 
 	if deal.StartEpoch > epoch {
-		return
+		return amountSlashed, 0
```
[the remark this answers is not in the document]
whyrusleeping: Actually, i'd say theres nothing to do even at the start epoch, we probably want to schedule this to run again 100 blocks after the start epoch.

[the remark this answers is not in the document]
whyrusleeping: And I also think this case can never be hit, as we are explicitly scheduling the update to happen at StartEpoch, this shouldnt be able to happen.

actors/builtin/market/market_actor.go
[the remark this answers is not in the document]
anorth: Ack, ok.

Not necessarily in this PR, but I'm increasingly thinking we should use an AMT for things indexed by epoch, so that we can traverse in order. The same goes for the cron callback table in the power actor. I think there's decent scope for bugs leading to missed events and junk state left behind otherwise, especially given that cron doesn't necessarily execute on the epoch scheduled (due to null rounds).

In this case, with the values as integers, it could be an AMT of bitfields; the power actor needs to store more metadata tho.

anorth:
> We ended up with the cron actor's fixed-at-construction table not in this repo, but I guess replicated in each impl. The default parameters to the cron constructor probably belong here somewhere (and the market actor needs to be added).

See #326

actors/builtin/market/market_actor.go
zixuanzh: this is no longer an issue because we are only withdrawing available funds that is not in the locked table. We should also test this.

actors/builtin/market/market_actor.go
zixuanzh: this is no longer an issue because it doesnt try to move funds out of lock table. We should also test the market actor more extensively.

actors/builtin/market/market_actor.go
```diff
@@ -301,7 +285,7 @@ func (a Actor) VerifyDealsOnSectorProveCommit(rt Runtime, params *VerifyDealsOnS
 		}
 
 		for _, dealID := range params.DealIDs {
-			deal, err := states.Get(dealID)
+			deal, _, err := states.Get(dealID)
```
anorth, line 288: Not safe to ignore this `found` retval (it was already unsafe b/c deal could be nil).

actors/builtin/market/market_actor.go
```diff
@@ -402,7 +386,7 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
 			}
 			Assert(deal.Provider == minerAddr)
 
-			state, err := states.Get(dealID)
+			state, _, err := states.Get(dealID)
```
anorth, line 389: Ditto

actors/builtin/market/market_actor.go
```diff
@@ -427,78 +411,139 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
 	return nil
 }
 
-type HandleExpiredDealsParams struct {
-	Deals []abi.DealID // TODO: RLE
-}
+func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
```
anorth, line 414: Add this to `cron.BuiltInEntries` (new since you started this change)

actors/builtin/market/market_actor.go
```diff
@@ -427,78 +411,139 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
...
+
+				if !found {
+					return nil
+				}
+
+				deal := st.mustGetDeal(rt, dealID)
+
+				if state.SectorStartEpoch == epochUndefined {
+					// Not yet appeared in proven sector; check for timeout.
+					if rt.CurrEpoch() > deal.StartEpoch {
```
anorth, line 459: The expiry check is scheduled for `deal.StartEpoch` exactly, and cron runs at the end of the epoch, so I think this should be `>=`

actors/builtin/market/market_actor.go
```diff
@@ -427,78 +411,139 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
...
+						return nil
 					}
-					newlySlashed := st.processDealInitTimedOut(rt, dealID)
-					big.Add(slashed, newlySlashed)
-				} else {
-					// All deals must have timed out.
-					rt.Abortf(exitcode.ErrIllegalArgument, "not all deals have timed out: %d", dealID)
+
+					// This should not be able to happen
+					rt.Abortf(exitcode.ErrIllegalState, "invalid deal state, unstarted, not timed out")
```
anorth, line 471: I agree it shouldn't, but handling it is tricky. This abort will mean state is rolled back and on the next epoch the cron actor will call again with the same result, and never progress past here. Would it be better to `Assert` and halt execution entirely? Alternatively, don't abort here but let all subsequent processing happen (and like, that other place recently, we should look into logging facilities for the runtime). Or re-schedule it for the StartEpoch, thus failing toward correct behaviour.

actors/builtin/market/market_actor.go
```diff
@@ -427,78 +411,139 @@ func (a Actor) OnMinerSectorsTerminate(rt Runtime, params *OnMinerSectorsTermina
...
 			builtin.VerifiedRegistryActorAddr,
 			builtin.MethodsVerifiedRegistry.RestoreBytes,
 			&verifreg.RestoreBytesParams{
-				Address:  deal.Client,
-				DealSize: big.NewIntUnsigned(uint64(deal.PieceSize)),
+				Address:  d.Client,
+				DealSize: big.NewIntUnsigned(uint64(d.PieceSize)),
 			},
 			abi.NewTokenAmount(0),
 		)
```
anorth, line 540: @zixuanzh  we should change this RestoreBytes to support batching, taking a list of (client, size) pairs as params.

actors/builtin/market/market_state.go
```diff
@@ -47,72 +48,26 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyMSetCid cid.Cid) *State {
 		EscrowTable:    emptyMapCid,
 		LockedTable:    emptyMapCid,
 		NextID:         abi.DealID(0),
-		DealIDsByParty: emptyMSetCid,
+		DealOpsByEpoch: emptyMSetCid,
```
anorth, line 51: LastCron should technically be  -1 so it will run in epoch 0.

actors/builtin/market/market_test.go
```diff
@@ -79,7 +79,7 @@ func TestMarketActor(t *testing.T) {
 		assert.Equal(t, emptyMap, state.EscrowTable)
 		assert.Equal(t, emptyMap, state.LockedTable)
 		assert.Equal(t, abi.DealID(0), state.NextID)
-		assert.Equal(t, emptyMultiMap, state.DealIDsByParty)
+		assert.Equal(t, emptyMultiMap, state.DealOpsByEpoch)
```
anorth, line 82: Check LastCron too

actors/builtin/market/market_actor.go
whyrusleeping: I'm leaning towards an assert for now, we can leave a TODO to look into it again later. Hopefully it never shows up

actors/builtin/market/market_actor.go
whyrusleeping: filed an issue for this: https://github.com/filecoin-project/specs-actors/issues/341

anorth:
Thanks. While approved, I request you don't land this for a couple of days so as not to introduce any uncertainty of un-exercised code into our current interop efforts. We may still need to push through some quick fixes that we want to consume quickly and minimally.

If you object to that, or if this lasts more than a few days, we can pick up the effort of maintaining a release branch for interop (which we will do later anyway).

PR #311, opened 2020-04-21
Increase miner vesting test coverage

frrist:
Needs a rebase on #312 once it lands in master.

### DIFF
--- a/actors/builtin/miner/miner_state_test.go
+++ b/actors/builtin/miner/miner_state_test.go
@@ -3,7 +3,6 @@ package miner_test
 import (
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/filecoin-project/go-bitfield"
--- a/actors/builtin/miner/miner_state_test.go
+++ b/actors/builtin/miner/miner_state_test.go
@@ -3,6 +3,7 @@ package miner_test
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/filecoin-project/go-bitfield"
@@ -399,6 +400,7 @@ func TestPostSubmissionsBitfield(t *testing.T) {
 }
 
 func TestVestingFunds_AddLockedFunds(t *testing.T) {
+	// TODO(frrist): merge these top-level methods into fewer
 	t.Run("LockedFunds increases with sequential calls", func(t *testing.T) {
 		harness := constructStateHarness(t, abi.ChainEpoch(0))
 		vspec := &miner.VestSpec{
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -798,6 +798,7 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
 	if err != nil {
 		return err
 	}
+	st.LockedFunds = vestingSum
```
frrist, line 801: we need to set this value else `Unlock*` methods will [always](https://github.com/filecoin-project/specs-actors/pull/311/files#diff-2580e4b82238370ee854a52e8d5428e7R852) [abort](https://github.com/filecoin-project/specs-actors/pull/311/files#diff-2580e4b82238370ee854a52e8d5428e7R896)

actors/builtin/miner/miner_state.go
```diff
@@ -821,8 +822,11 @@ func (st *State) UnlockUnvestedFunds(store adt.Store, currEpoch abi.ChainEpoch,
 	err = vestingFunds.ForEach(&lockedEntry, func(k int64) error {
 		if amountUnlocked.LessThan(target) {
 			if k >= int64(currEpoch) {
-				unlockAmount := big.Max(big.Sub(target, amountUnlocked), lockedEntry)
+				unlockAmount := big.Min(big.Sub(target, amountUnlocked), lockedEntry)
```
frrist, line 825: take `Min` here else `lockedEntry` can be negative, and thus never deleted.

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+		vspec := &miner.VestSpec{
+			InitialDelay: 0,
+			VestPeriod:   1,
+			StepDuration: 1,
+			Quantization: 1,
+		}
+		vestStart := abi.ChainEpoch(10)
+		vestSum := abi.NewTokenAmount(100)
+		harness.addLockedFunds(vestStart, vestSum, vspec)
+		amountVested := harness.unlockVestedFunds(vestStart + 2)
```
anorth, line 459: Please demonstrate what happens at `vestStart`? What about `vestStart + 1`?

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+		assert.Equal(t, vestSum, amountVested)
+		assert.Zero(t, harness.s.LockedFunds.Int64())
```
anorth, line 461: This assertion is more useful of we assert it's not zero before vesting

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+			} else {
+				assert.Equal(t, abi.NewTokenAmount(0), amountVested)
+			}
+		}
+		assert.Equal(t, vestSum, totalVested)
+		assert.Zero(t, harness.s.LockedFunds.Int64())
+		assert.True(t, harness.vestingFundsStoreEmpty())
+	})
+
+	t.Run("Unlock unvested funds leaving bucket with modified amount", func(t *testing.T) {
```
anorth, line 616: Move tests primarily about _un_vested funds to a different method. All the complex specs are pretty much irrelevant to them.

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+	})
+	t.Run("Sum of all entries in vesting store is equal to vested sum", func(t *testing.T) {
+		harness := constructStateHarness(t, abi.ChainEpoch(0))
+		vspec := &miner.VestSpec{
+			InitialDelay: 0,
+			VestPeriod:   10,
+			StepDuration: 1,
+			Quantization: 2,
+		}
+		vestStart := abi.ChainEpoch(10)
```
anorth, line 476: This just makes all the numbers below hard to understand. We need at least one test that demonstrates that vestStart is treated correctly, but that test should be very focussed on that. Just use zero elsewhere.

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+		amountVested := harness.unlockVestedFunds(vestStart + 2)
+		assert.Equal(t, vestSum, amountVested)
+		assert.Zero(t, harness.s.LockedFunds.Int64())
+		assert.True(t, harness.vestingFundsStoreEmpty())
+		// subsequent calls to unlock return 0
+		amountVested = harness.unlockVestedFunds(vestStart + 2)
+		assert.Zero(t, amountVested)
+
+	})
+	t.Run("Sum of all entries in vesting store is equal to vested sum", func(t *testing.T) {
```
anorth, line 468: Whoah, this goes from really simple to eye-watering complicated very quickly. Please try to write some tests between the first one and this one, each one of which changes _one_ thing about the schedule:
- vest all at once after non-zero cliff
- period = 2, step = 1,  Q=1 -> [50, 50]
- period = 2, step = 1, Q=2 ->  [0, 100]
- period = 3 , step= 1,  Q=1 -> [33, 33, 34] ???
- period = 3 , step= 1,  Q=2 -> [0, 66, 0, 34] ???
- period = 2, step = 2,  Q=1 -> [0, 100]
- period = 5, step = 2, Q=1 -> [0, 40, 0, 40, 0, 20] ???
- period = 5, step = 2, Q=2 -> [0, 40, 0, 40, 0, 20] ???
- period = 5, step = 3, Q=1 -> [0, 0, 60, 0, 0, 40] ???

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+		// Entry: 12 Funds: 20
+		// Entry: 14 Funds: 20
+		// Entry: 16 Funds: 20
+		// Entry: 18 Funds: 20
+		// Entry: 20 Funds: 20
+
+		totalVested := abi.NewTokenAmount(0)
+		for e := vestStart; e <= vestStart+vspec.VestPeriod+1; e++ {
+			amountVested := harness.unlockVestedFunds(e)
+			if e == 13 || e == 15 || e == 17 || e == 19 || e == 21 {
```
anorth, line 488: These are pretty gross. Manually calculated numbers for the amount vested are ok, but we can do better here. 

if you use relatively small numbers, I think it would be better to just write out the expected vesting schedules per epoch directly, like I have done in the comment above.

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+			Entry: 128 Funds: 100
+			Entry: 136 Funds: 80
+			Entry: 146 Funds: 100
+			Entry: 154 Funds: 80
+			Entry: 164 Funds: 100
+			Entry: 172 Funds: 80
+			Entry: 182 Funds: 100
+			Entry: 190 Funds: 80
+			Entry: 200 Funds: 100
+		*/
```
anorth, line 595: These comments are just saying exactly the same thing as the code, and don't help.

actors/builtin/miner/miner_state_test.go
```diff
@@ -440,6 +442,252 @@ func TestVestingFundsStore(t *testing.T) {
...
+		assert.True(t, harness.vestingFundsStoreEmpty())
+	})
+
+	t.Run("Unlock unvested funds leaving bucket with modified amount", func(t *testing.T) {
+		harness := constructStateHarness(t, abi.ChainEpoch(0))
+		vspec := &miner.VestSpec{
+			InitialDelay: 0,
+			VestPeriod:   100,
+			StepDuration: 9,
+			Quantization: 2,
```
anorth, line 622: Use a much simpler scheme for these tests - the quant handling isn't the point.

actors/builtin/miner/miner_state_test.go
```diff
@@ -398,6 +398,155 @@ func TestPostSubmissionsBitfield(t *testing.T) {
...
+		},
+		{
+			desc: "vest funds with period=5 step=2",
+			vspec: &miner.VestSpec{
+				InitialDelay: 0,
+				VestPeriod:   5,
+				StepDuration: 2,
+				Quantization: 1,
+			},
+			vepocs: []int64{0, 0, 0, 40, 0, 40, 0, 20, 0},
```
anorth, line 476: It would be interesting to see this one also with an initial delay of 1 (which I expect to offset the whole thing by 1)

actors/builtin/miner/miner_state_test.go
```diff
@@ -398,6 +398,155 @@ func TestPostSubmissionsBitfield(t *testing.T) {
...
+		},
+		{
+			desc: "vest funds with delay=5 period=5 step=1 quantization=1",
+			vspec: &miner.VestSpec{
+				InitialDelay: 5,
+				VestPeriod:   5,
+				StepDuration: 1,
+				Quantization: 1,
+			},
+			vepocs: []int64{0, 0, 20, 20, 20, 20, 20, 0},
```
anorth, line 526: This one looks weird because it looks like the delay is being ignored. I suggest removing the  `+tc.vspec.InitialDelay` on line 538, and putting in the leading zeros in this result.

Alternatively break out a new set of tests for non-zero delays.

actors/builtin/miner/miner_state_test.go
frrist: added thes test below this one.

actors/builtin/miner/miner_state_test.go
frrist: good idea, adjusted in latest commit.

PR #312, opened 2020-04-22
fix bugs in miner vesting

frrist:
fix: bug in miner LockedFunds and add test
- LockedFunds was not updated when AddLockedFunds was called.

fix: bug in UnlockUnvestedFunds and add test
- unlockAmount was incorrectly caluculated such that entries were
negative and never removed from vestingStore.
- vestingFunds was not updated when its entires changes.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -823,15 +823,15 @@ func (st *State) UnlockUnvestedFunds(store adt.Store, currEpoch abi.ChainEpoch, 
 		if amountUnlocked.LessThan(target) {
 			if k >= int64(currEpoch) {
 				unlockAmount := big.Min(big.Sub(target, amountUnlocked), lockedEntry)
-
-				lockedEntry = big.Sub(lockedEntry, unlockAmount)
-				if err := vestingFunds.Set(uint64(k), &lockedEntry); err != nil {
-					return err
-				}
 				amountUnlocked = big.Add(amountUnlocked, unlockAmount)
+				lockedEntry = big.Sub(lockedEntry, unlockAmount)
 
 				if lockedEntry.IsZero() {
 					toDelete = append(toDelete, uint64(k))
+				} else {
+					if err = vestingFunds.Set(uint64(k), &lockedEntry); err != nil {
+						return err
+					}
 				}
 			}
 			return nil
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -821,8 +822,12 @@ func (st *State) UnlockUnvestedFunds(store adt.Store, currEpoch abi.ChainEpoch,
 	err = vestingFunds.ForEach(&lockedEntry, func(k int64) error {
 		if amountUnlocked.LessThan(target) {
 			if k >= int64(currEpoch) {
-				unlockAmount := big.Max(big.Sub(target, amountUnlocked), lockedEntry)
+				unlockAmount := big.Min(big.Sub(target, amountUnlocked), lockedEntry)
+
 				lockedEntry = big.Sub(lockedEntry, unlockAmount)
+				if err := vestingFunds.Set(uint64(k), &lockedEntry); err != nil {
+					return err
+				}
```
anorth, line 830: Let's put this inside an `else` clause after line 833, so we either update or delete it.

actors/builtin/miner/miner_state_test.go
```diff
@@ -398,13 +398,76 @@ func TestPostSubmissionsBitfield(t *testing.T) {
...
+		VestPeriod:   1,
+		StepDuration: 1,
+		Quantization: 1,
+	}
+
+	vestStart := abi.ChainEpoch(10)
+	vestSum := abi.NewTokenAmount(100)
+
+	harness.addLockedFunds(vestStart, vestSum, vspec)
+	unvestedFunds := harness.unlockUnvestedFunds(vestStart, abi.NewTokenAmount(49))
```
anorth, line 436: `unlocked` rather than `unvestedFunds`

actors/builtin/miner/miner_state_test.go
```diff
@@ -398,13 +398,76 @@ func TestPostSubmissionsBitfield(t *testing.T) {
...
+
+		harness.addLockedFunds(vestStart, vestSum, vspec)
+		assert.Equal(t, big.Mul(vestSum, big.NewInt(2)), harness.s.LockedFunds)
+	})
+
+}
+
+func TestVestingFundsStore(t *testing.T) {
+	harness := constructStateHarness(t, abi.ChainEpoch(0))
+	vspec := &miner.VestSpec{
```
zixuanzh, line 425: lets try other values here for the vesting spec in another PR? InitialDelay and VestPeriod will be much longer than StepDuration which is longer than Quantization. Thanks!

anorth: @frrist has many more tests coming in #311, this is just a quick one to fix the bugs found so far.

zixuanzh: great!

PR #313, opened 2020-04-22
Merge mock syscalls with runtime and set proper expectations for signature verification.

anorth:
This is pre-work to setting mock expectations for seal/post verification for miner tests. Unfortunately it turned into giant dive into the payment channel tests which were doing some weird things.

This PR sets proper mock expectations for signature verification, to be followed by other syscalls.


actors/builtin/paych/paych_test.go
```diff
@@ -29,19 +29,20 @@ func TestExports(t *testing.T) {
...
 	payerAddr := tutil.NewIDAddr(t, 101)
+	payeeAddr := tutil.NewIDAddr(t, 102)
 	callerAddr := tutil.NewIDAddr(t, 102)
 
+	actor := pcActorHarness{Actor{}, t, paychAddr, payerAddr, payeeAddr}
+
 	t.Run("can create a payment channel actor", func(t *testing.T) {
 		builder := mock.NewBuilder(ctx, paychAddr).
 			WithCaller(callerAddr, builtin.InitActorCodeID).
-			WithActorType(paychAddr, builtin.AccountActorCodeID).
```
anorth, line 41: This was setting the payment channel actor as having an account actor code CID 😕. I think due to (ab)using the actor as the channel's payee, which is invalid and only worked here due to hacking the runtime like this.

actors/builtin/paych/paych_test.go
```diff
@@ -407,42 +421,56 @@ func TestActor_UpdateChannelStateMergeFailure(t *testing.T) {
...
+		rt.ExpectVerifySignature(*ucp.Sv.Signature, actor.payee, voucherBytes(t, &ucp.Sv), nil)
 		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
 			rt.Call(actor.UpdateChannelState, ucp)
 		})
+		rt.Verify()
 	})
 }
 
 func TestActor_UpdateChannelStateExtra(t *testing.T) {
-	rt1, actor1, sv1 := requireCreateChannelWithLanes(t, context.Background(), 1)
```
anorth, line 417: Constructing the runtime outside of the `t.Run()` blocks below is invalid since it leads to the wrong `testing.T` being used. There are other places that do this too, but the tests still happen to pass so I didn't push further.

actors/builtin/paych/paych_test.go
```diff
@@ -721,13 +757,14 @@ func requireCreateChannelWithLanes(t *testing.T, ctx context.Context, numLanes i
 }
 
 func requireAddNewLane(t *testing.T, rt *mock.Runtime, actor *pcActorHarness, params laneParams) *SignedVoucher {
-	sig := &crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte("doesn't matter")}
+	sig := &crypto.Signature{Type: crypto.SigTypeBLS, Data: []byte{0, 1, 2, 3, 4, 5, 6, 7}}
```
anorth, line 760: This is just easier to recognize in logs/debugger.

actors/builtin/paych/paych_test.go
frrist: Probably worth leaving a `TODO/FIXME` flagging this

anorth: Turned out there was only one more case, so I fixed it.

PR #314, opened 2020-04-22
Create README.md

anorth:
Explains versioning.

Note that the major version will change quite rapidly before mainnet, but then much more slowly thereafter (usually corresponding to a network upgrade). I don't think there's much value attempting to align major version 1 with mainnet launch, here or in other upstream libraries. E.g. proofs will be something like v26.


anorth:
Thanks for landing this. The versioning description is inaccurate now (doesn't play nice with Go), but I'll make an update.

PR #315, opened 2020-04-22
Increase placeholder parameter values

arajasek:
While testing deal workflow with large files, we run into the provider collateral max bound, so I'm raising it (along with client collateral max and deal price per epoch max).


actors/builtin/market/policy.go
```diff
@@ -8,15 +8,15 @@ func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.C
 }
 
 func dealPricePerEpochBounds(size abi.PaddedPieceSize, duration abi.ChainEpoch) (min abi.TokenAmount, max abi.TokenAmount) {
-	return abi.NewTokenAmount(0), abi.NewTokenAmount(1 << 40) // PARAM_FINISH
+	return abi.NewTokenAmount(0), abi.NewTokenAmount(1 << 60) // PARAM_FINISH
```
anorth, line 11: A good value for all of these BS values would be the total money supply: 2billion * 10^18. How about a constant for that perhaps in `abi`, and then replace all of these with that.

@zixuanzh these un-thought-through placeholders are a real pain.

zixuanzh: ack

arajasek: Yeah, that's a good idea...Change implemented

PR #321, opened 2020-04-24
Power actor proceeds after failure of cron callback.

anorth:
This behaviour is worth paying attention to. We don't expect miner cron callbacks to abort, ever. Previously the signal was very loud when they did. Now the network will proceed, but it'll be easy to miss this unexpected and probably-major error.

(This PR doesn't fix the miner cron, which fails to handle being invoked after its scheduled round, if it was scheduled on a null).

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -428,18 +428,19 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
 		}
 
 		st.LastEpochTick = rtEpoch
-
 		return nil
 	})
 
 	for _, event := range cronEvents {
-		_, code := rt.Send(
+		_, _ = rt.Send(
 			event.MinerAddr,
 			builtin.MethodsMiner.OnDeferredCronEvent,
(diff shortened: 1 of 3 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -428,18 +428,19 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
...
+		_, _ = rt.Send(
 			event.MinerAddr,
 			builtin.MethodsMiner.OnDeferredCronEvent,
 			vmr.CBORBytes(event.CallbackPayload),
 			abi.NewTokenAmount(0),
 		)
-		builtin.RequireSuccess(rt, code, "failed to defer cron event")
+		// The exit code is ignored. If a callback fails, this actor continues to invoke other callbacks
+		// and persists state removing the failed event from the event queue. It won't be tried again.
+		// A log message would really help here, though.
```
magik6k, line 443: It's fairly easy to create tooling which tracks this, I can look into this

magik6k:
(rebased on master)

PR #322, opened 2020-04-24
Store proving period start in state explicitly and fix cron to handle invocation off expected schedule

anorth:
Handling the possibility of the cron callback coming after the scheduled epoch means we can no longer infer the proving period from the current epoch. Instead, it's stored in state and updated for each new period. This does make some things more directly observable from the state, at least.

Apologies @magik6k that this may require tweaks to the wpost worker integration.

### DIFF
--- a/actors/builtin/miner/deadlines.go
+++ b/actors/builtin/miner/deadlines.go
@@ -49,7 +49,7 @@ func (d *DeadlineInfo) NextPeriodStart() abi.ChainEpoch {
 func ComputeProvingPeriodDeadline(periodStart, currEpoch abi.ChainEpoch) *DeadlineInfo {
 	periodProgress := currEpoch - periodStart
 	deadlineIdx := uint64(periodProgress / WPoStChallengeWindow)
-	if deadlineIdx < 0 { // Period not yet started.
+	if periodProgress < 0 { // Period not yet started.
 		deadlineIdx = 0
 	}
 	deadlineOpen := periodStart + (abi.ChainEpoch(deadlineIdx) * WPoStChallengeWindow)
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -688,9 +684,9 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 			if contains {
 				// This could happen if attempting to declare a fault for a deadline that's already passed,
 				// detected and added to Faults above.
-				// Wait for the fault detection at proving period end, or submit again omitting deadlines that have
-				// passed.
-				// Alternatively, we could subtract the detected faults from new faults.
+				// The miner must for the fault detection at proving period end, or submit again omitting
```
magik6k, line 687:
```suggestion
				// The miner must wait for the fault detection at proving period end, or submit again omitting
```

actors/builtin/miner/deadlines.go
```diff
@@ -41,22 +45,13 @@ func (d *DeadlineInfo) NextPeriodStart() abi.ChainEpoch {
...
-	} else {
-		periodProgress = WPoStProvingPeriod - (boundary - currModulus)
-	}
-
-	periodStart := currEpoch - periodProgress
+// Returns deadline-related calculations for a proving period start and current epoch.
+func ComputeProvingPeriodDeadline(periodStart, currEpoch abi.ChainEpoch) *DeadlineInfo {
+	periodProgress := currEpoch - periodStart
 	deadlineIdx := uint64(periodProgress / WPoStChallengeWindow)
+	if periodProgress < 0 { // Period not yet started.
```
magik6k, line 52: `deadlineIdx` is an uint, so it couldn't be negative here

anorth:
> One edge case I'd like to ask about is what happens in the rather unlikely, and quite catastrophic case of no blocks being produced for > 1 proving period (say 2 days of null blocks)? Realistically I don't think we need to handle this case, but it may be worth thinking about.

I have absolutely been thinking about this. It's on my personal backlog, I'll transfer it to an issue.

I think the miner actor is now close to handling this appropriately (well, at least consistently), but I'm not sure the power actor's cron callback table is robust to it.

PR #323, opened 2020-04-24
clear new sectors bitfield when assigning sectors

whyrusleeping:
please ignore my being OCD about the err assignments.

Line 1008 is the fix, I think this is the only place this is needed

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1005,9 +1005,7 @@ func handleProvingPeriod(rt Runtime) {
 				err = st.SaveDeadlines(store, deadlines)
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to store new deadlines")
 
-				if err := st.ClearNewSectors(); err != nil {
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1004,6 +1004,8 @@ func handleProvingPeriod(rt Runtime) {
 				// Store updated deadline state.
 				err = st.SaveDeadlines(store, deadlines)
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to store new deadlines")
+
+				st.NewSectors = abi.NewBitField()
```
magik6k, line 1008: Should be
```suggestion
				if err := st.ClearNewSectors(); err != nil {
					t.Abortf(exitcode.ErrIllegalState, "failed to clear newsectors %v", err)
				}
```

magik6k: (It does exactly the same thing, but also returns nil)

anorth:
Thanks for the fix. 
I love OCD, but _I'm_ OCD about removing all possible warnings, and most of those err assignments were that way because the `err` variable shadows another var of the same name when redeclared in that way, which some static analysis tools (incl GoLand) can detect.

PR #325, opened 2020-04-25
Resolve key addresses to ID addresses in multisig validation

arajasek:
We currently allow the approved signers of a multisig to not be ID addresses. However, when we try to validate that a caller is an approved signer, we only have the ID address of the caller. This causes an issue when checking if a caller is an approved signer.

We probably *don't* want to enforce that approved signers must be ID addresses, since these signers might not be registered at the time of multisig creation.

This PR changes the `isSigner` method to (attempt to) resolve any non-ID addresses in the approved signers, and to replace a non-ID address if this resolution is successful.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -132,7 +132,10 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	var txnID TxnID
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		a.validateSigner(rt, &st, callerAddr)
+		if !isSigner(rt, &st, callerAddr) {
+			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", callerAddr)
+		}
+
 		txnID = st.NextTxnID
 		st.NextTxnID += 1
 
@@ -153,6 +156,8 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	var emptyArray []byte
 	applied, ret, code := a.approveTransaction(rt, txnID, emptyArray, false)
 
+	// Note: this transaction ID may not be stable across chain re-orgs.
+	// The proposal hash may be provided as a stability check when approving.
 	return &ProposeReturn{
 		TxnID:   txnID,
 		Applied: applied,
@@ -180,7 +185,9 @@ func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *ApproveReturn {
 	callerAddr := rt.Message().Caller()
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		a.validateSigner(rt, &st, callerAddr)
+		if !isSigner(rt, &st, callerAddr) {
+			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", callerAddr)
+		}
 		return nil
 	})
 	approved, ret, code := a.approveTransaction(rt, params.ID, params.ProposalHash, true)
@@ -197,7 +204,10 @@ func (a Actor) Cancel(rt vmr.Runtime, params *TxnIDParams) *adt.EmptyValue {
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		a.validateSigner(rt, &st, callerAddr)
+		if !isSigner(rt, &st, callerAddr) {
+			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", callerAddr)
+		}
+
 		txn, err := st.getPendingTransaction(adt.AsStore(rt), params.ID)
 		if err != nil {
 			rt.Abortf(exitcode.ErrNotFound, "failed to get transaction for cancel: %v", err)
@@ -225,7 +235,7 @@ func (a Actor) Cancel(rt vmr.Runtime, params *TxnIDParams) *adt.EmptyValue {
 }
 
 type AddSignerParams struct {
-	Signer   addr.Address // must be an ID protocol address.
+	Signer   addr.Address
 	Increase bool
 }
 
@@ -235,8 +245,11 @@ func (a Actor) AddSigner(rt vmr.Runtime, params *AddSignerParams) *adt.EmptyValu
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		if a.isSigner(rt, &st, params.Signer) {
-			rt.Abortf(exitcode.ErrIllegalArgument, "party is already a signer")
+		// FIXME: if params.Signer is a pubkey address, but the corresponding ID address is already a signer,
+		// this won't resolve to detect that and the logically same account can end up present twice.
+		// This needs a test.
+		if isSigner(rt, &st, params.Signer) {
+			rt.Abortf(exitcode.ErrIllegalArgument, "%s is already a signer", params.Signer)
 		}
 		st.Signers = append(st.Signers, params.Signer)
 		if params.Increase {
@@ -248,7 +261,7 @@ func (a Actor) AddSigner(rt vmr.Runtime, params *AddSignerParams) *adt.EmptyValu
 }
 
 type RemoveSignerParams struct {
-	Signer   addr.Address // must be an ID protocol address.
+	Signer   addr.Address
 	Decrease bool
 }
 
@@ -258,7 +271,9 @@ func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.Emp
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		a.validateSigner(rt, &st, params.Signer)
+		if !isSigner(rt, &st, params.Signer) {
+			rt.Abortf(exitcode.ErrNotFound, "%s is not a signer", params.Signer)
+		}
 
 		if len(st.Signers) == 1 {
 			rt.Abortf(exitcode.ErrForbidden, "cannot remove only signer")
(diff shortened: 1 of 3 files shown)


actors/builtin/multisig/multisig_state.go
```diff
@@ -33,11 +34,21 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
...
+func (st *State) isSigner(rt vmr.Runtime, party address.Address) bool {
+	for i, ap := range st.Signers {
 		if party == ap {
 			return true
 		}
+
+		if ap.Protocol() != address.ID {
+			idAddr, found := rt.ResolveAddress(ap)
+			if found {
+				st.Signers[i] = idAddr
```
whyrusleeping, line 46: if we're going to do this, why not do the resolution in the constructor?

arajasek: The multisig actor's constructor? Because the address might not be registered yet, no?

anorth:
@arajasek I think you are trying to resolve #184. There is some prior art to review in #195.

The resolution is in [this comment thread](https://github.com/filecoin-project/specs-actors/pull/195#issuecomment-590570144) (linked from the issue).

The simplicity gains from resolving all address at construction are great. The only reason I didn't do it immediately when resolving addresses elsewhere was that I wanted to check with @whyrusleeping whether I was missing an important use case. I think we're all aligned that the only address not to resolve is the target for a specific transaction (which we don't need to validate anyway).

actors/builtin/multisig/multisig_state.go
```diff
@@ -33,11 +34,21 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 	return big.Mul(unitLocked, big.Sub(big.NewInt(int64(st.UnlockDuration)), big.NewInt(int64(elapsedEpoch))))
 }
 
-func (st *State) isSigner(party address.Address) bool {
-	for _, ap := range st.Signers {
+func (st *State) isSigner(rt vmr.Runtime, party address.Address) bool {
```
anorth, line 37: It is a rule in actor code that the state object must never use the runtime object, since this makes such methods impossible to call from outside actor code. Application logic like resolving addresses belongs in the actor itself.

(There are a few lingering violations of this rule)

arajasek:
Thanks for supplying the additional context, @anorth, I didn't realise this was a known issue. 

I don't much like the idea that we should resolve addresses only at construction (which seems to be the conclusion in #195). I don’t have an important use-case in mind, but I think a Filecoin address can meaningfully exist in meatspace without having been registered.

And in this case, all we “should” want is that the address plausibly exist in the real world — it’s not a Filecoin account yet, and doesn’t have to be, it’s just some address that conforms to our siggy schemes.

Happy to implement the resolution at construction if that's what we decide, though

anorth:
I concur that all addresses logically exist. The question is whether we want the complexity of allowing such addresses to be recorded as signers of a multisig. The address certainly won't be able to sign anything until it exists (has value transferred to it), because it won't be able to pay for a message.

arajasek:
Yeah, and the obvious argument is that the address needing to sign anything could come well after the wallet itself needing to be created. But the use case I'm describing might be entirely artificial.

actors/builtin/multisig/multisig_state.go
whyrusleeping: Ah right, I remember that now. Maybe instead of resolving addresses to ID addresses, we should resolve them all to their non-id variants for comparison and storage?

actors/builtin/multisig/multisig_state.go
whyrusleeping: @arajasek i'd just make this a method on the actor then, pass in the state as well, and move this function to the actor file.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -229,7 +228,7 @@ func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.Emp
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		if !st.isSigner(params.Signer) {
+		if !a.isSigner(rt, st, params.Signer) {
```
anorth, line 231: Use validateSigner?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -264,11 +263,11 @@ func (a Actor) SwapSigner(rt vmr.Runtime, params *SwapSignerParams) *adt.EmptyVa
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		if !st.isSigner(params.From) {
+		if !a.isSigner(rt, st, params.From) {
 			rt.Abortf(exitcode.ErrNotFound, "Party not found")
 		}
```
anorth, line 268: Use validateSigner?

actors/builtin/multisig/multisig_state.go
anorth: @whyrusleeping are you explicitly changing your mind from [this statement](https://github.com/filecoin-project/specs-actors/pull/195#issuecomment-594206413)?

Note that there's no mechanism at present to resolve actor ID addresses to their reorg-stable forms; we can do it for account actors only because the pk address is stored explicitly in state.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -368,11 +367,30 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
 }
 
 func (a Actor) validateSigner(rt vmr.Runtime, st *State, address addr.Address) {
-	if !st.isSigner(address) {
+	if !a.isSigner(rt, *st, address) {
 		rt.Abortf(exitcode.ErrForbidden, "party not a signer")
 	}
 }
 
+func (a Actor) isSigner(rt vmr.Runtime, st State, address addr.Address) bool {
```
anorth, line 375: Pass the state by pointer, not by copy. The mutation below will have no effect otherwise.

arajasek: D'oh, sorry!

arajasek:
Some feedback from a partner regarding whether we should be doing this:

> Personally I think it’s more user friendly not to have to rely on an account to have been created before using it for a multisig.
> I imagine that it will be  common for wallets and exchanges to make multisigs for users without first using the required internal addresses.
> I think most of the time, people that want to make multisig wallets will be doing it fresh without providing/using the input addresses

frrist:
Doing some housekeeping. @arajasek this PR aims to close https://github.com/filecoin-project/specs-actors/issues/184 right? If so im going to close out https://github.com/filecoin-project/specs-actors/pull/195 in favor of this one.

arajasek:
> #184

yessir

> Doing some housekeeping. @arajasek this PR aims to close #184 right? If so im going to close out #195 in favor of this one.

yessir

actors/builtin/multisig/multisig_state.go
whyrusleeping: @anorth Yeah, thinking about it in this context, the UX of forcing those addresses to exist is pretty annoying. Resolving on check like this seems acceptable

whyrusleeping:
Maybe we can resolve in the constructor still, but resolve to the pubkey or stable addresses? I'm on the fence for the right solution, but think that accounts probably shouldnt have to exist to be declared as the signer of a multisig

anorth:
I'm ok with the requirement that addresses shouldn't have to exist, so let's work to that.

I don't think we can resolve to pubkey/stable addresses because there is no mapping from ID->stable address for actors. The _reverse_ exists in the init actor, and account actors have a workaround of storing their pubkey in state, but other actors don't.

So I think this means keeping the ID address as canonical and resolving late if necessary.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -368,11 +363,30 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
 }
 
 func (a Actor) validateSigner(rt vmr.Runtime, st *State, address addr.Address) {
-	if !st.isSigner(address) {
+	if !a.isSigner(rt, st, address) {
 		rt.Abortf(exitcode.ErrForbidden, "party not a signer")
 	}
 }
 
+func (a Actor) isSigner(rt vmr.Runtime, st *State, address addr.Address) bool {
```
anorth, line 371: Nit: neither this nor `validateSigner` need to be methods of `Actor` since they don't reference it at all. This is the general case for all actors: methods are either exported or not methods.

Prefer to make these free functions, and also move them to the bottom of this file so all the exported methods come first.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -423,21 +417,35 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
...
+		case addr.ID:
+			if candidateIdAddr == ap {
+				return true
+			}
+		default:
 			idAddr, found := rt.ResolveAddress(ap)
 			if found {
 				st.Signers[i] = idAddr
-				if idAddr == address {
+				if idAddr == candidateIdAddr {
```
anorth, line 442: I think this is correct, but it's hard to mentally check because there are three different places where the addresses are compared. I think it would be simpler for the algo to be:

- candidate = resolve(candidate)
- for each signer:
  - signer = resolve(signer)
  - if candidate == signer: return true

... where `resolve()` include the checking the protocol etc as it does already here.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -423,21 +417,35 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
 }
 
 func isSigner(rt vmr.Runtime, st *State, address addr.Address) bool {
+	candidateIdAddr := address
```
anorth, line 420: It won't necessarily resolve to an ID address, so `candidateResolved` would be slightly more accurate.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -407,10 +416,28 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
...
 
-func (a Actor) validateSigner(rt vmr.Runtime, st *State, address addr.Address) {
-	if !st.isSigner(address) {
-		rt.Abortf(exitcode.ErrForbidden, "party not a signer")
+func isSigner(rt vmr.Runtime, st *State, address addr.Address) bool {
+	candidateResolved := resolve(rt,address)
+
+	for _, ap := range st.Signers {
+		signerResolved := resolve(rt, ap)
+		if signerResolved ==  candidateResolved {
```
anorth, line 424: Extra space after `==`

PR #328, opened 2020-04-27
Remove unused SectorSize from VerifyDealsOnSectorProveCommitParams

anorth:
FYI @magik6k


anorth:
I don't quite understand your request, but I think it's orthogonal to removing this unused field.

The market actor already confirms that each deal expires no later than the sector expiry, so the "time" part is already satisfied. I think the simple fact that the size of the deals sums to less than the size of the sector is implicit in the computation of the data commitment from the deal CIDs and sizes. The sector size is implicit in the sector proof type already passed as a parameter to that syscall.

PR #334, opened 2020-04-28
require success in reward actor send reward

frrist:


frrist:
whoops, this is already in

Issue #335, opened 2020-04-28
Fix registered proof type in miner actor at construction, remove from proof submission messages


anorth:
At present, each SectorPreCommitInfo carries a RegisteredProofType, which is replicated to the sector info stored on chain etc, generally implying that a miner could have sectors with different proof types (though the sector _size_ is fixed)

However, PoSt scheduling depends on the proof type (which determines partition size). It doesn't make sense for a miner to have sectors with different proof types. This flexibility is unnecessary, costs chain space, and may lead to errors.

- Fix the seal proof type (from which the PoSt proof type may be derived) along with the sector size at miner actor creation.
- Remove RegisteredProof from miner `SectorPreCommitInfo`, `OnChainSectorInfo`
- Remove RegisteredProof from `OnChainSealVerifyInfo`
- Remove RegisteredProof from `abi.SectorInfo`, put it in `Window/WinningPoStVerifyInfo` instead.

See a related clean-up of these structure requested by #276. Consider doing both together, so we can update filecoin-ffi just once.

anorth:
We have fixed the proof type in the miner actor ctor, but not yet removed it from all the messages. Leaving it there does give us some flexibility. TBD.

anorth:
Remaining work (uncertain) deferred to  #506.

PR #336, opened 2020-04-29
Fix seal proof type in miner actor and parameterize WPoSt partition size by it

anorth:
This is part of the way to #335. The registered proof types in all the sector states and messages are now redundant.

FYI @magik6k, @acruikshank

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -94,8 +94,16 @@ type MinerInfo struct {
 	// Libp2p identity that should be used when connecting to this miner.
 	PeerId peer.ID
 
-	// Amount of space in each sector committed to the network by this miner.
+	// The proof type used by this miner for sealing sectors.
+	SealProofType abi.RegisteredProof
+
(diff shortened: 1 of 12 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -94,8 +94,16 @@ type MinerInfo struct {
...
+	// The proof type used by this miner for sealing sectors.
+	SealProofType abi.RegisteredProof
+
+	// Amount of space in each sector committed by this miner.
+	// This is computed from the proof type and represented here redundantly.
 	SectorSize abi.SectorSize
+
+	// The number of sectors in each Window PoSt partition (proof).
+	// This is computed from the proof type and represented here redundantly.
+	WindowPoStPartitionSectors uint64
```
whyrusleeping, line 106: I guess its fine to put this here, since we never rewrite this.

anorth: You could convince me to take it out pretty easily, but if it [the code that derives it] ever changed we'd be in trouble, and this helps make the state transparent.

Issue #317, opened 2020-04-23
Fix reward actor technical debt


anorth:
The reward actor is not ready for prime time.

- No checks that the actor's actual balance is compatible with the reward schedule
- In case of a null round, the reward paid out will be the reward corresponding to the null epoch, not for an actually mined epoch.
- semi-done: No documentation of high level intent or what any of the state variables mean or functions do

- ~No tests~
- ~Initial state should reflect genesis configuration, not zeros (#374)~
- ~Logic doesn't handle being called more than once each round (multiple blocks)~
- ~Logic doesn't handle not being called every round (for null tipsets), undercounts~
- ~Doesn't divide by expected number of leaders when computing reward for a miner~
- ~The `Simple/BaselineSupply` state variables account for expected reward, not actual reward paid; money supply will always be an overestimate~
- ~Total allocation values hardcoded as constants, which are wrong; should be inferred from actual balance~

- ~`BlockTimeSeconds` duplicated and inconsistent with miner actor~
- ~`Lambda` parameters totally opaque and undocumented~
- ~Uninitialized `LastPerEpochReward` state~
- ~Unused argument in ConstructState~
- ~Unused `TicketCount` in `AwardBlockRewardParams` should be removed~

anorth:
FYI @Kubuxu

anorth:
Merging #329 into this:
> Existing code does an inverse mapping of realizedCumsum to effectiveNetworkTime but the integer math is always returning zero unless the daily power exceeds the baseline. This is not the intention and we should introduce a notion of precision here. If baseline is 100PiB on the first day and only 50PiB shows up, network time advances by half a day.

anorth:
Thanks for your great work here @Kubuxu.

I'd be happy to close this issue after:
- Add checks that the actor's actual balance is compatible with the reward schedule
- Document in-code the fact that "In case of a null round, the reward paid out will be the reward corresponding to the null epoch, not for an actually mined epoch." and why that's the case

@ZenGround0 would you mind making those final tweaks for us?

ZenGround0:
> Document in-code the fact that "In case of a null round, the reward paid out will be the reward corresponding to the null epoch, not for an actually mined epoch." and why that's the case

[...]

PR #338, opened 2020-05-01
document reward-minting taylorSeriesExpansion

davidad:
Hoping to pay down my share of the technical debt mentioned in #317, by cleaning up and adding many paragraphs of documentation to the algorithm I devised for computing f(t)=1-(1/2)^(t/6yr) to atto-level precision. Please let me know if there are further questions about what's going on here.


actors/builtin/reward/reward_state.go
```diff
@@ -49,31 +40,174 @@ func ConstructState(emptyMultiMapCid cid.Cid) *State {
...
+//
+//   When we have the numerator and denominator for a given term set up, we
+//   compute their fixed-point fraction by left-shifting the numerator before
+//   performing integer division.
+//
+//   Finally, at the end of each loop, we remove unnecessary bits of precision
+//   from both the numerator and denominator accumulators to control the
+//   computational complexity of the bigint multiplications.
+
+// Fixed-point precision used internally and for output
```
anorth, line 134:
```suggestion
// Fixed-point precision (in bits) used internally and for output
```

actors/builtin/reward/reward_state.go
```diff
@@ -49,31 +40,174 @@ func ConstructState(emptyMultiMapCid cid.Cid) *State {
...
+const FixedPoint = 97
+
+// Used in the definition of λ
+const BlockTimeSeconds = 30
+const SecondsInYear = 31556925
+
+// The following are the numerator and denominator of -ln(1/2), represented as
```
anorth, line 141: OMG that helps so much, thank you. Could we name it something like `NegLogHalf` too?

actors/builtin/reward/reward_state.go
```diff
@@ -49,31 +40,174 @@ func ConstructState(emptyMultiMapCid cid.Cid) *State {
...
+//
+//   Finally, at the end of each loop, we remove unnecessary bits of precision
+//   from both the numerator and denominator accumulators to control the
+//   computational complexity of the bigint multiplications.
+
+// Fixed-point precision used internally and for output
+const FixedPoint = 97
+
+// Used in the definition of λ
+const BlockTimeSeconds = 30
```
anorth, line 138: FYI this number is wrong, and has a different value elsewhere in this repo. But that's tracked as a thing to fix (with a structural fix, not just changing the literal here)

davidad: Yeah, I saw that mentioned in #317, and did some grepping around, but I was unable to find the other value.

actors/builtin/reward/reward_state.go
davidad: Sure!

davidad: Sure!

PR #339, opened 2020-05-04
miner: fix deadline calc after missed proving period

magik6k:


actors/builtin/miner/deadlines.go
```diff
@@ -57,7 +57,7 @@ func ComputeProvingPeriodDeadline(periodStart, currEpoch abi.ChainEpoch) *Deadli
 	return &DeadlineInfo{
 		CurrentEpoch: currEpoch,
 		PeriodStart:  periodStart,
-		Index:        deadlineIdx,
+		Index:        deadlineIdx % WPoStPeriodDeadlines,
```
anorth, line 60: The reasoning for this operation is a bit opaque (tho I understand it). I think clearer would be to cap `periodProgress` to `WPoStProvingPeriod - 1`, and then the rest of the arithmetic will work out.

Although if the period has been missed, I think `Index == WPoStPeriodDeadlines` is correct. I don't think this index should be inconsistent with the PeriodStart and other fields of this struct. The caller shouldn't be using if if it's expired. So now maybe I don't think we should change it, except to ensure it doesn't _exceed_ WPoStPeriodDeadlines.

anorth:
Subsumed by #349

PR #340, opened 2020-05-04
rewire porep verification to support bulk verifies

whyrusleeping:
being able to verify all the porep proofs in the same syscall will let us do some optimization that will let us get over the hump perf wise.

Currently, we spend >80% of our sync time running porep verifications. At the very least, we can run them in parallel with this change. Maybe we can even do faster vectorized things, or gpu acceleration. but any sort of speedup will require this change.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -58,6 +58,7 @@ func (a Actor) Exports() []interface{} {
 		14:                        a.AddLockedFund,
 		15:                        a.ReportConsensusFault,
 		16:                        a.WithdrawBalance,
+		17:                        a.ConfirmSectorProofsValid,
 	}
 }
 
@@ -346,7 +347,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 				rt.Abortf(exitcode.ErrIllegalArgument, "sector %v already committed with deals", params.SectorNumber)
 			} else {
 				// Committed Capacity sector upgrade.
-				if params.Expiration < sectorInfo.Info.Expiration  {
+				if params.Expiration < sectorInfo.Info.Expiration {
 					rt.Abortf(exitcode.ErrIllegalArgument, "upgraded sector %v expires before original expiration", params.SectorNumber)
 				}
 			}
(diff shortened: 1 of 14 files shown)


whyrusleeping:
This is probably ready for a once over, I'll integrate it into lotus and run some tests in a bit.

actors/builtin/power/power_actor.go
```diff
@@ -403,6 +442,85 @@ func (a Actor) computeInitialPledge(rt Runtime, desc *SectorStorageWeightDesc) a
...
+		}
+
+		ret, code := rt.Send(
+			m,
+			builtin.MethodsMiner.ConfirmSectorProofsValid,
+			&builtin.ConfirmSectorProofsParams{successful},
+			abi.NewTokenAmount(0),
+		)
+		builtin.RequireSuccess(rt, code, "failed to confirm sector proofs valid") // should never happen...
+		_ = ret
```
magik6k, line 518: There's a fairly large amount of abort cases in that call, I'm not 100% convinced that this can't break things in some way

whyrusleeping: Yeah, i'm slightly concerned about it, I'll see if i can clean that up.

actors/builtin/miner/miner_actor.go
```diff
@@ -444,91 +445,121 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
+			// Verify locked funds are are at least the sum of sector initial pledges.
+			verifyPledgeMeetsInitialRequirements(rt, &st)
 
-		if err := st.PutSector(store, newSectorInfo); err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to prove commit: %v", err)
-		}
+			// Lock up initial pledge for new sector.
+			availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
+			if availableBalance.LessThan(initialPledge) {
+				rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds for initial pledge requirement %s, available: %s", initialPledge, availableBalance)
```
anorth, line 527: This can definitely fail, unless we ensure that the precommit deposit to be unlocked covers the initial pledge. As it happens, I think that's exactly what research/cryptoecon will propose for #424. It's complicated by the fact that both those numbers change over time, so the deposit/pledge posted at precommit may no longer cover the pledge when the proof arrives. 

Maybe we can simplify to lock in the calculation at pre-commit. If so, in combination with moving `verifyPledgeMeetsInitialRequirements`, we can probably make this all safe.

actors/builtin/power/power_actor.go
anorth: Even if we're pretty sure `ConfirmSectorProofsValid` won't abort, I think we should be defensive here and let one miner fail without aborting this whole thing, which will bork all commitments forever since state will roll back.

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/libp2p/go-libp2p-core v0.3.0
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
 	github.com/multiformats/go-multihash v0.0.13
+	github.com/perlin-network/life v0.0.0-20191203030451-05c0e0f7eaea
```
anorth, line 19: Where's this coming from?

PR #343, opened 2020-05-05
DeadlineInfo handles expired proving period

anorth:
This provides a `PeriodElapsed()` predicate and fixes the other ones to indicate that no challenge window is open after a proving period has elapsed and before the proving period start has been updated to the next period.

Also changes assertions from using a too-high deadline index into errors.

### DIFF
--- a/actors/builtin/miner/deadlines.go
+++ b/actors/builtin/miner/deadlines.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/filecoin-project/specs-actors/actors/abi"
-	. "github.com/filecoin-project/specs-actors/actors/util"
 )
 
 // Deadline calculations with respect to a current epoch.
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/deadlines.go
```diff
@@ -14,40 +13,66 @@ import (
...
 	FaultCutoff  abi.ChainEpoch // First epoch at which a fault declaration is rejected (< Open).
 }
 
+// Whether the proving period has begun.
 func (d *DeadlineInfo) PeriodStarted() bool {
 	return d.CurrentEpoch >= d.PeriodStart
 }
 
+// Whether the proving period has elapsed.
+func (d *DeadlineInfo) PeriodElapsed() bool {
```
magik6k, line 29: If this returns true from `ComputeProvingPeriodDeadline(minerStateFor(ts).periodStart, ts.Height())` in PoSt scheduling logic, am I supposed to abort, should the code just abort, or is there anything smarter it can do?

anorth: I think aborting is pretty much it. You probably just have to wait for cron, and if the cron doesn't happen any proof will be rejected anyway.

Possibly you could compute the next period start that you know cron will set it to in order to eagerly begin, but I don't think that complexity is warranted at the moment.

anorth:
FYI @acruikshank we probably need a similar check in go-filecoin.

PR #345, opened 2020-05-05
Fix epoch reward computation math

whyrusleeping:
I'm not 100% sure on this, but it seems like 'newSimpleSupply' will always be smaller than the old supply, right?

### DIFF
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -107,8 +107,8 @@ func (a Actor) LastPerEpochReward(rt vmr.Runtime, _ *adt.EmptyValue) *abi.TokenA
 
 func (a Actor) computePerEpochReward(st *State, clockTime abi.ChainEpoch, networkTime abi.ChainEpoch, ticketCount int64) abi.TokenAmount {
 	// TODO: PARAM_FINISH
-	newSimpleSupply := big.Rsh(big.Mul(SimpleTotal, taylorSeriesExpansion(clockTime)), FixedPoint)
-	newBaselineSupply := big.Rsh(big.Mul(BaselineTotal, taylorSeriesExpansion(networkTime)), FixedPoint)
+	newSimpleSupply := big.Rsh(big.Mul(SimpleTotal, taylorSeriesExpansion(clockTime)), FixedPoint-90)
+	newBaselineSupply := big.Rsh(big.Mul(BaselineTotal, taylorSeriesExpansion(networkTime)), FixedPoint-90)
 
 	newSimpleMinted := big.Max(big.Sub(newSimpleSupply, st.SimpleSupply), big.Zero())
 	newBaselineMinted := big.Max(big.Sub(newBaselineSupply, st.BaselineSupply), big.Zero())


actors/builtin/reward/reward_actor.go
```diff
@@ -110,8 +110,8 @@ func (a Actor) computePerEpochReward(st *State, clockTime abi.ChainEpoch, networ
 	newSimpleSupply := big.Rsh(big.Mul(SimpleTotal, taylorSeriesExpansion(clockTime)), FixedPoint)
 	newBaselineSupply := big.Rsh(big.Mul(BaselineTotal, taylorSeriesExpansion(networkTime)), FixedPoint)
 
-	newSimpleMinted := big.Max(big.Sub(newSimpleSupply, st.SimpleSupply), big.Zero())
-	newBaselineMinted := big.Max(big.Sub(newBaselineSupply, st.BaselineSupply), big.Zero())
+	newSimpleMinted := big.Max(big.Sub(st.SimpleSupply, newSimpleSupply), big.Zero())
+	newBaselineMinted := big.Max(big.Sub(st.BaselineSupply, newBaselineSupply), big.Zero())
```
anorth, line 114: This doesn't look right to me. If `newSimpleMinted` is positive, it would imply that at 116 St.SimpleSupply will be set to a smaller number than it used to be, meaning supply is contracting.

In interpret `st.SimpleSupply` as the total amount has has been minted so far, so should be increasing.

whyrusleeping: I think that makes sense to me...

Though I think I've found the problem now

zixuanzh: newSimpleSupply should be greater than than st.SimpleSupply

actors/builtin/reward/reward_actor.go
```diff
@@ -107,8 +107,8 @@ func (a Actor) LastPerEpochReward(rt vmr.Runtime, _ *adt.EmptyValue) *abi.TokenA
 
 func (a Actor) computePerEpochReward(st *State, clockTime abi.ChainEpoch, networkTime abi.ChainEpoch, ticketCount int64) abi.TokenAmount {
 	// TODO: PARAM_FINISH
-	newSimpleSupply := big.Rsh(big.Mul(SimpleTotal, taylorSeriesExpansion(clockTime)), FixedPoint)
-	newBaselineSupply := big.Rsh(big.Mul(BaselineTotal, taylorSeriesExpansion(networkTime)), FixedPoint)
+	newSimpleSupply := big.Rsh(big.Mul(SimpleTotal, taylorSeriesExpansion(clockTime)), FixedPoint-90)
```
whyrusleeping, line 110: If you look at the state_test.go file, this is what its doing to get the expected values out.

anorth: @davidad @zx that 90 looks pretty magic. Can we encapsulate whatever the Rsh is doing in a function?

davidad: This is not how the function is meant to be used. The reason the test is doing this is to examine the 90 most significant bits of the output. Another way of phrasing it is that the test is like
```go
big.Rsh(big.Mul(2^90, taylorSeriesExpansion(x)),FixedPoint)
```
where 2^90 is taking the place of SimpleTotal. The reason it's 90 is that the dynamic range from attoFIL to a billion FIL is about 2^90, so the test is checking that it's exact to the appropriate precision (while leaving some slack for the implementation is changed, including the `FixedPoint` parameter).

davidad: I have cleaned up and documented the test here: https://github.com/filecoin-project/specs-actors/pull/348/files

I don't know why or how newSimpleSupply is less than st.SimpleSupply, assuming that st.SimpleSupply is indeed the newSimpleSupply from the previous epoch. The minting function is monotonic. In any case, I am pretty confident this isn't the fix.

PR #348, opened 2020-05-06
document reward minting function tests

davidad:
To prevent [this confusion](https://github.com/filecoin-project/specs-actors/pull/345#discussion_r420477068) from coming up in the future, I'm adding documentation to the test vectors for the reward function.

I'm also flagging that the **test vectors depend on SecondsPerEpoch=30**, which I understand from @anorth is now incorrect. When this is being fixed, please feel free to ping me to generate a new set of ground-truth `mintingTestVectors`.


davidad:
> Could you please generate them now for 25-second epochs so I have them on hand when changing this?

I will resolve this in a new PR by pulling out LambdaNum and LambdaDen as arguments that can be supplied separately for the test, alongside other changes to the minting function.

PR #350, opened 2020-05-06
Fix failure to detect faults of exactly-full top partition

anorth:
I discovered this while looking for the "failed to load fault sectors" error you observed. I don't think this fixes that problem, though.


actors/builtin/miner/miner_actor.go
```diff
@@ -1073,25 +1073,22 @@ func computeFaultsFromMissingPoSts(st *State, deadlines *Deadlines, beforeDeadli
...
 
-		for i := uint64(0); i < partitionCount; i++ {
-			if !submissions[deadlineFirstPartition+i] {
+		for dlPartIdx := uint64(0); dlPartIdx < dlPartCount; dlPartIdx++ {
+			if !submissions[deadlineFirstPartition+dlPartIdx] {
 				// No PoSt received in prior period.
-				firstSector := i * partitionSize
-				sectorCount := partitionSize
-				if i == partitionCount-1 {
-					sectorCount = dlSectorCount % partitionSize
```
anorth, line 1088: This was wrong if the final partition was exactly full.

PR #352, opened 2020-05-06
Temporary param update

zixuanzh:
Updating parameters for unincentivized testnet, we expect further updates before testnet incentives and mainnet launch and even after mainnet.

### DIFF
--- a/actors/abi/epoch.go
+++ b/actors/abi/epoch.go
@@ -5,3 +5,4 @@ package abi
 const EpochDurationSeconds = 25
 const SecondsInYear = 31556925
 const SecondsInDay = 86400
+const SecondsInHour = 3600
--- a/actors/abi/epoch.go
+++ b/actors/abi/epoch.go
@@ -1,0 +1,7 @@
+package abi
+
+// The duration of a chain epoch.
+// This is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
+const EpochDurationSeconds = 25
+const SecondsInYear = 31556925
+const SecondsInDay = 86400
--- a/actors/abi/epoch.go
+++ b/actors/abi/epoch.go
@@ -1,8 +1,0 @@
-package abi
-
-// The duration of a chain epoch.
-// This is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
-const EpochDurationSeconds = 25
-const SecondsInYear = 31556925
-const SecondsInDay = 86400
-const SecondsInHour = 3600
--- a/actors/builtin/market/policy.go
+++ b/actors/builtin/market/policy.go
@@ -10,7 +10,9 @@ const DealUpdatesInterval = 100
 
 // Bounds (inclusive) on deal duration
 func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.ChainEpoch) {
-	return abi.ChainEpoch(0), abi.ChainEpoch(builtin.EpochsInYear) // min 0, max 1y, PARAM_FINISH
+	// Cryptoeconomic modelling to date has used an assumption of a maximum deal duration of up to one year.
+	// It very likely can be much longer, but we're not sure yet.
+	return abi.ChainEpoch(0), abi.ChainEpoch(1 * builtin.EpochsInYear) // PARAM_FINISH
 }
 
 func dealPricePerEpochBounds(size abi.PaddedPieceSize, duration abi.ChainEpoch) (min abi.TokenAmount, max abi.TokenAmount) {
--- a/actors/builtin/market/policy.go
+++ b/actors/builtin/market/policy.go
@@ -1,13 +1,16 @@
 package market
 
-import "github.com/filecoin-project/specs-actors/actors/abi"
+import (
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+)
 
 // DealUpdatesInterval is the number of blocks between payouts for deals
 const DealUpdatesInterval = 100
 
 // Bounds (inclusive) on deal duration
 func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.ChainEpoch) {
-	return abi.ChainEpoch(0), abi.ChainEpoch(abi.SecondsInYear / abi.EpochDurationSeconds) // min 0, max 1y, PARAM_FINISH
+	return abi.ChainEpoch(0), abi.ChainEpoch(builtin.EpochsInYear) // min 0, max 1y, PARAM_FINISH
 }
 
 func dealPricePerEpochBounds(size abi.PaddedPieceSize, duration abi.ChainEpoch) (min abi.TokenAmount, max abi.TokenAmount) {
--- a/actors/builtin/market/policy.go
+++ b/actors/builtin/market/policy.go
@@ -7,7 +7,7 @@ const DealUpdatesInterval = 100
 
 // Bounds (inclusive) on deal duration
 func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.ChainEpoch) {
-	return abi.ChainEpoch(0), abi.ChainEpoch(10000) // PARAM_FINISH
+	return abi.ChainEpoch(0), abi.ChainEpoch(abi.SecondsInYear / abi.EpochDurationSeconds) // min 0, max 1y, PARAM_FINISH
 }
 
 func dealPricePerEpochBounds(size abi.PaddedPieceSize, duration abi.ChainEpoch) (min abi.TokenAmount, max abi.TokenAmount) {
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -4,14 +4,15 @@ import (
 	"fmt"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 )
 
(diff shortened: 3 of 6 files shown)


actors/abi/epoch.go
```diff
@@ -0,0 +1,8 @@
+package abi
+
+// The duration of a chain epoch.
+// This is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
+const EpochDurationSeconds = 25
```
davidad, line 5: is this the variable we should use instead of [`BlockTimeSeconds` in `reward_state.go`](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/reward/reward_state.go#L138) soon?

zixuanzh: I believe so, will let @anorth confirm

actors/abi/epoch.go
```diff
@@ -0,0 +1,8 @@
+package abi
```
anorth, line 1: Thanks for extracting these. I don't think `abi` is the right package for the constants (it's for types). Please move to `builtin/network.go`

actors/abi/epoch.go
anorth: Yes

actors/abi/epoch.go
```diff
@@ -0,0 +1,8 @@
...
+
+// The duration of a chain epoch.
+// This is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
+const EpochDurationSeconds = 25
+const SecondsInYear = 31556925
+const SecondsInDay = 86400
+const SecondsInHour = 3600
```
anorth, line 8: I think all of the uses of these are for epochs, so precompute EpochsInDay, etc, here.

actors/builtin/market/policy.go
```diff
@@ -7,7 +7,7 @@ const DealUpdatesInterval = 100
 
 // Bounds (inclusive) on deal duration
 func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.ChainEpoch) {
-	return abi.ChainEpoch(0), abi.ChainEpoch(10000) // PARAM_FINISH
+	return abi.ChainEpoch(0), abi.ChainEpoch(abi.SecondsInYear / abi.EpochDurationSeconds) // min 0, max 1y, PARAM_FINISH
```
anorth, line 10: Why 1 year? Why is there a max? Why not a millenium? If there's some concrete justification please write it here in a comment, otherwise use a "never going to happen" large number.

actors/builtin/miner/policy.go
```diff
@@ -130,10 +124,10 @@ type VestSpec struct {
...
 
 var PledgeVestingSpec = VestSpec{
-	InitialDelay: abi.ChainEpoch(SecondsInYear / EpochDurationSeconds),    // 1 year, PARAM_FINISH
-	VestPeriod:   abi.ChainEpoch(SecondsInYear / EpochDurationSeconds),    // 1 year, PARAM_FINISH
-	StepDuration: abi.ChainEpoch(7 * SecondsInDay / EpochDurationSeconds), // 1 week, PARAM_FINISH
-	Quantization: SecondsInDay / EpochDurationSeconds,                     // 1 day, PARAM_FINISH
+	InitialDelay: abi.ChainEpoch(7 * abi.SecondsInDay / abi.EpochDurationSeconds),   // 1 week for testnet, PARAM_FINISH
+	VestPeriod:   abi.ChainEpoch(7 * abi.SecondsInDay / abi.EpochDurationSeconds),   // 1 week for testnet, PARAM_FINISH
+	StepDuration: abi.ChainEpoch(12 * abi.SecondsInHour / abi.EpochDurationSeconds), // 12 hours for testnet, PARAM_FINISH
+	Quantization: abi.SecondsInHour / abi.EpochDurationSeconds,                      // 1 hour for testnet, PARAM_FINISH
```
anorth, line 130: This will mean we see 24x more expensive vesting than is representative. Set the quantization up to at least 6 hours, maybe higher. Consider setting the step duration to 1 day to leave room.

actors/builtin/reward/reward_actor.go
```diff
@@ -122,15 +122,16 @@ func (a Actor) computePerEpochReward(st *State, clockTime abi.ChainEpoch, networ
 	return perEpochReward
 }
 
+const baseline_power = 1 << 50 // 1PiB for testnet, PARAM_FINISH
```
anorth, line 125: `baselinePower`

actors/builtin/reward/reward_state.go
```diff
@@ -23,9 +23,9 @@ type State struct {
 
 type AddrKey = adt.AddrKey
 
-// These numbers are placeholders, but should be in units of attoFIL
-var SimpleTotal, _ = big.FromString("1000000000000000000000000")
-var BaselineTotal, _ = big.FromString("1000000000000000000000000")
+// These numbers are placeholders, but should be in units of attoFIL, 10^-18 FIL
+var SimpleTotal, _ = big.FromString("100000000000000000000000000") // 100M for testnet, PARAM_FINISH
```
anorth, line 27: Use `big.Mul(big.NewInt(100e6), big.NewInt(1e18))` and then you wont need any comments about what they mean.

actors/abi/epoch.go
whyrusleeping: can we make some of these vars? that way we can change them live in the implementations for testing things.

anorth: it would pain me, but in lieu of taking time to make them properly configurable, I'm on board.

actors/builtin/market/policy.go
whyrusleeping: +1

actors/builtin/miner/policy.go
```diff
@@ -130,10 +124,10 @@ type VestSpec struct {
 }
 
 var PledgeVestingSpec = VestSpec{
-	InitialDelay: abi.ChainEpoch(SecondsInYear / EpochDurationSeconds),    // 1 year, PARAM_FINISH
-	VestPeriod:   abi.ChainEpoch(SecondsInYear / EpochDurationSeconds),    // 1 year, PARAM_FINISH
-	StepDuration: abi.ChainEpoch(7 * SecondsInDay / EpochDurationSeconds), // 1 week, PARAM_FINISH
-	Quantization: SecondsInDay / EpochDurationSeconds,                     // 1 day, PARAM_FINISH
+	InitialDelay: abi.ChainEpoch(7 * abi.SecondsInDay / abi.EpochDurationSeconds),   // 1 week for testnet, PARAM_FINISH
+	VestPeriod:   abi.ChainEpoch(7 * abi.SecondsInDay / abi.EpochDurationSeconds),   // 1 week for testnet, PARAM_FINISH
+	StepDuration: abi.ChainEpoch(12 * abi.SecondsInHour / abi.EpochDurationSeconds), // 12 hours for testnet, PARAM_FINISH
```
whyrusleeping, line 129: what happens if EpochDurationSeconds doesnt evenly divide secondsInHour?

actors/builtin/market/policy.go
zixuanzh: we are expecting to upgrade PoRep fairly often anyways, a year seems fine? We dont have the feature to extend deals yet

actors/builtin/reward/reward_actor.go
zixuanzh: ack syntax dissonance

actors/builtin/miner/policy.go
zixuanzh: it will be truncated which i think it's fine, we will have to add test cases here to make sure

actors/builtin/market/policy.go
arajasek: I think it's more a question of what the number "means" -- I feel like the only good numbers for placeholders are 0 and infinity (regardless of whether our schedule means something like 1 year is "practically infinity").

actors/builtin/miner/policy.go
zixuanzh: got it, makes sense

actors/abi/epoch.go
zixuanzh: will let you guys make the change here in a different PR

zixuanzh:
@anorth I have set stepDuration to 1 day and quantization to 12 hours. I think a max deal duration of 1 year is reasonable, at least for testnet and even mainnet. Let us know what you think.

actors/builtin/market/policy.go
davidad: It’s not just a placeholder, it’s an assumption we have been making for a while in cryptoecon that we only consider sector lifetimes up to 1 year. After 1 year it may become rational to terminate (probably not, but I‘d want to think about it more if we want to increase this bound).

actors/builtin/miner/policy.go
davidad: As long as the remainder `12*SecondsInHour % EpochDurationSeconds` is less than, say, 1 percent of `12*SecondsInHour`, it’s not so bad. But it would be really nice if the epoch duration is a divisor of 3600. There are so many: 1 2 3 4 5 6 8 9 10 12 15 16 18 20 24 25 30 36 40 45 48 50 60 72 75 80 90 100...

actors/builtin/market/policy.go
anorth: Ok, I will document that assumption in the code.

Note that this deal bound does not bound sector lifetimes, though.

anorth:
I am going to make some cosmetic changes and then approve and land this.

actors/abi/epoch.go
anorth: I tried, but the change must flow on to all the other constants that are derived from this value, which need to read the live-changed value rather than the initial one set here, and thus be functions. I opted not to force this on us so close to testnet-2.

PR #354, opened 2020-05-08
First miner actor Window PoSt test

anorth:
This test demonstrates a single window post. That's not much, but it means all the helpers and mock/expectations are now in place for testing the many and varied scenarios we want to. I expect some of the in-line code to be factored out of this test for re-use from higher level test "scripts".

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -245,6 +245,8 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		// of penalties if locked pledge drops too low.
 		detectedFaultSectors, penalty = checkMissingPoStFaults(rt, &st, store, deadlines, deadline.PeriodStart, deadline.Index, currEpoch)
 
+		// TODO WPOST (follow-up): process Skipped as faults
+
 		// Work out which sectors are due in the declared partitions at this deadline.
 		partitionsSectors, err := ComputePartitionsSectors(deadlines, partitionSize, deadline.Index, params.Partitions)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to compute partitions sectors at deadline %d, partitions %s",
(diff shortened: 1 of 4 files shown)


anorth:
+2.73% absolute! Twenty more of those and we're done!

actors/builtin/miner/miner_actor.go
```diff
@@ -342,6 +318,16 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	if params.Expiration <= rt.CurrEpoch() {
 		rt.Abortf(exitcode.ErrIllegalArgument, "sector expiration %v must be after now (%v)", params.Expiration, rt.CurrEpoch())
 	}
+	if params.SealRandEpoch >= rt.CurrEpoch() {
```
whyrusleeping, line 321: @magik6k is going to want to work this into the fsm

actors/builtin/miner/miner_test.go
```diff
@@ -446,15 +529,99 @@ func (h *actorHarness) proveCommitSector(rt *mock.Runtime, precommit *miner.Sect
...
+		}
 	}
 	rt.Call(h.a.ProveCommitSector, params)
 	rt.Verify()
 }
 
-func (h *actorHarness) onProvingPeriodCron(rt *mock.Runtime, expectedEnrollment abi.ChainEpoch) {
+// Pre-commits and then proves a number of sectors.
+// The runtime epoch will be moved forward to the epoch of commitment proofs.
+func (h *actorHarness) commitAndProveSectors(rt *mock.Runtime, n int, expiration abi.ChainEpoch, pledge abi.TokenAmount) []*miner.SectorPreCommitInfo {
```
whyrusleeping, line 543: Nice.

anorth:
Ok, please land it for me when ready

actors/builtin/miner/miner_actor.go
anorth: As I understand it, `SealRandEpoch` is the epoch from which the challenge in seal phase 1 is sampled, producing the CommR which is also presented in this params. So it would make no sense for the challenge to be sampled after the precommit (which is the current epoch).  

The randomness isn't actually _checked_ until ProveCommit, though. At ProveCommit we check that it is no older than some window, but if the miner could arbitraily forward-date this that check wouldn't be doing anything.

But if I have that wrong, would love to be explained otherwise (and you can just drop this precondition).

actors/builtin/miner/miner_actor.go
```diff
@@ -253,34 +255,8 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
-		}
-
-		// Select a non-faulty sector as a substitute for faulty ones.
-		goodSectorNo, err := nonFaults.First()
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get first good sector")
-
-		// Load sector infos for proof
-		sectorInfos, err := st.LoadSectorInfosWithFaultMask(store, provenSectors, expectedFaults, abi.SectorNumber(goodSectorNo))
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sector infos")
+		sectorInfos, declaredRecoveries, err := st.LoadSectorInfosForProof(store, provenSectors)
```
anorth, line 258: This is a straight-up move of logic out of the middle of this  method to a place where things that are submitting window posts can reach it. @acruikshank and @magik6k should be able to delete some copy-pasta from the window post scheduling and just call this method, now.

actors/builtin/miner/miner_actor.go
magik6k: This should be fine

PR #360, opened 2020-05-10
Fix nil verified deal weight

### LINKED ISSUE #359: Nil SectorStorageWeightDesc.VerifiedDealWeight causes panic
https://github.com/filecoin-project/go-filecoin/issues/4094 reports a panic in state evaluation which comes down to nil `VerifiedDealWeight` from `AsStorageWeightDesc`.

anorth:
This was found in production.

I haven't gone to effort to cement this with regression tests because this whole SectoStorageWeightDesc and back-and-forth with the power actor is technical debt holdover from when the power actor owned these calculations, and I intend to remove most of it.

Closes #359, https://github.com/filecoin-project/go-filecoin/issues/4094


arajasek:
Do we know why this caused a panic in gfc, but not Lotus?

PR #361, opened 2020-05-11
fix calculate number of single wdpost proofs

cerasusland:


actors/builtin/miner/miner_actor.go
```diff
@@ -195,8 +195,9 @@ type SubmitWindowedPoStParams struct {
 
 // Invoked by miner's worker address to submit their fallback post
 func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) *adt.EmptyValue {
+	proofNum := len(params.Proofs[0].ProofBytes) / WdPoStSingleProofLen
 	if len(params.Partitions) != len(params.Proofs) {
-		rt.Abortf(exitcode.ErrIllegalArgument, "proof count %d must match partition count %", len(params.Proofs), len(params.Partitions))
+		rt.Abortf(exitcode.ErrIllegalArgument, "proof count %d must match partition count %d", proofNum, len(params.Partitions))
```
anorth, line 200: Thanks for catching the missing `%d`.

The message no longer matches the condition that's tested immediately above, and that condition is the right one to test. Could you please restore `len(params.Proofs)` here?

Are you suggesting that you're observing that `SubmitWindowedPoSt` is being called with a single entry in `params.Proofs` of `2*192` bytes, rather than two entries of `192` bytes each? That signals an error in the proof submission that we should track down to that source. cc @laser @magik6k 

The actor code in general does not want to lock in the proof byte-length here, even though it's unlikely to change. If future proofs have a different length, nothing should break here so long as the generation and verification of proofs have the same expectations.

cerasusland: agree with you, but that will be abort verify here when we have >=2 partitions in one deadline(half an hour), especially with  512MiB sector size. for 32GiB, when a miner's sealed sectors number reached at 48*2350+1, it will aborting.

anorth: I don't think we understand each other yet.

With >= 2 partitions, the submission should have two partition indexes and two proofs.
```
{
	Deadline: 0,
    Partitions: [0, 1],
    Proofs: [
        {RegisteredProof: 2, ProofBytes: "abcd"},
        {RegisteredProof: 2, ProofBytes: "efgh"},
    ],
}
```

cerasusland: It should be what you discribed.
but actual is 

{
	Deadline: 0,
    Partitions: [0, 1],
    Proofs: [
        {RegisteredProof: 2, ProofBytes: "abcdefgh"}
    ],
}

am i wrong?

cerasusland: image

giving the error log pic

magik6k: That's likely a bug in rust proofs, or at least in how we are calling it

```
GenerateWindowPoSt(ctx context.Context, minerID abi.ActorID, sectorInfo []abi.SectorInfo, randomness abi.PoStRandomness) ([]abi.PoStProof, error)
```
This claims to be returning an array of proofs, the `sectorInfo` we call it with matches `deadlines.Due[di.Index]`, but apparently it's just returning one concatenated proof

I could work around this by calling PoSt functions separately for each partition, but that feels like a bug in proofs lib / ffi

cc @dignifiedquire @laser

dignifiedquire: The fix in this PR is correct, according to the assumptions of the proofs libararies/interfaces. 

There is a slice returned in order to support multiple different versions of PoSt. With an imaginary `V2` sector type a return could be 

```
{
    Proofs: [
        {RegisteredProof: 2, ProofBytes: "foobar"},
        {RegisteredProof: 13, ProofBytes: "baz"},
    ],
}
```

But for now as there is only one available proof version, the slice will always have length `1`, which should likely be checked in addition.

anorth:
Oooh ok. It looks like I misunderstood what the `WindowPoStVerifyInfo.Proofs`  was supposed to mean.

In that case, I don't think the actor code should have any check that the proof matches the declared number of partitions. We should just pass it through to the proofs library to verify. Coding the expected proof length here is coupling to something the actor code has no control over. We get no value from checking it, but it's more brittle should it ever change.

I propose simply removing the check. We should also update the documentation of the `SubmitWindowPoStParams`, replacing "Parallel array of proofs corresponding to the partitions" with "Array of proofs, one per distinct registered proof type present in the sectors being proven. In the usual case of a single proof type, this array will always have a single element".

PR #362, opened 2020-05-11
Fix/wdpost proof len

magik6k:
Based on changes from https://github.com/filecoin-project/specs-actors/pull/361, after getting input from @dignifiedquire, this seems to be the correct way to count proven partitions


anorth:
https://github.com/filecoin-project/specs-actors/pull/361#issuecomment-627049527

anorth:
I merged #361 after it was changed to remove this check.

PR #363, opened 2020-05-11
miner: Restrict supported proof types in miner ctor

magik6k:


actors/builtin/miner/policy.go
```diff
@@ -49,6 +49,12 @@ const NewSectorsPerPeriodMax = 128 << 10
 // An approximation to chain state finality (should include message propagation time as well).
 const ChainFinalityish = abi.ChainEpoch(500) // PARAM_FINISH
 
+// List of proof types which can be used when creating new miner actors
+var SupportedProofTypes = map[abi.RegisteredProof]struct{}{
+	abi.RegisteredProof_StackedDRG32GiBSeal: {},
+	abi.RegisteredProof_StackedDRG64GiBSeal: {},
+}
```
anorth, line 56: FYI @acruikshank, @ZenGround0 when we integrate this we'll need to patch this value for tests, devnets using smaller sectors.

PR #366, opened 2020-05-12
update block reward target from KPI event, use that value to pay block rewards

whyrusleeping:


actors/builtin/reward/reward_actor.go
```diff
@@ -142,15 +139,20 @@ func (a Actor) UpdateNetworkKPI(rt vmr.Runtime, currRealizedPower *abi.StoragePo
...
-		newBaselinePower := a.newBaselinePower(&st)
-		st.BaselinePower = newBaselinePower
-		st.CumsumBaseline = big.Add(st.CumsumBaseline, st.BaselinePower)
+		for i := st.LastKPIUpdate; i < rt.CurrEpoch(); i++ {
+			newBaselinePower := a.newBaselinePower(&st)
+			st.BaselinePower = newBaselinePower
+			st.CumsumBaseline = big.Add(st.CumsumBaseline, st.BaselinePower)
+
+			st.RealizedPower = *currRealizedPower
+			st.CumsumRealized = big.Add(st.CumsumRealized, *currRealizedPower)
```
zixuanzh, line 148: We dont want to add `currRealizedPower` to `CumsumRealized` unless `i == rt.CurrEpoch` or `i == rt.CurrEpoch - 1`. In other words, for a null round, we are only adding the previous `st.RealizedPower` to the cumsum as no new power was proven.

actors/builtin/reward/reward_actor.go
```diff
@@ -71,21 +71,18 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
...
 	var penalty abi.TokenAmount
 	var st State
-	rewardPayable := rt.State().Transaction(&st, func() interface{} {
-		blockReward := a.computePerEpochReward(&st, rt.CurrEpoch(), st.EffectiveNetworkTime, params.TicketCount)
-		totalReward := big.Add(blockReward, params.GasReward)
-
-		// Cap the penalty at the total reward value.
-		penalty = big.Min(params.Penalty, totalReward)
+	rt.State().Readonly(&st)
+	blockReward := st.LastPerEpochReward
```
zixuanzh, line 75: We should divide this by the expected number of leaders.

actors/builtin/reward/reward_policy.go
```diff
@@ -0,0 +1,3 @@
+package reward
+
+var ExpectedLeadersPerRound = int64(5)
```
anorth, line 3: This might belong in `builtin/network.go` (but nbd if here for now)

actors/builtin/reward/reward_state.go
```diff
@@ -20,6 +20,8 @@ type State struct {
 	BaselineSupply abi.TokenAmount // current supply
 
 	LastPerEpochReward abi.TokenAmount
+
+	LastKPIUpdate abi.ChainEpoch
```
anorth, line 24: This should probably be initialized to -1 in ConstructState so that the first epoch doesn't ignore the reported power

actors/builtin/reward/reward_actor.go
```diff
@@ -142,15 +139,20 @@ func (a Actor) UpdateNetworkKPI(rt vmr.Runtime, currRealizedPower *abi.StoragePo
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		newBaselinePower := a.newBaselinePower(&st)
-		st.BaselinePower = newBaselinePower
-		st.CumsumBaseline = big.Add(st.CumsumBaseline, st.BaselinePower)
-
 		st.RealizedPower = *currRealizedPower
 		st.CumsumRealized = big.Add(st.CumsumRealized, *currRealizedPower)
```
anorth, line 143: I think this needs to be added as many times as there are iterations in the loop below.

actors/builtin/reward/reward_actor.go
```diff
@@ -142,15 +139,20 @@ func (a Actor) UpdateNetworkKPI(rt vmr.Runtime, currRealizedPower *abi.StoragePo
...
 
-		st.EffectiveNetworkTime = a.getEffectiveNetworkTime(&st, st.CumsumBaseline, st.CumsumRealized)
+		for i := st.LastKPIUpdate; i < rt.CurrEpoch(); i++ {
+			newBaselinePower := a.newBaselinePower(&st)
+			st.BaselinePower = newBaselinePower
+			st.CumsumBaseline = big.Add(st.CumsumBaseline, st.BaselinePower)
+
+			st.EffectiveNetworkTime = a.getEffectiveNetworkTime(&st, st.CumsumBaseline, st.CumsumRealized)
+
+			a.computePerEpochReward(&st, i, st.EffectiveNetworkTime, 1)
```
anorth, line 152: I would think we can do this just once at the end, though I could be missing a side-effect that's important.

actors/builtin/reward/reward_actor.go
anorth: I think @whyrusleeping misinterpreted this. We should add _something_ (`st.RealizedPower`), just not the new `currRealizedPower`

actors/builtin/reward/reward_actor.go
zixuanzh: what @anorth is saying is correct but not exactly right because of the edge cases of null rounds. If there is a null round and UpdateNetworkKPI is called in the round after that, we need to first add the RealizedPower at the null round (before the update) to CumsumRealized to account for the null round, and then update RealizedPower to currRealizedPower and then add that to CumsumRealized to account for the change in power in this round.

actors/builtin/reward/reward_actor.go
zixuanzh: Yes, doing it once at the end sgtm unless im missing something too.

actors/builtin/reward/reward_policy.go
zixuanzh: For consistency, `ExpectedLeadersPerEpoch`?

actors/builtin/reward/reward_actor.go
anorth: I was wrong, it has important side effects on the supply state

anorth: Actually I'm not so sure. Since no reward is actually minted on those null rounds, updating the supply as if there was also seems wrong.

anorth:
@whyrusleeping based on our discussion, I have pushed commits here that simplify the minting function to only count non-empty tipsets, since those are the only ones it pays out. The actor no longer references `rt.CurrentEpoch()` at all.

There are still a bunch of problems, most notably supply over-estimation, but this should pay non-zero rewards reliably (except on the first epoch 😭)

PR #376, opened 2020-05-13
handle last cron tick in market actor properly

whyrusleeping:


actors/builtin/market/set_multimap.go
```diff
@@ -101,7 +102,10 @@ func (mm *SetMultimap) PutMany(epoch abi.ChainEpoch, vs []abi.DealID) error {
 func (mm *SetMultimap) RemoveAll(key abi.ChainEpoch) error {
 	err := mm.mp.Delete(adt.UIntKey(uint64(key)))
 	if err != nil {
-		return xerrors.Errorf("failed to delete set key %v: %w", key, err)
+		if !xerrors.Is(err, hamt.ErrNotFound) {
+			return xerrors.Errorf("failed to delete set key %v: %w", key, err)
+		}
+		return nil
```
anorth, line 108: Slightly simpler as `if err != nil && !xerrors.Is(...) { `, and without line 108

PR #378, opened 2020-05-13
decouple SealVerifyInfo from OnChainSealVerifyInfo (and rename to SealVerifyParams)

### LINKED ISSUE #276: Decouple SealVerifyInfo from OnChainSealVerifyInfo
SealVerifyInfo is the structure passed to proof verification. OnChainSealVerifyInfo is the message parameter object for seal verification. The former embeds the latter, which includes redundant information like the sector number, randomness epochs.

The OnChainXxxVerifyInfo structures should be renamed as XxParams and moved to the miner actor after other things don't embed them directly.

This needs some coordinate with the FFI library which uses these types directly (but discards some unneeded info).

laser:
## Why does this PR exist?

Fixes #276.

- [x] The OnChainXxxVerifyInfo structures should be renamed as XxParams 
- [x] and moved to the miner actor after other things don't embed them directly.
- [x] This needs some coordinate with the FFI library which uses these types directly (but discards some unneeded info) - [see downstream PR](https://github.com/filecoin-project/filecoin-ffi/pull/89).

## What's in this PR?

- `abi.SealVerifyInfo` no longer embeds the `abi.OnChainSealVerifyInfo` (it now mirrors the required fields from the original struct).
- `OnChainSealVerifyInfo` renamed to `SealVerifyParams`.
- `SealVerifyParams` moved to `miner` package.


actors/abi/sector.go
```diff
@@ -209,27 +209,17 @@ type InteractiveSealRandomness Randomness
...
 type SealVerifyInfo struct {
+	RegisteredProof
 	SectorID
-	OnChain               OnChainSealVerifyInfo // TODO: don't embed this https://github.com/filecoin-project/specs-actors/issues/276
-	Randomness            SealRandomness
+
+	DealIDs               []DealID
 	InteractiveRandomness InteractiveSealRandomness
+	Proof                 []byte
+	Randomness            SealRandomness
```
anorth, line 218: Nit: move Randomness above InteractiveRandomness

actors/builtin/miner/miner_state.go
```diff
@@ -135,6 +135,20 @@ type SectorOnChainInfo struct {
 	VerifiedDealWeight abi.DealWeight // Integral of active verified deals over sector lifetime
 }
 
+// SealVerifyParams is the structure of information that must be sent with a
+// message to commit a sector. Most of this information is not needed in the
+// state tree but will be verified in sm.CommitSector. See SealCommitment for
+// data stored on the state tree for each sector.
+type SealVerifyParams struct {
```
anorth, line 142: Since this ends up not being the params to an exported actor method or stored in state, but just an internal blob of data, please:
- rename to SealVerifyStuff
- move to miner_actor.go (perhaps just above the `verifySeal` helper that uses it)
- remove cbor-gen serialization

Maybe we can refactor to make the naming less stuffy later.

laser: > Maybe we can refactor to make the naming less stuffy later.

Sounds good.

actors/abi/sector.go
laser: Done.

laser:
@anorth 

> Please don't land this, but wait for me to do so after we have a release branch for testnet-2

Sounds good. Please do let me know when it's safe to merge (or just merge it yourself) and I'll handle the filecoin-ffi side of things.

PR #379, opened 2020-05-14
call validate caller

whyrusleeping:


actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -90,6 +90,7 @@ func (a Actor) AddVerifiedClient(rt vmr.Runtime, params *AddVerifiedClientParams
 	if params.Allowance.LessThanEqual(MinVerifiedDealSize) {
 		rt.Abortf(exitcode.ErrIllegalArgument, "Allowance %d below MinVerifiedDealSize for add verified client %v", params.Allowance, params.Address)
 	}
+	rt.ValidateImmediateCallerAcceptAny()
```
zixuanzh, line 93: this looks like a temporary fix for launch but caller verification should be moved up from L99-104, thanks!

whyrusleeping: @zixuanzh unfortunately that doesnt count as 'caller validation' for the purposes of the runtime. Its some weird restriction the spec team wants here.

PR #380, opened 2020-05-18
README extracted from go-filecoin documentation

schomatis:
Taken from https://github.com/filecoin-project/go-filecoin/blob/master/CODEWALK.md#actors, this is not a proper README and some of the contents might be outdated but it's better than nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,38 @@
+# Actors
+
+Actors are Filecoin’s notion of smart contracts. 
+They are not true smart contracts—with bytecode running on a VM—but instead implemented in Go. 
+It is expected that other implementations will match the behaviour of the Go actors exactly. 
+An ABI describes how inputs and outputs to the VM are encoded. 
+Future work will replace this implementation with a "real" VM.
+
+The [Actor](https://github.com/filecoin-project/lotus/blob/master/chain/types/actor.go) struct is the base implementation of actors, with fields common to all of them.
+
+- `Code` is a CID identifying the actor code, but since these actors are implemented in Go, is actually some fixed bytes acting as an identifier. 
+This identifier selects the kind of actor implementation when a message is sent to its address.
+- `Head` is the CID  of this actor instance’s state.
+- `Nonce` is a counter of #messages received *from* that actor. 
+It is only set for account actors (the actors from which external parties send messages); messages between actors in the VM don’t use the nonce.
+- `Balance` is FIL balance the actor controls. 
+
+Some actors are singletons (e.g. the storage market) but others have multiple instances (e.g. storage miners). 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,38 @@
+# Actors
+
+Actors are Filecoin’s notion of smart contracts. 
+They are not true smart contracts—with bytecode running on a VM—but instead implemented in Go. 
+It is expected that other implementations will match the behaviour of the Go actors exactly. 
+An ABI describes how inputs and outputs to the VM are encoded. 
+Future work will replace this implementation with a "real" VM.
+
+The [Actor](https://github.com/filecoin-project/lotus/blob/master/chain/types/actor.go) struct is the base implementation of actors, with fields common to all of them.
```
arajasek, line 9: IMO, it's a little weird for specs-actors to refer back to an implementation. Maybe rephrase this to say something like "see, for example, <Lotus actor link> and <go-filecoin actor link>"?

README.md
```diff
@@ -0,0 +1,38 @@
...
+
+The [account](./actors/builtin/account/account_actor.go) actor doesn’t have any special behavior or state other than a balance. 
+Everyone who wants to send messages (transactions) has an account actor, and it is from this actor’s address that they send messages.
+
+Every storage miner has an instance of a [miner](./actors/builtin/miner/miner_actor.go) actor. 
+The miner actor plays a role in the storage protocol, for example it pledges space and collateral for storage, posts proofs of storage, etc. 
+A miner actor’s state is located in the state tree at its address; the value found there is an Actor structure. 
+The head CID in the actor structure points to that miner’s state instance (encoded).
+
+Other built-in actors include the [payment broker](./actors/builtin/paych/paych_actor.go),
```
arajasek, line 31: I'd take this paragraph out, I'm not sure that it adds much value.

README.md
anorth: Yes, remove this. The linked concept does not exist in this repo, but is abstracted. Ditto for the four bullet points below. In this repo those are better thought of as abstractions provided by the runtime, but it's not nearly a complete list of those abstractions so better to omit it. Linking to the [runtime.go](https://github.com/filecoin-project/specs-actors/blob/master/actors/runtime/runtime.go) file would be better - it has decent doc-comments.

README.md
anorth: Yes, I would delete this and the two preceding paragraphs. The information is lacking context and will likely rot.

ZenGround0:
Hey @schomatis are you still working on this or would you like to hand it off?

schomatis:
@ZenGround0 Happy to hand it off, thanks.

PR #384, opened 2020-05-19
add Unencodable Return method to puppet actor

frrist:


actors/puppet/puppet.go
```diff
@@ -25,6 +25,7 @@ func (a Actor) Exports() []interface{} {
 		builtin.MethodConstructor: a.Constructor,
 		2:                         a.Send,
 		3:                         a.SendMarshalCBORFailure,
+		4:                         a.UnencodableReturn,
```
anorth, line 28: Nit: consistency of these names would be good. Consider what other methods will be coming in this group (e.g. store get/put) and then let's rename so they sound related.

PR #386, opened 2020-05-20
add BuiltinActors method

frrist:
Returns list of all builtin actors with exported functions and code cids


actors/actors.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/filecoin-project/specs-actors/actors/builtin/multisig"
+	"github.com/filecoin-project/specs-actors/actors/builtin/paych"
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/actors/builtin/reward"
+	"github.com/filecoin-project/specs-actors/actors/builtin/system"
+	"github.com/filecoin-project/specs-actors/actors/builtin/verifreg"
+)
+
+type BuiltinActor struct {
+	exports []interface{}
```
anorth, line 21: Rather than call Exports() for this list, reference a `Invokee` reference (which is an `x.Actor{}` object), which will be more flexible if we ever need to extend that.

And in fact BuiltinActor could then implement `Invokee`.

actors/actors.go
```diff
@@ -0,0 +1,82 @@
+package actors
```
anorth, line 1: I think this should be in the `builtin` package, since that's definitely what it's defining. That might prompt a rename of `BuiltinActor` to avoid stutter: maybe `Exported` or something?

anorth:
> combine all of those things into one list of builtin-actor-metadata

... in a follow-up PR. Don't derail this one, but let's unify after it lands.

actors/actors.go
frrist: Almost all actors (account actor as an easy example) import the `builtin` package (since their addresses and codeIDs are defined there) leading to a circular dependency, which is why I put it here -- although I agree this is probably not the spot either. We could break the circular decency by:
- moving actor addresses and codes to a child package of builtin.
- move this file to a child package of builtin.
- something else?

Whatever solution we decide on I will land first and rebase this onto it.

anorth: Ok. The builtin package is referenced fairly extensively in the impls, so moving the codes/addresses there would be a bit of merge pain. Which suggests option 2. 

Note we might move things again in attempting to [unify the lists](https://github.com/filecoin-project/specs-actors/pull/405#pullrequestreview-422322432).

frrist:
> Looking at this list and the various ones in codes.go, singletons.go it seems very profitable to me for us to combine all of those things into one list of builtin-actor-metadata, which will be much easier to maintain

If we want to do that I think we may need to draw more separation between singletons and the other actors like account, multisig, and payment channel since the latter 3 don't have a single address.

PR #388, opened 2020-05-21
Don't clear recovery declarations for future deadlines on PP boundary cron

### LINKED ISSUE #387: Declared recoveries are cleared at PP boundary even if recovery deadline has not happened yet
A miner must declare recoveries ahead of the challenge window for deadline in which they will then be proven (for similar reasons that faults must be pre-declared to avoid punitive penalties). At the end of the proving period, the cron handler checks partitions for the prior period and penalises missed PoSts, including those expected for recovered sectors. Missed PoSts and recoveries are then treated as new faults, in particular being subtracted from the declared recoveries set.

Once a deadline D has passed in some proving period, a fault or recovery declaration is taken as applying to the corresponding deadline in the _next_ proving period. A recovery declaration for that deadline should _not_ be cleared at the PP boundary. The same fault detection logic is in fact executed on every PoSt and fault or recovery declaration and would have the same problem. If a recovery is declared for D during D2>D (thus intending to apply to the next period), a PoSt for D3 should not "detect a new fault" for sectors due at D simply because they are declared recovered.

@magik6k suggested a solution could be to have 2 bitfields for recoveries - one for declaring recoveries, and the other for proven recoveries, only the second would be cleared by the cron tick. I don't think this quite works. Firstly, declared recoveries are already removed immediately upon the corresponding PoSt - the challenge here is distinguishing declarations that haven't been proven but should have been from those that should not yet have been. Also we need to charge a higher penalty for declaring a recovery but missing it compared with simply not declaring recovery; recovery declarations can't simply stick around if unproven.

I think the solution comes down to keeping a pointer/index into the deadline sequence and only processing faults for deadlines that have not been processed yet. Since declaring a recovery processes all prior deadlines in the PP, a recovery at deadline D2 targeting D1 will move the pointer up past D1, such that subsequent declarations and the PP cron will all ignore it. But then if the next PP elapses without the expected proof arriving, the next cron (or other fault detection trigger) will have the second iteration of D1 in scope and enact the penalty.

anorth:
Fixes the erroneous clearing of declared recoveries during the proving period cron (or other detect-faults hooks). See the linked issue for background.

I'm painfully aware that this behaviour doesn't have unit tests yet.

Fixes #387

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -215,19 +215,21 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		if uint64(len(params.Partitions)) > submissionPartitionLimit {
 			rt.Abortf(exitcode.ErrIllegalArgument, "too many partitions %d, limit %d", len(params.Partitions), submissionPartitionLimit)
 		}
-		deadline := st.DeadlineInfo(currEpoch)
-		if !deadline.PeriodStarted() {
-			rt.Abortf(exitcode.ErrIllegalArgument, "proving period %d not yet open at %d", deadline.PeriodStart, currEpoch)
+		currDeadline := st.DeadlineInfo(currEpoch)
+		deadlines, err := st.LoadDeadlines(adt.AsStore(rt))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
+
+		// Traverse earlier submissions and enact detected faults.
+		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
+		// of penalties if locked pledge drops too low.
+		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines)
+
+		if !currDeadline.PeriodStarted() {
+			rt.Abortf(exitcode.ErrIllegalArgument, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
 		}
-		if deadline.PeriodElapsed() {
-			// A cron event has not yet processed the previous proving period and established the next one.
-			// This is possible in the first non-empty epoch of a proving period if there was an empty tipset on the
-			// last epoch of the previous period.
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -215,19 +215,21 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
-		deadline := st.DeadlineInfo(currEpoch)
-		if !deadline.PeriodStarted() {
-			rt.Abortf(exitcode.ErrIllegalArgument, "proving period %d not yet open at %d", deadline.PeriodStart, currEpoch)
-		}
-		if deadline.PeriodElapsed() {
-			// A cron event has not yet processed the previous proving period and established the next one.
-			// This is possible in the first non-empty epoch of a proving period if there was an empty tipset on the
-			// last epoch of the previous period.
-			rt.Abortf(exitcode.ErrIllegalState, "proving period at %d elapsed, next one not yet opened", deadline.PeriodStart)
+		currDeadline := st.DeadlineInfo(currEpoch)
```
anorth, line 218: This delta is just re-arrangement to make this case follow the same pattern as the others. No behaviour change intended. The `deadline.PeriodElapsed` check has moved inside `detectFaultsThisPeriod`

actors/builtin/miner/miner_actor.go
```diff
@@ -1084,7 +1116,7 @@ func computeFaultsFromMissingPoSts(st *State, deadlines *Deadlines, beforeDeadli
 
 	deadlineFirstPartition := uint64(0)
 	var fGroups, rGroups []*abi.BitField
-	for dlIdx := uint64(0); dlIdx < beforeDeadline; dlIdx++ {
+	for dlIdx := sinceDeadline; dlIdx < beforeDeadline; dlIdx++ {
```
anorth, line 1119: This is the heart of the fix.

anorth: Gah, this was buggy. `deadlineFirstPartition` is initialized above assuming this loop will iterate all partitions.

Issue #389, opened 2020-05-21
Consider leaving first Window PoSt proving deadline empty


anorth:
The first deadline within a proving period is not like the others. For most, the set of sectors to be included is known well ahead of time. But because new sectors are allocated to deadlines in the proving period boundary cron, and will include some non-determinism, the proving set for the first deadline isn't known until the deadline window has started, and a window for stability of the chain/state has passed. 

This means the first period has a shorter computation window and much shorter (<0) stable lead up time in which a mining operation could prepare sectors for sampling.

We could leave it empty, assigning sectors only to the other deadlines.

waynewyang:
oh, I think this is very necessary. now we can't set the recoveries  for window 0  in window 23 period. because the state will be reset in `handleProvingPeriod()`which is in `periodStart-1`

dtynn:
Besides, i think currently we could not declare fault sectors in the first proving deadline to be recovered. 
Since:

1. we could not make `DeclareFaultsRecovered` calls after the `FaultCutoff`
  [DeclareFaultsRecovered](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L815-L816)
  [validateFRDeclaration](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1762-L1764)
  so we have to declare the recoveries earlier, that would be within the last proving period

2. `CronEventProvingPeriod` always cleans up the `Recoveries`, removes all sectors included in the missing post partitions.
  [handleProvingPeriod](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1008-L1012)
  [processMissingPoStFaults](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1131-L1132)
  [computeFaultsFromMissingPoSts](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1185-L1186)

whyrusleeping:
@anorth I think this is technically resolved by #388 right?

anorth:
The problems @waynewyang and @dtynn are describing are resolved separately, yes.

anorth:
Closed by #395

PR #395, opened 2020-05-26
Leaving first Window PoSt proving deadline empty

waynewyang:
Consider leaving first Window PoSt proving deadline empty #389


anorth:
Thanks. I have no objection but would like to sit with this and/or hear from @whyrusleeping and @magik6k whether they think this is a good idea, given that the problems described in the comments on #389 (as opposed to the OP) are resolved.

PR #398, opened 2020-05-26
CC upgrade

zixuanzh:
This PR allows for CC sectors to be upgraded to sectors with deals and extend the sector lifetime at the same time. This step is crucial in storing useful data on Filecoin as miners can first prove capacity to the network in the absence of storage demand and later upgrade their sectors when demand arrives. Currently, this will involve resealing and go through the same PublishStorageDeals, PreCommit, and ProveCommit workflow. There are plans to make it more efficient and cost effective post Mainnet launch. 

Currently, miners will have to put up new pledge collateral. We would like to reduce the pledge collateral required here but that will require us landing penalty calculation first and we can piggyback on the calculation there.


zixuanzh:
@whyrusleeping thanks, just updated PR description, merging.

PR #401, opened 2020-05-28
make batch verification set cid be a pointer

whyrusleeping:


anorth:
Why is this better than using the CID of an empty collection? On the surface, that would seem easier to work-with and less likely to cause a nil pointer panic one day.

This is inconsistent with the many other empty collections in state, so I'd like to understand the reasoning in case it should also be applied elsewhere, or a consistent approach is possible.

anorth:
I have inferred from discussion elsewhere that one distinguishing property here is that this CID is not expected to ever point to a non-empty collection in a tipset's final state, because the cron job should empty it every time. This would still be the case if it were a non-nil CID of an empty collection, though.

I'm not really convinced that justifies introducing the risk of nil pointer dereference, but at least explains why this one might be different to all the other ones.

PR #405, opened 2020-06-01
Restrict exec permissions

### LINKED ISSUE #219: Restrict init actor exec's ability to create singleton actors

ZenGround0:
User messages shouldn't be creating verifiedRegistry, system or reward actors or unknown code cids.

This PR also flips the permission strategy to fail closed.  This gives us an advantage catching bugs introduced by rot.

Close #219

### DIFF
--- a/actors/builtin/init/init_actor.go
+++ b/actors/builtin/init/init_actor.go
@@ -88,13 +88,14 @@ func (a Actor) Exec(rt runtime.Runtime, params *ExecParams) *ExecReturn {
 
 func canExec(callerCodeID cid.Cid, execCodeID cid.Cid) bool {
 	switch execCodeID {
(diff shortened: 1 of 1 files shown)


ZenGround0:
My bad did not look through PRs before filing so missed 282.

anorth:
No worries. Please do land this, but follow up.

whyrusleeping:
If this ends up not being a blacklist, it will just have to get changed to one in the future.

anorth:
@whyrusleeping you mean because of user-defined actors? I see what you're saying, but we need something better than the current state, where a builtin developer usually achieves the wrong result by default (e.g. verified registry was missed here too). So the blacklist has to be automatic from a list of the builtin actors etc (#386). 

So I agree, but I think this is still a local step forward.

ZenGround0:
@whyrusleeping merging for now.  If this starts to give you trouble with testing let me know and I'll flip to blacklist.

Issue #425, opened 2020-06-03
Miner actor MinerInfo#PeerID filed should be a byte array


frrist:
Spawned from a conversation on [slack](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1591222055359400?thread_ts=1591221924.359100&cid=CHMNDCK9P) --`MinerInfo#PeerID` should be a byte array, this removes the dependency on `github.com/libp2p/go-libp2p-core/peer`.

anorth:
My motivation for this: the actor code itself doesn't care about the internal structure of this ID. It's just providing discovery for the miner operator so that deal clients can identify the provider off-chain. Yes, we absolutely expect this to be a Peer ID but nothing about the state machine or blockchain is broken if it's not a valid ID. We have no way to validate that the ID corresponds to a real key, an offline miner, etc etc.

But we're pulling in an 8000-line repository of code into a consensus-critical area. While we don't use many (any?) methods of the Peer.ID struct (yet), if we did, any change in their behaviour could constitute a forking change. Our idea of version compatibility differs from the traditional Go versioning approach, "backwards compatible" changes are not compatible for us if they change behaviour _at all_, say to allow or correctly handle something previously disallowed or broken. Go-mod's version handling does not permit this repo to pin a specific version of the libp2p code to enforce compatibility.

When we do versioning for chain upgrades (#385) we'll want to effectively vendor all the actor code, including transitive dependencies. Every dependent repo here represents a potential world of pain.

We should store the peer ID (and addresses, #403) as plain uninterpreted byte arrays and leave the interpretation to off-chain code, which will need to handle various ways it could be invalid anyway. Drop code dependencies on libp2p. This is not a great burden to place on the users.

anorth:
(Dependencies on things like CID have similar potential pain, but are so much more fundamental that the trade-off is much better).

PR #427, opened 2020-06-03
add ChangeMultiaddrs to miner actor

### LINKED ISSUE #403: Add an "addresses" field to the MinerInfo data structure
This would reduce reliance on the DHT to find miners. Miners could include addresses like `/dnsaddr/super-big-miner.me`, or even direct IP addresses like `/ip4/a.b.c.d/tcp/1234` if their IP doesn't change frequently.

frrist:
closes #403

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
 	cid "github.com/ipfs/go-cid"
@@ -59,6 +58,7 @@ func (a Actor) Exports() []interface{} {
 		15:                        a.ReportConsensusFault,
 		16:                        a.WithdrawBalance,
 		17:                        a.ConfirmSectorProofsValid,
+		18:                        a.ChangeMultiaddrs,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


actors/builtin/shared.go
```diff
@@ -43,6 +43,10 @@ type MinerAddrs struct {
 	Worker addr.Address
 }
 
+// This type is required to construct a miner (in power_actor) and change a miners multiaddress (in miner_actor)
+// the type is defined here to work around a circular dependency between the actors.
+type MultiAddress = []byte
```
Stebalien, line 48: Why not just call it `Multiaddr`?

anorth, line 48: I think we'd be ok just using `[]byte` in place, since actor code never inspects or cares about the internals.

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/ipfs/go-log v1.0.0 // indirect
 	github.com/libp2p/go-libp2p-core v0.3.0
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
+	github.com/multiformats/go-multiaddr v0.2.0
```
anorth, line 18: See [this discussion](https://github.com/filecoin-project/specs-actors/issues/425#issuecomment-638521129) about why I really want to avoid this.

actors/builtin/miner/miner_actor.go
```diff
@@ -59,6 +59,7 @@ func (a Actor) Exports() []interface{} {
 		15:                        a.ReportConsensusFault,
 		16:                        a.WithdrawBalance,
 		17:                        a.ConfirmSectorProofsValid,
+		18:                        a.ChangeMultiaddrs,
```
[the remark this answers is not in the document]
Stebalien: How would that work in practice? Either they're always multiaddrs or we'd need something to indicate what they are.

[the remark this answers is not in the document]
anorth: Probably hackily. I'm not saying it would be good, just that the actors don't care. If for some unlikely reason something like that was wanted, no code here would change but the names would be misleading. 

Taken to the extreme, the field might as well be called "arbitrary provider metadata" and embed JSON. But "addresses" is fine and this is not a blocking issue for me.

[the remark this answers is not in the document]
frrist: I've prefixed with `Multi` to avoid confusion with methods that use `address.Address`s -- `ControlAddresses` & `ChangeWorkerAddress`. Having a method called `ChangeAddresses` seems confusing given that it operates on something completely different that `ControlAddresses`.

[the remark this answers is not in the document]
Stebalien: At that point, I'd add additional fields. Although, multiaddrs are extensible enough (literally multicodec + argument) that I'd expect us to always use mutliaddrs here (even if it means assigning new multicodecs for new multiaddrs).

actors/builtin/shared.go
frrist: I made it a separate type as a means of providing documentation, i'm a bit surprised by this comment coming from you :smile:

frrist: No good reason, will call it that instead :+1:

PR #428, opened 2020-06-04
Change WPoSt deadline duration to 40 minutes

anorth:
We believe proofs are now fast enough again that a ~half-hour~ 40 minute deadline duration is reasonable.

See #373 where this was changed to an hour.

### DIFF
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -12,7 +12,7 @@ import (
 const WPoStProvingPeriod = abi.ChainEpoch(builtin.EpochsInDay) // 24 hours
 
 // The duration of a deadline's challenge window, the period before a deadline when the challenge is available.
-const WPoStChallengeWindow = abi.ChainEpoch(3600 / builtin.EpochDurationSeconds) // An hour (=24 per day)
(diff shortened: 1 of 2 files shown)


whyrusleeping:
our miners are currently at 25 minutes to pull off a windowed post, when you mix in winning posts due to winning a block, and other block validation processing work, i'm still a bit concerned about the deadline of 30 minutes.

anorth:
Sure, that sounds tight. I've changed this to 40 minutes, which also is a multiple of 25s and evenly divides 24hr.

If you'd still rather not take it, no worries.

actors/builtin/miner/policy.go
```diff
@@ -12,7 +12,7 @@ import (
 const WPoStProvingPeriod = abi.ChainEpoch(builtin.EpochsInDay) // 24 hours
 
 // The duration of a deadline's challenge window, the period before a deadline when the challenge is available.
-const WPoStChallengeWindow = abi.ChainEpoch(3600 / builtin.EpochDurationSeconds) // An hour (=24 per day)
+const WPoStChallengeWindow = abi.ChainEpoch(40*60 / builtin.EpochDurationSeconds) // Half an hour (=48 per day)
```
magik6k, line 15: Isn't this actually 36 per day?

PR #433, opened 2020-06-04
Replace peerID with byte array

### LINKED ISSUE #425: Miner actor MinerInfo#PeerID filed should be a byte array

frrist:
closes #425

Also includes a fix (aa68cf7) to add the multiaddresses added in #427 to the power actors miner-constructor-params.


actors/builtin/miner/miner_actor.go
```diff
@@ -76,12 +75,6 @@ type ConstructorParams = power.MinerConstructorParams
 func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.InitActorAddr)
 
-	// Sanity check that we've been given a valid peer ID
-	_, err := peer.IDFromString(string(params.PeerId))
```
ZenGround0, line 80: Unless there is a reason to remove this a similar sanity check using `IDFromBytes` would still be helpful.

frrist: Reasoning from @anorth here: https://github.com/filecoin-project/specs-actors/issues/425#issuecomment-638521129

actors/builtin/shared.go
```diff
@@ -43,9 +43,10 @@ type MinerAddrs struct {
 	Worker addr.Address
 }
 
-// This type is required to construct a miner (in power_actor) and change a miners multiaddress (in miner_actor)
-// the type is defined here to work around a circular dependency between the actors.
+// These types are required to construct a miner (in power_actor) and change a miners multiaddress/peerId (in miner_actor)
+// these types are defined here to work around a circular dependency between the actors.
 type Multiaddrs = []byte
+type PeerID = []byte
```
anorth, line 49: Let's just move them to `abi`. You then don't need to talk about the dependency, just state what each one represents.

PR #434, opened 2020-06-04
return exitcode and retval from Approve & Propose

### LINKED ISSUE #113: Multisig tx ret code is ignored
Here https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go#L311. 

Return code and tx status are ignored. I propose add method `getTxStatus(TxId)` that returns `[Pending, Cancelled, InvocationOutput]`

frrist:
- closes #113

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -385,12 +385,12 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
 		ret, code = rt.Send(
 			txn.To,
 			txn.Method,
-			// pass the return value through uninterpreted with the expectation that serializing into a CBORBytes never fails since it just copies the bytes.
 			vmr.CBORBytes(txn.Params),
 			txn.Value,
 		)
 		applied = true
 
+		// pass the return value through uninterpreted with the expectation that serializing into a CBORBytes never fails since it just copies the bytes.
 		if err := ret.Into(&out); err != nil {
 			rt.Abortf(exitcode.ErrSerialization, "failed to deserialize result: %v", err)
 		}
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -154,9 +154,6 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	var emptyArray []byte
 	applied, ret, code := a.approveTransaction(rt, txnID, emptyArray, false)
 
-	// Note: this ID may not be stable across chain re-orgs.
-	// https://github.com/filecoin-project/specs-actors/issues/7
-
 	return &ProposeReturn{
 		TxnID:   txnID,
 		Applied: applied,
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -388,6 +388,7 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []by
 		ret, code = rt.Send(
 			txn.To,
 			txn.Method,
+			// pass the return value through uninterpreted with the expectation that serializing into a CBORBytes never fails since it just copies the bytes.
 			vmr.CBORBytes(txn.Params),
 			txn.Value,
 		)
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -118,10 +118,14 @@ type ProposeParams struct {
 }
 
 type ProposeReturn struct {
-	TxnID   cbg.CborInt
+	// TxnID is the ID of the proposed transaction
+	TxnID cbg.CborInt
+	// Applied indicates if the transaction was applied as opposed to proposed but not applied due to lack of approvals
 	Applied bool
-	Code    exitcode.ExitCode
-	Ret     vmr.CBORBytes
+	// Code is the exitcode of the transaction, if Applied is false this field should be ignored.
+	Code exitcode.ExitCode
+	// Ret is the return vale of the transaction, if Applied is false this field should be ignored.
+	Ret vmr.CBORBytes
 }
 
 func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
@@ -169,9 +173,12 @@ type TxnIDParams struct {
 }
 
 type ApproveReturn struct {
+	// Applied indicates if the transaction was applied as opposed to proposed but not applied due to lack of approvals
 	Applied bool
-	Code    exitcode.ExitCode
-	Ret     vmr.CBORBytes
+	// Code is the exitcode of the transaction, if Applied is false this field should be ignored.
+	Code exitcode.ExitCode
+	// Ret is the return vale of the transaction, if Applied is false this field should be ignored.
+	Ret vmr.CBORBytes
 }
 
 func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *ApproveReturn {
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -123,7 +123,7 @@ type ProposeReturn struct {
 	// Code is the exitcode of the transaction, if Applied is false this field should be ignored.
 	Code exitcode.ExitCode
 	// Ret is the return vale of the transaction, if Applied is false this field should be ignored.
-	Ret vmr.CBORBytes
+	Ret []byte
 }
 
 func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
@@ -173,7 +173,7 @@ type ApproveReturn struct {
 	// Code is the exitcode of the transaction, if Applied is false this field should be ignored.
 	Code exitcode.ExitCode
 	// Ret is the return vale of the transaction, if Applied is false this field should be ignored.
-	Ret vmr.CBORBytes
+	Ret []byte
 }
 
 func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *ApproveReturn {
@@ -336,7 +336,7 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
 	return nil
 }
 
-func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []byte, checkHash bool) (bool, vmr.CBORBytes, exitcode.ExitCode) {
+func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []byte, checkHash bool) (bool, []byte, exitcode.ExitCode) {
 	var st State
 	var txn Transaction
 	rt.State().Transaction(&st, func() interface{} {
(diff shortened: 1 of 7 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -110,7 +110,14 @@ type ProposeParams struct {
 	Params []byte
 }
 
-func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
+type ProposeReturn struct {
+	TxnID   cbg.CborInt
+	Applied bool
```
frrist, line 115: When a transaction is proposed but not executed/applied we need a way to indicate that to the caller. One way the caller can be made aware of this is by using a bool here. This is also done for `ApproveReturn` since approving a transaction does not guarantee it is executed.

support/mock/mockrt.go
```diff
@@ -803,6 +803,12 @@ type ReturnWrapper struct {
 
 func (r ReturnWrapper) Into(o runtime.CBORUnmarshaler) error {
 	b := bytes.Buffer{}
+	if r.V == nil {
```
frrist, line 806: When applying a message with no return value (e.g. transfer of value) the return value is an empty byte array. This change allows `Into` to be called on a `ReturnWrapper` that has an empty value, previously it would panic. This is inline with the lotus implementation which handles empty returns without failing: https://github.com/filecoin-project/lotus/blob/9b867a221a2babcfc97a43b4b9f24054976ef975/chain/vm/runtime.go#L338-L344

frrist: To fix this correctly we'll have to change all the call sites of ExpectSend (an example: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_test.go#L435) since we pass `nil` for the return value. I'll make that change in a seperate PR and rebase onto it.

anorth: Rather than changing the call sites, change ExpectSend to adapt nil to `EmptyValue` (which is what the real vm's (should) do).

actors/builtin/multisig/multisig_actor.go
```diff
@@ -110,7 +110,14 @@ type ProposeParams struct {
...
+	Code    exitcode.ExitCode
+	Ret     vmr.CBORBytes
```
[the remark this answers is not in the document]
ZenGround0: To make sure everyone is on the same page -- using `CBORbytes` like this gives the serialized `ProposeReturn` a different cbor shape depending on the cbor structure of the return value of the method the transaction is applied to.  So if propose hits a method with an int return there will be a literal cbor int in the `Ret` position in the `ProposeReturn` tuple, if it hits a method with a tuple return value that tuple value will be directly embedded in the `Ret` position etc.

You could argue that cbor viewers would handle this direct embedding better than an extra byte array decode step.  Either way you are going to need to do two decode steps -- one for `ProposeReturn` one for `Ret`.  Anyone decoding is going to need a type like `CBORBytes` that reads raw -- any good library should have it but its a bit annoying.

Biggest argument for using a cbor byte slice is that it is the standard way the protocol does things afaict.  For example `ProposeParams` in this file, `UnsignedMessage.Params`.   I'm new to a lot of specs-actors so I might be missing the full story.

frrist:
@anorth I've executed this test against the code here and confirmed it passes: https://gist.github.com/frrist/204c3373391976d3bae02efcacfb47d4

[Here](https://gist.github.com/frrist/204c3373391976d3bae02efcacfb47d4#file-multisig_test-go-L77) is where the assertion relating to `CBORBytes` in the returned structure is made.

support/mock/mockrt.go
frrist: done

actors/builtin/multisig/multisig_actor.go
```diff
@@ -110,7 +110,14 @@ type ProposeParams struct {
 	Params []byte
 }
 
-func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
+type ProposeReturn struct {
+	TxnID   cbg.CborInt
```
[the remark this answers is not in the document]
frrist: done

actors/builtin/multisig/multisig_test.go
```diff
@@ -1094,14 +1113,36 @@ func (h *msActorHarness) propose(rt *mock.Runtime, to addr.Address, value abi.To
...
+	if !ok {
+		h.t.Fatalf("unexpected type returned from call to Propose")
+	}
+	if proposeReturn.Applied {
+		// get the last message sent and ensure it matches the exit code
+		msg := rt.LastMessageSent()
+		assert.Equal(h.t, proposeReturn.Code, msg.ExitCode())
+		var out runtime.CBORBytes
+		assert.NoError(h.t, msg.SendReturn().Into(&out))
+		assert.Equal(h.t, proposeReturn.Ret, out)
```
anorth, line 1127: What I really want to see is decoding into the result type for the method that was called, which depends entirely on which method that was and so can't be generically done here. You could pass a `CBORUnmarshaller` parameter to receive the return value.

ZenGround0:
Note: this also closes #471

actors/builtin/multisig/multisig_actor.go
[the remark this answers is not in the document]
ZenGround0: Really sorry for throwing us off here.  It turns out `CBORBytes` does encode things as cbor major type 2, not directly like cbg.Deferred.  The above comment applies to cbg.Deferred.  [This gist](https://gist.github.com/ZenGround0/18e8ffcab6e47af7eb6e21e76620940d) goes into a little more detail.

anorth:
Ok, I'll approve after the test demonstrates decoding into the result type appropriate for the method that was called

frrist:
> demonstrates decoding into the result type appropriate for the method that was called

done in 622b3e224495efd75d355dcb96c7a67a428066a0

actors/builtin/multisig/multisig_test.go
frrist: done in 622b3e2

actors/builtin/multisig/multisig_test.go
```diff
@@ -1095,21 +1123,41 @@ func (h *msActorHarness) constructAndVerify(rt *mock.Runtime, numApprovalsThresh
...
+	ret := rt.Call(h.a.Propose, proposeParams)
+	proposeReturn, ok := ret.(*multisig.ProposeReturn)
+	if !ok {
+		h.t.Fatalf("unexpected type returned from call to Propose")
+	}
+	if proposeReturn.Applied {
+		// get the last message sent and ensure it matches the exit code
+		msg := rt.LastMessageSent()
+		assert.Equal(h.t, proposeReturn.Code, msg.ExitCode())
+		assert.NoError(h.t, msg.SendReturn().Into(out))
```
anorth, line 1142: This isn't quite right. We want to deserialize the proposeReturn into the output.

This observation was triggered by the addition of LastMessageSent() and exposing the ExpectedSend. I think we should be able to avoid that altogether.

actors/builtin/multisig/multisig_test.go
```diff
@@ -1095,21 +1123,41 @@ func (h *msActorHarness) constructAndVerify(rt *mock.Runtime, numApprovalsThresh
...
+	if !ok {
+		h.t.Fatalf("unexpected type returned from call to Approve")
+	}
+	if approveReturn.Applied {
+		// get the last message sent and ensure it matches the exit code
+		msg := rt.LastMessageSent()
+		assert.Equal(h.t, approveReturn.Code, msg.ExitCode())
+		var out runtime.CBORBytes
+		assert.NoError(h.t, msg.SendReturn().Into(&out))
+		assert.Equal(h.t, approveReturn.Ret, out)
```
anorth, line 1159: Ditto

support/mock/mockrt.go
```diff
@@ -71,14 +73,14 @@ type expectRandomness struct {
 	out abi.Randomness
 }
 
-type expectedMessage struct {
-	// expectedMessage values
+type ExpectedMessage struct {
```
anorth, line 76: Blocking: do not export this. I think we'll be able to avoid the LastSentMessage altogether, but if we do need something like that, do not return an objet of this type: return just the bits that are needed (which I think just the return value).

actors/builtin/multisig/multisig_actor.go
```diff
@@ -117,7 +115,18 @@ type ProposeParams struct {
...
-func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *cbg.CborInt {
+type ProposeReturn struct {
+	// TxnID is the ID of the proposed transaction
+	TxnID TxnID
+	// Applied indicates if the transaction was applied as opposed to proposed but not applied due to lack of approvals
+	Applied bool
+	// Code is the exitcode of the transaction, if Applied is false this field should be ignored.
+	Code exitcode.ExitCode
+	// Ret is the return vale of the transaction, if Applied is false this field should be ignored.
+	Ret vmr.CBORBytes
```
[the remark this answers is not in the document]
frrist: Correct, they are interchangeable as far as cbor generation goes. Here is a diff after running gen with `Ret` as a `[]byte` if you are curious:
<details>
  <summary>Click to expand</summary>

```diff
frrist@sequoia:~/s/g/f/specs-actors|fix/113✓
➤ go run gen/gen.go 
frrist@sequoia:~/s/g/f/specs-actors|fix/113⚡*
➤ git diff
diff --git a/actors/builtin/multisig/cbor_gen.go b/actors/builtin/multisig/cbor_gen.go
index 3064421..d3279c6 100644
--- a/actors/builtin/multisig/cbor_gen.go
+++ b/actors/builtin/multisig/cbor_gen.go
@@ -1234,7 +1234,7 @@ func (t *ProposeReturn) MarshalCBOR(w io.Writer) error {
                }
        }
 
-       // t.Ret (runtime.CBORBytes) (slice)
+       // t.Ret ([]uint8) (slice)
        if len(t.Ret) > cbg.ByteArrayMaxLen {
                return xerrors.Errorf("Byte array in field t.Ret was too long")
        }
@@ -1330,7 +1330,7 @@ func (t *ProposeReturn) UnmarshalCBOR(r io.Reader) error {
 
                t.Code = exitcode.ExitCode(extraI)
        }
-       // t.Ret (runtime.CBORBytes) (slice)
+       // t.Ret ([]uint8) (slice)
 
        maj, extra, err = cbg.CborReadHeader(br)
        if err != nil {
diff --git a/actors/builtin/multisig/multisig_actor.go b/actors/builtin/multisig/multisig_actor.go
index d2dd1eb..1e51af1 100644
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -123,7 +123,7 @@ type ProposeReturn struct {
        // Code is the exitcode of the transaction, if Applied is false this field should be ignored.
        Code exitcode.ExitCode
        // Ret is the return vale of the transaction, if Applied is false this field should be ignored.
-       Ret vmr.CBORBytes
+       Ret []byte
 }
```
</details>

actors/builtin/multisig/multisig_test.go
frrist: `LastMessageSent` was added to address the (now removed) TODO:
```
// TODO In a follow-up, this method should also verify the return value from Approve contains the exit code prescribed in ExpectSend.
// exercise both un/successful sends.
```

anorth: It could, but probably better would be to just return it from `rt.Call`, like the return value.

anorth: Wait, no. I mean it should be parallel with the return value, so proposeReturn.Code returned from this `propose` method. The caller (who set up the ExpectSend, if any) will know the expected value and can assert it if relevant, along with checking for the expected `out`.

frrist: How would that work? The mockrt's Call method doesn't catch aborts -- I think we'd need a new method for that.

PR #435, opened 2020-06-04
Move sector expiration in `ExtendSectorExpiration`.

### LINKED ISSUE #429: Fix ExtendSectorExpiration to actually update the expiration queue
Sectors are expired according to registration in the `SectorExpirations` queue. The `ExtendSectorExpiration` method doesn't change this, so the sector will expire at its original expiration anyway.

The expiration is redundantly represented in two places, leading to this error. Consider removing the expiration from the sector info, although this might make inspection of sector committed lifetimes a bit more difficult.

See also #394 and #430

acruikshank:
closes #429

### Motivation

The sector expiration is recorded in the sector itself and as an entry in a bitfield at the expiration epoch. `ExtendSectorExpiration` updates the former but not the latter, meaning the sector will not actually be expired at the appropriate time.

### Proposed Changes

1. Remove the sector expiry at the original expiration.
2. Add a new sector expiry at the new expiration.
3. Add tests for `ExtendSectorExpiration`.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -418,16 +418,9 @@ func TestExtendSectorExpiration(t *testing.T) {
 			NewExpiration: newExpiration,
 		}
 
-		func() {
-			defer func() {
-				r := recover()
-				require.NotNil(t, r)
-				assert.Contains(t, fmt.Sprintf("%s",r), "cannot reduce sector expiration")
-			}()
-
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
 			actor.extendSector(rt, sector, 0, params)
-		}()
-
+		})
 	})
 
 	t.Run("updates expiration with valid params", func(t *testing.T) {
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -385,6 +385,97 @@ func TestProvingPeriodCron(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -390,6 +390,97 @@ func TestProvingPeriodCron(t *testing.T) {
...
+
+		func() {
+			defer func() {
+				r := recover()
+				require.NotNil(t, r)
+				assert.Contains(t, fmt.Sprintf("%s",r), "cannot reduce sector expiration")
+			}()
+
+			actor.extendSector(rt, sector, 0, params)
+		}()
```
anorth, line 434: Use `rt.ExpectAbort` - you can find example uses in other tests.

actors/builtin/miner/miner_test.go
```diff
@@ -390,6 +390,97 @@ func TestProvingPeriodCron(t *testing.T) {
...
+
+	})
+
+	t.Run("updates expiration with valid params", func(t *testing.T) {
```
anorth, line 438: Since the set-up for these tests is lengthy, I think it's prudent to combine multiple variations into a single run, where the implicit state dependencies are minimal. That also makes the essential difference between the cases easier to spot. Since aborted calls don't change state at all, you can combine as many of them as you want before a state-changing call.

In this case I suggest combining with the case above. See TestCommitments as example.

acruikshank: I have a PR stacked on this one. I'll do this refactor in the followup.

PR #436, opened 2020-06-04
Bug/sector expiry checks new expiry is valid

### LINKED ISSUE #394: ExtendSectorExpiration func don't check sector.Info.Expiration
in PreCommitSector, I saw that 
``` go
// Check expiry is exactly *the epoch before* the start of a proving period.
periodOffset := st.ProvingPeriodStart % WPoStProvingPeriod
expiryOffset := (params.Expiration + 1) % WPoStProvingPeriod
```
but there is no check in ExtendSectorExpiration, maybe that's  forgotten

acruikshank:
closes #394

### Motivation

Expiration for a sector must be set to the epoch before the proving period start. This is validated in `PreCommitSector`, but not in `ExtendSectorExpiration`.

### Proposed Changes

1. Execute the same validation in `ExtendSectorExpiration` by extracting it as a function in `miner.Actor`.
2. Test.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -361,13 +361,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 			}
 		}
 
-		// Check expiry is exactly *the epoch before* the start of a proving period.
-		periodOffset := st.ProvingPeriodStart % WPoStProvingPeriod
-		expiryOffset := (params.Expiration + 1) % WPoStProvingPeriod
-		if expiryOffset != periodOffset {
-			rt.Abortf(exitcode.ErrIllegalArgument, "invalid expiration %d, must be immediately before proving period boundary %d mod %d",
-				params.Expiration, periodOffset, WPoStProvingPeriod)
-		}
+		validateExpiration(st, params.Expiration, rt)
 
 		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
@@ -608,6 +602,8 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
 	var st State
 	rt.State().Readonly(&st)
 	rt.ValidateImmediateCallerIs(st.Info.Worker)
+
+	validateExpiration(st, params.NewExpiration, rt)
 
 	store := adt.AsStore(rt)
 	sectorNo := params.SectorNumber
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1221,6 +1226,16 @@ func computeFaultsFromMissingPoSts(st *State, deadlines *Deadlines, sinceDeadlin
 	return
 }
 
+// Check expiry is exactly *the epoch before* the start of a proving period.
+func validateExpiration(st State, expiration abi.ChainEpoch, rt Runtime) {
```
anorth, line 1230: Pass `State` as a pointer, not value.

For consistency with other methods, pass `rt` as the first parameter. But better generally would be to return an error from this function and do the abort directly in the actor method, avoding the `rt` parameter. As a policy, it makes these methods callable without needing a runtime (but there are cases where it's pragmatic to ignore).

actors/builtin/miner/miner_test.go
```diff
@@ -390,6 +390,112 @@ func TestProvingPeriodCron(t *testing.T) {
...
+
+		func() {
+			defer func() {
+				r := recover()
+				require.NotNil(t, r)
+				assert.Contains(t, fmt.Sprintf("%s",r), "invalid expiration")
+			}()
+
+			actor.extendSector(rt, sector, uint64(extension), params)
+		}()
```
anorth, line 445: Use `rt.ExpectAbort`.

As my comment in #435 , combine this with other test cases to reduce the set-up.

actors/builtin/miner/miner_test.go
```diff
@@ -390,6 +390,112 @@ func TestProvingPeriodCron(t *testing.T) {
...
+	st := getState(rt)
+	deadline := st.DeadlineInfo(rt.Epoch())
+	expiration := deadline.PeriodEnd() + 10*miner.WPoStProvingPeriod
+	sectorInfo := a.commitAndProveSectors(rt, 1, expiration, big.Zero())
+
+	sector, found, err := getState(rt).GetSector(rt.AdtStore(), sectorInfo[0].SectorNumber)
+	require.NoError(t, err)
+	require.True(t, found)
+	return rt, st, expiration, sector
+}
```
anorth, line 497: I don't love this approach, because it makes these setup functions easy to abuse as a blunt way to set up other, unrelated tests, which leads to unwanted coupling between the tests and general maintenance pain. This is based on my experience working with existing uses of the pattern in go-filecoin.

I think there probably is room to factor out some generic set-up code more than is already done, but I would want to do so in a way that is explicitly not coupled to any particular tests, probably some kind of builder pattern.

In this case, I'd prefer to share the setup code by just testing multiple conditions in the one test case. E.g. see TestCommitSector.

PR #438, opened 2020-06-05
Load balance new sectors to less-full partitions

### LINKED ISSUE #402: Partition assignment logic is buggy
Somehow miners are getting more sectors than the limit assigned to a given deadline. For example, in block bafy2bzaceamzy2qpoanxusxwzorsz2kp4zjrn43mfkgi7me2ickyuyyoi6vgk of the testnet, miner t0212888 has 2386 sectors in deadline 0.

### LINKED ISSUE #409: Implement load balancing of partitions among a miner's deadlines
Follow-on work from the initial Window PoSt implementation is to assign new sectors to deadlines so as to balance the load for a single miner between its deadlines. 

A simple sort (as suggested in #402) is attractive, but not correct if there are a larger number of partitions to be filled than deadlines of equally-minimal weight. The full solution probably involves building a heap or maintaining a sorted list in order to repeatedly add the next partition to a minimal deadline, taking into account partitions just added.

See also #402 and #432.

anorth:
Fixes the assignment of new sectors to deadlines to open new partitions in one of the least-full deadlines, rather than a simple round robin always starting from the first (which led the first deadline to get ~all the sectors in practice).

See discussion in the comments about efficiency trade-offs made.

Closes #409, closes #402.

### DIFF
--- a/actors/builtin/miner/deadlines.go
+++ b/actors/builtin/miner/deadlines.go
@@ -239,7 +239,7 @@ func AssignNewSectors(deadlines *Deadlines, partitionSize uint64, newSectors []u
 
 	sortDeadlines := func() {
 		// Order deadline indexes by corresponding partition count (then secondarily by index) to form a queue.
-		sort.Slice(dlIdxs, func(i, j int) bool {
+		sort.SliceStable(dlIdxs, func(i, j int) bool {
 			idxI, idxJ := dlIdxs[i], dlIdxs[j]
 			countI, countJ := deadlinePartitionCounts[idxI], deadlinePartitionCounts[idxJ]
 			if countI == countJ {
--- a/actors/builtin/miner/deadlines.go
+++ b/actors/builtin/miner/deadlines.go
@@ -2,6 +2,8 @@ package miner
 
 import (
 	"fmt"
+	"math"
+	"sort"
 
 	"github.com/filecoin-project/specs-actors/actors/abi"
 )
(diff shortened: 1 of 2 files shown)


anorth:
There's one more test I'd like to write, but since I probably can't do it before the end of week, @magik6k please land this if you want it for a devnet/testnet reset.

actors/builtin/miner/deadlines.go
```diff
@@ -209,18 +223,45 @@ func AssignNewSectors(deadlines *Deadlines, partitionSize uint64, newSectors []u
...
+	// TODO: randomize assignment among equally-full deadlines https://github.com/filecoin-project/specs-actors/issues/432
+
+	dlIdxs := make([]uint64, WPoStPeriodDeadlines)
+	for i := range dlIdxs {
+		dlIdxs[i] = uint64(i)
+	}
+
+	sortDeadlines := func() {
+		// Order deadline indexes by corresponding partition count (then secondarily by index) to form a queue.
+		sort.Slice(dlIdxs, func(i, j int) bool {
```
magik6k, line 242: I'd feel more comfortable if this was
```suggestion
		sort.SliceStable(dlIdxs, func(i, j int) bool {
```

actors/builtin/miner/deadlines_test.go
```diff
@@ -443,17 +556,57 @@ func deadlinesWithFullPartitions(t *testing.T, n uint64) *miner.Deadlines {
...
+}
+
+// A builder for initialising a Deadlines with sectors assigned.
+type DeadlinesBuilder struct {
+	Deadlines     *miner.Deadlines
+	NextSectorIdx uint64
+	t             *testing.T
+}
+
+// Creates a new builder with, with optional initial sector counts.
```
ZenGround0, line 570: nit: "with, with"

magik6k: fixed

actors/builtin/miner/deadlines.go
anorth: I did consider that, but note that ordering secondarily by deadline index makes this a total ordering, so stability is irrelevant.

PR #443, opened 2020-06-05
Enforce useable approval number

### LINKED ISSUE #92: Multisig actor should prevent impossible approval combinations
Right now it's possible to instantiate an actor with more approvals required than it has signing keys. We should run a check to make sure you can only require as many approvals as there are keys on the actor, and no more.

ZenGround0:
To prevent devops199ing yourself by accident
closes #92

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -78,6 +78,10 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 		rt.Abortf(exitcode.ErrIllegalArgument, "must have at least one signer")
 	}
 
+	if params.NumApprovalsThreshold > int64(len(params.Signers)) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "must not require more approvals than signers")
+	}
+
 	var signers []addr.Address
 	for _, sa := range params.Signers {
 		signers = append(signers, sa)
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -82,18 +82,13 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 		rt.Abortf(exitcode.ErrIllegalArgument, "must not require more approvals than signers")
 	}
 
-	var signers []addr.Address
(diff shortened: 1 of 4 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -78,6 +78,10 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 		rt.Abortf(exitcode.ErrIllegalArgument, "must have at least one signer")
 	}
 
+	if params.NumApprovalsThreshold > int64(len(params.Signers)) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "must not require more approvals than signers")
+	}
+
 	var signers []addr.Address
 	for _, sa := range params.Signers {
 		signers = append(signers, sa)
```
frrist, line 87: While were here, I think we can remove this too.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -78,6 +78,10 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 		rt.Abortf(exitcode.ErrIllegalArgument, "must have at least one signer")
 	}
 
+	if params.NumApprovalsThreshold > int64(len(params.Signers)) {
```
frrist, line 81: A couple other cases we should consider checking:
- Negative `NumApprovalsThreshold`
- Negative `UnlockDuation`
- Duplicate `Signers` -- e.g. `signers = addr.Address{"alice", "bob", "bob", bob"}`

ZenGround0: Both the negative checks are taken care of by using cbor and int64 typed fields.  Our cbor decoder should fail reading params bytes into the ConstructorParams struct.  If we want to prohibit duplicate signers you should file another issue.  It's not clear to me that we want to enforce that and that design discussion should be separate from this PR.

ZenGround0: int64 means cbor doesn't help

anorth: Changing to `uint64` in `ConstructorParams` is ok. The days of "int64 only" are passed. However, I think we actually want to require that approvals >= 1, so check that explicitly. Please also check the `UnlockDuration` and file an issue for validating signers.

PR #444, opened 2020-06-05
miner actor does not get randomness from future

### LINKED ISSUE #406: Miner actor can try to acquire randomness from a future epoch
This can happen when trying to setup the first proving period in `handleProvingPeriod`, since we do:

```
assignmentSeed := rt.GetRandomness(crypto.DomainSeparationTag_WindowedPoStDeadlineAssignment, deadline.PeriodEnd(), nil)
```

(https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1074)

Based on a quick chat with @anorth, it will probably be sufficient to use `min(deadline.PeriodEnd(), currEpoch - <lookback constant>)` when doing this. The lookback constant can probably just be 1.

acruikshank:
closes #406

### Motivation

In the first proving period, the miner actor may attempt to retrieve randomness from the future when it handles the proving period cron task. In this case it should retrieve the randomness from the current epoch minus a lookback.

### Proposed Changes

1. Set the randomness epoch to be the min of of the proving period end and the current epoch - `ElectionLookback`. It is unclear whether this deserves a new constant or not. The thinking is that whatever would cause us to change `ElectionLookback` would drive us to change this value as well, so we might as well reuse it.
2. Add a test that runs `handleProvingPeriod` in scenarios where the miner would have required randomness from the future and another where it would get it from the proving period end.
3. Add the ability for the actor harness to submit PoSt on deadlines so we can run `handleProvingPeriod` with sectors in proving periods other than the first without incurring faults. This is a bit heavyweight for this test, but I suspect we'll need it later.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1084,7 +1084,9 @@ func handleProvingPeriod(rt Runtime) {
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expand new sectors")
 
 			if len(newSectors) > 0 {
-				assignmentSeed := rt.GetRandomness(crypto.DomainSeparationTag_WindowedPoStDeadlineAssignment, deadline.PeriodEnd(), nil)
+				//if period end is not in the future, get randomness from previous epoch
+				randomnessEpoch := minEpoch(deadline.PeriodEnd(), rt.CurrEpoch()-ElectionLookback)
+				assignmentSeed := rt.GetRandomness(crypto.DomainSeparationTag_WindowedPoStDeadlineAssignment, randomnessEpoch, nil)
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1084,7 +1084,9 @@ func handleProvingPeriod(rt Runtime) {
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expand new sectors")
 
 			if len(newSectors) > 0 {
-				assignmentSeed := rt.GetRandomness(crypto.DomainSeparationTag_WindowedPoStDeadlineAssignment, deadline.PeriodEnd(), nil)
+				//if period end is not in the future, get randomness from previous epoch
```
ZenGround0, line 1087: @acruikshank should this say "if period end is not in the past"?

PR #445, opened 2020-06-05
Feat/reuse vesting funds

whyrusleeping:
This is an old set of commits i wrote to reuse the vesting funds object multiple times across a single method call. I never got around to putting it up as a PR, but if anyone wants to revive it, here it is. (please also feel free to start from scratch, i wont be offended)


anorth:
Thanks, I'mma keep this around for a bit for inspiration, but probably not first priority.

ZenGround0:
The ideas live on in #678

PR #446, opened 2020-06-06
miner: Get correct randomness for new sector assignments

magik6k:
In the first non-complete PP deadline.PeriodEnd() is in the future, this works around the problem, but we may want a nicer solution.


magik6k:
So #444 exists

Issue #353, opened 2020-05-07
Network parameter and policy configuration


anorth:
The miner/power/market actor policy files hard-code a bunch of values that are ultimately network parameters. For test networks, it would be really handy to be able to change them. Right now, node implementations monkey-patch some of them at boot.

anorth:
A cheap but effective thing might be to wrap all these values up in a struct, set in a `var`, that can be replaced wholesale. And a `DefaultPolicy` instance with our mainnet values.

@nonsense would that help for Oni? @magik6k would this be reasonable for Lotus to absorb (effectively a rename of any of those constants you reference directly)?

anorth:
I bumped into this today wanting to write unit tests using the 2KiB sector size, which are not in the allowed list for miners.

I think a good solution here would be to wrap up these policy values and functions into a struct that we set on the `Actor` code object itself [here](https://github.com/filecoin-project/specs-actors/blob/51145815bc17542f4731e1ac08d914d9d39a81ed/actors/builtin/miner/miner_actor.go#L49). In Lotus, this value would need to be set with the mainnet values (which would still be provided from specs-actors) when the dispatch table is constructed. But it could also be replaced with different values for testnets, and replaced in every individual unit test without having any kind of global/static effects.

wcgcyx:
Hello, I came across this today.

It seems to me that at the moment, the local dev-net 2k has a non-configurable epoch time of 4 seconds while the actor policy assumed the epoch time to be 30 seconds. 

As a result, you can't really test the storage miner using local dev-net, because the client uses 4 seconds to covert days to epoch while the actor uses 30 seconds and the minimum 180 days easily exceed the maximum allowed epoch range.

PR #449, opened 2020-06-08
Change ExpectedLeadersPerEpoch to a const

austinabell:
Just switch to see why it was a var, and figured might as well PR in even though very trivial change


ZenGround0:
@whyrusleeping does lotus rely on setting this var during testing or is the move to const preferred from lotus's POV?

whyrusleeping:
I would prefer keeping this as a var, we may want to be able to play with changing the value for different devnets or tests

anorth:
Thanks, see also https://github.com/filecoin-project/specs-actors/issues/353

PR #457, opened 2020-06-11
Refactor RegisteredProof to agree with proofs API.

porcuquine:
As of https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/24, the actual enumerated values of registered seal proofs are significant to the protocol. In order to emphasize the non-arbitrary nature of these types/values, and to simplify keeping them in lockstep through the repos, I would like to mirror the types as closely as possible. I consulted @laser and mentioned my intention to @anorth before undertaking this.

To support this work, we will need a third PR — this one in filecoin-ffi. Hopefully once `specs-actors` and `rust-filecoin-proofs-api` are in sync, the job of adapting between them will be simpler. So that code will need some refactoring, hopefully to something more straightforward than currently exists.

I will leave this PR in draft until the work in `filecoin-ffi` is done (depending on this branch). Review can begin prior to that, though. I am not very familiar with `specs-actors`, so it's entirely possible that I overlooked something important while making these changes. I did not intentionally change anything that wasn't required to make tests pass after making the intended changes in `sector.go`.

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -109,16 +109,16 @@ func (p RegisteredPoStProof) RegisteredSealProof() (RegisteredSealProof, error) 
 
 func (p RegisteredSealProof) SectorSize() (SectorSize, error) {
 	switch p {
-	case RegisteredSealProof_StackedDrg64GiBV1:
-		return 2 * (32 << 30), nil
-	case RegisteredSealProof_StackedDrg32GiBV1:
-		return 32 << 30, nil
 	case RegisteredSealProof_StackedDrg2KiBV1:
 		return 2 << 10, nil
 	case RegisteredSealProof_StackedDrg8MiBV1:
 		return 8 << 20, nil
 	case RegisteredSealProof_StackedDrg512MiBV1:
 		return 512 << 20, nil
+	case RegisteredSealProof_StackedDrg32GiBV1:
+		return 32 << 30, nil
+	case RegisteredSealProof_StackedDrg64GiBV1:
+		return 2 * (32 << 30), nil
 	default:
 		return 0, errors.Errorf("unsupported proof type: %v", p)
 	}
@@ -184,18 +184,6 @@ func (p RegisteredSealProof) RegisteredWinningPoStProof() (RegisteredPoStProof, 
 	}
 }
 
-// RegisteredWinningPoStProof produces the PoSt-specific RegisteredProof corresponding
-// to the receiving RegisteredProof.
-func (p RegisteredPoStProof) RegisteredWinningPoStProof() (RegisteredPoStProof, error) {
-	// Resolve to seal proof and then compute Winning PoSt from that.
-	sp, err := p.RegisteredSealProof()
-	if err != nil {
-		return 0, err
-	}
-	return sp.RegisteredWinningPoStProof()
-}
-
-
 // RegisteredWindowPoStProof produces the PoSt-specific RegisteredProof corresponding
 // to the receiving RegisteredProof.
 func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, error) {
@@ -215,17 +203,6 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 	}
 }
 
-// RegisteredWindowPoStProof produces the PoSt-specific RegisteredProof corresponding
-// to the receiving RegisteredProof.
-func (p RegisteredPoStProof) RegisteredWindowPoStProof() (RegisteredPoStProof, error) {
-	// Resolve to seal proof and then compute Window PoSt from that.
-	sp, err := p.RegisteredSealProof()
-	if err != nil {
-		return 0, err
-	}
-	return sp.RegisteredWindowPoStProof()
-}
-
 ///
 /// Sealing
 ///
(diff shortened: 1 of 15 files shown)


actors/abi/sector.go
```diff
@@ -64,140 +64,166 @@ func NewStoragePower(n int64) StoragePower {
...
+	case RegisteredSealProof_StackedDrg512MiBV1:
+		return RegisteredPoStProof_StackedDrgWinning512MiBV1, nil
 	default:
 		return 0, errors.Errorf("unsupported mapping from %+v to PoSt-specific RegisteredProof", p)
 	}
 }
 
+// RegisteredWinningPoStProof produces the PoSt-specific RegisteredProof corresponding
+// to the receiving RegisteredProof.
+func (p RegisteredPoStProof) RegisteredWinningPoStProof() (RegisteredPoStProof, error) {
```
anorth, line 189: We probably don't need this method, providing PoStProof->PoStProof. It's only a convenience and I can't think of an immediate need given the SealProof->WinningPoStProof method.

(The others are also now only convenience, but we understand a use case for them all so it makes sense to provide it).

actors/abi/sector.go
```diff
@@ -64,140 +64,166 @@ func NewStoragePower(n int64) StoragePower {
...
-		return RegisteredProof_StackedDRG32GiBSeal, nil
-	case RegisteredProof_StackedDRG2KiBSeal, RegisteredProof_StackedDRG2KiBPoSt, RegisteredProof_StackedDRG2KiBWindowPoSt, RegisteredProof_StackedDRG2KiBWinningPoSt:
-		return RegisteredProof_StackedDRG2KiBSeal, nil
-	case RegisteredProof_StackedDRG8MiBSeal, RegisteredProof_StackedDRG8MiBPoSt, RegisteredProof_StackedDRG8MiBWindowPoSt, RegisteredProof_StackedDRG8MiBWinningPoSt:
-		return RegisteredProof_StackedDRG8MiBSeal, nil
-	case RegisteredProof_StackedDRG512MiBSeal, RegisteredProof_StackedDRG512MiBPoSt, RegisteredProof_StackedDRG512MiBWindowPoSt, RegisteredProof_StackedDRG512MiBWinningPoSt:
-		return RegisteredProof_StackedDRG512MiBSeal, nil
-	default:
-		return 0, errors.Errorf("unsupported mapping from %+v to seal-specific RegisteredProof", p)
+func (p RegisteredPoStProof) RegisteredWindowPoStProof() (RegisteredPoStProof, error) {
```
anorth, line 220: Similarly I don't think we need this one, if we have the SealProof->WindowPoStProof one.

actors/abi/sector.go
```diff
@@ -209,7 +235,7 @@ type InteractiveSealRandomness Randomness
 
 // Information needed to verify a seal proof.
 type SealVerifyInfo struct {
-	RegisteredProof
+	RegisteredSealProof
```
anorth, line 238: For brevity, I'd mildly prefer `SealProof RegisteredSealProof`, and similarly for the others below.

actors/abi/sector.go
```diff
@@ -64,140 +64,166 @@ func NewStoragePower(n int64) StoragePower {
...
+	RegisteredSealProof_StackedDrg32GiBV1 = RegisteredSealProof(3);
+	RegisteredSealProof_StackedDrg64GiBV1 = RegisteredSealProof(4);
+)
 
-	RegisteredProof_StackedDRG64GiBWinningPoSt = RegisteredProof(18)
-	RegisteredProof_StackedDRG64GiBWindowPoSt  = RegisteredProof(19)
+type RegisteredPoStProof RegisteredProof
+const(
+ 	RegisteredPoStProof_StackedDrgWinning2KiBV1 = RegisteredPoStProof(0);
+   	RegisteredPoStProof_StackedDrgWinning8MiBV1 = RegisteredPoStProof(1);
```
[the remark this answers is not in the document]
porcuquine: Where? (I haven't been writing much Go, so I may just be blind to it.)

actors/abi/sector.go
```diff
@@ -64,140 +64,166 @@ func NewStoragePower(n int64) StoragePower {
...
+	case RegisteredSealProof_StackedDrg32GiBV1:
 		return 32 << 30, nil
-	case RegisteredProof_StackedDRG2KiBSeal:
+	case RegisteredSealProof_StackedDrg2KiBV1:
 		return 2 << 10, nil
-	case RegisteredProof_StackedDRG8MiBSeal:
+	case RegisteredSealProof_StackedDrg8MiBV1:
 		return 8 << 20, nil
-	case RegisteredProof_StackedDRG512MiBSeal:
+	case RegisteredSealProof_StackedDrg512MiBV1:
```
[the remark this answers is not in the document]
porcuquine: Thanks. Left it as such to ease first-round review. Fixed.

actors/abi/sector.go
porcuquine: Done.

actors/abi/sector.go
porcuquine: Done.

actors/abi/sector.go
porcuquine: Done.

actors/abi/sector.go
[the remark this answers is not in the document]
anorth: I think a browser extension I have is making it more obvious (or misleading me).

Screen Shot 2020-06-12 at 12 14 16 pm

Issue #419, opened 2020-06-03
Move sector power calculations to miner actor, simplify miner/power interactions


anorth:
The calculation of sector quality and power is in the power actor for legacy reasons. This causes awkward and unnecessary exchanges between the miner and actor when computing sector power. Move the `SectorStorageWeightDesc` and its uses to the miner actor and exchange plain power quantities in messages like `OnFaultBegin/End`.

The calculation of pledge requirements need to continue to consult the power actor, as it depends on the total already pledged.

PR #469, opened 2020-06-11
Document issues for existing TODOs

### LINKED ISSUE #412: Audit TODOs
A number of TODOs remain in the actor code a residual from porting them from a non-executable form. Some of them represent significant shortcomings. Audit them all and ensure important outstanding items are represented in issues here.

ZenGround0:
Adding github issue links for all TODOs in the code.  I defaulted to filing an issue for every TODO even though some appear to be easy and maybe an issue is overkill.  If some TODOs are totally done let me know and I'll close down and remove from PR.

Please search for appearances of issue 411 in this patch and verify that this issue is correctly scoped to cover the 4 instances where it is linked.  It seemed ok but I don't understand the scaling problem well enough to know if we should file distinct issues for separate TODOs.

Closes #412

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1341,7 +1341,7 @@ func checkPrecommitExpiry(rt Runtime, sectors *abi.BitField) {
 }
 
 // TODO: red flag that this method is potentially super expensive
-// https://github.com/filecoin-project/specs-actors/issues/411
+// https://github.com/filecoin-project/specs-actors/issues/483
 func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power.SectorTermination) {
 	empty, err := sectorNos.IsEmpty()
 	if err != nil {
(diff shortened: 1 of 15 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1334,6 +1341,7 @@ func checkPrecommitExpiry(rt Runtime, sectors *abi.BitField) {
 }
 
 // TODO: red flag that this method is potentially super expensive
+// https://github.com/filecoin-project/specs-actors/issues/411
```
anorth, line 1344: This one deserves a distinct issue for the fact that it loads all the deal IDs into a slice and makes a method call with it to terminate them.

actors/builtin/miner/miner_actor.go
```diff
@@ -1507,6 +1516,7 @@ func requestTerminateDeals(rt Runtime, dealIDs []abi.DealID) {
 func requestTerminateAllDeals(rt Runtime, st *State) {
 	// TODO: red flag this is an ~unbounded computation.
 	// Transform into an idempotent partial computation that can be progressed on each invocation.
+	// https://github.com/filecoin-project/specs-actors/issues/411
```
anorth, line 1519: This one is the same as the comment above, and distinct from the power/fee calculation issue.

actors/builtin/power/power_actor.go
```diff
@@ -217,7 +217,7 @@ func (a Actor) OnSectorProveCommit(rt Runtime, params *OnSectorProveCommitParams
 
 type OnSectorTerminateParams struct {
 	TerminationType SectorTermination
-	Weights         []SectorStorageWeightDesc // TODO: replace with power if it can be computed by miner
+	Weights         []SectorStorageWeightDesc // TODO: replace with power if it can be computed by miner, https://github.com/filecoin-project/specs-actors/issues/466
```
anorth, line 220: Use #419

support/mock/mockrt.go
```diff
@@ -800,6 +800,7 @@ func (rt *Runtime) failTestNow(msg string, args ...interface{}) {
 }
 
 func (rt *Runtime) TotalFilCircSupply() abi.TokenAmount {
+	// https://github.com/filecoin-project/specs-actors/issues/467
 	panic("todo crypto econ")
```
anorth, line 804: I think we could remove this TODO and just wait for the code that will use this to be written.

PR #472, opened 2020-06-11
Terminate all miner deals when a consensus fault is detected

### LINKED ISSUE #279: Terminate deals when miner is reported with consensus fault
Deals aren't handled at all when a miner is terminated for a consensus fault.

Ideally, centralise miner termination logic regardless of cause (consensus fault, repeated storage fault, voluntary termination).

acruikshank:
closes #279

### Motivation

When a miner is deleted due to a consensus fault, all of it's deals need to be properly terminated.

### Changes in this PR

1. Extract the miner termination logic out of `ReportConsensusFault` into a new `terminateMiner` method.
2. Call `requestTerminateAllDeals` from within that method.
3. Add a method to call `ReportConsensusFault` within the miner test actor harness.
4. Add logic to expect a `VerifyConsensusFault` and `DeleteActor` to the mock runtime.
5. Add the ability to specify deal IDs when creating sectors for miner tests so that we can observe the deal termination logic.
6. Add happy path test for `ReportConsensusFault` that confirms that it sends a `OnMinerSectorsTerminate` message to the storage market actor whose params contain every deal id referenced by all the miner's sectors.

### Work left to be done

1. Issue #279 requests that deal ids also be removed when a miner is removed voluntarily or in response to a storage fault. The logic for these cases does not seem to exist, but can use the new `terminateMiner` function when they do.
2. There are TODOs in the code base suggesting the `requestTerminateAllDeals` method could be too expensive to run. The proposed alternative did not seem sufficiently specified to deal with in this PR.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -910,7 +910,7 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 	}
 
 	// Reward reporter with a share of the miner's current balance.
-	slasherReward := rewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
+	slasherReward := RewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
 	_, code := rt.Send(reporter, builtin.MethodSend, nil, slasherReward)
 	builtin.RequireSuccess(rt, code, "failed to reward reporter")
 
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -909,27 +909,14 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid fault epoch %v ahead of current %v", fault.Epoch, rt.CurrEpoch())
 	}
 
-	var st State
-	rt.State().Readonly(&st)
-
-	// Notify power actor with lock-up total being removed.
-	_, code := rt.Send(
-		builtin.StoragePowerActorAddr,
-		builtin.MethodsPower.OnConsensusFault,
-		&st.LockedFunds,
-		abi.NewTokenAmount(0),
-	)
-	builtin.RequireSuccess(rt, code, "failed to notify power actor on consensus fault")
-
-	// TODO: terminate deals with market actor, https://github.com/filecoin-project/specs-actors/issues/279
-
 	// Reward reporter with a share of the miner's current balance.
 	slasherReward := rewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
-	_, code = rt.Send(reporter, builtin.MethodSend, nil, slasherReward)
+	_, code := rt.Send(reporter, builtin.MethodSend, nil, slasherReward)
 	builtin.RequireSuccess(rt, code, "failed to reward reporter")
 
-	// Delete the actor and burn all remaining funds
-	rt.DeleteActor(builtin.BurntFundsActorAddr)
+	// kill power, close deals and burn funds
+	terminateMiner(rt)
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1619,6 +1606,26 @@ func getVerifyInfo(rt Runtime, params *SealVerifyStuff) *abi.SealVerifyInfo {
 	}
 }
 
+// Closes down this miner by erasing its power, terminating all its deals and burning its funds
+func terminateMiner(rt Runtime) {
+	var st State
+	rt.State().Readonly(&st)
+
+	// Notify power actor with lock-up total being removed.
+	_, code := rt.Send(
```
ZenGround0, line 1615: It would be nice to extract miner termination into a function like this, but only in so much as we use it for other (yet to be impled/specced out) cases.  Including the consensus fault specific message send inside `terminateMiner` seems to defeat the purpose since we can't use this in other code paths.

I would either move the send up and leave only the two other calls, or defer wrapping miner termination into its own function at all.

acruikshank: Good catch.

actors/builtin/miner/miner_test.go
```diff
@@ -808,7 +850,35 @@ func (h *actorHarness) extendSector(rt *mock.Runtime, sector *miner.SectorOnChai
 		exitcode.Ok,
 	)
 	rt.Call(h.a.ExtendSectorExpiration, params)
+}
+
+func (h *actorHarness) reportConsensusFault(rt *mock.Runtime, from addr.Address, params *miner.ReportConsensusFaultParams, dealIDs []abi.DealID) {
+	rt.Reset()
```
anorth, line 856: This mock reset is worrying. I expect reset to be used only in the test code itself, rather than here. On the other hand, there should be a `rt.Verify()` call after the last `rt.Call` in this method (line 881), which will have the effect of resetting after verifying expectations.

I looked into other uses of `Reset` that might have guided you and saw that I missed this in your tests earlier for extending sector expirations. I will submit a PR to fix up the patterns there.

support/mock/mockrt.go
```diff
@@ -306,12 +318,18 @@ func (rt *Runtime) CreateActor(codeId cid.Cid, address addr.Address) {
...
 		rt.Abortf(exitcode.SysErrorIllegalActor, "side-effect within transaction")
 	}
-	panic("implement me")
+	if rt.expectDeleteActor == nil {
+		rt.failTestNow("Unexpected call to delete actor")
+	}
+
+	if *rt.expectDeleteActor != addr {
+		rt.failTestNow("Attempt to delete wrong actor. Expected %s, got %s.", rt.expectDeleteActor.String(), addr.String())
+	}
```
anorth, line 332: Finally, clear the expectation with 
```
rt.expectDeleteActor = nil
```

support/mock/mockrt.go
```diff
@@ -306,12 +318,18 @@ func (rt *Runtime) CreateActor(codeId cid.Cid, address addr.Address) {
...
 	if rt.inTransaction {
 		rt.Abortf(exitcode.SysErrorIllegalActor, "side-effect within transaction")
 	}
-	panic("implement me")
+	if rt.expectDeleteActor == nil {
+		rt.failTestNow("Unexpected call to delete actor")
+	}
+
+	if *rt.expectDeleteActor != addr {
+		rt.failTestNow("Attempt to delete wrong actor. Expected %s, got %s.", rt.expectDeleteActor.String(), addr.String())
```
anorth, line 331: Nit: lowercase messages for consistency with the existing ones.

support/mock/mockrt.go
```diff
@@ -306,12 +318,18 @@ func (rt *Runtime) CreateActor(codeId cid.Cid, address addr.Address) {
...
 
-func (rt *Runtime) DeleteActor(_ addr.Address) {
+func (rt *Runtime) DeleteActor(addr addr.Address) {
 	rt.requireInCall()
 	if rt.inTransaction {
 		rt.Abortf(exitcode.SysErrorIllegalActor, "side-effect within transaction")
 	}
-	panic("implement me")
+	if rt.expectDeleteActor == nil {
+		rt.failTestNow("Unexpected call to delete actor")
```
anorth, line 327: Include the `addr` in message.

support/mock/mockrt.go
```diff
@@ -499,7 +517,24 @@ func (rt *Runtime) VerifyPoSt(vi abi.WindowPoStVerifyInfo) error {
...
+			rt.failTest("block header 1 does not equal expected block header 1 (%v != %v)", h1, rt.expectVerifyConsensusFault.BlockHeader1)
+		}
+		if !bytes.Equal(h2, rt.expectVerifyConsensusFault.BlockHeader2) {
+			rt.failTest("block header 2 does not equal expected block header 2 (%v != %v)", h2, rt.expectVerifyConsensusFault.BlockHeader2)
+		}
+		if !bytes.Equal(extra, rt.expectVerifyConsensusFault.BlockHeaderExtra) {
+			rt.failTest("block header extra does not equal expected block header extra (%v != %v)", extra, rt.expectVerifyConsensusFault.BlockHeaderExtra)
+		}
+	}
+
```
anorth, line 536: `rt.expectVerifyConsensusFault = nil`

support/mock/mockrt.go
```diff
@@ -648,6 +687,17 @@ func (rt *Runtime) ExpectVerifyPoSt(post abi.WindowPoStVerifyInfo, result error)
...
+		BlockHeader1:        h1,
+		BlockHeader2:        h2,
+		BlockHeaderExtra:    extra,
+		Fault:               result,
+		Err:                 resultErr,
+	}
+}
+
 // Verifies that expected calls were received, and resets all expectations.
 func (rt *Runtime) Verify() {
```
anorth, line 702: Add checks for the new expectations for `DeleteActor` and `VerifyConsensusFault` to ensure the expected calls arrived.

actors/builtin/miner/miner_actor.go
acruikshank: I removed the call to storage market, but left the method so at least deal termination will be tied to miner termination.

PR #481, opened 2020-06-12
Add explanatory comments and fix some unhandled errors.

anorth:
Thanks @austinabell for [pointing out](https://github.com/austinabell/specs-actors/commit/23f9f7b4596fd5278948ea20997a31b84d6315f0) some of the unhandled errors.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -355,6 +355,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 				rt.Abortf(exitcode.ErrIllegalArgument, "sector %v already committed with deals", params.SectorNumber)
 			} else {
 				// Committed Capacity sector upgrade.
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -980,7 +990,9 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
 	case CronEventProvingPeriod:
 		handleProvingPeriod(rt)
 	case CronEventPreCommitExpiry:
-		checkPrecommitExpiry(rt, payload.Sectors)
+		if payload.Sectors != nil {
```
ZenGround0, line 993: Does this improve correctness?  It looks like it is just an optimization and if so it's not clear its worth doing.

anorth: Accessing it if nil is definitely wrong. I don't have reason to believe it will be, but this is just on the side of worthwhile defensiveness for me.

PR #484, opened 2020-06-13
Fix up testing patterns for extending sector expirations

anorth:
- Fixes incorrect building of runtime outside of the `t.Run()` blocks that use it, because this passes the wrong `testing.T` instance.
- Removes `rt.Reset()` and adds missing `rt.Verify()`.

The tests will pass after #482 lands, which cleans up empty sector expiration epochs.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -433,20 +433,24 @@ func TestExtendSectorExpiration(t *testing.T) {
 		WithHasher(fixedHasher(uint64(periodOffset))).
 		WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
 
-	rt := builder.Build(t)
-	actor.constructAndVerify(rt, periodOffset)
-	precommitEpoch := abi.ChainEpoch(1)
-	rt.SetEpoch(precommitEpoch)
-	st := getState(rt)
-	deadline := st.DeadlineInfo(rt.Epoch())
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -484,25 +494,21 @@ func TestExtendSectorExpiration(t *testing.T) {
...
-		assert.True(t, exists)
-
-		// assert that the expiration has been removed from the old epoch
-		expirations, err = st.GetSectorExpirations(rt.AdtStore(), expiration)
-		require.NoError(t, err)
-		exists, err = expirations.IsSet(uint64(sector.Info.SectorNumber))
-		require.NoError(t, err)
-		assert.False(t, exists)
+		// assert the only expiration is for the new target epoch
+		err = st.ForEachSectorExpiration(rt.AdtStore(), func(expiry abi.ChainEpoch, sectors *abi.BitField) error {
```
acruikshank, line 504: This will be a false negative if the code removes the expiration and fails to add it at the new epoch. Adding a length check outside the iteration would cover this, but I still think the previous version is more clear in its intent.

anorth: The motivation for the change was that the `expiration` var was no longer in scope, and this was attempting to assert the complete state of the expirations. You're right about the false negative, though. I can get it from the old sector info and will change back.

Issue #487, opened 2020-06-15
Unify redundant power-accounting methods


anorth:
The power actor contains a number of methods that essentially just update the power table:
- OnSectorProveCommit
- OnSectorTerminate
- OnFaultBegin
- OnFaultEnd
- OnSectorModifyWeightDesc

The only differences between these are whether the power is added or removed, and including calculation of pledge collateral.

The collateral calculation in OnSectorModifyWeightDesc is unused and can be removed. This means that all but the first of these methods can be combined into an `UpdateClaimedPower` method (c.f. `UpdatePledgeTotal`).

Then `OnSectorProveCommit` can possibly be combined if pledge calculation can be moved to the miner actor (see #468, #476)

PR #488, opened 2020-06-15
Move sector quality and power calculations into miner actor

### LINKED ISSUE #419: Move sector power calculations to miner actor, simplify miner/power interactions

anorth:
The calculation of sector quality and power was once in the power actor, but has no need to be. Moving the calculations to the miner actor simplifies the interactions.

- Removed `SectorStorageWeightDesc` struct
- Moved quality and power calculation functions to miner, in terms of `SectorOnChainInfo`
- Use power quantities rather than weight desc in parameters to power-accounting methods

Closes #419

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -192,116 +192,125 @@ func (a Actor) DeleteMiner(rt Runtime, params *DeleteMinerParams) *adt.EmptyValu
 }
 
 type OnSectorProveCommitParams struct {
-	Weight SectorStorageWeightDesc
+	RawBytePower         abi.StoragePower
+	QualityAdjustedPower abi.StoragePower
 }
 
-// Returns the initial pledge collateral requirement.
+// Adds power for the calling actor and returns the initial pledge requirement for committing that
+// much power at this epoch.
+// May only be invoked by a miner actor.
 func (a Actor) OnSectorProveCommit(rt Runtime, params *OnSectorProveCommitParams) *abi.TokenAmount {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
-	initialPledge := a.computeInitialPledge(rt, &params.Weight)
+	builtin.RequireParam(rt, params.RawBytePower.GreaterThanEqual(big.Zero()), "negative raw power %s", params.RawBytePower)
+	builtin.RequireParam(rt, params.QualityAdjustedPower.GreaterThanEqual(big.Zero()), "negative QA power %s", params.RawBytePower)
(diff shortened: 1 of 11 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -192,116 +192,125 @@ func (a Actor) DeleteMiner(rt Runtime, params *DeleteMinerParams) *adt.EmptyValu
...
 }
 
-// Returns the initial pledge collateral requirement.
+// Adds power for the calling actor and returns the initial pledge requirement for committing that
+// much power at this epoch.
+// May only be invoked by a miner actor.
 func (a Actor) OnSectorProveCommit(rt Runtime, params *OnSectorProveCommitParams) *abi.TokenAmount {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
-	initialPledge := a.computeInitialPledge(rt, &params.Weight)
+	builtin.RequireParam(rt, params.RawBytePower.GreaterThanEqual(big.Zero()), "negative raw power %s", params.RawBytePower)
```
acruikshank, line 204: Question: I'm curious why these are being added. I get that positive power is an invariant, but it's a weak one. The real invariant is that the power must be correctly computed, and for that the power actor is dependent on the miner.

Is it to catch bugs in the miners calculation? If so, should it happen on the miner side where there is more context?
Is it to prevent improper usage? I can see the potential for the method to be misused to remove power.

This will set a precedent for future assertions in actor code so I'd like to understand the reasoning.

actors/builtin/power/power_actor.go
```diff
@@ -192,116 +192,125 @@ func (a Actor) DeleteMiner(rt Runtime, params *DeleteMinerParams) *adt.EmptyValu
 }
 
 type OnSectorProveCommitParams struct {
-	Weight SectorStorageWeightDesc
```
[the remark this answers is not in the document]
anorth: Deleted in next PR

actors/builtin/power/power_actor.go
anorth: Right, this is all cleaned up in #487, where they are removed again for a method that is actually expecting negative values. Yes, it was to catch bugs or improper usage on the miner side.

PR #490, opened 2020-06-15
Unify sector power accounting methods

### LINKED ISSUE #487: Unify redundant power-accounting methods

anorth:
Deletes redundant power actor exported methods that are left over from when much more logic lived in the power actor.

- [x] Rebase after #488 lands

FYI @magik6k. Closes #487.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -301,14 +301,10 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		return nil
 	})
 
-	// Remove power for new faults, and burn penalties.
-	requestBeginFaults(rt, st.Info.SectorSize, detectedFaultSectors)
+	// Restore power for recovered sectors. Remove power for new faults.
+	requestUpdateSectorPower(rt, st.Info.SectorSize, recoveredSectors, detectedFaultSectors)
+	// Burn penalties.
 	burnFundsAndNotifyPledgeChange(rt, penalty)
-
-	// Restore power for recovered sectors.
-	if len(recoveredSectors) > 0 {
-		requestEndFaults(rt, st.Info.SectorSize, recoveredSectors)
-	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1411,12 +1405,11 @@ func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power
 		return nil
 	})
 
-	// End any fault state before terminating sector power.
-	// TODO: could we compress the three calls to power actor into one sector termination call?
+	// TODO: could we compress the multiple calls to power actor into one sector termination call?
 	// https://github.com/filecoin-project/specs-actors/issues/478
-	requestEndFaults(rt, st.Info.SectorSize, faultySectors)
+	// Copmute the power delta as if recovering all the currently-faulty sectors before terminating all of them.
```
ZenGround0, line 1410: nit: compute

actors/builtin/power/power_actor.go
```diff
@@ -214,106 +212,26 @@ func (a Actor) OnSectorProveCommit(rt Runtime, params *OnSectorProveCommitParams
...
-}
-
-// Removes power and replaces for the calling actor.
-// May only be invoked by a miner actor.
-// Returns new initial pledge, now committed in place of the old.
-// This is just a composition of OnSectorTerminate and OnSectorProveCommit.
-// TODO: remove duplication of OnSectorTerminate, https://github.com/filecoin-project/specs-actors/issues/487
-func (a Actor) OnSectorModifyWeightDesc(rt Runtime, params *OnSectorModifyWeightDescParams) *abi.TokenAmount {
-	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
-	builtin.RequireParam(rt, params.PrevRawBytePower.GreaterThanEqual(big.Zero()), "negative prev raw power %s", params.PrevRawBytePower)
```
ZenGround0, line 299: Have these checks (and the others removed from other methods) become deprecated?  Were they never useful?  Is there anywhere we should consider moving them?

actors/builtin/power/power_actor.go
```diff
@@ -37,15 +37,12 @@ func (a Actor) Exports() []interface{} {
...
-		5:                         a.OnSectorTerminate,
-		6:                         a.OnFaultBegin,
-		7:                         a.OnFaultEnd,
-		8:                         a.OnSectorModifyWeightDesc,
-		9:                         a.EnrollCronEvent,
-		10:                        a.OnEpochTickEnd,
-		11:                        a.UpdatePledgeTotal,
-		12:                        a.OnConsensusFault,
-		13:                        a.SubmitPoRepForBulkVerify,
+		5:                         a.UpdateClaimedPower,
```
acruikshank, line 40: It's a little disturbing that this change affects no tests.

actors/builtin/power/power_actor.go
anorth: I added them only the the previous PR. Previously there were different methods for adding xor removing power for different scenarios. Now we're down to just two, and the new `UpdateClaimedPower` is intended to handle both positive and negative deltas, without being opinionated about why it's being called.

actors/builtin/power/power_actor.go
anorth: Right. Because The `OnSectorProveCommit` is still a distinct method for the power actor, all the test scenarios committing new sectors weren't touched. We need to cover the lifecycle events of the removed methods (faults, recovery, termination).

PR #493, opened 2020-06-15
Add ChargeGas method to Runtime

Kubuxu:
It will be used in the current effort of setting sane values for gas.
Gas is split into compute gas and storage gas to allow better analisys
of computation time in relation to gas.
Both gas types are latter combined into one and this is charged for the
execution.

Virtual gas is added for observing how changes to gas numbers affect
the balance without affecting the stateroot of the chain.

### DIFF
--- a/actors/runtime/runtime.go
+++ b/actors/runtime/runtime.go
@@ -95,6 +95,13 @@ type Runtime interface {
 
 	// Starts a new tracing span. The span must be End()ed explicitly, typically with a deferred invocation.
 	StartSpan(name string) TraceSpan
+
+	// ChargeGas charges specified amount of `gas` for execution.
+	// `name` provides information about gas charging point
+	// `virtual` sets virtual amount of gas to charge, this amount is not counted
+	// toward execution cost. This functionality is used for observing global changes
+	// in total gas charged if amount of gas charged was to be changed.
+	ChargeGas(name string, gas int64, virtual int64)
 }
 
 // Store defines the storage module exposed to actors.
--- a/support/mock/mockrt.go
+++ b/support/mock/mockrt.go
@@ -808,6 +808,8 @@ func (rt *Runtime) failTestNow(msg string, args ...interface{}) {
 	rt.t.FailNow()
 }
 
+func (rt *Runtime) ChargeGas(_ string, _, _ int64) {}
+
 type ReturnWrapper struct {
 	V runtime.CBORMarshaler
 }


actors/runtime/gas.go
```diff
@@ -0,0 +1,48 @@
+package runtime
+
+// Gas describes amount of gas to charge
+type Gas struct {
+	// Name describes gas charging point.
+	// For example: OnIpldPut, OnHashing, OnRLEDecode
+	// The Name should be stable.
+	Name string
+	// Info contains additiona information about gas charge.
```
ZenGround0, line 9: nit "additional"

actors/runtime/gas.go
```diff
@@ -0,0 +1,48 @@
...
+	// For example it might be integer if given length of data is processed
+	// or string if operation has multiple types.
+	Info interface{}
+
+	// Compute is amount of gas charged for computation.
+	Compute int64
+	// Storage is amount of gas charged for permament strage.
```
ZenGround0, line 16: nit: permanent storage.

actors/runtime/gas.go
```diff
@@ -0,0 +1,48 @@
...
+	Storage int64
+
+	// VirtualCompute is gas for computation that is no counted towards total gas
```
ZenGround0, line 19: nit: not

Kubuxu:
What can I say, this is my spelling ability at midnight without spellcheck.

actors/runtime/runtime.go
```diff
@@ -87,6 +87,10 @@ type Runtime interface {
 
 	// Starts a new tracing span. The span must be End()ed explicitly, typically with a deferred invocation.
 	StartSpan(name string) TraceSpan
+
+	// ChargeGas charges specified amount of gas for execution.
+	// Should be used before the operation the charge is for.
+	ChargeGas(gas Gas)
```
anorth, line 93: This is a pretty big structure to pass by value all the time. Consider taking a pointer here, and changing the constructor/builder methods to also return/receive pointers.

Kubuxu: It is 64bytes, so not that big, by being copied on the stack we are quite possibly saving on a lot of heap allocations (as we are passing it to interface it has to be assumed escaped).
64 bytes is also the size of a cache line on most CPU. Meaning that copy should be very fast if it happens.

I also prefer the immutable semantics of how it is built. So unless you are feeling strongly about it, I won't be changing it.

Kubuxu:
@anorth 
It is what Lotus is using internally right now, we can either put it here and delete our version, or we have to duplicate it and convert between these two because in lotus internally we need storage gas accounting.

>  Does this code imply a proposal to make all gas charging explicit?

No,  you can probably in the initial wave expect some `ChargeGas` with 0 gas cost so we can better understand what is taking the time and in future limited few additions of gas charging, probably in some loops that do a lot of computation.

actors/runtime/runtime.go
anorth: It looks to me like 4 int64s + 2 pointers.

anorth:
If we really needed all the complexity on this side of the runtime boundary, sure. But other implementations will have other gas accounting structures and I think decoupling them as much as we can is worthwhile. The design principle behind the runtime interface is to be as narrow as we can possibly make it (although that still ends up rather wide). More coupling means more friction for implementations to ingest changes, or resistance to the actors making improvements.

If you ack that there's no concrete present use for the storage/compute split in the actors at present, why not just a simple method on the runtime:

```
  ChargeGas(name string, amt int64)
```

You can easily convert this into the larger structure on the inside. Maybe we'll need to add another parameter later either to distinguish real/virtual, or compute/storage.

actors/runtime/runtime.go
Kubuxu: Slice is 3 uint64s, so 1 pointer, 4 int64 and slice header (3), 8x uint64.

Kubuxu:
We for sure need the compute/virtual compute split, otherwise we are no able to use current tooling for observing how gas changes affect existing chains.

Issue #176, opened 2020-02-18
Switch from pkg/errors to xerrors


anorth:
The pkg/errors package is reaching end of life since Go 1.13 has defined a new mechanism for wrapping errors.

See discussion here https://github.com/filecoin-project/specs-actors/pull/174#discussion_r380790404

We don't *have* to change, but I request that we have a single, consistent pattern throughout this whole module.

PR #498, opened 2020-06-16
Correct declared and undeclared fault computations

### LINKED ISSUE #437: Implement fault and termination penalty calculations
The fault fee and sector penalty depend on the instantaneous block reward value (from reward actors) and the total committed power (from power actor) during the epoch the fault occurs. Each is a simple multiple of these.

- `BR(t) = CurrEpochReward(t) * SectorQualityAdjustedPower * EpochsInDay  / TotalNetworkQualityAdjustedPower(t)`
- `FF(t) = a*BR(t)`, with `a` = 2.14.
- `SP(t) = 5BR(t)`

The sector termination fee additionally depends on the initial pledge calculated _when the sector was committed_. The intent is to calculate the block reward that was earned by the sector during its life (up to 180 days max), if the block reward had been fixed at the value when the sector was committed. This is then bounded below by `SP` calculated at the termination epoch.

- `TF = max(SP(t), InitialPledge + BR(StartEpoch)*min(SectorAgeInDays, 180))`

Note that this requires recording the block reward value when the sector is committed. We can do this in two ways:
- record the raw BR
- record the initial pledge, from which BR might be calculable if it's a strict function of BR

For now, I propose recording *both*. It will be easy to remove one if it turns out to be redundant.

Consider doing #419 first to ease implementation.

### LINKED ISSUE #468: Implement InitialPledgeForWeight calculation
Currently builtin/power/policy.go's `InitialPledgeForWeight` is a stub.

The calculation should be `20*BR(t)`, for now, where `BR(t)` is the share of block reward that the sector would earn in one day.

The implementation has parameters for circulating supply and total committed pledge, but does not use them for now. Leave these here to keep the plumbing in place for the quite-possible case that they will be part of the computation.

Consider #420 and #489 first.

acruikshank:
partially addresses #437
closes #468

### Motivation

We have not been using the right calculation for fault penalties. This PR:

1. Moves the calculation logic into the monies.go file.
2. Extracts `ExpectedDayRewardForSector` as the basis for all penalty calculations.
3. Implents `pledgePenaltyForSectorDeclaredFault`, `pledgePenaltyForSectorUndeclaredFault`, and `InitialPledgeForPower` according to the latest understanding from slash the world.
4. Adds calls to get `epochReward` from reward actor and `totalPwr` from power actor to miner calls that now need them to compute penalties.
5. Modify penalty calls in miner actor to use the new logic.
6. Modify tests to expect new extra calls to reward and power actor.

### Work left undone

Issue #437 talks of both fault fees and sector termination fees. This work only covers the former. Sector termination will be covered in a future PR.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -9,7 +9,6 @@ import (
 	"github.com/filecoin-project/go-bitfield"
 	cid "github.com/ipfs/go-cid"
 	cbg "github.com/whyrusleeping/cbor-gen"
-	"golang.org/x/xerrors"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
@@ -1804,45 +1803,34 @@ func validateFRDeclaration(deadlines *Deadlines, deadline *DeadlineInfo, declare
 }
 
 // qaPowerForSectors sums the quality adjusted power of all sectors
-func qaPowerForSectors(sectors []*SectorOnChainInfo) (abi.StoragePower, error) {
+func qaPowerForSectors(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo) abi.StoragePower {
 	power := big.Zero()
 	for _, s := range sectors {
-		sectorSize, err := s.Info.SealProof.SectorSize()
-		if err != nil {
-			return abi.NewTokenAmount(0), xerrors.Errorf("could not get sector size for sector: %w", err)
-		}
 		power = big.Add(power, QAPowerForSector(sectorSize, s))
 	}
-	return power, nil
+	return power
 }
 
 func unlockDeclaredFaultPenalty(st *State, store adt.Store, currEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, sectors []*SectorOnChainInfo) (abi.TokenAmount, error) {
-	totalQAPower, err := qaPowerForSectors(sectors)
-	if err != nil {
-		return abi.NewTokenAmount(0), err
-	}
-	fee := PledgePenaltyForSectorDeclaredFault(epochTargetReward, networkQAPower, totalQAPower)
+	sectorSize := st.Info.SectorSize
+	totalQAPower := qaPowerForSectors(sectorSize, sectors)
+	fee := PledgePenaltyForDeclaredFault(epochTargetReward, networkQAPower, totalQAPower)
 	return st.UnlockUnvestedFunds(store, currEpoch, fee)
 }
 
 func unlockUndeclaredFaultPenalty(st *State, store adt.Store, currEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, sectors []*SectorOnChainInfo) (abi.TokenAmount, error) {
-	totalQAPower, err := qaPowerForSectors(sectors)
-	if err != nil {
-		return abi.NewTokenAmount(0), err
-	}
-	fee := pledgePenaltyForSectorUndeclaredFault(epochTargetReward, networkQAPower, totalQAPower)
+	sectorSize := st.Info.SectorSize
+	totalQAPower := qaPowerForSectors(sectorSize, sectors)
+	fee := PledgePenaltyForUndeclaredFault(epochTargetReward, networkQAPower, totalQAPower)
 	return st.UnlockUnvestedFunds(store, currEpoch, fee)
 }
 
 func unlockTerminationPenalty(st *State, store adt.Store, curEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, sectors []*SectorOnChainInfo) (abi.TokenAmount, error) {
 	totalFee := big.Zero()
+	sectorSize := st.Info.SectorSize
 	for _, s := range sectors {
-		sectorSize, err := s.Info.SealProof.SectorSize()
-		if err != nil {
-			return abi.NewTokenAmount(0), xerrors.Errorf("could not get sector size for sector: %w", err)
-		}
 		sectorPower := QAPowerForSector(sectorSize, s)
-		fee := pledgePenaltyForSectorTermination(s, epochTargetReward, networkQAPower, sectorPower)
+		fee := pledgePenaltyForTermination(s, epochTargetReward, networkQAPower, sectorPower)
 		totalFee = big.Add(fee, totalFee)
 	}
 	return totalFee, nil
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"github.com/pkg/errors"
 
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
@@ -216,6 +217,10 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 	var detectedFaultSectors []*SectorOnChainInfo
 	var recoveredSectors []*SectorOnChainInfo
 	penalty := abi.NewTokenAmount(0)
+
+	epochReward := requestCurrentEpochBlockReward(rt)
+	pwrTotal := requestCurrentTotalPower(rt)
+
 	rt.State().Transaction(&st, func() interface{} {
 		rt.ValidateImmediateCallerIs(st.Info.Worker)
 
@@ -231,7 +236,7 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		// Traverse earlier submissions and enact detected faults.
 		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
 		// of penalties if locked pledge drops too low.
-		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines)
+		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
 
 		if !currDeadline.PeriodStarted() {
 			rt.Abortf(exitcode.ErrIllegalArgument, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
@@ -675,6 +680,9 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 	var detectedFaultSectors []*SectorOnChainInfo
 	penalty := abi.NewTokenAmount(0)
 
+	epochReward := requestCurrentEpochBlockReward(rt)
+	pwrTotal := requestCurrentTotalPower(rt)
+
 	rt.State().Transaction(&st, func() interface{} {
 		rt.ValidateImmediateCallerIs(st.Info.Worker)
 
@@ -685,7 +693,7 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 		// Traverse earlier submissions and enact detected faults.
 		// This is necessary to move the NextDeadlineToProcessFaults index past the deadline that this recovery
 		// is targeting, so that the recovery won't be declared failed next time it's checked during this proving period.
-		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines)
+		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
 
 		var decaredSectors []*abi.BitField
 		for _, decl := range params.Faults {
@@ -747,7 +755,7 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load fault sectors")
 
 			// Unlock penalty for declared faults.
-			declaredPenalty, err := unlockPenalty(&st, store, currEpoch, declaredFaultSectors, pledgePenaltyForSectorDeclaredFault)
+			declaredPenalty, err := unlockPenalty(&st, store, currEpoch, declaredFaultSectors, computeDeclaredFaultPenalty(epochReward, pwrTotal.QualityAdjPower))
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to charge fault fee")
 			penalty = big.Add(penalty, declaredPenalty)
 		}
@@ -787,6 +795,9 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 	var detectedFaultSectors []*SectorOnChainInfo
 	penalty := abi.NewTokenAmount(0)
 
+	epochReward := requestCurrentEpochBlockReward(rt)
+	pwrTotal := requestCurrentTotalPower(rt)
+
 	currEpoch := rt.CurrEpoch()
 	store := adt.AsStore(rt)
 	var st State
@@ -800,7 +811,7 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 		// Traverse earlier submissions and enact detected faults.
 		// This is necessary to move the NextDeadlineToProcessFaults index past the deadline that this recovery
 		// is targeting, so that the recovery won't be declared failed next time it's checked during this proving period.
-		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines)
+		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
 
 		var declaredSectors []*abi.BitField
 		for _, decl := range params.Recoveries {
@@ -976,6 +987,10 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
 // Invoked at the end of each proving period, at the end of the epoch before the next one starts.
 func handleProvingPeriod(rt Runtime) {
 	store := adt.AsStore(rt)
+
+	epochReward := requestCurrentEpochBlockReward(rt)
+	pwrTotal := requestCurrentTotalPower(rt)
+
 	var st State
 	{
 		// Vest locked funds.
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"github.com/pkg/errors"
```
anorth, line 7: FYI we'll be switching to xerrors, so just use that now for new things (or at least use the `%w` format specifier for wrapping). The current state is a mix of both. #176

actors/builtin/miner/miner_actor.go
```diff
@@ -1140,7 +1158,8 @@ func processMissingPoStFaults(rt Runtime, st *State, store adt.Store, deadlines
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load failed recovery sectors")
 
 	// Unlock sector penalty for all undeclared faults.
-	penalty, err := unlockPenalty(st, store, currEpoch, append(detectedFaultSectors, failedRecoverySectors...), pledgePenaltyForSectorUndeclaredFault)
+	penaltyCalculator := computeUndeclaredFaultPenalty(epochReward, currentTotalPower)
```
anorth, line 1161: Pass the penalty calculator as the parameter, instead of the two values. You can move and re-use the one currently at 1066.

actors/builtin/miner/miner_actor.go
```diff
@@ -1140,7 +1158,8 @@ func processMissingPoStFaults(rt Runtime, st *State, store adt.Store, deadlines
...
+	penalty, err := unlockPenalty(st, store, currEpoch, append(detectedFaultSectors, failedRecoverySectors...), penaltyCalculator)
```
anorth, line 1162: As discussed, consider returning the detectedFaultSectors _and_ failedRecoverySectors from this method, and processing the penalty from the caller.

actors/builtin/miner/miner_actor.go
```diff
@@ -1777,14 +1796,32 @@ func validateFRDeclaration(deadlines *Deadlines, deadline *DeadlineInfo, declare
...
+			return abi.NewTokenAmount(0), errors.Wrap(err, "could not get sector size for sector")
+		}
+		fee = big.Add(fee, feeCalc(sectorSize, s))
 	}
 	return st.UnlockUnvestedFunds(store, currEpoch, fee)
 }
 
+func computeDeclaredFaultPenalty(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower) func(abi.SectorSize, *SectorOnChainInfo) abi.TokenAmount {
+	return func(size abi.SectorSize, s *SectorOnChainInfo) abi.TokenAmount {
+		qaPower := QAPowerForSector(size, s)
```
anorth, line 1813: I'm pretty sure penalties should be linear with power, so for multiple sectors one could sum all the power first, and then compute the penalty for that power. That might give cleaner control flow.

actors/builtin/miner/miner_test.go
```diff
@@ -730,6 +730,14 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
 	rt.SetCaller(h.worker, builtin.AccountActorCodeID)
 	rt.ExpectValidateCallerAddr(h.worker)
 
+	reward := big.NewIntUnsigned(1e18)
+	rt.ExpectSend(builtin.RewardActorAddr, builtin.MethodsReward.LastPerEpochReward, nil, big.Zero(), &reward, exitcode.Ok)
+
+	pwrTotal := power.CurrentTotalPowerReturn{
+		QualityAdjPower: big.NewIntUnsigned(1000),
```
anorth, line 737: This value is unrealistically small. I used `1 << 50` elsewhere.

actors/builtin/miner/monies.go
```diff
@@ -3,15 +3,49 @@ package miner
 import (
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
 )
 
+const InitialPledgeFactor = int64(20)
+const ExpectedFaultFactorNum = int64(214)
+const ExpectedFaultFactorDenom = int64(1000)
+const UnexpectedFaultFactor = int64(5)
```
anorth, line 12: Please document these. 

For consistency with naming elsewhere:
- Expected -> Declared
- Unexpected -> Undeclared

Also, let's make the big.Ints for these all up front rather than re-allocating at every use.

actors/builtin/miner/monies.go
```diff
@@ -3,15 +3,49 @@ package miner
...
+func pledgePenaltyForSectorDeclaredFault(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
+	return big.Div(
+		big.Mul(big.NewInt(ExpectedFaultFactorNum), ExpectedDayRewardForSector(epochTargetReward, networkQAPower, qaSectorPower)),
+		big.NewInt(ExpectedFaultFactorDenom))
+}
+
+// This is the SP(t) penalty for a newly faulty sector that has not been declared.
+// SP(t) = UnexpectedFaultFactor * BR(t)
+func pledgePenaltyForSectorUndeclaredFault(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
+	return big.Mul(big.NewInt(UnexpectedFaultFactor), ExpectedDayRewardForSector(epochTargetReward, networkQAPower, qaSectorPower))
```
anorth, line 34: Even though redundant, could you add the denominator = 1 so these follow exactly the same arithmetic flow, and are easy to change constants later?

actors/builtin/miner/miner_actor.go
```diff
@@ -1783,19 +1803,53 @@ func validateFRDeclaration(deadlines *Deadlines, deadline *DeadlineInfo, declare
...
-// The fee computation is a parameter.
-func unlockPenalty(st *State, store adt.Store, currEpoch abi.ChainEpoch, sectors []*SectorOnChainInfo,
-	feeCalc func(info *SectorOnChainInfo) abi.TokenAmount) (abi.TokenAmount, error) {
-	fee := big.Zero()
+// qaPowerForSectors sums the quality adjusted power of all sectors
+func qaPowerForSectors(sectors []*SectorOnChainInfo) (abi.StoragePower, error) {
+	power := big.Zero()
 	for _, s := range sectors {
-		fee = big.Add(fee, feeCalc(s))
+		sectorSize, err := s.Info.SealProof.SectorSize()
```
anorth, line 1810: The miner has a single fixed seal proof, and the sectorsize already computed from it at construction and stored in miner info. Use that constant value (pass it as a parameter here). See `PowerForSectors`.

If we ever change that, this will be only one of many places that need updating.

actors/builtin/miner/miner_actor.go
```diff
@@ -1783,19 +1803,53 @@ func validateFRDeclaration(deadlines *Deadlines, deadline *DeadlineInfo, declare
...
+		return abi.NewTokenAmount(0), err
 	}
+	fee := pledgePenaltyForSectorUndeclaredFault(epochTargetReward, networkQAPower, totalQAPower)
 	return st.UnlockUnvestedFunds(store, currEpoch, fee)
 }
 
+func unlockTerminationPenalty(st *State, store adt.Store, curEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, sectors []*SectorOnChainInfo) (abi.TokenAmount, error) {
+	totalFee := big.Zero()
+	for _, s := range sectors {
+		sectorSize, err := s.Info.SealProof.SectorSize()
```
anorth, line 1840: Ditto

actors/builtin/miner/miner_actor.go
```diff
@@ -1783,19 +1803,53 @@ func validateFRDeclaration(deadlines *Deadlines, deadline *DeadlineInfo, declare
...
+		if err != nil {
+			return abi.NewTokenAmount(0), xerrors.Errorf("could not get sector size for sector: %w", err)
+		}
+		sectorPower := QAPowerForSector(sectorSize, s)
+		fee := pledgePenaltyForSectorTermination(s, epochTargetReward, networkQAPower, sectorPower)
```
anorth, line 1845: This one is going to end up being more complicated, including reference to the sector's stored initial pledge.

actors/builtin/miner/miner_actor.go
```diff
@@ -1783,19 +1803,53 @@ func validateFRDeclaration(deadlines *Deadlines, deadline *DeadlineInfo, declare
...
+	}
+	return totalFee, nil
+}
+
 // Returns the sum of the raw byte and quality-adjusted power for sectors.
-func powerForSectors(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo) (rawBytePower, qaPower big.Int) {
+func PowerForSectors(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo) (rawBytePower, qaPower big.Int) {
 	rawBytePower = big.Mul(big.NewIntUnsigned(uint64(sectorSize)), big.NewIntUnsigned(uint64(len(sectors))))
 	qaPower = big.Zero()
 	for _, s := range sectors {
```
anorth, line 1855: This is redundant with the added `qaPowerForSectors`. Re-arrange so that the computation is only present once.

actors/builtin/miner/miner_test.go
```diff
@@ -424,6 +425,51 @@ func TestProvingPeriodCron(t *testing.T) {
...
+		st = getState(rt)
+		rt.SetEpoch(deadline.NextPeriodStart())
+		info, found, err := st.GetSector(rt.AdtStore(), sectorNumber)
+		require.NoError(t, err)
+		require.True(t, found)
+
+		// Declare the sector as faulted
+		ss, err := info.Info.SealProof.SectorSize()
+		require.NoError(t, err)
+		_, sectorQAPower := miner.PowerForSectors(ss, []*miner.SectorOnChainInfo{info})
```
anorth, line 464: Use `miner.QAPowerForSector` for just one sector

actors/builtin/miner/miner_test.go
```diff
@@ -424,6 +425,51 @@ func TestProvingPeriodCron(t *testing.T) {
...
+		expectedReward := big.Exp(big.NewInt(10), big.NewInt(18)) // 1 FIL
```
anorth, line 465: `big.NewInt(1e18)`

actors/builtin/miner/miner_test.go
```diff
@@ -731,6 +777,14 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
 	rt.SetCaller(h.worker, builtin.AccountActorCodeID)
 	rt.ExpectValidateCallerAddr(h.worker)
 
+	reward := big.NewIntUnsigned(1e18)
```
anorth, line 780: Aside: we might move some of these "network" properties into the harness so they can easily be consistent between calls.

actors/builtin/miner/miner_test.go
```diff
@@ -781,6 +835,76 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
 	rt.Verify()
 }
 
+func (h *actorHarness) declareFaults(rt *mock.Runtime, expectedReward abi.TokenAmount, totalQAPower abi.StoragePower, fee abi.TokenAmount, faultSectorInfos ...*miner.SectorOnChainInfo) {
+	rt.SetCaller(h.worker, builtin.AccountActorCodeID)
+	rt.ExpectValidateCallerAddr(h.worker)
+
+	ss, err := faultSectorInfos[0].Info.SealProof.SectorSize()
```
anorth, line 842: This is needed frequently enough that we should just put the miner-constant proof type and sector size into the harness object to save the verbosity of all these lookups

actors/builtin/miner/miner_test.go
```diff
@@ -781,6 +835,76 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
...
+	require.NoError(h.t, err)
+	expectedRawDelta, expectedQADelta := miner.PowerForSectors(ss, faultSectorInfos)
+	expectedRawDelta = big.Mul(big.NewInt(-1), expectedRawDelta)
```
anorth, line 845: `expectedRawDelta.Neg()`

actors/builtin/miner/miner_test.go
```diff
@@ -781,6 +835,76 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
...
+		builtin.BurntFundsActorAddr,
+		builtin.MethodSend,
+		nil,
+		fee,
+		nil,
+		exitcode.Ok,
+	)
+
+	// expect pledge update
+	pledgeDelta := big.Mul(big.NewInt(-1), fee)
```
anorth, line 891: `fee.Neg()`

actors/builtin/miner/miner_test.go
```diff
@@ -848,6 +972,13 @@ func (h *actorHarness) onProvingPeriodCron(rt *mock.Runtime, expectedEnrollment
 		rt.ExpectGetRandomness(crypto.DomainSeparationTag_WindowedPoStDeadlineAssignment, randEpoch, nil, bytes.Repeat([]byte{0}, 32))
 	}
 	// Re-enrollment for next period.
+	reward := big.NewIntUnsigned(1e18)
+	rt.ExpectSend(builtin.RewardActorAddr, builtin.MethodsReward.LastPerEpochReward, nil, big.Zero(), &reward, exitcode.Ok)
+
+	pwrTotal := power.CurrentTotalPowerReturn{
+		QualityAdjPower: big.NewIntUnsigned(1 << 50),
+	}
+	rt.ExpectSend(builtin.StoragePowerActorAddr, builtin.MethodsPower.CurrentTotalPower, nil, big.Zero(), &pwrTotal, exitcode.Ok)
```
anorth, line 981: This whole pattern happens more than once, so break it out into a helper `expectQueryNetworkInfo(qaPower, epochReward)`.

actors/builtin/miner/miner_test.go
```diff
@@ -892,11 +1023,40 @@ func makeProveCommit(sectorNo abi.SectorNumber) *miner.ProveCommitSectorParams {
 	}
 }
 
+func makePowerClaimUpdate(rawDelta, qaDelta abi.StoragePower) *power.UpdateClaimedPowerParams {
```
anorth, line 1026: I'd rather just see this inline at the call site.

actors/builtin/miner/miner_test.go
```diff
@@ -892,11 +1023,40 @@ func makeProveCommit(sectorNo abi.SectorNumber) *miner.ProveCommitSectorParams {
...
+func makeFaultParamsFromFaultingSectors(t testing.TB, st *miner.State, store adt.Store, faultSectorInfos []*miner.SectorOnChainInfo) *miner.DeclareFaultsParams {
+	deadlines, err := st.LoadDeadlines(store)
+	require.NoError(t, err)
+	faultAtDeadline := make(map[uint64][]uint64)
+	for _, sectorInfo := range faultSectorInfos {
+		dl, err := miner.FindDeadline(deadlines, sectorInfo.Info.SectorNumber)
+		require.NoError(t, err)
+		if _, ok := faultAtDeadline[dl]; !ok {
+			faultAtDeadline[dl] = []uint64{uint64(sectorInfo.Info.SectorNumber)}
+		}
```
anorth, line 1042: This appears redundant to me. The line below should work anyway.

actors/builtin/miner/miner_test.go
```diff
@@ -892,11 +1023,40 @@ func makeProveCommit(sectorNo abi.SectorNumber) *miner.ProveCommitSectorParams {
...
+		RawByteDelta:         rawDelta,
+		QualityAdjustedDelta: qaDelta,
+	}
+}
+
+func makeFaultParamsFromFaultingSectors(t testing.TB, st *miner.State, store adt.Store, faultSectorInfos []*miner.SectorOnChainInfo) *miner.DeclareFaultsParams {
+	deadlines, err := st.LoadDeadlines(store)
+	require.NoError(t, err)
+	faultAtDeadline := make(map[uint64][]uint64)
+	for _, sectorInfo := range faultSectorInfos {
```
anorth, line 1037: Please add an explanatory comment, like "find the deadline for each faulty sector, which must be provided with the fault declaration".

actors/builtin/miner/monies.go
```diff
@@ -3,15 +3,59 @@ package miner
...
+// FF(t) = DeclaredFaultFactor * BR(t)
+func PledgePenaltyForSectorDeclaredFault(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
+	return big.Div(
+		big.Mul(DeclaredFaultFactorNum, ExpectedDayRewardForSector(epochTargetReward, networkQAPower, qaSectorPower)),
+		DeclaredFaultFactorDenom)
+}
+
+// This is the SP(t) penalty for a newly faulty sector that has not been declared.
+// SP(t) = UndeclaredFaultFactor * BR(t)
+func pledgePenaltyForSectorUndeclaredFault(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
```
anorth, line 40: Export this too

actors/builtin/miner/monies.go
```diff
@@ -3,15 +3,59 @@ package miner
...
+}
+
 // Computes the pledge requirement for committing new quality-adjusted power to the network, given the current
 // total power, total pledge commitment, epoch block reward, and circulating token supply.
 // In plain language, the pledge requirement is a multiple of the block reward expected to be earned by the
 // newly-committed power, holding the per-epoch block reward constant (though in reality it will change over time).
 // The network total pledge and circulating supply parameters are currently unused, but may be included in a
 // future calculation.
-func InitialPledgeForPower(newQAPower abi.StoragePower, networkQAPower abi.StoragePower, networkTotalPledge abi.TokenAmount, epochTargetReward abi.TokenAmount, networkCirculatingSupply abi.TokenAmount) abi.TokenAmount {
+func InitialPledgeForPower(qaSectorPower abi.StoragePower, networkQAPower abi.StoragePower, networkTotalPledge abi.TokenAmount, epochTargetReward abi.TokenAmount, networkCirculatingSupply abi.TokenAmount) abi.TokenAmount {
```
anorth, line 58: This function is explicitly ignorant of where the power comes from, so I don't think adding `sector` to the name helps.

actors/builtin/miner/monies.go
```diff
@@ -3,15 +3,59 @@ package miner
...
+// FF = (DeclaredFaultFactorNum / DeclaredFaultFactorDenom) * BR(t)
+var DeclaredFaultFactorDenom = big.NewInt(100)
+
+// SP = UndeclaredFaultFactor * BR(t)
+var UndeclaredFaultFactor = big.NewInt(5)
+
+// This is the BR(t) value of the given sector for the current epoch.
+// It is the expected reward this sector would pay out over a one day period.
+// BR(t) = CurrEpochReward(t) * SectorQualityAdjustedPower * EpochsInDay / TotalNetworkQualityAdjustedPower(t)
+func ExpectedDayRewardForSector(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
```
anorth, line 24: This function isn't about sectors, it's just about power. Rename to something like `ExpectedDayRewardForPower`.

actors/builtin/miner/monies.go
```diff
@@ -3,15 +3,59 @@ package miner
...
+	expectedRewardForProvingPeriod := big.Mul(big.NewInt(builtin.EpochsInDay), epochTargetReward)
+	return big.Div(big.Mul(qaSectorPower, expectedRewardForProvingPeriod), networkQAPower)
+}
+
+// This is the FF(t) penalty for a sector expected to be in the fault state either because the fault was declared or because
+// it has been previously detected by the network.
+// FF(t) = DeclaredFaultFactor * BR(t)
+func PledgePenaltyForSectorDeclaredFault(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
```
anorth, line 32: Ah, none of these take info about a single sector! Rename to talk just about power instead, so that it's clear we could re-arrange the arithmetic to sum the power first and then compute fees.

actors/builtin/miner/miner_actor.go
ZenGround0: Yes we'll be sure to add that parameter in the next PR

ZenGround0:
@acruikshank fyi merging this, happy to talk over anything unclear + suggestions I left for later.

Issue #489, opened 2020-06-15
Move initial pledge calculation into miner actor


anorth:
The initial pledge calculation for a sector is currently performed by the power actor while processing the power update for a newly committed sector. It's here because it depends on the currently committed power and pledge collateral. However this gets in the way of efficient batch processing of sector commitments and complicates the interactions between the miner and power actor. If the pledge were instead computed in the miner actor:
- the `OnSectorProveCommit` method would disappear, redundant given the other power accounting methods
- the miner could call the reward actor just once per batch of new sectors, rather than once per sector
- the miner could call the runtime for circulating supply just once per batch
- both power and pledge accounting could happen just once per batch

Rather than compute the pledge, the power actor should implement a method returning the total QA power and total pledge collateral, for use in calculations external.

See also #419

Issue #424, opened 2020-06-03
Implement pre-commit deposit calculation and rollover to pledge


anorth:
The pre-commit deposit is currently a stub, always zero.

The deposit should be equal to the initial pledge for the sector (#468). This initial pledge value should be calculated and stored in the pre-commit state. The initial pledge does not need to be recomputed when the proof is verified. Instead, the precommit deposit should be rolled directly into the locked pledge funds.

Since the precommit deposit value is already stored in the state, the most straightforward thing might be to rename `precommit.PreCommitDeposit` to `precommit.InitialPledge`, and simply treat it as a deposit during the precommit phase.

anorth:
Consider doing #419 first.

acruikshank:
@anorth There might an issue computing initial pledge on precommit. `IP=20*BR(t)` and BR depends on QA weight of the sector. It's not clear QA weight is fully defined prior to prove commit. We only lock QA weight into the SectorOnChainInfo at prove commit. The market actor computes the weight and marks deals active at the same time, so we currently don't have a means to compute QA weight prior to prove commit.

The block reward and total storage power are also a function of time, so there really can be no guarantees about pre-commit deposit equaling initial pledge unless we define initial pledge to equal the pre-commit deposit.

If we are confident that deals must be published prior to precommit (so QA weight cannot change between pre- and prove- commit) we can add the ability to compute QA power without altering state.

Otherwise we can compute an upper bound on QA weight based on sector size, but this will make precommit deposit >= initial pledge.

Thoughts?

acruikshank:
Note that the above assumes #468 the existing initial pledge calculation can be implemented this way easily.

acruikshank:
A conversation with cryptoecon suggests that we'd rather compute the initial pledge at precommit time than deal with inconsistencies between precommit deposit and initial pledge. Since initial pledge correlates to activation time of the sector, we'll use the longest possible duration between precommit and commit to perform the calculation to compute the worst case value for pledge.

Plan approved by @zixuanzh:

> My preference would be to
1. Split VerifyDealsOnSectorProveCommit into two functions. One that computes deal weight based on a minimum activation epoch and another that actually activates the deals at prove commit.
2. Compute deal weight at precommit, use it to compute initial pledge with assuming a worst case activation time. Use this initial pledge as the precommit deposit.
3. On prove commit, transfer precommit deposit to lock funds as the initial pledge. Activate all deals.

anorth:
[...]

PR #499, opened 2020-06-17
Restructure weight and power calculations to be callable from outside

### LINKED ISSUE #420: Simplify initial pledge collateral calculations from state
At present, it's really awkward for miners to compute the pledge collateral requirement for a sector they're about to commit. This is because the calculation code is spread through the miner/power actors and not available as free functions or state methods. Implementations must work through hijinks like simulating a sector prove-commit call to the power actor in order to get an indicative value.

Rework this code to make the pledge calculation a relatively simple and exported function of power actor state. Probably do #419 and #489 first. Consider whether computing the pledge in the miner actor, rather than power actor, would simplify things.

See also completing these calculations: #468

anorth:
Miner operators need to compute the initial pledge for a sector before attempting to (prove-) commit it. The pledge calculation logic was previously buried inside power actor methods and so Lotus goes through some [gymnastics involving simulating side-effecting actor messages](https://github.com/filecoin-project/lotus/blob/741825a04ee11b5ef68f32a7aee3608fca99911e/node/impl/full/state.go#L685) to the market and power actor in order to get at the expected pledge amount.

#489 moved the pledge amount calculation from the power actor to the miner, so that method won't work any more. This PR rearranges the code to export some free functions with which a miner operator can make the calculations directly (after plucking some values like the LastEpochBlockReward and TotalPower from the state). This is an API-breaking change for nodes, but with the intention to make a much easier API.

After this, instead of simulating the messages, Lotus should:
- Load the reward actor and power actor state to get the block reward and power constants
- Call `market.ValidateDealsForActivation()` to get the deal weight for deals in the precommit
- Call `miner.QAPowerForWeight()` with the weights to calculate the power the sector will have
- Call `miner.InitialPledgeForPower()` with the power and network values to calculate the pledge

Note that the pledge calculation is set to move from ProveCommitSector to PreCommitSector, where it will be the pre-commit deposit (#424) so the point at which the node needs to invoke this will shortly change too.

Closes #420

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -1,8 +1,9 @@
 package market
 
 import (
-	addr "github.com/filecoin-project/go-address"
+	"fmt"
 
+	addr "github.com/filecoin-project/go-address"
 	cbg "github.com/whyrusleeping/cbor-gen"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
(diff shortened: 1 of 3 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -556,22 +532,57 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
...
+	}
+	return totalDealSpaceTime, totalVerifiedSpaceTime, nil
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 // Checks
 ////////////////////////////////////////////////////////////////////////////////
 
-func validateDealCanActivate(rt Runtime, minerAddr addr.Address, sectorExpiration abi.ChainEpoch, proposal *DealProposal) {
+func validateDealCanActivate(proposal *DealProposal, minerAddr addr.Address, sectorExpiration abi.ChainEpoch, currEpoch abi.ChainEpoch) error {
```
acruikshank, line 575: Nit: sectorExpiration and currEpoch could be combined into `sectorExpiration, currEpoch abi.ChainEpoch`.

actors/builtin/market/policy.go
```diff
@@ -31,3 +32,11 @@ func dealClientCollateralBounds(pieceSize abi.PaddedPieceSize, duration abi.Chai
 func collateralPenaltyForDealActivationMissed(providerCollateral abi.TokenAmount) abi.TokenAmount {
 	return providerCollateral // PARAM_FINISH
 }
+
+// Computes the weight for a deal proposal, which is a function of its size and duration.
+func DealWeight(proposal *DealProposal) abi.DealWeight {
```
acruikshank, line 37: ❤️

Issue #266, opened 2020-04-02
Fix calculation of total network power in small networks


anorth:
#236 and #238 changed the value of the power actor's TotalNetworkPower to be only the sum of power of those actors that individually meet the consensus power minimum (b/c that's the denominator desired for elections, etc). This breaks calculations when no miner is above the minimum and hence total power is zero.

By chance, things are working at the moment because both implementations write non-zero total power into the power actor state in genesis even when no miner meets the minimum, and both also fail to validate that block miners have power exceeding this minimum. See `minerNominalPowerMeetsConsensusMinimum` for how that was intended to work.

Two possible solution paths:
- track both the total power of above-threshold actors and the total sum of claims, and use the "right" one as denominator depending on whether some small number of miners have crossed the threshold
- detect the threshold crossing on-chain and recalculate the total network power when that threshold is crossed, excluding the below-threshold power

anorth:
Consider also tracking the count of all miners and miners above the threshold separately.

PR #500, opened 2020-06-17
Add basic tests for power actor accounting

anorth:


actors/builtin/power/power_test.go
```diff
@@ -67,6 +68,78 @@ func TestConstruction(t *testing.T) {
...
+		require.Equal(t, claim1.QualityAdjPower, mul(powerUnit, 2))
+
+		claim2, found, err := st.GetClaim(rt.AdtStore(), miner2)
+		require.NoError(t, err)
+		require.True(t, found)
+		require.Equal(t, claim2.RawBytePower, powerUnit)
+		require.Equal(t, claim2.QualityAdjPower, powerUnit)
+
+		// Subtract power and some pledge for miner2
+		actor.updateClaimedPower(rt, miner2, powerUnit.Neg(), powerUnit.Neg())
```
acruikshank, line 134: This test is missing an assertion that miner2's Claim should have zero raw and quality adjusted power after the second update.

anorth:
Yes, lots missing around `ConsensusMinerMinPower`  but that's because the logic isn't done yet: #266.

PR #501, opened 2020-06-17
track pending deal proposals to ensure deals can't be duplicately published

### LINKED ISSUE #364: PublishStorageDeals on Storage Market Actor should reject deals that are already published
If a client sends a deal proposal to a miner, and signs the deal proposal with their signature, then sends the data for the piece to the miner, what prevents the miner from submitting the proposal to the chain lots of times, storing it, and getting paid multiple times for it? (other than the total escrow a client has in the SMA)

I believe answer to that question at the moment is nothing.

In thinking through possible solutions to the problem, there are two issues that I believe need to be solved:
1. Deal proposals are not unique as long as they are for the same pieceCID, start epoch, end epoch etc
2. Even if we added some unique identifier to them I'm not clear how you could query the existing deals to see if a proposal was unique in a way that is performant.

Maybe there is a different approach I am not thinking of?

whyrusleeping:
I think that this works, but it has one caveat that i think is okay.

If a miner publishes a deal, seals the sector, publishes the prove commit, and terminates the sector *all before the deal start date* the miner could then resubmit the deal. I think that this is fine because the miner isnt paid for the deal until the deal start date anyways, so theres no way to get any extra money out of this.

Closes #364

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -5,6 +5,7 @@ import (
 
 	addr "github.com/filecoin-project/go-address"
 	cbg "github.com/whyrusleeping/cbor-gen"
+	xerrors "golang.org/x/xerrors"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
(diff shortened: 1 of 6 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -467,7 +529,7 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 					// Not yet appeared in proven sector; check for timeout.
 					AssertMsg(rt.CurrEpoch() >= deal.StartEpoch, "if sector start is not set, we must be in a timed out state")
 
-					slashed := st.processDealInitTimedOut(rt, et, lt, dealID, deal, state)
+					slashed := st.processDealInitTimedOut(rt, et, lt, pending, dealID, deal, state)
```
anorth, line 532: In combination with my comment above, we need to delete from pending regardless of whether is has a sector start epoch.

actors/builtin/market/market_actor.go
```diff
@@ -308,6 +356,10 @@ func (a Actor) VerifyDealsOnSectorProveCommit(rt Runtime, params *VerifyDealsOnS
 				rt.Abortf(exitcode.ErrIllegalState, "set deal %v", err)
 			}
 
+			if err := pending.Delete(adt.CidKey(propc)); err != nil {
+				rt.Abortf(exitcode.ErrIllegalState, "failed to delete pending proposal: %v", err)
+			}
```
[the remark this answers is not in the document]
whyrusleeping: I left the bit that checks that the pending proposal exists, can remove if you prefer but it feels like a decent sanity check.

go.mod
```diff
@@ -6,14 +6,13 @@ require (
 	github.com/filecoin-project/go-address v0.0.2-0.20200218010043-eb9bb40ed5be
 	github.com/filecoin-project/go-amt-ipld/v2 v2.0.1-0.20200424220931-6263827e49f2
 	github.com/filecoin-project/go-bitfield v0.0.1
+	github.com/filecoin-project/go-cbor-util v0.0.0-20191219014500-08c40a1e63a2
```
anorth, line 9: Please remove this now it's no longer used.

PR #512, opened 2020-06-22
Denormalize sector precommit info into sector on-chain info

anorth:
Previously, the `SectorOnChainInfo` (which lives in chain state for the life of the sector) had the `SectorPreCommitInfo` embedded in it, I think primarily for expediency in initial implementation. Bytes of `SectorOnChainInfo` are expensive, because we expect 10s-100s of million of them for exabyte-scale network. The pre-commit info contains the `SealRandEpoch`, which is not relevant after the sector is proven, wasting a few (hundred mega-)bytes. 

I am about to add another pre-commit-only field, so decided to clean this up first.

This changes inlines the relevant sector pre-commit information into SectorOnChainInfo, dropping the `SealRandEpoch` and saving another byte on the CBOR nested struct delimiter.

(Aside: it should be possible to remove the SectorNumber from this struct too, but require some awkward data flow in one place that we use it, and is useful for manual state inspection).

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -118,24 +118,31 @@ type WorkerKeyChange struct {
 	EffectiveAt abi.ChainEpoch
 }
 
+// Information provided by a miner when pre-committing a sector.
 type SectorPreCommitInfo struct {
 	SealProof     abi.RegisteredSealProof
 	SectorNumber  abi.SectorNumber
 	SealedCID     cid.Cid // CommR
 	SealRandEpoch abi.ChainEpoch
 	DealIDs       []abi.DealID
-	Expiration    abi.ChainEpoch // Sector Expiration
+	Expiration    abi.ChainEpoch
 }
 
+// Information stored on-chain for a pre-committed sector.
 type SectorPreCommitOnChainInfo struct {
 	Info             SectorPreCommitInfo
 	PreCommitDeposit abi.TokenAmount
 	PreCommitEpoch   abi.ChainEpoch
 }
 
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -118,24 +118,31 @@ type WorkerKeyChange struct {
...
+// Information stored on-chain for a proven sector.
 type SectorOnChainInfo struct {
-	Info               SectorPreCommitInfo
-	ActivationEpoch    abi.ChainEpoch // Epoch at which SectorProveCommit is accepted
+	SectorNumber       abi.SectorNumber
+	SealProof          abi.RegisteredSealProof // The seal proof type implies the PoSt proof/s
+	SealedCID          cid.Cid                 // CommR
+	DealIDs            []abi.DealID
+	Activation         abi.ChainEpoch // Epoch during which the sector proof was accepted
+	Expiration         abi.ChainEpoch // Epoch during which the sector expires
```
Stebalien, line 145: nit: You may want to consider putting these fields into an embedded struct so we can share a common datastructure (easier to add common fields). But I'm not sure if it's worth it in this case.

anorth: Thanks. I'm not going to in this case because (a) it cost a (hundred-mega) byte when encoded, and (b) I fear will encourage unwanted re-used of that structure which led to this kind of thing in the first place.

Stebalien: Hm. I guess our CBOR library doesn't act like `encoding/json` (`encoding/json` will inline embedded structs).

PR #514, opened 2020-06-22
Implement replacement of committed-capacity sector by a new one

### LINKED ISSUE #485: Implement committed capacity sector upgrade with new sector
We wish to allow a miner to replace a "committed capacity" sector (with no deals) with a sector that does have deals, without penalty for terminating the CC sector early. An [earlier attempt](https://github.com/filecoin-project/specs-actors/pull/398) structured this as upgrading the sector in-place, re-using the sector ID, but this introduced a bunch of complications.

It would be simpler to actually just use another sector. We can treat the "upgrade" as a license to terminate an empty sector before due without paying any penalties. This results in

- only one sector lifecycle flow, with no cycles
- static correspondence between sector id and CommR/CommP, deals, and on-disk data

In implementation, this amounts to attaching the ID of a CC sector to terminate when a new deal-ful sector is PreCommitted, and then actually terminating that sector as an on-time (no penalty) expiry when the new PoRep is verified.

Constraints etc:
- the old sector must have no deals
- the new sector must have at least one deal
- both sectors must be the same size (trivially true in present implementation)
- the new sector's expiry must not be sooner than the old sector's expiry
- the miner must post a new initial pledge for the new sector, based on it's (higher) power
- the pledge amount _recorded_ for the new sector should be max(old sector pledge, new sector pledge)

anorth:
Adds a `ReplaceSector` parameter to sector pre-commit, allowing a sector with deals to replace a committed-capacity sector. There are some restrictions on the replacing/replacement pair to ensure total pledge is not reduced. Termination of the replaced sector is deferred until the end of the proving period (not yet implemented).

This draft is FYI @zixuanzh @magik6k. I won't seek real review until writing a bunch of tests.

- [x] Rebase after #512 lands

Closes #485.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -325,6 +325,9 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		// when it arrives.
 		rt.Abortf(exitcode.ErrIllegalArgument, "seal challenge epoch %v too old, must be after %v", params.SealRandEpoch, challengeEarliest)
 	}
+	if params.ReplaceCapacity && len(params.DealIDs) == 0 {
+		rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector without committing deals")
+	}
 
 	// gather information from other actors
 	epochReward := requestCurrentEpochBlockReward(rt)
@@ -337,23 +340,45 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	newlyVestedAmount := rt.State().Transaction(&st, func() interface{} {
 		rt.ValidateImmediateCallerIs(st.Info.Worker)
 		if params.SealProof != st.Info.SealProofType {
-			rt.Abortf(exitcode.ErrIllegalArgument, "wrong proof type")
+			rt.Abortf(exitcode.ErrIllegalArgument, "sector seal proof %v must match miner seal proof type %d", params.SealProof, st.Info.SealProofType)
 		}
 
-		if _, found, err := st.GetPrecommittedSector(store, params.SectorNumber); err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to check precommit %v: %v", params.SectorNumber, err)
-		} else if found {
-			// Sector is currently precommitted but still not proven.
-			rt.Abortf(exitcode.ErrIllegalArgument, "sector %v already precommitted", params.SectorNumber)
+		_, preCommitFound, err := st.GetPrecommittedSector(store, params.SectorNumber)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to check pre-commit %v", params.SectorNumber)
+		if preCommitFound {
+			rt.Abortf(exitcode.ErrIllegalArgument, "sector %v already pre-committed", params.SectorNumber)
 		}
 
-		if found, err := st.HasSectorNo(store, params.SectorNumber); err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to check sector %v: %v", params.SectorNumber, err)
-		} else if found {
+		sectorFound, err := st.HasSectorNo(store, params.SectorNumber)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to check sector %v", params.SectorNumber)
+		if sectorFound {
 			rt.Abortf(exitcode.ErrIllegalArgument, "sector %v already committed", params.SectorNumber)
 		}
 
 		validateExpiration(rt, &st, rt.CurrEpoch(), params.Expiration, params.SealProof)
+
+		if params.ReplaceCapacity {
+			replaceSector, found, err := st.GetSector(store, params.ReplaceSector)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sector %v", params.SectorNumber)
+			if !found {
+				rt.Abortf(exitcode.ErrNotFound, "no such sector %v to replace", params.ReplaceSector)
+			}
+			if len(replaceSector.DealIDs) > 0 {
+				rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector %v which has deals", params.ReplaceSector)
+			}
+			// The sectors must be the same size, but that is always true because each miner actor supports
+			// only one sector size.
+			if params.Expiration < replaceSector.Expiration {
+				rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector %v expiration %v with sooner expiration %v",
+					params.ReplaceSector, replaceSector.Expiration, params.Expiration)
+			}
+			faulty, err := st.Faults.IsSet(uint64(params.ReplaceSector))
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to check fault for sector %v", params.ReplaceSector)
+			if faulty {
+				// Note: the sector might still fault later, in which case we should not license its termination.
+				rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace faulty sector %v", params.ReplaceSector)
+			}
+		}
 
 		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -126,6 +126,7 @@ type SectorPreCommitInfo struct {
 	SealRandEpoch abi.ChainEpoch
 	DealIDs       []abi.DealID
 	Expiration    abi.ChainEpoch
+	ReplaceSector abi.SectorNumber // "Committed capacity" no-deal sector to replace, valid only with non-empty deals
```
magik6k, line 129: `0` is a valid `SectorNumber`, this should probably be a pointer

anorth: Elsewhere in this PR I have defined `0` to not be a valid sector number. I really do not want to add nil-pointer possibilities here for want of a sentinel value. Can we make `0` invalid? Alternatively, we could make it signed so I can use `-1`, but that would be error prone since the default value will be to replace a sector.

anorth: If you feel really strongly about `0` being a valid sector number, I will add a boolean field here in addition, rather than use a pointer.

actors/builtin/miner/miner_actor.go
```diff
@@ -332,33 +335,61 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
+			if len(replaceSector.DealIDs) > 0 {
+				rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector %v which has deals", params.ReplaceSector)
+			}
+			// The sectors must be the same size, but that is always true because each miner actor supports
+			// only one sector size.
+			if params.Expiration < replaceSector.Expiration {
+				rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector %v expiration %v with sooner expiration %v",
+					params.ReplaceSector, replaceSector.Expiration, params.Expiration)
+			}
+			faulty, err := st.Faults.IsSet(uint64(params.ReplaceSector))
```
nicola, line 377: This check should ensure that we are not in the two cases: the sector is faulty or that the sector is active but declared faulted.

actors/builtin/miner/miner_actor.go
```diff
@@ -508,6 +541,24 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			VerifiedDealWeight: dealWeights.VerifiedDealWeight,
 		}
 
+		replaceSector := precommit.Info.ReplaceSector
+		if replaceSector != 0 {
+			info, exists, err := st.GetSector(store, replaceSector)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to check sector %v", replaceSector)
+			faulty, err := st.Faults.IsSet(uint64(replaceSector))
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to check sector %v fault state", replaceSector)
```
nicola, line 549: We should check this at ProveCommit to avoid performing verification of proofs for ProveCommits that are not valid.

actors/builtin/miner/miner_actor.go
```diff
@@ -562,6 +613,31 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 
 		notifyPledgeChanged(rt, big.Sub(initialPledge, newlyVestedAmount))
 	}
+
+	// This transaction should replace the one inside the loop above.
+	// Migrate state mutations into here.
+	rt.State().Transaction(&st, func() interface{} {
+		// Mark replaced sectors for on-time expiration at the end of the current proving period.
+		// It's can't be removed right now because it may yet be challenged for Window PoSt in this period,
+		// and the deadline assignments can't be changed mid-period.
```
nicola, line 622: If it will be challenged, shouldn't the new one be challenged instead?

E.g. we update the previous as "replaced" and when it's challenged the new sector should be challenged.

It's ok if this is not the case, we will run in a setting where both sectors can be challenged (fine from theory), just trying to understand.

actors/builtin/miner/miner_actor.go
anorth: These are the same. However a sector ends up in a fault state, it's set here.

actors/builtin/miner/miner_actor.go
anorth: The new sector proof is valid even if the replaced sector is faulty. We just _won't remove the replaced sector_ if it's faulty.

actors/builtin/miner/miner_actor.go
anorth: The new one won't be challenged until it gets assigned to a deadline at PP end. it's not "replaced", just "license to terminate early".

actors/builtin/miner/miner_actor.go
```diff
@@ -564,6 +612,41 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 
 		notifyPledgeChanged(rt, big.Sub(initialPledge, newlyVestedAmount))
 	}
+
+	// This transaction should replace the one inside the loop above.
+	// Migrate state mutations into here.
+	rt.State().Transaction(&st, func() interface{} {
+		// Mark replaced sectors for on-time expiration at the end of the current proving period.
```
acruikshank, line 619: consider extracting this as a function.

actors/builtin/miner/miner_test.go
```diff
@@ -274,15 +268,77 @@ func TestCommitments(t *testing.T) {
 		// TODO: test insufficient funds when the precommit deposit is set above zero
 	})
 
+	t.Run("validate CC upgrade", func(t *testing.T) {
```
acruikshank, line 271: What does CC stand for? Maybe better to spell it out here?

actors/builtin/miner/miner_test.go
```diff
@@ -274,15 +268,77 @@ func TestCommitments(t *testing.T) {
...
+
+		//deadline := actor.deadline(rt)
+
+		// Demonstrate good upgrade request is accepted.
+		goodUpgrade := makePreCommit(111, challengeEpoch, oldSectors[0].Expiration, []abi.DealID{1})
+		goodUpgrade.ReplaceCapacity = true
+		goodUpgrade.ReplaceSector = oldSectors[0].SectorNumber
+		actor.preCommitSector(rt, goodUpgrade, networkPower)
+
+		//upgrade := makePreCommit(112, challengeEpoch, oldSectors[1].Expiration, []abi.DealID{2})
```
acruikshank, line 330: Remove, skip or finish this test.

actors/builtin/miner/miner_test.go
```diff
@@ -885,6 +938,34 @@ func (h *actorHarness) constructAndVerify(rt *mock.Runtime) {
...
+	return pc
+}
+
+func (h *actorHarness) getSector(rt *mock.Runtime, sno abi.SectorNumber) *miner.SectorOnChainInfo {
+	st := getState(rt)
+	sector, found, err := st.GetSector(rt.AdtStore(), sno)
+	require.NoError(h.t, err)
+	require.True(h.t, found)
+	return sector
+}
```
acruikshank, line 963: ❤️

actors/builtin/miner/miner_actor.go
```diff
@@ -337,30 +340,55 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
+		if params.ReplaceCapacity {
+			replaceSector, found, err := st.GetSector(store, params.ReplaceSector)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sector %v", params.SectorNumber)
+			if !found {
+				rt.Abortf(exitcode.ErrNotFound, "no such sector %v to replace", params.ReplaceSector)
+			}
+			if len(replaceSector.DealIDs) > 0 {
+				rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector %v which has deals", params.ReplaceSector)
+			}
+			// The sectors must be the same size, but that is always true because each miner actor supports
```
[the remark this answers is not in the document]
anorth: I've replaced the comment with code asserting that the seal proof types match.

actors/builtin/miner/miner_actor.go
```diff
@@ -337,30 +340,55 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
+		sectorFound, err := st.HasSectorNo(store, params.SectorNumber)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to check sector %v", params.SectorNumber)
+		if sectorFound {
 			rt.Abortf(exitcode.ErrIllegalArgument, "sector %v already committed", params.SectorNumber)
 		}
 
 		validateExpiration(rt, &st, rt.CurrEpoch(), params.Expiration, params.SealProof)
 
+		if params.ReplaceCapacity {
+			replaceSector, found, err := st.GetSector(store, params.ReplaceSector)
```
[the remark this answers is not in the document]
anorth: Done

actors/builtin/miner/miner_actor.go
anorth: Done

PR #516, opened 2020-06-22
compute initial pledge for precommit deposit

### LINKED ISSUE #424: Implement pre-commit deposit calculation and rollover to pledge

acruikshank:
closes #424

### Motivation

We need to collect a deposit from miners precommitting sectors to be return when the sectors are committed. It has been decided that this deposit should be equivalent to the initial pledge that is locked when the sector is proven.

Tying precommit deposit to initial pledge has a few consequences. Initial pledge is a function of time, so we've decided to move the activation time of the sector to precommit time. This means we can compute the precommit deposit as `IP(currEpoch)` and then simply transfer this amount to lock funds to cover the initial pledge. This will have a small (but predictable) effect on the values of initial pledge and quality adjusted power, but shouldn't affect the behavior of the system beyond that. Power will still be granted and deals will still be activated when the sector is proven.

### Proposed Changes

1. Previously we had been both computing the initial pledge and activating deals in `ConfirmSectorProofsValid` so we had a `VerifyDealsOnProveCommit` method on the market actor that activated deals and returned the sum of their weight. Since we now actually compute initial pledge in precommit, this method has been split into a `VerifyDealsForActivation` that verifies the deals and computes their weights, and an `ActivateDeals` method that simply updates the activation state for a set of deals.
2. Compute precommit deposit in `miner.PreCommitSector` by computing `initialPledge` with parameters derived from the current epoch. Set this value to `SectorPreCommitOnChainInfo` and add it to `state.PreCommitDeposits`. 
3. Set intermediate values of deposit calculation, `DealWeight` and `VerifiedDealWeight`, into new fields in the `SectorPreCommitOnChainInfo` so they can be transferred to `SectorOnChainInfo` rather than being recomputed at a different type (resulting in slightly different values).
4. Add new "valid precommit then provecommit" test to assert precommit and prove commit make expected modifications to state on the happy path.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -26,10 +26,11 @@ func (a Actor) Exports() []interface{} {
 		2:                         a.AddBalance,
 		3:                         a.WithdrawBalance,
 		4:                         a.PublishStorageDeals,
-		5:                         a.VerifyDealsOnSectorProveCommit,
-		6:                         a.OnMinerSectorsTerminate,
-		7:                         a.ComputeDataCommitment,
-		8:                         a.CronTick,
+		5:                         a.VerifyDeals,
+		6:                         a.ActivateDealsOnSectorProveCommit,
+		7:                         a.OnMinerSectorsTerminate,
+		8:                         a.ComputeDataCommitment,
+		9:                         a.CronTick,
 	}
 }
 
@@ -248,20 +249,45 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 	return &PublishStorageDealsReturn{newDealIds}
 }
 
-type VerifyDealsOnSectorProveCommitParams struct {
+type VerifyDealsParams struct {
 	DealIDs      []abi.DealID
 	SectorExpiry abi.ChainEpoch
+	SectorStart  abi.ChainEpoch
 }
 
-type VerifyDealsOnSectorProveCommitReturn struct {
+type VerifyDealsReturn struct {
 	DealWeight         abi.DealWeight
 	VerifiedDealWeight abi.DealWeight
 }
 
-// Verify that a given set of storage deals is valid for a sector currently being ProveCommitted,
-// update the market's internal state accordingly, and return DealWeight of the set of storage deals given.
+// Verify that a given set of storage deals is valid for a sector currently being PreCommitted
+// and return DealWeight of the set of storage deals given.
 // The weight is defined as the sum, over all deals in the set, of the product of deal size and duration.
-func (a Actor) VerifyDealsOnSectorProveCommit(rt Runtime, params *VerifyDealsOnSectorProveCommitParams) *VerifyDealsOnSectorProveCommitReturn {
+func (A Actor) VerifyDeals(rt Runtime, params *VerifyDealsParams) *VerifyDealsReturn {
+	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
+	minerAddr := rt.Message().Caller()
+
+	var st State
+	rt.State().Readonly(&st)
+	store := adt.AsStore(rt)
+
+	dealWeight, verifiedWeight, err := ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, params.SectorStart)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to validate proposals for activation")
+
+	return &VerifyDealsReturn{
+		DealWeight:         dealWeight,
+		VerifiedDealWeight: verifiedWeight,
+	}
+}
+
+type ActivateDealsOnSectorProveCommitParams struct {
+	DealIDs      []abi.DealID
+	SectorExpiry abi.ChainEpoch
+}
+
+// Verify that a given set of storage deals is valid for a sector currently being ProveCommitted,
+// update the market's internal state accordingly.
+func (a Actor) ActivateDealsOnSectorProveCommit(rt Runtime, params *ActivateDealsOnSectorProveCommitParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 	minerAddr := rt.Message().Caller()
 	currEpoch := rt.CurrEpoch()
@@ -270,14 +296,9 @@ func (a Actor) VerifyDealsOnSectorProveCommit(rt Runtime, params *VerifyDealsOnS
 	store := adt.AsStore(rt)
 
 	// Update deal states.
-	dealWeight := big.Zero()
-	verifiedWeight := big.Zero()
-	var err error
 	rt.State().Transaction(&st, func() interface{} {
-		dealWeight, verifiedWeight, err = ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, currEpoch)
-		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to validate proposals for activation: %v", err)
-		}
+		_, _, err := ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, currEpoch)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to validate proposals for activation")
 
 		states, err := AsDealStateArray(store, st.States)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deal states")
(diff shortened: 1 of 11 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -609,13 +610,42 @@ func (h *actorHarness) controlAddresses(rt *mock.Runtime) (owner, worker addr.Ad
 	return ret.Owner, ret.Worker
 }
 
-func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.SectorPreCommitInfo, pledgeDelta abi.TokenAmount) {
+func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.SectorPreCommitInfo, networkPower abi.StoragePower, pledgeDelta abi.TokenAmount) {
+	epochReward := big.Mul(big.NewIntUnsigned(100), big.NewIntUnsigned(1e18))
+	networkPledge := big.Mul(epochReward, big.NewIntUnsigned(1000))
```
anorth, line 615: Move these into harness variables (I think some have have been already)

actors/builtin/miner/miner_test.go
```diff
@@ -609,13 +610,42 @@ func (h *actorHarness) controlAddresses(rt *mock.Runtime) (owner, worker addr.Ad
...
+
 	rt.SetCaller(h.worker, builtin.AccountActorCodeID)
 	rt.ExpectValidateCallerAddr(h.worker)
-	if !pledgeDelta.IsZero() {
-		rt.ExpectSend(builtin.StoragePowerActorAddr, builtin.MethodsPower.UpdatePledgeTotal, &pledgeDelta, big.Zero(), nil, exitcode.Ok)
+
+	// Prepare for and receive call to ConfirmSectorProofsValid at the end of the same epoch.
```
anorth, line 620: I think this comment was moved, is now out of place

acruikshank: deleted

actors/builtin/miner/miner_actor.go
```diff
@@ -359,7 +368,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
-		depositReq := precommitDeposit(st.GetSectorSize(), params.Expiration-rt.CurrEpoch())
+		duration := params.Expiration - maxSectorStart
+
+		// maximum sector weigh is maximum deal weight assuming it's verified
+		qaSectorWeight := QAPowerForWeight(st.GetSectorSize(), duration, big.Zero(), dealWeight.VerifiedDealWeight)
```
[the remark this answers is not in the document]
acruikshank: Good catch. Thanks!

ZenGround0:
> we've decided to move the activation time of the sector to precommit time

It looks like this is inaccurate.  This PR keeps activation at provecommit time and moves deal weight calculation and verification to precommit time.

actors/builtin/market/market_actor.go
```diff
@@ -248,20 +249,45 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+	dealWeight, verifiedWeight, err := ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, params.SectorStart)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to validate proposals for activation")
+
+	return &VerifyDealsForActivationReturn{
+		DealWeight:         dealWeight,
+		VerifiedDealWeight: verifiedWeight,
+	}
+}
+
+type ActivateDeals struct {
```
anorth, line 283: `ActivateDealsParams`

actors/builtin/miner/miner_actor.go
```diff
@@ -359,7 +365,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
-		depositReq := precommitDeposit(st.GetSectorSize(), params.Expiration-rt.CurrEpoch())
+		duration := params.Expiration - rt.CurrEpoch()
+
+		// maximum sector weight is maximum deal weight assuming it's verified
```
anorth, line 370: Remove this comment

actors/builtin/market/market_actor.go
```diff
@@ -270,14 +296,9 @@ func (a Actor) VerifyDealsOnSectorProveCommit(rt Runtime, params *VerifyDealsOnS
...
-	dealWeight := big.Zero()
-	verifiedWeight := big.Zero()
-	var err error
 	rt.State().Transaction(&st, func() interface{} {
-		dealWeight, verifiedWeight, err = ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, currEpoch)
-		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to validate proposals for activation: %v", err)
-		}
+		_, _, err := ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, currEpoch)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to validate proposals for activation")
```
anorth, line 301: I think we can probably remove this. Leave it for now, I'll file an issue to discuss.

actors/builtin/miner/miner_test.go
```diff
@@ -145,15 +145,87 @@ func TestCommitments(t *testing.T) {
...
+		assert.Equal(t, big.NewInt(int64(sectorSize/2)), onChainPrecommit.VerifiedDealWeight)
+
+		qaPower := miner.QAPowerForWeight(sectorSize, precommit.Expiration-precommitEpoch, onChainPrecommit.DealWeight, onChainPrecommit.VerifiedDealWeight)
+		expectedDeposit := miner.InitialPledgeForPower(qaPower, networkPower, actor.networkPledge, actor.epochReward, rt.TotalFilCircSupply())
+		assert.Equal(t, expectedDeposit, onChainPrecommit.PreCommitDeposit)
+
+		// expect total precommit deposit to equal our new deposit
+		assert.Equal(t, expectedDeposit, st.PreCommitDeposits)
+
+		// Sector pre-commitment missing.
```
anorth, line 192: This comment is confusing me

actors/builtin/miner/miner_test.go
```diff
@@ -590,6 +661,9 @@ func (h *actorHarness) constructAndVerify(rt *mock.Runtime, provingPeriodStart a
 		PeerId:        testPid,
 	}
 
+	h.epochReward = big.Mul(big.NewIntUnsigned(100), big.NewIntUnsigned(1e18))
+	h.networkPledge = big.Mul(h.epochReward, big.NewIntUnsigned(1000))
```
anorth, line 665: This seems redundant given the code in `newHarness`.

support/mock/mockrt.go
```diff
@@ -722,6 +722,9 @@ func (rt *Runtime) ExpectAbort(expected exitcode.ExitCode, f func()) {
 		}
 		// Roll back state change.
 		rt.state = prevState
+
+		// clear expectations unfulfilled due to abort
+		rt.Reset()
```
anorth, line 727: Hmm, I'm a bit torn about this. On the one hand, test code that's expecting an abort can set up just the right expectations for things that happen before the abort point and we should verify them.

On the other hand, a function that factors out some boilerplate might not know that it's expecting an abort, and so set up expectations for success. I'm guessing something like that motivated this here. We could do the reset for those cases in the caller, instead of here.

If it's possible to revert this without much churn, please try. But otherwise ok.

anorth:
@acruikshank you might have seen a change of mind from cryptoecon team. I propose that we land this now anyway, and file a follow-up issue to recalculate the weights and power at prove commit later.

support/mock/mockrt.go
acruikshank: I was also a little torn. You correctly identified the problem with the need to consolidate expectation setup in the harness to avoid too much boiler plate. The problem is that we have tests that make make multiple calls to mockrt where the earlier ones expect aborts. In this case, we never verify the expectations, but the expectations unmet due to the abort linger and alter the expectation sequence for the next call.

This scenario is not trivial to debug. I added it to the expect abort call to try to head off this difficulty in the future. If we were going to verify expectations after an abort, I can see how this would be even more confusing. 

I moved the reset up into the caller as you asked. We can work on the pattern for expected aborts from there.

PR #518, opened 2020-06-23
Quantize vesting with offset 

### LINKED ISSUE #301: Offset pledge vesting quantization boundaries
Miner actor pledge vesting is quantized to reduce the aggregate processing cost and state impact. The quanta is currently 1 day (for vesting periods measured in months).

The quantization boundaries for all miners are currently aligned with the chain epoch clock, which means that all vesting, for all miners, will happen in the same epoch.

We need to offset different miners' quantization to spread the load over the whole day, similar to how miners have different proving period boundaries. Maybe we can use the randomized PP boundary mod 1-day  to set the offset.

ZenGround0:
Use proving period start as entropy for setting offset in quantization boundaries for vesting.

Closes #301

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -1079,7 +1079,12 @@ func quantizeUp(e abi.ChainEpoch, unit abi.ChainEpoch, offsetSeed abi.ChainEpoch
 
 	remainder := (e - offset) % unit
 	quotient := (e - offset) / unit
+	// Don't round if epoch falls on a quantization epoch
 	if remainder == 0 {
+		return unit*quotient + offset
+	}
+	// Negative truncating division rounds up
+	if e-offset < 0 {
 		return unit*quotient + offset
 	}
 	return unit*(quotient+1) + offset
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -868,7 +868,7 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
 
 	vestedSoFar := big.Zero()
 	for e := vestBegin + spec.StepDuration; vestedSoFar.LessThan(vestingSum); e += spec.StepDuration {
-		vestEpoch := quantizeUp(e, spec.Quantization)
+		vestEpoch := quantizeUpWithOffset(e, spec.Quantization, st.ProvingPeriodStart)
 		elapsed := vestEpoch - vestBegin
 
 		targetVest := big.Zero() //nolint:ineffassign
@@ -1068,6 +1068,11 @@ func deleteMany(arr *adt.Array, keys []uint64) error {
 		}
 	}
 	return nil
+}
+
+func quantizeUpWithOffset(e abi.ChainEpoch, unit, offsetSeed abi.ChainEpoch) abi.ChainEpoch {
+	offset := offsetSeed % unit
+	return quantizeUp(e, unit) + offset
 }
 
 // Rounds e to the nearest exact multiple of the quantization unit, rounding up.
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -566,6 +566,49 @@ func TestExtendSectorExpiration(t *testing.T) {
...
+	workerKey := tutil.NewBLSAddr(t, 0)
+	actor := newHarness(t, tutil.NewIDAddr(t, 1000), owner, worker, workerKey)
+	periodOffset := abi.ChainEpoch(99)
+	builder := mock.NewBuilder(context.Background(), actor.receiver).
+		WithActorType(owner, builtin.AccountActorCodeID).
+		WithActorType(worker, builtin.AccountActorCodeID).
+		WithHasher(fixedHasher(uint64(periodOffset))).
+		WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID).
+		WithBalance(big.Mul(big.NewInt(1000), big.NewInt(1e18)), big.Zero())
+	t.Run("funds vested with offset", func(t *testing.T) {
```
acruikshank, line 581: ❤️

actors/builtin/miner/miner_test.go
```diff
@@ -566,6 +566,49 @@ func TestExtendSectorExpiration(t *testing.T) {
 	})
 }
 
+func TestAddLockedFund(t *testing.T) {
+	owner := tutil.NewIDAddr(t, 100)
+	worker := tutil.NewIDAddr(t, 101)
+	workerKey := tutil.NewBLSAddr(t, 0)
+	actor := newHarness(t, tutil.NewIDAddr(t, 1000), owner, worker, workerKey)
+	periodOffset := abi.ChainEpoch(99)
```
acruikshank, line 574: A better test would be to make this greater than the Quantization to ensure the periodOffset won't push the vesting schedule off by more than the quantization interval (it appears to do this but it's not exercised in the test). Better still would be to test it both ways.

ZenGround0: ok, to test this it sounds like you also want something like `assert.LessThan(t, newVestingFunds[0], AddFundsEpoch + PledgeVestingSpec.InitialDelay + PledgeVestingSpec.Quantization)` (first vesting epoch < one quantization interval away from initial delay)

actors/builtin/miner/miner_state.go
```diff
@@ -1068,6 +1068,11 @@ func deleteMany(arr *adt.Array, keys []uint64) error {
 	return nil
 }
 
+func quantizeUpWithOffset(e abi.ChainEpoch, unit, offsetSeed abi.ChainEpoch) abi.ChainEpoch {
+	offset := offsetSeed % unit
+	return quantizeUp(e, unit) + offset
```
anorth, line 1073: This is the only caller of `quantizeUp`. Please inline it here so that no-one else introduces the same kind of problem by using `quantizeUp` directly.

actors/builtin/miner/miner_test.go
```diff
@@ -566,6 +566,49 @@ func TestExtendSectorExpiration(t *testing.T) {
 	})
 }
 
+func TestAddLockedFund(t *testing.T) {
```
anorth, line 569: While we should have some test for the miner actor's `AddLockedFund`, I think you should be adding tests to `miner_state_test` instead, which already has a lot of tests for the vesting table. Please add to `TestVesting_AddLockedFunds_Table` instead, perhaps pulling the period boundary into the table entries.

anorth:
Ping if you want me to review the new tests, but good to go if you're confident after moving them.

ZenGround0:
In case anyone is wondering what's holding this up I've discovered through testing that the existing calculation is not quite right. It maps epoch values that are aligned with an offset quantization unit epoch to the next epoch when the offset != 0, giving different offsets different properties.  The `quantizeUp` function we want is `unit * ceil(epoch - normalizedOffset / unit) + offset`.  Still working out how to do this without resorting to float or bigints and only using mod as in the current implementation

anorth:
Thanks. It's probably worth testing the new `quantizeUp` directly in miner_internal_test, since it's tricky to get right.

PR #524, opened 2020-06-24
Multisig: Approve transaction if threshold already met

### LINKED ISSUE #71: MultiSig | what happens when threshold is reduced below the number of approvers an existing transaction already has

aarshkshah1992:
Closes #71

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -141,7 +141,7 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 
 	var txnID TxnID
 	var st State
-	var txn Transaction
+	var txn *Transaction
 	rt.State().Transaction(&st, func() interface{} {
 		if !isSigner(rt, &st, callerAddr) {
 			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", callerAddr)
@@ -149,7 +149,7 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 
 		txnID = st.NextTxnID
 		st.NextTxnID += 1
-		txn = Transaction{
+		txn = &Transaction{
 			To:       params.To,
 			Value:    params.Value,
 			Method:   params.Method,
@@ -193,7 +193,7 @@ func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *ApproveReturn {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	callerAddr := rt.Message().Caller()
 	var st State
-	var txn Transaction
+	var txn *Transaction
 	rt.State().Transaction(&st, func() interface{} {
 		if !isSigner(rt, &st, callerAddr) {
 			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", callerAddr)
@@ -372,7 +372,7 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
 	return nil
 }
 
-func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn Transaction) (bool, []byte, exitcode.ExitCode) {
+func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn *Transaction) (bool, []byte, exitcode.ExitCode) {
 	var st State
 	// abort duplicate approval
 	for _, previousApprover := range txn.Approved {
@@ -394,7 +394,7 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn Transaction) 
 	return executeTransactionIfApproved(rt,st, txnID, txn)
 }
 
-func (a Actor) getTransaction(rt vmr.Runtime, st State, txnID TxnID, proposalHash []byte, checkHash bool) Transaction {
+func (a Actor) getTransaction(rt vmr.Runtime, st State, txnID TxnID, proposalHash []byte, checkHash bool) *Transaction {
 	var txn Transaction
 
 	// get transaction from the state trie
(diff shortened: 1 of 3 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -357,42 +366,59 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
...
+		if err := st.putPendingTransaction(adt.AsStore(rt), txnID, txn); err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "failed to put transaction for approval: %v", err)
 		}
 		return nil
 	})
 
+	return executeTransactionIfApproved(rt,st, txnID, txn)
+}
+
+func (a Actor) getTransaction(rt vmr.Runtime, txnID TxnID, proposalHash []byte, checkHash bool) Transaction {
```
anorth, line 391: This factoring results in an additional state loading, of state that's already loaded before this. We want to minimise the number of state transactions or reads.

In `Approve `, move the call to this method back inside the state transaction that method already has, and pass the state here rather than re-load it. Similarly in `approveTransaction`. In fact, `Approve` ends up loading the same transaction twice because it calls `approveTransaction`.

Try to refactor so all the flows have the shape:
- in first state transaction, load all necessary data, call isSigner (which may write a newly-resolved address to state), mutate approvers
- outside of a state transaction check whether txn is ready to execute and execute if so
- iff it was executed, in a second state transaction, remove it from state

aarshkshah1992: @anorth What is the cost of a State transaction ? Does it have to load the entire state from disk ?

actors/builtin/multisig/multisig_test.go
```diff
@@ -637,6 +637,119 @@ func TestApprove(t *testing.T) {
...
+		rt.ExpectSend(chuck, fakeMethod, fakeParams, sendValue, nil, 0)
+		rt.SetBalance(sendValue)
+		rt.SetCaller(bob, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerType(builtin.AccountActorCodeID, builtin.MultisigActorCodeID)
+		actor.approveOK(rt, txnID, proposalHashData, nil)
+		rt.Verify()
+
+		// Transaction should be removed from actor state after send
+		actor.assertTransactions(rt)
+	})
```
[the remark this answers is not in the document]
aarshkshah1992: These are done.


I put the "fourth case" in it's own test for clarity.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: Also, I've finished this refactoring.

actors/builtin/multisig/multisig_test.go
```diff
@@ -637,6 +637,119 @@ func TestApprove(t *testing.T) {
...
+		rt.Verify()
+
+		// even if anne calls for an approval again(duplicate approval), transaction is executed because the threshold has been met.
+		proposalHashData := makeProposalHash(t, &multisig.Transaction{
+			To:       chuck,
+			Value:    sendValue,
+			Method:   fakeMethod,
+			Params:   fakeParams,
+			Approved: []addr.Address{anne},
+		})
```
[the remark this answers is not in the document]
anorth: Oh, I see that `checkHash` is hard-coded to `true` in that path. Ok.

In that case, move the verbosity out of this test body by returning the proposal hash from  `proposeOk`.

actors/builtin/multisig/multisig_test.go
```diff
@@ -637,6 +637,163 @@ func TestApprove(t *testing.T) {
...
+
+		actor.approveOK(rt, txnID, proposalHashData, nil)
+		rt.Verify()
```
anorth, line 756: Nit: this `Verify` should really appear immediately after the `rt.Call` inside `approve`. In a follow-up PR, could you please make sure that every `Call` is immediately verified, and then remove unnecessary verifies from the test body code.

actors/builtin/multisig/multisig_test.go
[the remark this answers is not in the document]
anorth: I would actually prefer they be combined to save the 20-lines of set-up code, which we will have to refactor if anything changes. More code is more friction.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -355,44 +364,57 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
...
+		if err := st.putPendingTransaction(adt.AsStore(rt), txnID, txn); err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "failed to put transaction for approval: %v", err)
 		}
 		return nil
 	})
 
+	return executeTransactionIfApproved(rt,st, txnID, txn)
+}
+
+func (a Actor) getTransaction(rt vmr.Runtime, st State, txnID TxnID, proposalHash []byte, checkHash bool) Transaction {
```
anorth, line 389: There's a bit much copying of `Transaction` structs now. Return a pointer (never nil) from here, and other places where a Transaction is returned or taken as parameter. Change `st.putPendingTransaction` to accept a pointer too.

actors/builtin/multisig/multisig_actor.go
anorth: We don't quite control what it does. Maybe the encoded block is cached. Maybe the decoded block is even cached too, but probably not. The cost is primarily en/decoding, and maybe disk access.

actors/builtin/multisig/multisig_test.go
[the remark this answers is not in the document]
aarshkshah1992: This is a good idea and has been done for all tests that this PR introduces.

actors/builtin/multisig/multisig_test.go
aarshkshah1992: Will do.

actors/builtin/multisig/multisig_test.go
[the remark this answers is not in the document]
aarshkshah1992: Makes sense. The fourth test has been rolled into the third test.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: This has been done.

PR #525, opened 2020-06-24
Multisig: Explicitly decrease threshold in RemoveSigner

### LINKED ISSUE #114: MultiSig automatic threshold decrease ignores input parameter
When `removeSigner` called with `decrease` parameter set as `false` and number of signers after removal is greater than threshold, the `decrease` parameter is ignored and threshold decreased implicitly. I think error ErrIllegalArgument should be raised.

https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go#L212

Test
https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_test.go#L687

aarshkshah1992:
Closes #114

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -292,7 +292,15 @@ func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.Emp
 				newSigners = append(newSigners, s)
 			}
 		}
-		if params.Decrease || uint64(len(st.Signers)-1) < st.NumApprovalsThreshold {
+
+		// if the number of signers is below the threshold after removing the given signer,
+		// we should decrease the threshold by 1. This means that decrease should NOT be set to false
+		// in such a scenario.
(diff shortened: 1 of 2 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -292,7 +292,15 @@ func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.Emp
 				newSigners = append(newSigners, s)
 			}
 		}
-		if params.Decrease || uint64(len(st.Signers)-1) < st.NumApprovalsThreshold {
+
+		// if the number of signers is below the threshold after removing the given signer,
+		// we should decrease the threshold by 1. This means that decrease should NOT be set to false
+		// in such a scenario.
+		if !params.Decrease && uint64(len(st.Signers)-1) < st.NumApprovalsThreshold {
+			rt.Abortf(exitcode.ErrIllegalArgument, "decrease flag should be true as number of signers below threshold")
```
anorth, line 300: Add as much info to the message as we can.

`"can't reduce signers to %d below threshold %d with decrease=false",  len(st.Signers)-1, st.NumApprovalsThreshold`

PR #526, opened 2020-06-24
Add WinCount to reward actor

Kubuxu:
It is necessary as we transition toward Poisson Sortition for the leader
election.

Q: How well does the reward actor handle over issuing rewards? We've currently set max `WinCount` of `15` in lotus but there can be also multiple blocks in a tipset.


Kubuxu:
Makes sense. The question was theoretical what if: we end up giving 50 rewards in one epoch and reward actor is running low on funds.

Issue #411, opened 2020-06-03
Power/penalty calculation for large fault sets or termination is too expensive


anorth:
The penalty for a sector missing a PoSt depends on per-sector state (specifically, sector quality/deal weight). When a miner goes AWOL and misses many PoSts, calculating the penalty for thousands or millions of sectors is an unreasonable performance cost and blows out block validation time. This is similarly a problem if a large number of sectors terminate at the same epoch, which is a likely, even desired outcome otherwise, since expiries are constrained to proving period boundaries (see #357).

A similar problem happens when miner is terminated. Also similarly, power calculations depend on the same information. While detected faults are the scariest case, this also limits ability to declare faults.

- Recording the per-partition quality/power on chain may help
- For termination, there may be some amortization possible to spread the load.

davidad:
I discussed with @nicola and others in Slack on April 16 a solution approach based on a (relatively) straightforward modification to the PoSt aggregation SNARK. Essentially, the faulting miner does the work of calculating how much power they've lost (rather than all the chain verifiers doing that computation independently), and SNARKing it makes it easy to verify without digging through lots of data.
> Currently the verifier is getting from the SNARK a bitmap of which sectors were proven, and you want to do a popcnt and multiply by a fixed fee per sector. The upgrade would be to (also) get from the SNARK a total power number, and multiply by a fixed fee per unit power.

nicola:
Possible ideas: bundle sectors with similar expiration with the same deadline and slash an entire deadline even if there is a single fault (very aggressive ideas here - do not consider if not useful)

Stebalien:
Discussed with @anorth in a call:

Observation: Faults will usually cover a few sectors, a few partitions, or an entire miner.

Proposal: Therefore, we can memoize per-sector, per-partition, and per-miner power and update these powers whenever the miner faults:

* When a miner declares a fault for a sector (or a few sectors), we can load state for those specific sectors. At the end of the proving period, we'll need to load `NUM_BAD_SECTORS * NUM_PARTITIONS` to update each partition's power.
* When a miner fails to submit a proof, we can fault the entire partition without loading the power for each sector.
* When a miner fails to submit _any_ proofs for the day, we can simply fault the entire miner's power without loading any additional state.

However, I'm going to need to dig a bit more. I assume we need to remove power immediately, but we also need to keep faulting these sectors some number of times before declaring them faulty, right?

davidad:
[...]

PR #528, opened 2020-06-24
Enforce locked funds do not fall below initial pledges

### LINKED ISSUE #415: Implement verification of locked pledge requirements on PoRep and PoSt
A miner must maintain locked pledge collateral equal to the sum of the initial pledge for all sectors (to cover potential fees from failing all PoSts and having sectors terminated, though because block reward is dynamic, this isn't a strong guarantee).  The pledge requirement should be checked upon the explicit miner worker messages of ~submitting a PoSt and~ proving a new sector commitment, and withdrawing unlocked balance.

There's a placeholder function unimplemented for this. Note that the call site for sector commitment was jumbled in the implementation of batched porep verifications, and should probably move back to be in the explicit miner-initiated call path (PreCommitSector), rather than in cron.

acruikshank:
closes #415

### Motivation

We need to enforce that a miner's locked funds exceeds the sum of their initial pledges to ensure the miner has enough funds to over potential fees. To do this we need to track the sum of initial pledges (and maintain that sum). We will check lock funds against initial pledges before a miner proves a sector and before the miner withdraws funds.

### Proposed Changes

The bulk of this PR breaks out into its 5 initial commits:

1. Add `InitialPledges` to `miner.State` and add to it every time a sector is proven.
2. Add `InitialPledge' to 'SectorOnChainInfo' and set it to the sector's initial pledge when proven.
3. Use the sector's `InitialPledge` to subtract from `miner.State.InitialPledges` when sectors are terminated.
4. Add a check in `ProveCommitSector` to ensure locked funds are sufficient relative to initial pledges.
5. Do the same in `WithdrawFunds`.

All of these changes are tested. This required some new infrastructure in miner test.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -323,20 +323,6 @@ func (st *State) DeleteSectors(store adt.Store, sectorNos *abi.BitField) error {
 	return err
 }
 
-func (st *State) DeductPledges(store adt.Store, sectorNos *abi.BitField) error {
-	return sectorNos.ForEach(func(sectorNo uint64) error {
-		sector, found, err := st.GetSector(store, abi.SectorNumber(sectorNo))
-		if err != nil {
-			return errors.Wrapf(err, "failed to get sector to deduct pledge %v", sectorNo)
-		}
-		if !found {
-			return errors.Errorf("failed to find sector to deduct pledge %v", sectorNo)
-		}
-		st.InitialPledgeRequirement = big.Sub(st.InitialPledgeRequirement, sector.InitialPledge)
-		return nil
-	})
-}
-
 func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) error {
 	sectors, err := adt.AsArray(store, st.Sectors)
 	if err != nil {
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -174,9 +174,9 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyDeadlinesCid cid.Cid, owner
 		},
 
 		PreCommitDeposits:        abi.NewTokenAmount(0),
-		InitialPledgeRequirement: abi.NewTokenAmount(0),
 		LockedFunds:              abi.NewTokenAmount(0),
 		VestingFunds:             emptyArrayCid,
+		InitialPledgeRequirement: abi.NewTokenAmount(0),
 
 		PreCommittedSectors: emptyMapCid,
 		Sectors:             emptyArrayCid,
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -637,6 +682,94 @@ func TestExtendSectorExpiration(t *testing.T) {
...
+		precommitEpoch := abi.ChainEpoch(1)
+		rt.SetEpoch(precommitEpoch)
+		st := getState(rt)
+		deadline := st.DeadlineInfo(rt.Epoch())
+		expiration := deadline.PeriodEnd() + 100*miner.WPoStProvingPeriod
+		sectorInfo := actor.commitAndProveSectors(rt, 1, expiration, 0, nil)
+
+		sector, found, err := getState(rt).GetSector(rt.AdtStore(), sectorInfo[0].SectorNumber)
+		require.NoError(t, err)
+		require.True(t, found)
```
anorth, line 709: This looks inconvenient. We should just change commitAndProveSectors to return the SectorOnChainInfos.

support/mock/mockrt.go
```diff
@@ -585,6 +585,10 @@ func (rt *Runtime) GetState(o runtime.CBORUnmarshaler) {
 	}
 }
 
+func (rt *Runtime) SetStateForTesting(o runtime.CBORMarshaler) {
```
anorth, line 588: Move this method into the "mocking facilities" section. 

Since it's a mock the `ForTesting` is probably implicit. Maybe `ReplaceState`? or `OverwriteState`?

actors/builtin/miner/miner_state.go
```diff
@@ -320,6 +323,20 @@ func (st *State) DeleteSectors(store adt.Store, sectorNos *abi.BitField) error {
 	return err
 }
 
+func (st *State) DeductPledges(store adt.Store, sectorNos *abi.BitField) error {
+	return sectorNos.ForEach(func(sectorNo uint64) error {
+		sector, found, err := st.GetSector(store, abi.SectorNumber(sectorNo))
```
anorth, line 328: Blocking: Unfortunately this is hiding a potentially huge performance issue inside the state class. Loading an arbitrary number of sector infos is unreasonable. This will need to be replaced as part of #411, using a similar technique, and the logic will need to move. In order to make that easier later, please let's not put it inside the state class. cc @Stebalien 

- Remove this method, do the work in the actor code and then use `AddInitialPledge(pledgeToRemove.Neg())`.
- Remove the call site from `removeTerminatedSectors` and pull up to `terminateSectors`. That method _already_ loads all these infos, so accumulate the pledge delta there and then call the cheap state mutation

actors/builtin/miner/miner_test.go
```diff
@@ -405,6 +409,49 @@ func TestWindowPost(t *testing.T) {
 	})
 }
 
+func TestProveCommit(t *testing.T) {
+	periodOffset := abi.ChainEpoch(100)
+	actor := newHarness(t, periodOffset)
+	builder := builderForHarness(actor).
+		WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID).
```
anorth, line 416: `constructAndVerify` does this now, so you shouldn't need this line.

PR #530, opened 2020-06-25
SubmitWindowedPoSt cleanups

Stebalien:
This patch set contains a set of cleanups/comments/questions from reading SubmitWindowedPoSt. None of these are critical, but they provide a nice way to ask questions.

### DIFF
--- a/actors/builtin/miner/deadlines.go
+++ b/actors/builtin/miner/deadlines.go
@@ -153,8 +153,8 @@ func ComputePartitionsSectors(d *Deadlines, partitionSize uint64, deadlineIndex 
 
 	// Work out which sector numbers the partitions correspond to.
 	deadlineSectors := d.Due[deadlineIndex]
-	var partitionsSectors []*abi.BitField
-	for _, pIdx := range partitions {
+	partitionsSectors := make([]*abi.BitField, len(partitions))
+	for i, pIdx := range partitions {
 		if pIdx < deadlineFirstPartition || pIdx >= deadlineFirstPartition+deadlinePartitionCount {
 			return nil, fmt.Errorf("invalid partition %d at deadline %d with first %d, count %d",
 				pIdx, deadlineIndex, deadlineFirstPartition, deadlinePartitionCount)
@@ -168,7 +168,7 @@ func ComputePartitionsSectors(d *Deadlines, partitionSize uint64, deadlineIndex 
 			return nil, fmt.Errorf("failed to slice deadline %d, size %d, offset %d, count %d",
 				deadlineIndex, deadlineSectorCount, sectorOffset, sectorCount)
 		}
-		partitionsSectors = append(partitionsSectors, partitionSectors)
+		partitionsSectors[i] = partitionSectors
 	}
 	return partitionsSectors, nil
 }
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -250,11 +250,6 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		deadlines, err := st.LoadDeadlines(adt.AsStore(rt))
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
 
-		// Traverse earlier submissions and enact detected faults.
-		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
-		// of penalties if locked pledge drops too low.
-		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
-
 		// Double check that the current proving period has started. This should only happen if the cron actor wasn't invoked.
 		if !currDeadline.PeriodStarted() {
 			rt.Abortf(exitcode.ErrIllegalState, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
@@ -263,6 +258,12 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 			rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d at epoch %d, expected %d",
 				params.Deadline, currEpoch, currDeadline.Index)
 		}
+
+		// Detect and mark faults from any missed deadlines (doesn't including the current deadline).
+		// Traverse earlier submissions and enact detected faults.
+		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
+		// of penalties if locked pledge drops too low.
+		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
 
 		// TODO WPOST (follow-up): process Skipped as faults
 		// https://github.com/filecoin-project/specs-actors/issues/410
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -255,8 +255,9 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		// of penalties if locked pledge drops too low.
 		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
 
+		// Double check that the current proving period has started. This should only happen if the cron actor wasn't invoked.
 		if !currDeadline.PeriodStarted() {
-			rt.Abortf(exitcode.ErrIllegalArgument, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
+			rt.Abortf(exitcode.ErrIllegalState, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
 		}
 		if params.Deadline != currDeadline.Index {
 			rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d at epoch %d, expected %d",
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -233,6 +233,7 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 	var recoveredSectors []*SectorOnChainInfo
 	penalty := abi.NewTokenAmount(0)
 
+	// Get the total power/reward. We need these to compute penalties.
 	epochReward := requestCurrentEpochBlockReward(rt)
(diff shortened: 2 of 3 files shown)


actors/builtin/miner/deadlines.go
```diff
@@ -153,8 +153,8 @@ func ComputePartitionsSectors(d *Deadlines, partitionSize uint64, deadlineIndex
 
 	// Work out which sector numbers the partitions correspond to.
 	deadlineSectors := d.Due[deadlineIndex]
-	var partitionsSectors []*abi.BitField
-	for _, pIdx := range partitions {
+	partitionsSectors := make([]*abi.BitField, len(partitions))
```
Stebalien, line 156: I noticed we tend to avoid pre-allocating. Is this a style thing?

actors/builtin/miner/miner_actor.go
```diff
@@ -217,34 +217,42 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
 
-		// Traverse earlier submissions and enact detected faults.
-		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
-		// of penalties if locked pledge drops too low.
-		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
-
+		// Double check that the current proving period has started. This should be impossible in practice (XXX: check this?)
 		if !currDeadline.PeriodStarted() {
-			rt.Abortf(exitcode.ErrIllegalArgument, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
+			rt.Abortf(exitcode.ErrIllegalState, "proving period %d not yet open at %d", currDeadline.PeriodStart, currEpoch)
```
Stebalien, line 241: Is this an illegal argument or illegal state?

actors/builtin/miner/miner_actor.go
```diff
@@ -217,34 +217,42 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
 		if params.Deadline != currDeadline.Index {
 			rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d at epoch %d, expected %d",
 				params.Deadline, currEpoch, currDeadline.Index)
 		}
 
+		// Detect and mark faults from any missed deadlines (doesn't including the current deadline).
+		// Traverse earlier submissions and enact detected faults.
+		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
+		// of penalties if locked pledge drops too low.
+		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
```
Stebalien, line 254: I moved this down after the argument checking. It should be fine either way, but we might as well save some time with easy checks first.

actors/builtin/miner/miner_actor.go
```diff
@@ -263,6 +271,8 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 		provenSectors, err := abi.BitFieldUnion(partitionsSectors...)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to union %d partitions of sectors", len(partitionsSectors))
 
+		// Load sector infos, substituting a known-good sector for known-faulty sectors.
+		// XXX: ErrIllegalState isn't quite correct. This can fail if the user submits a proof for a partition with only faulty sectors.
```
Stebalien, line 275: ?

actors/builtin/miner/miner_state.go
```diff
@@ -738,40 +738,39 @@ func (st *State) LoadSectorInfosForProof(store adt.Store, provenSectors *abi.Bit
...
 	}
 
 	// Load the sector infos, masking out fault sectors with a good one.
-	var sectorInfos []*SectorOnChainInfo
+	sectorInfos := make([]*SectorOnChainInfo, sectorCount)
 	err = sectors.ForEach(func(i uint64) error {
-		sector := standInInfo
-		faulty := faultSet[i]
-		if !faulty {
-			sectorOnChain, found, err = st.GetSector(store, abi.SectorNumber(i))
```
Stebalien, line 766: Was there a reason to re-use the outer variables?

actors/builtin/miner/deadlines.go
anorth: No, thanks for improving it.

actors/builtin/miner/miner_actor.go
```diff
@@ -217,34 +217,42 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
 		currDeadline := st.DeadlineInfo(currEpoch)
 		deadlines, err := st.LoadDeadlines(adt.AsStore(rt))
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
 
-		// Traverse earlier submissions and enact detected faults.
-		// This isn't strictly necessary, but keeps the power table up to date eagerly and can force payment
-		// of penalties if locked pledge drops too low.
-		detectedFaultSectors, penalty = detectFaultsThisPeriod(rt, &st, store, currDeadline, deadlines, epochReward, pwrTotal.QualityAdjPower)
-
+		// Double check that the current proving period has started. This should be impossible in practice (XXX: check this?)
```
anorth, line 239: No, this can happen when cron doesn't happen exactly on the final PP epoch, because it's an empty tipset. Or if the cron has borked generally (which is bad, but currently happens when bugs).

actors/builtin/miner/miner_actor.go
anorth: Well, the problem is that the actor state is not yet updated (missed cron). There is no valid argument that could be provided right now to this call that would work. I think IllegalState is better.

actors/builtin/miner/miner_actor.go
anorth: Yeah, this is an awkwardness we have currently that errors propagated up from below can't indicate what the appropriate exit code is (but, OTOH, sometimes the bottom level doesn't know and the decision should be made here).

A solution probably involves a custom error type that can carry an exit code and message, and using that exit code if provided. Or some other error-inspection mechanism.

This isn't the only place, and I'd welcome generally your thoughts on how to do this better. P2 though.

actors/builtin/miner/miner_state.go
anorth: Static analysis warnings about shadowing (in IDEs). I don't know of a great way to resolve this in Go. We've had bugs caused by shadowing, but also many shadowings that are difficult to remove without some gymnastics.

actors/builtin/miner/miner_state.go
```diff
@@ -738,40 +738,39 @@ func (st *State) LoadSectorInfosForProof(store adt.Store, provenSectors *abi.Bit
...
-			}
-			sector = sectorOnChain
+		if faulty := faultSet[i]; faulty {
+			sectorInfos[i] = standInInfo
+			return nil
+		}
+
+		sector, found, err := st.GetSector(store, abi.SectorNumber(i))
+		if err != nil {
+			return fmt.Errorf("failed to load sector %d: %v", i, err)
```
anorth, line 769: `%w` while you're here, and `xerrors` too.

actors/builtin/miner/miner_actor.go
ZenGround0: This is probably well out of our control at this point but have we thought about invoking cron from the validator on every preceding empty tipset before running a tipset's messages?  Haven't thought about how that could go wrong but it feels like a bad coupling of blockchain details with state machine to require logic in the state machine for handling these edge cases.

actors/builtin/miner/miner_actor.go
```diff
@@ -280,7 +290,7 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 		err = st.AddPoStSubmissions(postedPartitions)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to record submissions for partitions %s", params.Partitions)
 
-		// If the PoSt was successful, the declared recoveries should be restored
+		// If the PoSt was successful, the restore declared recoveries.
```
ZenGround0, line 293: nit: did you mean "then restore" or am I just failing to parse the new phrasing?

actors/builtin/miner/miner_actor.go
Stebalien: Wouldn't a block that missed a cron invocation be invalid?

actors/builtin/miner/miner_state.go
Stebalien: Are we still supporting old go versions? Or should we switch to `fmt`?

actors/builtin/miner/miner_actor.go
Stebalien: We can introduce a custom error interface:

```go
type ExitError interface {
    ErrorCode() ExitCode
}
```

Then, in `builtin.RequireNoError`, we'd use the error code argument as the _default_ error code, first trying to extract an error code from the error itself. Of course, that could cause it's own problems and bugs.

Ideally, we'd be able to check arguments up-front and avoid this mess, but that's not always possible.

Stebalien:
Meta: do we add commits, or rebase and fix commits, or fix commits at the end, etc?

actors/builtin/miner/miner_state.go
```diff
@@ -738,34 +738,33 @@ func (st *State) LoadSectorInfosForProof(store adt.Store, provenSectors *abi.Bit
...
 	}
-	faultSet, err := faults.AllMap(faultMax)
+	faultSet, err := faults.AllMap(sectorCount)
 	if err != nil {
 		return nil, fmt.Errorf("failed to expand faults: %w", err)
 	}
 
 	// Load the sector infos, masking out fault sectors with a good one.
-	var sectorInfos []*SectorOnChainInfo
+	sectorInfos := make([]*SectorOnChainInfo, 0, sectorCount)
```
Stebalien, line 760: I've reverted most of my changes to this function as I misunderstood how `ForEach` worked.

actors/builtin/miner/miner_actor.go
anorth: Cron isn't invoked per block, but per tipset, implicitly by the vm. There is no way for a block to fail to invoke cron because block producers don't know which the last block in the tipset is going to be.

actors/builtin/miner/miner_state.go
anorth: @whyrusleeping requested xerrors. I'm happy with whatever y'all decide on, I just want there to be one consistent thing for us to use.

anorth:
> Meta: do we add commits, or rebase and fix commits, or fix commits at the end, etc

Up to the implementer, so long as you squash-merge (no merge commit) with a cleaned-up commit message at the end.

I frequently rebase+squash as I go, sometimes leave the individual commits if I think they're useful to the reviewer.

PR #533, opened 2020-06-25
Multisig: Use correct error codes

### LINKED ISSUE #473: Multisig test match exitcode instead of abort
From builtin/multisig/multisig_test.go:TestApprove:"fail approve transaction more than once"
`// TODO replace with correct exit code when multisig actor breaks the AbortStateMsg pattern.`

I believe this is related to #434 getting closed out, but something potentially more complicated is going on.  Needs more investigation.

aarshkshah1992:
Closes #473 

@anorth On second thoughts, replacing the `AbortMsg` pattern in the plumbing methods looks like a larger refactor that we can defer to later. Do we have an issue for it or should I create one ?


anorth:
Thanks. There's no issue, but I don't think we need one. Just all of us to keep changing things in the right direction. Stopping to fix it all up now isn't worth it.

PR #534, opened 2020-06-25
Multisig Tests: Remove verifies that are not required

aarshkshah1992:
Followup to https://github.com/filecoin-project/specs-actors/pull/524/files#r445244817.

@anorth There's potentially a lot of other calls for other methods where we can do this but this is a start.


anorth:
Thanks. This is great so far. Could you please do more in this file? I'm concerned that tests might be invalid if missing a `Verify`.

aarshkshah1992:
@anorth Sure, I'll keep raising these.

PR #538, opened 2020-06-25
validateExpiration() checks current extension and total sector lifetime are not too long

### LINKED ISSUE #454: Enforce sector maximum lifetime, dependent on seal proof type
Sectors have a maximum lifetime, governed by security of the PoRep. `ExtendSectorExpiration` does not yet enforce this.

- The sector maximum lifetime should be made a function of the seal proof (as is window post partition size).
- Initial committed lifetime should have a maximum value, for now = 1 year (in epochs)
- Each extension should also have a maximum of 1 year, bounded by the sector maximum lifetime.

This is probably an appropriate moment to refactor the "things that depend on seal proof" into a structure, rather than maintain many parallel mappings.

acruikshank:
closes  #454

### Motivation

Extending a sector's expiration more than a year or extending a sector past the maximum lifetime for its seal proof type decreases security (in ways unknown to the implementer).

### Proposed Changes

1. Introduce `sealProof.SectorMaximumLifetime` method to limit the total lifetime of a sector after multiple extensions (currently always returns 5 years).
2. Introduce `MaxSectorExpirationExtension` as the maximum single shot expiration.
3. Augment `validateExpiration()` to check that `MaxSectorExpirationExtension` and `sealProof.SectorMaximumLifetime` are not exceeded.
4. Test.

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -68,33 +68,35 @@ type RegisteredProof = int64
 
 // This ordering, defines mappings to UInt in a way which MUST never change.
 type RegisteredSealProof RegisteredProof
+
 const (
-	RegisteredSealProof_StackedDrg2KiBV1 = RegisteredSealProof(0);
-	RegisteredSealProof_StackedDrg8MiBV1 = RegisteredSealProof(1);
-	RegisteredSealProof_StackedDrg512MiBV1 = RegisteredSealProof(2);
-	RegisteredSealProof_StackedDrg32GiBV1 = RegisteredSealProof(3);
-	RegisteredSealProof_StackedDrg64GiBV1 = RegisteredSealProof(4);
+	RegisteredSealProof_StackedDrg2KiBV1   = RegisteredSealProof(0)
+	RegisteredSealProof_StackedDrg8MiBV1   = RegisteredSealProof(1)
+	RegisteredSealProof_StackedDrg512MiBV1 = RegisteredSealProof(2)
+	RegisteredSealProof_StackedDrg32GiBV1  = RegisteredSealProof(3)
+	RegisteredSealProof_StackedDrg64GiBV1  = RegisteredSealProof(4)
 )
 
 type RegisteredPoStProof RegisteredProof
-const(
- 	RegisteredPoStProof_StackedDrgWinning2KiBV1 = RegisteredPoStProof(0);
-   	RegisteredPoStProof_StackedDrgWinning8MiBV1 = RegisteredPoStProof(1);
-	RegisteredPoStProof_StackedDrgWinning512MiBV1 = RegisteredPoStProof(2);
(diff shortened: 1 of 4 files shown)


actors/abi/sector.go
```diff
@@ -203,6 +205,14 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 	}
 }
 
+// SectorMaximumLifetime is the maximum duration a sector sealed with this proof may exist between activation and expiration
+func (p RegisteredSealProof) SectorMaximumLifetime() ChainEpoch {
+	// currently all proofs set to around 5 years
+	epochsPerYear := 1_262_277
```
anorth, line 211: Use the existing `builtin.EpochsInYear` which will change appropriately if epoch duration changes.

Also I can clarify the comment for you: "For all Stacked DRG sectors, the max is 5 years".

actors/builtin/miner/miner_actor.go
```diff
@@ -1236,7 +1236,20 @@ func computeFaultsFromMissingPoSts(st *State, deadlines *Deadlines, sinceDeadlin
...
+func validateExpiration(rt Runtime, st *State, activation, expiration abi.ChainEpoch, sealProof abi.RegisteredSealProof) {
+	// expiration cannot exceed MaxSectorExpirationExtension from now
+	if expiration > rt.CurrEpoch()+MaxSectorExpirationExtension {
+		rt.Abortf(exitcode.ErrIllegalArgument, "invalid expiration %d, cannot be more than %d past current epoch %d",
+			expiration, MaxSectorExpirationExtension, rt.CurrEpoch())
+	}
+
+	// total sector lifetime cannot exceed SectorMaximumLifetime for the sector's seal proof
+	if expiration-activation > sealProof.SectorMaximumLifetime() {
+		rt.Abortf(exitcode.ErrIllegalArgument, "invalid expiration %d, total sector lifetime (%d) cannot exceed %d",
```
anorth, line 1248: Nit: "cannot exceed %d after activation %d"

actors/builtin/miner/miner_test.go
```diff
@@ -245,16 +245,26 @@ func TestCommitments(t *testing.T) {
 		rt.Reset()
 
 		// Expires before current epoch
+		expiration := deadline.PeriodEnd()
 		rt.SetEpoch(deadline.PeriodEnd() + 1)
```
anorth, line 249:
```suggestion
		rt.SetEpoch(expiration + 1)
```

actors/abi/sector.go
acruikshank: I would love to, but sector.go is in `abi`, so referencing `builtin` would be circular.

anorth: Gah, ok. We should probably move this logic out of `abi` then, so it returns to just types.

PR #540, opened 2020-06-25
Feat/termination fees

### LINKED ISSUE #437: Implement fault and termination penalty calculations

acruikshank:
closes #437

### Motivation

While the infrastructure for charging sector termination fees exists, the actual calculation has not been implemented. This PR implements that calculation and tests.

### Proposed Changes

1. Implement the termination fee formula based on the initial pledge of the sector.
2. Augment test to show correct fee has been sent to burnt funds and pledge has been updated in power actor
3. Clean up miner actor harness by adding `networkQAPower` and `networkRawPower` as instance variables with defaults.

### DIFF
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -46,13 +46,13 @@ func PledgePenaltyForUndeclaredFault(epochTargetReward abi.TokenAmount, networkQ
 // Penalty to locked pledge collateral for the termination of a sector before scheduled expiry.
 func PledgePenaltyForTermination(s *SectorOnChainInfo, currEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
 	// max(SP(t), InitialPledge + BR(StartEpoch)*min(SectorAgeInDays, 180)) where BF(StartEpoch)=InitialPledge/InitialPledgeFactor
-	sectorAge := int64(min64(uint64(currEpoch-s.Activation), 180*builtin.EpochsInDay))
+	sectorAge := big.NewInt(int64(minEpoch(currEpoch-s.Activation, 180*builtin.EpochsInDay)))
 	return big.Max(
 		PledgePenaltyForUndeclaredFault(epochTargetReward, networkQAPower, qaSectorPower),
 		big.Add(
 			InitialPledgeFactor,
 			big.Div(
-				big.Mul(s.InitialPledge, big.NewInt(sectorAge)),
+				big.Mul(s.InitialPledge, sectorAge),
 				InitialPledgeFactor)))
 }
 
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -44,9 +44,16 @@ func PledgePenaltyForUndeclaredFault(epochTargetReward abi.TokenAmount, networkQ
 }
 
 // Penalty to locked pledge collateral for the termination of a sector before scheduled expiry.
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/monies.go
```diff
@@ -44,9 +44,16 @@ func PledgePenaltyForUndeclaredFault(epochTargetReward abi.TokenAmount, networkQ
 }
 
 // Penalty to locked pledge collateral for the termination of a sector before scheduled expiry.
-func pledgePenaltyForTermination(s *SectorOnChainInfo, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
-	//TODO #437 compute correct termination fee using initial pledge value from SectorOnChainInfo
-	return big.Zero() // PARAM_FINISH
+func PledgePenaltyForTermination(s *SectorOnChainInfo, currEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
+	// max(SP(t), InitialPledge + BR(StartEpoch)*min(SectorAgeInDays, 180)) where BF(StartEpoch)=InitialPledge/InitialPledgeFactor
```
ZenGround0, line 48: nit: BF should be BR I think?

actors/builtin/miner/monies.go
```diff
@@ -44,9 +44,16 @@ func PledgePenaltyForUndeclaredFault(epochTargetReward abi.TokenAmount, networkQ
...
+	sectorAge := big.NewInt(int64(minEpoch(currEpoch-s.Activation, 180*builtin.EpochsInDay)))
+	return big.Max(
+		PledgePenaltyForUndeclaredFault(epochTargetReward, networkQAPower, qaSectorPower),
+		big.Add(
+			InitialPledgeFactor,
```
ZenGround0, line 53: From the comment above I would have thought this would be `s.InitialPledge`, not `InitialPledgeFactor`

anorth: I have been meaning to add unit tests specifically for all these calculation functions. Let's start with this one.

Could you add some basic tests outlining the calculation results? Especially useful might be border cases like 1 epoch, 180days, >180days, and some right-order-of-magnitude numbers for pledge, power etc.

acruikshank:
Good catch @ZenGround0. I've refactored the function a bit to make testing easier and test the major cases. In the process I realized I think I'd made another more subtle mistake. BR represents block reward per day, and I was multiplying it by the sector age in epochs which should be too much by a factor of ~3000. The new version divides by `EpochsInDay` at the end to maintain precision. Please take another looks @anorth and @ZenGround0 if you get a chance.

PR #543, opened 2020-06-26
Check for nil SendReturns before serializing

arajasek:
Some gofmt changes got thrown in, can remove them

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -203,7 +203,7 @@ func (a Actor) Approve(rt vmr.Runtime, params *TxnIDParams) *ApproveReturn {
 	})
 
 	// if the transaction already has enough approvers, execute it without "processing" this approval.
-	approved,ret, code := executeTransactionIfApproved(rt, st, params.ID, txn)
+	approved, ret, code := executeTransactionIfApproved(rt, st, params.ID, txn)
 	if !approved {
 		// if the transaction hasn't already been approved, let's "process" this approval
 		// and see if we can execute the transaction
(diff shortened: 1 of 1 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -439,10 +439,12 @@ func executeTransactionIfApproved(rt vmr.Runtime, st State, txnID TxnID, txn *Tr
 		)
 		applied = true
 
-		// Pass the return value through uninterpreted with the expectation that serializing into a CBORBytes never fails
-		// since it just copies the bytes.
-		if err := ret.Into(&out); err != nil {
-			rt.Abortf(exitcode.ErrSerialization, "failed to deserialize result: %v", err)
+		if ret != nil {
```
frrist, line 442: `ret` in an interface, right? This line concerns me since an interface value that holds a nil concrete value (lotus uses bytes, I think?) is itself non-nil. Maybe im just paranoid, but do we have test coverage here?

arajasek: Yeah, Lotus returns an actual `nil` interface (both type and value `nil`), as opposed to an interface with the wrapper type, and `nil` value.

This can be changed if one is objectively wrong-er than the other, I lack the judgment to make a call here.

arajasek: But, yes, we do have a chain-val test that hits this and fails (when using `master`s of Lotus and specs-actors).

frrist: Im not sure which is better here either, gut check says what lotus is doing is probably fine.

The test fails (?) or passes by triggering a failure via abort?

anorth: Lotus should be returning `adt.Empty` here (which is a typed nil). The chain-validation test is doing its job. The documentation on Runtime.Send is explicit:

> 	// The result is never a bare nil, but may be (a wrapper of) adt.Empty.

The nil-check belongs just once inside the VM, rather than every actor that uses `Send` having to do this (some of which will doubtless be forgotten).

Here is [where go-filecoin does it](https://github.com/filecoin-project/go-filecoin/blob/c0a783f847012ecf78e3428bd1c2cb5c0f016cc1/internal/pkg/vm/internal/vmcontext/invocation_context.go#L225).

arajasek: Sounds good, thanks

PR #545, opened 2020-06-26
Feat/cidify minerinfo

### LINKED ISSUE #422: MinerInfo becomes a cid
MinerInfo is a field in the miner actor state.  The field should be converted to the cid of the corresponding ipld cbor node of the existing struct.

ZenGround0:
Closes #422

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1858,7 +1858,8 @@ func commitWorkerKeyChange(rt Runtime) *adt.EmptyValue {
 
 		info.Worker = info.PendingWorkerKey.NewWorker
 		info.PendingWorkerKey = nil
-		st.SaveInfo(adt.AsStore(rt), info)
+		err = st.SaveInfo(adt.AsStore(rt), info)
+		builtin.RequireNoErr(rt, err, exitcode.ErrSerialization, "failed to save miner info")
 
 		return nil
 	})
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -102,7 +102,25 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
 	periodStart := nextProvingPeriodStart(currEpoch, offset)
 	Assert(periodStart > currEpoch)
 
-	state, err := ConstructState(emptyArray, emptyMap, emptyDeadlinesCid, owner, worker, params.PeerId, params.Multiaddrs, params.SealProofType, periodStart)
+	sectorSize, err := params.SealProofType.SectorSize()
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "no sector size for seal proof type %d: %w", params.SealProofType)
+
+	partitionSectors, err := params.SealProofType.WindowPoStPartitionSectors()
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "no partition size for seal proof type %d", params.SealProofType)
+
+	info := MinerInfo{
+		Owner:                      owner,
+		Worker:                     worker,
+		PendingWorkerKey:           nil,
+		PeerId:                     params.PeerId,
+		Multiaddrs:                 params.Multiaddrs,
+		SealProofType:              params.SealProofType,
+		SectorSize:                 sectorSize,
+		WindowPoStPartitionSectors: partitionSectors,
+	}
+	infoCid := rt.Store().Put(&info)
+
+	state, err := ConstructState(emptyArray, emptyMap, emptyDeadlinesCid, infoCid, periodStart)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to construct state")
 	rt.State().Create(state)
 
@@ -126,9 +144,11 @@ func (a Actor) ControlAddresses(rt Runtime, _ *adt.EmptyValue) *GetControlAddres
 	rt.ValidateImmediateCallerAcceptAny()
 	var st State
 	rt.State().Readonly(&st)
+	info, err := st.GetInfo(adt.AsStore(rt))
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "could not read miner info")
 	return &GetControlAddressesReturn{
-		Owner:  st.Info.Owner,
-		Worker: st.Info.Worker,
+		Owner:  info.Owner,
+		Worker: info.Worker,
 	}
 }
 
@@ -140,17 +160,22 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 	var effectiveEpoch abi.ChainEpoch
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		rt.ValidateImmediateCallerIs(st.Info.Owner)
+		info, err := st.GetInfo(adt.AsStore(rt))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "could not read miner info")
+
+		rt.ValidateImmediateCallerIs(info.Owner)
 
 		worker := resolveWorkerAddress(rt, params.NewWorker)
 
 		effectiveEpoch = rt.CurrEpoch() + WorkerKeyChangeDelay
 
 		// This may replace another pending key change.
-		st.Info.PendingWorkerKey = &WorkerKeyChange{
+		info.PendingWorkerKey = &WorkerKeyChange{
 			NewWorker:   worker,
 			EffectiveAt: effectiveEpoch,
 		}
+		err = st.SaveInfo(adt.AsStore(rt), info)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "could not save miner info")
 		return nil
 	})
 
@@ -168,8 +193,13 @@ type ChangePeerIDParams struct {
 func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyValue {
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
-		rt.ValidateImmediateCallerIs(st.Info.Worker)
-		st.Info.PeerId = params.NewID
+		info, err := st.GetInfo(adt.AsStore(rt))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "could not read miner info")
+
+		rt.ValidateImmediateCallerIs(info.Worker)
+		info.PeerId = params.NewID
+		err = st.SaveInfo(adt.AsStore(rt), info)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "could not save miner info")
 		return nil
 	})
 	return nil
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -652,7 +698,9 @@ type TerminateSectorsParams struct {
 func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *adt.EmptyValue {
 	var st State
 	rt.State().Readonly(&st)
-	rt.ValidateImmediateCallerIs(st.Info.Worker)
+	info, err := st.GetInfo(adt.AsStore(rt))
```
acruikshank, line 701: There seems to be a number of places where we access info solely to validate the caller is worker. Consider extracting a function.

acruikshank: Seeing more cases, maybe just a function to get the info obj and do the error check.

actors/builtin/miner/miner_actor.go
```diff
@@ -1870,22 +1946,34 @@ func qaPowerForSectors(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo)
 }
 
 func unlockDeclaredFaultPenalty(st *State, store adt.Store, currEpoch abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, sectors []*SectorOnChainInfo) (abi.TokenAmount, error) {
-	sectorSize := st.Info.SectorSize
+	info, err := st.GetInfo(store)
+	if err != nil {
+		return abi.NewTokenAmount(0), err
+	}
+	sectorSize := info.SectorSize
```
acruikshank, line 1953: Just pass `info.SectorSize` into `qaPowerForSectors`. This variable isn't necessary.

acruikshank: Here and in the next two functions.

actors/builtin/miner/miner_state.go
```diff
@@ -191,12 +172,37 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyDeadlinesCid cid.Cid, owner
 	}, nil
 }
 
-func (st *State) GetWorker() addr.Address {
-	return st.Info.Worker
+func (st *State) GetInfo(store adt.Store) (*MinerInfo, error) {
+	var info MinerInfo
+	if err := store.Get(store.Context(), st.Info, &info); err != nil {
+		return nil, errors.Wrapf(err, "failed to get miner info")
```
anorth, line 178: Don't extend the use of `pkg/errors`, use `%w` either with `fmt.Errorf` or `xerrors`.

actors/builtin/miner/miner_state.go
```diff
@@ -191,12 +172,37 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyDeadlinesCid cid.Cid, owner
...
+func (st *State) SaveInfo(store adt.Store, info *MinerInfo) error {
+	c, err := store.Put(store.Context(), info)
+	if err != nil {
+		return err
+	}
+	st.Info = c
+	return nil
+}
+
+func (st *State) GetWorker(store adt.Store) (addr.Address, error) {
```
anorth, line 192: I think we should not have these convenience methods because they encourage an inefficient pattern of loading the miner info more than once.

actors/builtin/miner/miner_actor.go
```diff
@@ -102,7 +102,25 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
...
+	info := MinerInfo{
+		Owner:                      owner,
+		Worker:                     worker,
+		PendingWorkerKey:           nil,
+		PeerId:                     params.PeerId,
+		Multiaddrs:                 params.Multiaddrs,
+		SealProofType:              params.SealProofType,
+		SectorSize:                 sectorSize,
+		WindowPoStPartitionSectors: partitionSectors,
+	}
```
anorth, line 120: Please make a `ConstructMinerInfo` function in miner_state.go to push a little of the noise into there.

actors/builtin/miner/miner_actor.go
```diff
@@ -102,7 +102,25 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
 	periodStart := nextProvingPeriodStart(currEpoch, offset)
 	Assert(periodStart > currEpoch)
 
-	state, err := ConstructState(emptyArray, emptyMap, emptyDeadlinesCid, owner, worker, params.PeerId, params.Multiaddrs, params.SealProofType, periodStart)
+	sectorSize, err := params.SealProofType.SectorSize()
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "no sector size for seal proof type %d: %w", params.SealProofType)
```
anorth, line 106: Note: `RequireNoErr` does the wrapping, don't append `%w`. But push this back into a function anyway (see below)

actors/builtin/miner/miner_actor.go
```diff
@@ -298,8 +334,10 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 		return nil
 	})
 
+	info, err := st.GetInfo(adt.AsStore(rt))
```
anorth, line 337: Redundant loading of info (from 258)

actors/builtin/miner/miner_actor.go
```diff
@@ -360,7 +400,9 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
 		duration := params.Expiration - rt.CurrEpoch()
 
-		sectorWeight := QAPowerForWeight(st.GetSectorSize(), duration, dealWeight.DealWeight, dealWeight.VerifiedDealWeight)
+		sectorSize, err := st.GetSectorSize(adt.AsStore(rt))
```
anorth, line 403: This redundantly loads the miner info

actors/builtin/miner/miner_actor.go
anorth: Such a function would need the `rt` as an argument, which is against the general policy. It's a decent example of a pragmatic exception, though, so 👍 .

I'd prefer caller validation remains explicitly in the top-level actor method wherever possible.

This should be a free function, rather than method on state. `info := getMinerInfo(rt, &st)`

actors/builtin/miner/miner_actor.go
```diff
@@ -779,7 +829,9 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 	})
 
 	// Remove power for new faulty sectors.
-	requestUpdateSectorPower(rt, st.Info.SectorSize, nil, append(detectedFaultSectors, declaredFaultSectors...))
+	info, err := st.GetInfo(adt.AsStore(rt))
```
anorth, line 832: Redundant loading.

actors/builtin/miner/miner_actor.go
```diff
@@ -850,7 +904,9 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 	})
 
 	// Remove power for new faulty sectors.
-	requestUpdateSectorPower(rt, st.Info.SectorSize, nil, detectedFaultSectors)
+	info, err := st.GetInfo(adt.AsStore(rt))
```
anorth, line 907: Ditto

actors/builtin/miner/miner_actor.go
```diff
@@ -1025,7 +1086,11 @@ func handleProvingPeriod(rt Runtime) {
 		var detectedFaultSectors []*SectorOnChainInfo
 		currEpoch := rt.CurrEpoch()
 		penalty := abi.NewTokenAmount(0)
+		var info *MinerInfo
```
anorth, line 1089: Move this to outer scope and re-use it. The fields accessed are immutable.

actors/builtin/miner/miner_actor.go
```diff
@@ -1175,12 +1244,12 @@ func processMissingPoStFaults(rt Runtime, st *State, store adt.Store, deadlines
 
 // Computes the sectors that were expected to be present in partitions of a PoSt submission but were not, in the
 // deadlines from sinceDeadline (inclusive) to beforeDeadline (exclusive).
-func computeFaultsFromMissingPoSts(st *State, deadlines *Deadlines, sinceDeadline, beforeDeadline uint64) (detectedFaults, failedRecoveries *abi.BitField, err error) {
+func computeFaultsFromMissingPoSts(info *MinerInfo, faults, recoveries, postSubmissions *abi.BitField, deadlines *Deadlines, sinceDeadline, beforeDeadline uint64) (detectedFaults, failedRecoveries *abi.BitField, err error) {
```
anorth, line 1247: Change the parameter to be `partitionSize` and don't pass the whole info here.

actors/builtin/miner/miner_actor.go
anorth: The call sites have already loaded the info, so pass the sector size as a parameter to `unlockXxxPenalty`

actors/builtin/miner/miner_test.go
```diff
@@ -365,7 +367,9 @@ func TestWindowPost(t *testing.T) {
 		actor.constructAndVerify(rt)
 		st := getState(rt)
 		store := rt.AdtStore()
-		partitionSize := st.Info.WindowPoStPartitionSectors
+		info, err := st.GetInfo(store)
+		require.NoError(t, err)
+		partitionSize := info.WindowPoStPartitionSectors
```
anorth, line 372: This is inconvenient enough that we should just move it to a method on the harness

actors/builtin/miner/miner_test.go
```diff
@@ -443,8 +447,9 @@ func TestProveCommit(t *testing.T) {
 		st := getState(rt)
 		st.LockedFunds = big.Div(st.LockedFunds, big.NewInt(2))
 		rt.ReplaceState(st)
-
-		rt.SetEpoch(precommitEpoch + miner.MaxSealDuration[st.Info.SealProofType] - 1)
+		info, err := st.GetInfo(rt.AdtStore())
+		require.NoError(t, err)
```
anorth, line 451: Add `getInfo` to the harness. (I don't care about redundant loads of the state/info in tests).

actors/builtin/miner/miner_test.go
```diff
@@ -1250,9 +1257,11 @@ func (h *actorHarness) extendSector(rt *mock.Runtime, sector *miner.SectorOnChai
 	rt.ExpectValidateCallerAddr(h.worker)
 
 	st := getState(rt)
+	info, err := st.GetInfo(rt.AdtStore())
+	require.NoError(h.t, err)
 	newSector := *sector
 	newSector.Expiration += extension
-	qaDelta := big.Sub(miner.QAPowerForSector(st.Info.SectorSize, &newSector), miner.QAPowerForSector(st.Info.SectorSize, sector))
+	qaDelta := big.Sub(miner.QAPowerForSector(info.SectorSize, &newSector), miner.QAPowerForSector(info.SectorSize, sector))
```
anorth, line 1264: As for partition size, we need sectorSize enough that the harness should expose it directly.

PR #553, opened 2020-06-29
Multisig: Test that ONLY the proposer can cancel a transaction

aarshkshah1992:
Closes #97

### DIFF
--- a/actors/builtin/multisig/multisig_test.go
+++ b/actors/builtin/multisig/multisig_test.go
@@ -919,6 +919,64 @@ func TestCancel(t *testing.T) {
 			Approved: []addr.Address{anne},
 		})
 	})
+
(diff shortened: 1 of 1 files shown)


actors/builtin/multisig/multisig_test.go
```diff
@@ -919,6 +919,64 @@ func TestCancel(t *testing.T) {
 			Approved: []addr.Address{anne},
 		})
 	})
+
+	t.Run("transaction can ONLY be cancelled by the proposer", func(t *testing.T) {
```
anorth, line 923: "...proposer who is still a signer"

aarshkshah1992: This comment has been addressed,

Issue #455, opened 2020-06-10
Remove runtime access from market actor state methods


anorth:
The Runtime interface is large and dependencies on it should be restricted in order to enable testability. Ideally, access should be restricted only to top-level exported actor methods, though there are some cases where its justified to plumb a little deeper. It is often passed merely for access to the `Abortf` method, as an alternative to canonical Go error propagation.

The `market.State` object is one such place with rampant use of the Runtime only for aborting. This makes the state mutations much harder to test, and indeed they are largely untested (14% coverage).

Purge use of the runtime from this code.

PR #555, opened 2020-06-29
Keep track of locked funds

### LINKED ISSUE #531: Keep track of TotalDealCollateral in actors
To compute TotalFILCirculatingSupply, we will also need to keep track of TotalDealCollateral in market actor (similar to TotalPledgeCollateral).

aarshkshah1992:
Closes #531 

@anorth Needs tests. Will put in along with the tests for the other PRs as they will share the test utilities.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -222,7 +222,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 
 			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientCollateral, ClientCollateral)
 			st.lockBalanceOrAbort(rt, client, deal.Proposal.TotalStorageFee(), ClientStorageFee)
-			st.lockBalanceOrAbort(rt, client, deal.Proposal.ProviderCollateral, ProviderCollateral)
+			st.lockBalanceOrAbort(rt, provider, deal.Proposal.ProviderCollateral, ProviderCollateral)
 
 			id := st.generateStorageDealID()
 
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -221,7 +221,11 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			deal.Proposal.Client = client
 
 			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientBalanceRequirement())
+			st.TotalLockedClientCollateral = big.Add(st.TotalLockedClientCollateral, deal.Proposal.ClientCollateral)
+			st.TotalLockedDealPayments = big.Add(st.TotalLockedDealPayments, deal.Proposal.TotalStorageFee())
+
 			st.lockBalanceOrAbort(rt, provider, deal.Proposal.ProviderBalanceRequirement())
+			st.TotalLockedProviderCollateral = big.Add(st.TotalLockedProviderCollateral, deal.Proposal.ProviderCollateral)
 
 			id := st.generateStorageDealID()
 
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -220,12 +220,9 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			deal.Proposal.Provider = provider
 			deal.Proposal.Client = client
 
-			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientBalanceRequirement())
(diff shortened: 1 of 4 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -217,7 +217,11 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			deal.Proposal.Client = client
 
 			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientBalanceRequirement())
+			st.TotalLockedClientCollateral = big.Add(st.TotalLockedClientCollateral, deal.Proposal.ClientCollateral)
+			st.TotalLockedDealPayments = big.Add(st.TotalLockedDealPayments, deal.Proposal.TotalStorageFee())
```
anorth, line 221: I think the state object should maintain the invariants, so this logic should be moved into the state methods. Probably that means changing the `lockBalanceOrAbort` to distinguish the two parts of client lock-up.

So how about `st.LockValue(party, amount, role)` where role is an enum of 3 options: provider collateral, client collateral, storage fee. Then we'll need three calls to that function here for all the lock-ups.

Note also that `lockBalanceOrAbort` is an example of the anti-pattern of passing the `rt` just for access to `Abortf`. Change it to return `(ok, error)` instead. The precondition check that the proposal's values are positive (`ErrIllegalArgument`) should move out to the caller, here. Then `ok=false` indicates `ErrInsufficientFunds`.

actors/builtin/market/market_state.go
```diff
@@ -22,6 +22,12 @@ const epochUndefined = abi.ChainEpoch(-1)
 // activate deal (miner)
 // end deal (miner terminate, expire(no activation))
 
+type role int
+const (
+	client role = iota
+	provider
```
anorth, line 28: `CAPITALIZE` these please, they're otherwise confusing to see in the code (where client or provider usually refers to some address)

aarshkshah1992: This is done.

actors/builtin/market/market_actor.go
aarshkshah1992: @anorth I think I see hat you mean. I've introduced a `BalanceLockingReason` enum with values as 

* ClientCollateral
* ClientStorageFee
* ProviderCollateral

Both `st.LockFunds()` and `st.UnlockFunds` now take this enum so as to add/subtract to the corresponding state variables. 

Please note that I do not want to make the change of not aborting in "helper functions" as we plan to do it in #455 .

aarshkshah1992:
@anorth I've made the changes. Please take a look.

actors/builtin/market/market_actor.go
```diff
@@ -220,8 +220,9 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			deal.Proposal.Provider = provider
 			deal.Proposal.Client = client
 
-			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientBalanceRequirement())
-			st.lockBalanceOrAbort(rt, provider, deal.Proposal.ProviderBalanceRequirement())
+			st.lockBalanceOrAbort(rt, client, deal.Proposal.ClientCollateral, ClientCollateral)
+			st.lockBalanceOrAbort(rt, client, deal.Proposal.TotalStorageFee(), ClientStorageFee)
+			st.lockBalanceOrAbort(rt, client, deal.Proposal.ProviderCollateral, ProviderCollateral)
```
anorth, line 225: Blocking:
```suggestion
			st.lockBalanceOrAbort(rt, provider, deal.Proposal.ProviderCollateral, ProviderCollateral)
```

actors/builtin/market/market_state.go
```diff
@@ -289,13 +318,23 @@ func (st *State) maybeLockBalance(rt Runtime, addr addr.Address, amount abi.Toke
...
 	err := lt.MustSubtract(addr, amount)
 	if err != nil {
 		return xerrors.Errorf("subtracting from locked balance: %v", err)
 	}
+
+	switch lockReason {
+	case ClientCollateral:
+		st.TotalLockedClientCollateral = big.Sub(st.TotalLockedClientCollateral, amount)
+	case ClientStorageFee:
+		st.TotalLockedDealPayments = big.Sub(st.TotalLockedDealPayments, amount)
```
anorth, line 333: Unify the terminology here between the reason and the state variable. I suggest `st.TotalClientStorageFee`.

And just for super consistency, how about `TotalClientLockedCollateral` and `TotalProviderLockedCollateral` too.

aarshkshah1992: This is done.

actors/builtin/market/market_actor.go
aarshkshah1992: Great catch. This is done.

PR #558, opened 2020-06-30
Slash deals immediately on sector termination

### LINKED ISSUE #462: Process deal slashing immediately when sectors terminate early
From builtin/market:OnMinerSectorsTerminate:
```
// Note: we do not perform the balance transfers here, but rather simply record the flag
// to indicate that processDealSlashed should be called when the deferred state computation
// is performed. // TODO: Do that here

```

aarshkshah1992:
Closes #462 

Also, do not slash deals that have expired.

@anorth Ofcourse, tests.

### DIFF
--- a/actors/builtin/market/market_state.go
+++ b/actors/builtin/market/market_state.go
@@ -2,6 +2,7 @@ package market
 
 import (
 	"bytes"
+	"fmt"
 
 	addr "github.com/filecoin-project/go-address"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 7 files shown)


actors/builtin/market/market_state.go
```diff
@@ -96,35 +90,37 @@ func (st *State) updatePendingDealState(rt Runtime, state *DealState, deal *Deal
...
-		}
-
-		// slash provider collateral
-		amountSlashed = deal.ProviderCollateral
-		if err := st.slashBalance(et, lt, deal.Provider, amountSlashed); err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "slashing balance: %s", err)
-		}
-
-		st.deleteDeal(rt, dealID)
-		return amountSlashed, epochUndefined
```
anorth, line 114: I think this introduces a bug. Previously, slashed deals would return here and not enter `processDealExpired` further down. This will now unlock the provider collateral twice.

We need to make sure the expiration processing doesn't do something redundant. I think that either means restoring a flag to the deal state, or deleting the deal state entirely when slashed. Minor preference for the first way.

This might mean that the issue description motivating this change, "Slash deals immediately", was a poor direction. We still need to fix the bug about not slashing deals that are finished, but maybe through a smaller change.

aarshkshah1992:
Closed in favour of #559 .

actors/builtin/market/market_state.go
aarshkshah1992: @anorth I agree with you. I think keeping the logic in one place also makes it easy to reason about. Have raised #559 .

PR #559, opened 2020-06-30
Do not slash expired deals and also delete deal state when deal proposal is deleted

### LINKED ISSUE #462: Process deal slashing immediately when sectors terminate early

aarshkshah1992:
Closes #462

@anorth 

Also, discovered a bug wherein we delete the deal proposal but not the deal state.

### DIFF
--- a/actors/builtin/market/market_state.go
+++ b/actors/builtin/market/market_state.go
@@ -2,6 +2,7 @@ package market
 
 import (
 	"bytes"
+	"fmt"
 
 	addr "github.com/filecoin-project/go-address"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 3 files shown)


actors/builtin/market/market_state.go
```diff
@@ -148,10 +148,32 @@ func (st *State) mutateDealProposals(rt Runtime, f func(*DealArray)) {
 	st.Proposals = rcid
 }
 
+func (st *State) mutateDealStates(rt Runtime, f func(*DealMetaArray)) {
+	states, err := AsDealStateArray(adt.AsStore(rt), st.States)
+	if err != nil {
+		rt.Abortf(exitcode.ErrIllegalState, "failed to load deal states array: %s", err)
+	}
```
anorth, line 155: Would prefer not to add more uses of `rt` in state: return an error instead

aarshkshah1992: This is done.

PR #560, opened 2020-06-30
Test for publishing a deal after activating a previous deal with a start epoch in the future

### LINKED ISSUE #265: Publish successive deals fails (possible incorrect assertion in SMA market_state)
Issue: Publishing deals fails because market.State.updatePendingDealStatesForParty fails, if a miner has an existing deal that has been sealed but has not reached the StartEpoch in the proposal.

Steps to reproduce:
1. Publish a deal for a miner, put StartEpoch far in future, activate storage deal after sealing, but before StartEpoch.
2. Publish a second deal for the miner, which calls market.State.updatePendingDealStatesForParty which fails on the assertion here: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_state.go#L110 -- which seems incorrect -- a sector may be proved and activated before the StartEpoch?

Lotus experiences has assembled a failing test based on this issue: https://github.com/filecoin-project/lotus/pull/1481

aarshkshah1992:
Closes #265

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -321,6 +321,42 @@ func TestMarketActor(t *testing.T) {
 	})
 }
 
+func TestPublishStorageDeals(t *testing.T) {
+	marketActor := tutil.NewIDAddr(t, 100)
+	owner := tutil.NewIDAddr(t, 101)
+	provider := tutil.NewIDAddr(t, 102)
+	worker := tutil.NewIDAddr(t, 103)
+	client := tutil.NewIDAddr(t, 104)
+
+	t.Run("publish a deal after activating a previous deal which has a start epoch far in the future", func(t *testing.T) {
+		startEpoch := abi.ChainEpoch(1000)
+		endEpoch := abi.ChainEpoch(2000)
+		publishEpoch := abi.ChainEpoch(1)
+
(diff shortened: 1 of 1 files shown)


actors/builtin/market/market_test.go
```diff
@@ -321,6 +321,42 @@ func TestMarketActor(t *testing.T) {
...
+		deal1 := generateDealProposal(client, provider, startEpoch, endEpoch)
+
+		// ensure client and provider have enough funds to lock for the deal
+		actor.addParticipantFunds(rt, client, deal1.ClientBalanceRequirement())
+		actor.addProviderFunds(rt, deal1.ProviderBalanceRequirement())
+
+		// publish the deal and activate it
+		rt.SetEpoch(publishEpoch)
+		deal1ID := actor.publishDeal(rt, deal1)
+		actor.activateDeals(rt, []abi.DealID{deal1ID}, endEpoch+1, provider)
```
anorth, line 346: Aside: is the `+1` necessary? I think we should be allowing a sector to terminate in the same epoch as it's longest deal. Could you follow up to investigate that? Remove the +1 now if possible.

aarshkshah1992: @anorth This is done.

aarshkshah1992: I'll investigate this.

PR #562, opened 2020-06-30
Defensive cron error handling

### LINKED ISSUE #452: Defend against broken cron callback in miner
If the miner code doesn't have any bugs, then the world is a happy place. However, if there are bugs transitively reached by the cron handler such that the handler aborts (and rolls back state), a miner is frozen. This is bad because the miner can't operate, but worse because they also maintain power while being freed from the requirement of submitting PoSts.

This is an area where defensive programming is called for. While we should strive to eliminate abort cases from the miner's cron handler, the power actor should probably remove the miner from the power table if something does go wrong, so at least it fails conservatively.

ZenGround0:
Closes #452

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -154,7 +154,6 @@ type UpdateClaimedPowerParams struct {
 func (a Actor) UpdateClaimedPower(rt Runtime, params *UpdateClaimedPowerParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 	minerAddr := rt.Message().Caller()
-
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
 		err := st.AddToClaim(adt.AsStore(rt), minerAddr, params.RawByteDelta, params.QualityAdjustedDelta)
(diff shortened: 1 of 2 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -472,15 +471,39 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
...
 			abi.NewTokenAmount(0),
 		)
-		// The exit code is ignored. If a callback fails, this actor continues to invoke other callbacks
+		// If a callback fails, this actor continues to invoke other callbacks
 		// and persists state removing the failed event from the event queue. It won't be tried again.
-		// A log message would really help here, though.
+		// Failures are unexpected here but will result in removal of miner power
+		// A log message would really help here.
+		if code != exitcode.Ok {
+			rt.State().Transaction(&st, func() interface{} {
```
anorth, line 485: Accumulate the failed miner addresses and then do just one state transaction, with a loop _inside_ it.

actors/builtin/power/power_actor.go
```diff
@@ -472,15 +471,39 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
...
+				store := adt.AsStore(rt)
+				// Remove power and leave miner frozen
+				claim, found, err := st.GetClaim(store, event.MinerAddr)
+				if err != nil {
+					// TODO #564 log this, failing without deducting power: bad
+					return nil
+				}
+				if !found {
+					// TODO #564 log this, failing without deducting power, but power doesn't exist so maybe ok?
+					return nil
```
anorth, line 495: I can totally see this happening if a miner is deleted. Yes we should log but it's probably fine (the exit code should indicate this).

anorth:
FYI @Stebalien

actors/builtin/power/power_actor.go
```diff
@@ -472,15 +471,39 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
...
 			builtin.MethodsMiner.OnDeferredCronEvent,
 			vmr.CBORBytes(event.CallbackPayload),
 			abi.NewTokenAmount(0),
 		)
-		// The exit code is ignored. If a callback fails, this actor continues to invoke other callbacks
+		// If a callback fails, this actor continues to invoke other callbacks
 		// and persists state removing the failed event from the event queue. It won't be tried again.
-		// A log message would really help here, though.
+		// Failures are unexpected here but will result in removal of miner power
+		// A log message would really help here.
```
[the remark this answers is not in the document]
Stebalien: Ah, given https://github.com/filecoin-project/specs-actors/pull/562#discussion_r448070056 we might not want to do that.

PR #567, opened 2020-06-30
Skipped Faults

### LINKED ISSUE #410: Implement Window PoSt "skipped" faults
A follow-on from the initial Window PoSt implementation is to implement handling of faults that are declared concurrently with a PoSt submission. Such faults pay the same per-sector penalty as if the fault was undeclared and the PoSt missed, but allow the PoSt for other sectors in the same partition to succeed.

If we don't do this, then when a miner fails to declare a single sector fault in a partition before the cutoff, the whole partition is penalised.

acruikshank:
closes #410

### Motivation

The `Skipped` field in `SubmitWindowedPoSt` is currently ignored. This field is important because it allows miners to signal which sectors have faulted late without losing the entire PoSt. This PR adds the functionality to penalize skipped faults and add them to declared faults so that the proof verification will not fail.

### Proposed Changes

1. Introduce `processSkippedFaults` call into `SubmitWindowedPoSt` to handle the `Skipped` parameter.
2. Add logic to handle intersection with recoveries.
3. Add logic to add skipped faults to `Faults` and the current `FaultEpoch`.
4. Add logic to compute an undeclared fault penalty for the skipped faults, unlock that amount and burn it.
5. Add logic to compute power delta for unlocked faults and notify power actor of the change.
6. Test.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1230,7 +1230,7 @@ func processSkippedFaults(rt Runtime, st *State, store adt.Store, currDeadline *
 	}
 
 	currEpoch := rt.CurrEpoch()
-	var skippedFaultSectors []*SectorOnChainInfo
+	var newFaultSectors []*SectorOnChainInfo
 
 	// Check that the declared sectors are actually due at the deadline.
 	deadlineSectors := deadlines.Due[currDeadline.Index]
@@ -1252,24 +1252,24 @@ func processSkippedFaults(rt Runtime, st *State, store adt.Store, currDeadline *
 	err = st.AddFaults(store, newFaults, currDeadline.PeriodStart)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add skipped faults")
 
-	// Load info for sectors.
-	skippedFaultSectors, err = st.LoadSectorInfos(store, newFaults)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load fault sectors")
-
-	// Unlock undeclared penalty for faults (skipped faults including retracted recoveries)
-	retractedRecoveryInfos, err := st.LoadSectorInfos(store, retractedRecoveries)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load retracted recovery secrtors")
-
 	// Remove faulty recoveries
 	err = st.RemoveRecoveries(retractedRecoveries)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to remove recoveries")
 
+	// Load info for sectors.
+	newFaultSectors, err = st.LoadSectorInfos(store, newFaults)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load fault sectors")
+
+	// load info for retractions
+	retractedRecoveryInfos, err := st.LoadSectorInfos(store, retractedRecoveries)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load retracted recovery secrtors")
+
 	// Penalize new skipped faults and retracted recoveries
-	penalizeFaultSectors := append(skippedFaultSectors, retractedRecoveryInfos...)
+	penalizeFaultSectors := append(newFaultSectors, retractedRecoveryInfos...)
 	penalty, err := unlockUndeclaredFaultPenalty(st, store, minerInfo.SectorSize, currEpoch, epochReward, currentTotalPower, penalizeFaultSectors)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to charge fault fee")
 
-	return skippedFaultSectors, penalty
+	return newFaultSectors, penalty
 }
 
 // Detects faults from missing PoSt submissions that did not arrive by some deadline, and moves
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -274,7 +274,7 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		newFaultSectors = append(newFaultSectors, detectedFaultSectors...)
 
 		// Add skipped as faults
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -857,6 +861,59 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
...
+	}
+
+	// Find all skipped faults that have been labeled recovered
+	retractedRecoveries, err := bitfield.IntersectBitField(st.Recoveries, skipped)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to intersect sectors with recoveries")
+
+	// Ignore skipped faults that are already faults but have not been declared recovered
+	newFaults, err := bitfield.SubtractBitField(skipped, st.Faults)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to subtract existing faults from skipped")
+	penalizableFaults, err := bitfield.MergeBitFields(newFaults, retractedRecoveries)
```
anorth, line 895: Blocking: As noted in the issue, I think we'll need to charge penalty tokens for retracted recoveries, but not deduct power, so we'll need to track these carefully separately. 

The return value `skippedFaultSectors` currently contains all of these. We need to load sector info for retractions to calculate fee, but exclude from power.

actors/builtin/miner/miner_test.go
```diff
@@ -417,6 +413,68 @@ func TestWindowPost(t *testing.T) {
...
+
+		partitions := make([]uint64, partitionCount)
+		for i := uint64(0); i < partitionCount; i++ {
+			partitions[i] = firstPartIdx + i
+		}
+
+		partitionsSectors, err := miner.ComputePartitionsSectors(deadlines, partitionSize, deadline.Index, partitions)
+		require.NoError(t, err)
+		provenSectors, err := abi.BitFieldUnion(partitionsSectors...)
+		require.NoError(t, err)
```
anorth, line 451: There's probably some part of the above code to break out to helpers functions that we'll use in other tests too.

actors/builtin/miner/miner_test.go
```diff
@@ -417,6 +413,68 @@ func TestWindowPost(t *testing.T) {
...
+		cfg := &skipConfig{
+			skipped:               *skipped,
+			expectedRawPowerDelta: rawPower.Neg(),
+			expectedQAPowerDelta:  qaPower.Neg(),
+			expectedPenalty:       expectedPenalty,
+		}
+
+		actor.submitWindowPost(rt, deadline, partitions, infos, cfg)
+
+	})
```
anorth, line 477: Please also test the handling for skipping a recovery.

actors/builtin/miner/miner_test.go
```diff
@@ -1083,15 +1143,30 @@ func (h *actorHarness) commitAndProveSectors(rt *mock.Runtime, n int, lifetimePe
...
+	st := getState(rt)
+	deadline := st.DeadlineInfo(rt.Epoch())
+	rt.SetEpoch(deadline.PeriodEnd())
+	nextCron := deadline.NextPeriodStart() + miner.WPoStProvingPeriod - 1
+	h.onProvingPeriodCron(rt, nextCron, true)
+	rt.SetEpoch(deadline.NextPeriodStart())
+}
+
+type skipConfig struct {
+	skipped               bitfield.BitField
```
anorth, line 1156: Use a pointer for bitfields, they don't copy safely

actors/builtin/miner/miner_actor.go
```diff
@@ -857,6 +861,59 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
...
+	if !contains {
+		rt.Abortf(exitcode.ErrIllegalArgument, "skipped faults contains sectors not due in deadline %d", deadline.Index)
+	}
+
+	// Find all skipped faults that have been labeled recovered
+	retractedRecoveries, err := bitfield.IntersectBitField(st.Recoveries, skipped)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to intersect sectors with recoveries")
+
+	// Ignore skipped faults that are already faults but have not been declared recovered
+	newFaults, err := bitfield.SubtractBitField(skipped, st.Faults)
```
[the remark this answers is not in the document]
acruikshank: The highlighted line is only half the operation. The next line restores the retracted recoveries. It's `(S ∖ F) ∪ (S ∩ R)` instead of `S ∖ (F ∖ R)`. Since `R ⊆ F` they're equivalent, and we already have the right side. It's moot though since we need to track skipped faults and skipped recoveries separately.

actors/builtin/miner/miner_test.go
```diff
@@ -1404,18 +1552,64 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
...
 		Proofs:     proofs,
-		Skipped:    bitfield.BitField{},
+		Skipped:    skipped,
 	}
 
 	rt.Call(h.a.SubmitWindowedPoSt, &params)
 	rt.Verify()
 }
 
+func (h *actorHarness) computePartitions(rt *mock.Runtime, deadlines *miner.Deadlines, deadline *miner.DeadlineInfo) ([]*miner.SectorOnChainInfo, []uint64) {
```
anorth, line 1588: `Index` is the only use of `deadline`, so just take the index directly. That will also make it a bit clearer what this function does.

Please document it. I think it's something like "Computes  the set of all partition numbers, and all non-faulty sector infos included in those partitions, due at a single deadline".

actors/builtin/miner/miner_test.go
```diff
@@ -1384,15 +1513,34 @@ func (h *actorHarness) submitWindowPost(rt *mock.Runtime, deadline *miner.Deadli
 		rt.ExpectGetRandomness(crypto.DomainSeparationTag_WindowedPoStChallengeSeed, deadline.Challenge, buf.Bytes(), abi.Randomness(challengeRand))
 	}
 	{
+		var goodInfo *miner.SectorOnChainInfo
```
anorth, line 1516: A brief explanation of intent, please. "Find the first non-faulty sector in the post, which is to be substituted for all the skipped fault sector infos".

PR #568, opened 2020-06-30
fix: use MultiMerge from the bitfield library

### LINKED ISSUE #460: Optimize Bitfield Union
From abi/bitfield.go:BitFieldUnion: `// TODO: optimize me`

Stebalien:
This function employs the same strategy, but stream-merges instead of merging each pair, serializing, deserializing, merging the next pair, etc.

fixes #460

----

#

### DIFF
--- a/actors/abi/bitfield.go
+++ b/actors/abi/bitfield.go
@@ -21,25 +21,7 @@ func BitFieldUnion(bfs ...*BitField) (*BitField, error) {
 	if len(bfs) == 0 {
 		return NewBitField(), nil
 	}
-	// TODO: optimize me: https://github.com/filecoin-project/specs-actors/issues/460
-	for len(bfs) > 1 {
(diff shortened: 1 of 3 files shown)


actors/abi/bitfield.go
```diff
@@ -21,25 +21,7 @@ func BitFieldUnion(bfs ...*BitField) (*BitField, error) {
...
-			if err != nil {
-				return nil, err
-			}
-
-			next = append(next, merged)
-		}
-		bfs = next
-	}
-	return bfs[0], nil
+	return bitfield.MultiMerge(bfs...)
```
anorth, line 24: Thanks. Could you in fact remove `BitFieldUnion` and inline this (unless the former doesn't handle empty args)?

Stebalien: It doesn't handle empty args (yet).

PR #572, opened 2020-07-01
Handle already terminated sectors in termination logic

### LINKED ISSUE #513: Sector expiration doesn't handle already-terminated sectors
The miner actor maintains a queue of sector expirations by epoch. This queue is processed in the proving period cron callback, where the sector numbers to terminate are popped from the queue and passed to `terminateSectors`. The latter call expects the terminating sectors to exist in state, and will abort if they don't.

There are a number of ways that a sector might have been removed from state before expiring:
- terminated for being faulty for too long
- terminated by explicit request of the miner
- committed capacity sector replaced by a new one (#485)

To resolve this, either 
- all paths which remove sectors from state must also remove them from the expiration queue (which requires loading the sector info to look up the expiration), or
- termination of sectors must handle a sector already being gone

The second way seems more robust. Some defensive programming is not out of place here, though callers will need to be careful not to assume that the termination implies the sectors previously existed. In the case of very large sets of long-faulty sectors, loading all the sector-infos in order to know the expiration epochs from which to deregister the sectors is exactly the kind of thing we're trying to avoid in #411.

Stebalien:
Ignore the termination if:

1. The sector doesn't exist.
2. We're terminating it due to an expiration, but it hasn't actually expired.

fixes #513

TODO:

* [x] Tests
* [x] Address XXX.
* [x] Consider eagerly removing sector expirations.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1582,6 +1582,7 @@ func checkPrecommitExpiry(rt Runtime, sectors *abi.BitField) {
 // TODO: red flag that this method is potentially super expensive
 // https://github.com/filecoin-project/specs-actors/issues/483
 func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power.SectorTermination, epochReward abi.TokenAmount, currentTotalPower abi.StoragePower) {
+	currentEpoch := rt.CurrEpoch()
 	empty, err := sectorNos.IsEmpty()
 	if err != nil {
 		rt.Abortf(exitcode.ErrIllegalState, "failed to count sectors")
@@ -1594,7 +1595,7 @@ func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power
 	var st State
 
 	var dealIDs []abi.DealID
-	var allSectors []*SectorOnChainInfo
+	var terminatedSectors []*SectorOnChainInfo
 	var faultySectors []*SectorOnChainInfo
 	penalty := abi.NewTokenAmount(0)
 	var info *MinerInfo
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1541,18 +1542,39 @@ func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power
...
 			}
+
+			// If the sector wasn't found, skip termination. It may
+			// have already expired, have been terminated due to a
+			// faults, etc.
 			if !found {
-				rt.Abortf(exitcode.ErrNotFound, "no sector %v", sectorNo)
+				// XXX: Should we fail if the termination reason
+				// is "manual termination"? Or should we just
+				// let termination be idempotent?
```
anorth, line 1559: Let's just let it go. The manual declaration might have raced something else, potentially with a looong blockchain delay. We shouldn't deny any other terminations they're trying to achieve at the same time.

actors/builtin/miner/miner_actor.go
```diff
@@ -1541,18 +1542,39 @@ func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power
 			rt.Abortf(exitcode.ErrIllegalState, "failed to expand faults")
 		}
 
+		var skippedSectorNos []uint64 // sectors we're not terminating because they either don't exist or have expired.
```
anorth, line 1545: Perhaps "missing"? Just because "skipped" has a specific meaning wrt faults.

actors/builtin/miner/miner_actor.go
```diff
@@ -1541,18 +1542,39 @@ func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power
...
+				skippedSectorNos = append(skippedSectorNos, sectorNo)
+				return nil
+			}
+
+			// If we're terminating the sector because it expired,
+			// make sure it has actually expired.
+			//
+			// We can hit this case if the sector was changed
+			// somehow but the termination wasn't removed from the
+			// SectorExpirations queue.
```
anorth, line 1569: I'm leery of this one. Is there a concrete case you know of? If it wasn't removed from the queue, it most likely wasn't added back to the queue somewhere else, so will never expire. We need the two representations of expiration to be consistent.

OTOH, I guess we could take advantage of this to avoid re-writing the queue entries to remove entries when we extend an expiration? We'll be paying an unnecessary loading of sector info here though.

actors/builtin/miner/miner_actor.go
```diff
@@ -1541,18 +1542,39 @@ func terminateSectors(rt Runtime, sectorNos *abi.BitField, terminationType power
...
+			if terminationType == power.SectorTerminationExpired && currentEpoch < sector.Expiration {
+				// XXX: Should this be <=? Where is the boundary.
```
anorth, line 1571: It should expire during the expiration epoch, so `<` is correct here.

actors/builtin/miner/miner_actor.go
Stebalien: Sector number reuse? We don't currently (see TODOs) remove scheduled terminations when we terminate a sector early. If we then add a new sector with the same number, we could hit this case.

actors/builtin/miner/miner_actor.go
Stebalien: `notTerminated`? It's not quite "missing" because we also skip non-expired sectors.

Stebalien:
Based on #586 for now.

PR #573, opened 2020-07-02
Tests for PublishStorageDeals

### LINKED ISSUE #542: Market Actor: Write tests for PublishStorageDeals
It currently has NO tests.

@anorth Please could you give the appropriate labels to this and I'll prioritise this accordingly.

aarshkshah1992:
Closes #542 .

Tests for multiple calls to publish with different providers in the next PR.

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -258,7 +258,7 @@ func TestMarketActor(t *testing.T) {
 
 			// publish the deal so that client AND provider collateral is locked
 			rt.SetEpoch(publishEpoch)
-			actor.publishDeal(rt, deal)
+			actor.publishDeals(rt, []market.DealProposal{deal})
 			rt.GetState(&st)
 			require.Equal(t, deal.ProviderCollateral, st.GetLockedBalance(rt, provider))
 			require.Equal(t, deal.ClientBalanceRequirement(), st.GetLockedBalance(rt, client))
@@ -293,7 +293,7 @@ func TestMarketActor(t *testing.T) {
 
 			// publish the deal
 			rt.SetEpoch(publishEpoch)
-			dealID := actor.publishDeal(rt, deal)
+			dealID := actor.publishDeals(rt, []market.DealProposal{deal})[0]
 
 			// activate the deal
 			actor.activateDeals(rt, []abi.DealID{dealID}, endEpoch+1, provider)
@@ -342,7 +342,7 @@ func TestPublishStorageDeals(t *testing.T) {
 
 		// publish the deal and activate it
 		rt.SetEpoch(publishEpoch)
-		deal1ID := actor.publishDeal(rt, deal1)
+		deal1ID := actor.publishDeals(rt, []market.DealProposal{deal1})[0]
 		actor.activateDeals(rt, []abi.DealID{deal1ID}, endEpoch, provider)
 		st := actor.getDealState(rt, deal1ID)
 		require.EqualValues(t, publishEpoch, st.SectorStartEpoch)
@@ -352,7 +352,7 @@ func TestPublishStorageDeals(t *testing.T) {
 		actor.addParticipantFunds(rt, client, deal2.ClientBalanceRequirement())
 		actor.addProviderFunds(rt, deal2.ProviderBalanceRequirement())
 		rt.SetEpoch(publishEpoch + 1)
-		deal2ID := actor.publishDeal(rt, deal2)
+		deal2ID := actor.publishDeals(rt, []market.DealProposal{deal2})[0]
 		actor.activateDeals(rt, []abi.DealID{deal2ID}, endEpoch+1, provider)
 	})
 }
@@ -379,7 +379,7 @@ func TestMarketActorDeals(t *testing.T) {
 
 	// First attempt at publishing the deal should work
 	{
-		actor.publishDeal(rt, dealProposal)
+		actor.publishDeals(rt, []market.DealProposal{dealProposal})
 	}
 
 	// Second attempt at publishing the same deal should fail
@@ -400,7 +400,7 @@ func TestMarketActorDeals(t *testing.T) {
 
 	// Same deal with a different label should work
 	{
-		actor.publishDeal(rt, dealProposal)
+		actor.publishDeals(rt, []market.DealProposal{dealProposal})
 	}
 }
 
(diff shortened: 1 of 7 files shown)


actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
-		deal2ID := actor.publishDeal(rt, deal2)
+		deal2ID := actor.publishDeals(rt, deal2)[0]
 		actor.activateDeals(rt, []abi.DealID{deal2ID}, endEpoch+1, provider)
 	})
+
+	t.Run("publish multiple deals for different clients and ensure balances are correct", func(t *testing.T) {
+		rt, actor := basicMarketSetup(t, marketActor, owner, provider, worker, client)
+		client1 := tutil.NewIDAddr(t, 900)
+		client2 := tutil.NewIDAddr(t, 901)
+		client3 := tutil.NewIDAddr(t, 902)
```
anorth, line 365: I suggest pulling these out of the `t.Run` and re-using

actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
+
+		// generate first deal for
+		deal1 := generateDealProposal(client1, provider, abi.ChainEpoch(20), abi.ChainEpoch(50))
+		actor.addProviderFunds(rt, deal1.ProviderCollateral)
+		actor.addParticipantFunds(rt, client1, deal1.ClientBalanceRequirement())
```
anorth, line 370: Since this is so common, I suggest either:
- a harness method that does these three, or
- an  `addFundsAndPublishDeals` that adds the requisite funds first

actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
+
+		// generate second deal
+		deal2 := generateDealProposal(client2, provider, abi.ChainEpoch(25), abi.ChainEpoch(60))
```
anorth, line 373: Are the start/end epochs different for a reason? The differing numbers makes me think they are relevant to the test, but it appears not. If they're not, just use obviously-irrelevant numbers like `100`, 1000`, `42`, and the same for each deal.

actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
+
+		// generate third deal
+		deal3 := generateDealProposal(client3, provider, abi.ChainEpoch(30), abi.ChainEpoch(70))
+		actor.addProviderFunds(rt, deal3.ProviderCollateral)
+		actor.addParticipantFunds(rt, client3, deal3.ClientBalanceRequirement())
+
+		actor.publishDeals(rt, deal1, deal2, deal3)
+
+		// assert locked balance for all clients and provider
+		providerLocked := big.Add(big.Add(deal1.ProviderCollateral, deal2.ProviderCollateral), deal3.ProviderCollateral)
```
anorth, line 385: Let's write a `big.Sum(a, bs...)` method since this is common. The naive implementation is ok.

actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
+			"deal end after deal start": {
+				setup: func(_ *mock.Runtime, _ *marketActorTestHarness, d *market.DealProposal) {
+					d.StartEpoch = 10
+					d.EndEpoch = 9
+				},
+				exitCode: exitcode.ErrIllegalArgument,
+			},
+			"current epoch greater than start epoch": {
+				setup: func(rt *mock.Runtime, _ *marketActorTestHarness, d *market.DealProposal) {
+					rt.SetEpoch(d.StartEpoch + 1)
```
anorth, line 457: Rather than adjust the rt epoch somewhat haphazardly between tests, set the deal epoch relevant to the rt epoch, and leave the rt epoch fixed. That probably means you can use a constant for the rt epoch and remove `rt. from the signature here.

actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
+				setup: func(rt *mock.Runtime, a *marketActorTestHarness, d *market.DealProposal) {
+					a.addParticipantFunds(rt, client, big.Sub(d.ClientBalanceRequirement(), big.NewInt(1)))
+					a.addProviderFunds(rt, d.ProviderCollateral)
+				},
+				exitCode: exitcode.ErrInsufficientFunds,
+			},
+			"provider does not have enough balance for collateral": {
+				setup: func(rt *mock.Runtime, a *marketActorTestHarness, d *market.DealProposal) {
+					a.addParticipantFunds(rt, client, d.ClientBalanceRequirement())
+					a.addProviderFunds(rt, big.Sub(d.ProviderCollateral, big.NewInt(1)))
```
anorth, line 514: This is ok, but also consider doing this the other way too: set up the provider/participant with some funds, fixed for all cases, then change the deal to cost just too much.

actors/builtin/market/market_test.go
```diff
@@ -352,9 +354,289 @@ func TestPublishStorageDeals(t *testing.T) {
...
+
+				},
+				exitCode:                   exitcode.ErrIllegalArgument,
+				signatureVerificationError: errors.New("error"),
+			},
+			"no entry for client in locked  balance table": {
+				setup: func(rt *mock.Runtime, a *marketActorTestHarness, d *market.DealProposal) {
+					a.addProviderFunds(rt, d.ProviderCollateral)
+				},
+				exitCode: exitcode.ErrIllegalState,
```
anorth, line 535: This is unfortunate. Is it possible to get the code to return ErrInsufficientFunds here too? (same below)

actors/builtin/market/market_test.go
```diff
@@ -566,6 +876,24 @@ func (h *marketActorTestHarness) getDealState(rt *mock.Runtime, dealID abi.DealI
 	return s
 }
 
+func (h *marketActorTestHarness) dealsByEpochShouldContain(rt *mock.Runtime, epoch abi.ChainEpoch, dealId abi.DealID) {
```
anorth, line 879: I'd prefer not to do this kind of detailed state inspection. We should instead test that deals are processed correctly over time by exercising the passage of time, not checking in here. We should be able to change this impl a lot without breaking tests.

For now, I suggest omitting this and coming back to test deal lifecycle (which will exercise this)

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: I agree. I have removed this.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

PR #576, opened 2020-07-02
Market: Test Publish with different providers

aarshkshah1992:
@anorth Tagging you as this is a small continuation of the previous PR for market tests.
This PR introduces tests for `PublishStorageDeals` with different providers.

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -55,6 +55,8 @@ func TestMarketActor(t *testing.T) {
 	provider := tutil.NewIDAddr(t, 102)
 	worker := tutil.NewIDAddr(t, 103)
 	client := tutil.NewIDAddr(t, 104)
+	minerAddrs := &minerAddrs{owner, worker, provider}
+
 	var st market.State
 
 	t.Run("simple construction", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


actors/builtin/market/market_test.go
```diff
@@ -404,6 +407,36 @@ func TestPublishStorageDeals(t *testing.T) {
 
 		totalStorageFee = big.Sum(totalStorageFee, deal4.TotalStorageFee(), deal5.TotalStorageFee())
 		require.EqualValues(t, totalStorageFee, st.TotalClientStorageFee)
+
+		// PUBLISH DEALS with a different provider
+		owner := tutil.NewIDAddr(t, 108)
+		provider := tutil.NewIDAddr(t, 109)
+		worker := tutil.NewIDAddr(t, 110)
```
anorth, line 414: Only the provider address need to differ, so the test is slightly stronger if limited to that. Yes, it's realistic for providers to share owner/worker.

actors/builtin/market/market_test.go
```diff
@@ -404,6 +407,36 @@ func TestPublishStorageDeals(t *testing.T) {
...
+
+		// generate second deal for second provider
+		deal7 := actor.generateDealAndAddFunds(rt, client1, miner, abi.ChainEpoch(25), abi.ChainEpoch(60))
+
+		// publish both the deals for the second provider
+		actor.publishDeals(rt, miner, deal6, deal7)
+
+		// assertions
+		rt.GetState(&st)
+		provider2Locked := big.Add(deal6.ProviderCollateral, deal7.ProviderCollateral)
```
anorth, line 428: Check provider1Locked again too

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

Issue #475, opened 2020-06-11
Simplify sector confirmation message flow 


ZenGround0:
Currently `Power.OnSectorProveCommit` is hit by way of a Power -> Miner -> Power cycle.  From builtin/miner/miner_actor.go:ConfirmSectorProofsValid:

```
		// TODO: since this method gets called by the storage power actor
		// initially, can we just do this while we're there?
		// We can probably return the right information from this call to the caller, so it can update there
		// TODO: we should batch these calls...
```

We should evaluate whether it is better to return info needed for power.OnSectorProveCommit back up to the power cron tick per the first TODO

Per the second TODO we should eval doing batch send either in the current configuration or the new one if we end up addressing the first TODO.

anorth:
The flow here has been simplified a lot since the TODO was written. I don't think we should go so far as trying to return info to the power actor to save one call, but we should do all the batching we can.

anorth:
I think you probably need to do this alongside #476

ZenGround0:
Closed by #577

PR #577, opened 2020-07-02
Batch ConfirmSectorProofsValid updates

### LINKED ISSUE #476: Sector Confirmation batch miner pledge updates
From builtin/miner/miner_actor.go:ConfirmSectorProofsValid

```
// Add sector and pledge lock-up to miner state
// TODO: do this all at once after the loop
```

There is a lot going on underneath this code block so it's not obvious how much can be batched.  I think it is referring to the `notifyPledgeChanged` call which currently happens per sector and invokes a method on the storage power actor.  Perhaps it is also possible/good to batch the persector miner state mutations in the code block under this TODO.

ZenGround0:
Prefactor cleaning up `ConfirmSectorProofsValid` before making it more robust to individual sector failures.

Initial traversal over sector number now only does error prone operations over individual sectors and gathers precommits into a slice.  Upcoming robustness work will drop precommits that fail without failing `ConfirmSectorProofsValid`.  This restructure allows batching

1. Power and pledge update calls to power actor into one call
2. All state updates into one transaction (in a loop)
3. Fund accounting operations into one balance transfer per operation

This should close #476 and #475

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -521,7 +521,7 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 	// Committed-capacity sectors licensed for early removal by new sectors being proven.
 	var replaceSectors []*SectorOnChainInfo
 	replaceSectorKeys := map[abi.SectorNumber]struct{}{} // For de-duplication
-
+	var preCommits []*SectorPreCommitOnChainInfo
 	for _, sectorNo := range params.Sectors {
 		precommit, found, err := st.GetPrecommittedSector(store, sectorNo)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-committed sector %v", sectorNo)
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -565,65 +538,96 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
...
+				Expiration:         precommit.Info.Expiration,
+				Activation:         precommit.PreCommitEpoch,
+				DealWeight:         precommit.DealWeight,
+				VerifiedDealWeight: precommit.VerifiedDealWeight,
+				InitialPledge:      initialPledge,
 			}
-			st.AssertBalanceInvariants(rt.CurrentBalance())
+			newSectors = append(newSectors, &newSectorInfo)
 
 			if err := st.PutSector(store, &newSectorInfo); err != nil {
```
ZenGround0, line 586: @anorth it wasn't totally clear but from conversation in #476 it might be that you want to go further and take these four sector state operations out of this loop creating batch calls (`PutSectors`, `DeletePrecommittedSectors`, etc) on the state.  Let me know if that's the case.

actors/builtin/miner/miner_actor.go
```diff
@@ -580,65 +553,96 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
...
-			if err != nil {
-				rt.Abortf(exitcode.ErrIllegalState, "failed to vest new funds: %s", err)
-			}
+		preCommits = append(preCommits, precommit)
+	}
 
-			// Unlock deposit for successful proof, make it available for lock-up as initial pledge.
-			st.AddPreCommitDeposit(precommit.PreCommitDeposit.Neg())
-			st.AddInitialPledgeRequirement(initialPledge)
+	// This transaction should replace the one inside the loop above.
```
anorth, line 573: You can remove this comment now

actors/builtin/miner/miner_actor.go
anorth: I do also want to do that, but defer it to a follow up. It's probably worth a new issue for "batch miner state mutations for sectors, precommits"

actors/builtin/miner/miner_actor.go
```diff
@@ -580,65 +553,96 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
...
+				VerifiedDealWeight: precommit.VerifiedDealWeight,
+				InitialPledge:      initialPledge,
 			}
-			st.AssertBalanceInvariants(rt.CurrentBalance())
+			newSectors = append(newSectors, &newSectorInfo)
 
 			if err := st.PutSector(store, &newSectorInfo); err != nil {
-				rt.Abortf(exitcode.ErrIllegalState, "failed to prove commit: %v", err)
+				rt.Abortf(exitcode.ErrIllegalState, "failed to put sector %v: %v", newSectorInfo.SectorNumber, err)
 			}
```
anorth, line 603: Change these to `RequireNoErr` while you're here

actors/builtin/miner/miner_actor.go
```diff
@@ -580,65 +553,96 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
...
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to replace sector expirations")
 
-			// Lock up initial pledge for new sector.
-			availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
-			if availableBalance.LessThan(initialPledge) {
-				rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds for initial pledge requirement %s, available: %s", initialPledge, availableBalance)
-			}
-			if err := st.AddLockedFunds(store, rt.CurrEpoch(), initialPledge, &PledgeVestingSpec); err != nil {
-				rt.Abortf(exitcode.ErrIllegalState, "failed to add pledge: %v", err)
+		info = getMinerInfo(rt, &st)
```
anorth, line 582: Nit: may as well load this from the readonly state before the transaction.

Issue #578, opened 2020-07-02
Order of removing faults is not deterministic


Kubuxu:
We iterate over the map here:
https://github.com/filecoin-project/specs-actors/blob/a5da78919b5055d2eadc023c6918447544cae8fc/actors/builtin/miner/miner_state.go#L607

Causing the order of fault removal to be non-deterministic which in turn causes state mismatches.
See: https://gist.github.com/Kubuxu/408b35bbf42414032788c480b5f970d0

arajasek:
We probably need to do an audit to make sure this doesn't happen anywhere else in actors code. 

A very quick grep only pulled up [this instance](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go#L579-L586), which has a disclaimer about why it's okay (that probably holds, but could still afford to be moved to some ordered iterating to be safe).

Kubuxu:
The issue is, that case is not okay, with gas in the mix, it has a possibility of causing a state miss-match also.

anorth:
Right, the commented case is not ok, because it might abort at a different point.

whyrusleeping:
aborting at a different point here isnt actually the issue, aborting always causes the same gas usage. The potential issue in the code @arajasek linked is that the difference in insertion order might cause the structure we're writing to to sharder earlier or later than it otherwise might, causing a different number of objects to have to be written. But I think thats likely not a problem either, unless [this](https://github.com/ipfs/go-hamt-ipld/blob/master/hamt.go#L366) slightly suspicious put gets abused.

In any case, best to just force an ordered traversal.

Stebalien:
Reopening as we need to audit all map iteration.

PR #579, opened 2020-07-02
Deterministically update modified fault epochs in-order when removing faults

Stebalien:
See #578.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -575,7 +575,15 @@ func (st *State) RemoveFaults(store adt.Store, sectorNos *abi.BitField) error {
 		return err
 	}
 
-	epochsChanged := map[uint64]*abi.BitField{}
+	type change struct {
+		index uint64
+		value *abi.BitField
+	}
+
+	var (
+		epochsChanged []change
+		epochsDeleted []uint64
+	)
 
 	epochFaultsOld := &abi.BitField{}
 	err = arr.ForEach(epochFaultsOld, func(i int64) error {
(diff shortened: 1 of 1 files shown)


Stebalien:
Is there any way to test for deterministic state access in a unit test?

actors/builtin/miner/miner_state.go
```diff
@@ -594,8 +602,10 @@ func (st *State) RemoveFaults(store adt.Store, sectorNos *abi.BitField) error {
 			return err
 		}
 
-		if countOld != countNew {
-			epochsChanged[uint64(i)] = epochFaultsNew
+		if countNew == 0 {
+			epochsDeleted = append(epochsDeleted, uint64(i))
+		} else if countOld != countNew {
+			epochsChanged = append(epochsChanged, change{index: uint64(i), value: epochFaultsNew})
```
Stebalien, line 608: I assume that _deleting_ elements from an AMT while iterating isn't allowed, but is changing them allowed? That would allow us to update changed bit fields in-place.

Kubuxu:
Just FYI, [this](https://github.com/filecoin-project/specs-actors/commit/2cd72643a5cf921c3f4ff609191cd54cb396f09c) change ended up being compatible with the current chain (non-necessary for specs-actors 0.7.0.

The reason it worked out is that with small maps iteration has high probability of being in order.

Stebalien:
Ah, got it. FYI, you could just append to a single changes array to get the same result without using a map and/or sorting.

Kubuxu:
True, I just wanted to touch as little of that code as possible.

Stebalien:
@Kubuxu could I get a :+1: or a :-1:?

anorth:
I have updated the description to remove the auto-closure of #578, until we audit the other map iterations with a more critical eye than we have previously.

actors/builtin/miner/miner_state.go
anorth: AFAIK it is ok, since the key isn't altered. But no docs :-(

But I think the pattern you have now is more clear, less tricky. If AMT one day supports a batch set, we'll want this anyway.

PR #581, opened 2020-07-02
Robustify Bulk PoRep

### LINKED ISSUE #431: Robustify bulk porep verification to failures
There are a few scenarios where a deferred `ConfirmSectorProofsValid` can abort. Some of them I don't think can be avoided.
- Update `ConfirmSectorProofsValid` to handle some possible but irrecoverable errors (like failure to activate deals) without aborting and rolling back state - just drop the provecommit
    - Alternatively, find a way to move all those conditions to `ProveCommitSector`
- Rework the checking of pledge lock-up and precommit deposit so that any funding-related conditions are checked at `ProveCommitSector` and can't halt the `ConfirmSectorProofsValid`
- Change the power actor to defensively allow the callback for an individual miner to fail without aborting processing for all others (which will be irrecoverable), as it does for cron

See discussion on #340.

ZenGround0:
Closes #431 

Prove commits are validated independently up front.  If any fails validation then it is quietly dropped and we move onto the next prove commit.  After validating we terminate early if all prove commits have been dropped before moving onto batch update.

Keeping WIP because this is based off of #577 which is still in review

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -183,7 +183,7 @@ func TestCommitments(t *testing.T) {
 		// run prove commit logic
 		rt.SetEpoch(precommitEpoch + miner.PreCommitChallengeDelay + 1)
 		rt.SetBalance(big.Mul(big.NewInt(1000), big.NewInt(1e18)))
-		actor.proveCommitSector(rt, precommit, precommitEpoch, makeProveCommit(sectorNo), proveCommitConf{})
+		actor.proveCommitSectorAndConfirm(rt, precommit, precommitEpoch, makeProveCommit(sectorNo), proveCommitConf{})
 		st = getState(rt)
 
 		// expect precommit to have been removed
@@ -317,7 +317,7 @@ func TestCommitments(t *testing.T) {
 
 		// Prove new sector
 		rt.SetEpoch(upgrade.PreCommitEpoch + miner.PreCommitChallengeDelay + 1)
-		newSector := actor.proveCommitSector(rt, &upgrade.Info, upgrade.PreCommitEpoch,
+		newSector := actor.proveCommitSectorAndConfirm(rt, &upgrade.Info, upgrade.PreCommitEpoch,
 			makeProveCommit(upgrade.Info.SectorNumber), proveCommitConf{})
 
 		// Both sectors have pledge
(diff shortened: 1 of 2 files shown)


ZenGround0:
Needs rebase before review, will ping here when done

actors/builtin/miner/miner_test.go
```diff
@@ -758,14 +760,44 @@ func TestProveCommit(t *testing.T) {
...
+		info := actor.getInfo(rt)
+		rt.SetEpoch(precommitEpoch + miner.MaxSealDuration[info.SealProofType] - 1)
+
+		actor.proveCommitSector(rt, precommitA, precommitEpoch, makeProveCommit(sectorNoA))
+		actor.proveCommitSector(rt, precommitB, precommitEpoch, makeProveCommit(sectorNoB))
+
+		conf := proveCommitConf {
+			verifyDealsExit: make(map[abi.SectorNumber]exitcode.ExitCode),
+		}
+		conf.verifyDealsExit[sectorNoA] = exitcode.ErrIllegalArgument
```
anorth, line 799: Nit: compress this into a literal 
```
verifyDealsExit: map[abi.SectorNumber]exitcode.ExitCode{ 
	sectorNoA: ErrIllegalArgument
}
```

actors/builtin/miner/miner_test.go
```diff
@@ -1341,88 +1373,110 @@ func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.SectorPre
...
+			rt.ExpectSend(builtin.StoragePowerActorAddr, builtin.MethodsPower.SubmitPoRepForBulkVerify, &seal, abi.NewTokenAmount(0), nil, exitcode.Ok)
+		}
+		rt.SetCaller(h.worker, builtin.AccountActorCodeID)
+		rt.ExpectValidateCallerAny()
+		rt.Call(h.a.ProveCommitSector, params)
+		rt.Verify()
+}
 
+func (h *actorHarness) confirmSectorProofsValid(rt *mock.Runtime, conf proveCommitConf, precommitEpoch abi.ChainEpoch, precommits ...*miner.SectorPreCommitInfo) {
 	// Prepare for and receive call to ConfirmSectorProofsValid at the end of the same epoch.
```
anorth, line 1426:
```suggestion
	// Prepare for and receive call to ConfirmSectorProofsValid.
```

actors/builtin/miner/miner_test.go
```diff
@@ -1341,88 +1373,110 @@ func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.SectorPre
...
-	{
+	validPrecommits := make([]*miner.SectorPreCommitInfo, 0)
+	allSectorNumbers := make([]abi.SectorNumber, 0)
```
anorth, line 1428: Why empty rather than `nil`?

PR #582, opened 2020-07-02
Charge gas for SubmitPoRepForBulkVerify

### LINKED ISSUE #442: Account for gas fees in SubmitPoRepForBulkVerify
From power_actor.go:SubmitPoRepForBulkVerify:
"//TODO: charge a LOT of gas"

Kubuxu:
Closes #442

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -253,13 +253,14 @@ func (a Actor) OnConsensusFault(rt Runtime, pledgeAmount *abi.TokenAmount) *adt.
 	return nil
 }
 
+const GasOnSubmitVerifySeal = 132166313
+
 func (a Actor) SubmitPoRepForBulkVerify(rt Runtime, sealInfo *abi.SealVerifyInfo) *adt.EmptyValue {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 
 	minerAddr := rt.Message().Caller()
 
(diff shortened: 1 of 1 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -253,13 +253,14 @@ func (a Actor) OnConsensusFault(rt Runtime, pledgeAmount *abi.TokenAmount) *adt.
 	return nil
 }
 
+const GasOnSubmitVerifySeal = 132166313
```
ZenGround0, line 256: Is the meaning behind this number documented anywhere you an link?

Kubuxu: It is `10*time_in_ns` execution time increase of bulk seal verify per seal.
It is derived from lotus bench and our analisys of it:
```
BatchVerifySeals: incr by 0.9884~0.2162; tt 220288665.7433~81346279.4427
	 correll: 0.94, tt = 13216631.32 * extra + 69750883.26
	 covar: 439135248.40, extra: 11.39~5.76, tt2: 220288665.74~81346279.44, count 25929
```
You can see there that `tt` (time taken) has a 0.94 correlation coefficient for the number of seals it verifies, and further in that line there is a formula for the time taken by BatchVerifySeals where `extra` is number of seals to verify.

anorth: A brief note that "this number is empirically determined" would be nice.

PR #584, opened 2020-07-02
older undetected faults penalized 2x SP

### LINKED ISSUE #561: Set missed-PoSt fault fee to 2*SP, support declaring skipped fault for all sectors in partition
Because a fault from a missing PoSt might not be detected until nearly 24hr later, it must pay a higher fee compared to a fault declared at the time PoSt was due (a "skipped fault" attached to a PoSt submission).

If the fault is actually detected in the immediately subsequent deadline during processing of an on-time submission (which processes all faults in previous deadlines), the fee can remain 1*SP. But if more than one deadline late, 2*SP.

TBD: can we ramp the fee more slowly?

Similarly, the termination fee for faulty sectors must be `TF = max{2*SP, ComputedTF}`, assuming termination is delayed until end of proving period.

Ensure that PoSt submission supports declaring _all sectors_ in a partition faulty in this way, so we don't force the miner into a double fee.

We can revert this if we do #552.

cc @nicola

acruikshank:
closes #561

### Motivation

Undetected faults have likely lived in an undetected state much longer than declared faults. To compensate, these late undetected faults should be charged 2x the undeclared fault penalty. This includes skipped recoveries as well.

### Proposed changes

1. Add logic in `processMissingPoStFaults` to separate faults older than one dead line from newer undetected faults.
2. Charge the new late undeclared penalty (2xSP) for these.
3. Alter `processSkippedFaults` to charge late undeclared penalty for skipped recoveries.
4. Add test logic to test declared recoveries and undeclared faults in `handleProvingPeriod`.
5. Add tests for the new behavior.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -293,9 +293,12 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		sectorInfos, declaredRecoveries, err := st.LoadSectorInfosForProof(store, provenSectors)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load proven sector info")
 
-		// Verify the proof.
-		// A failed verification doesn't immediately cause a penalty; the miner can try again.
-		verifyWindowedPost(rt, currDeadline.Challenge, sectorInfos, params.Proofs)
+		// Skip verification if all sectors are faults
+		if len(sectorInfos) > 0 {
+			// Verify the proof.
+			// A failed verification doesn't immediately cause a penalty; the miner can try again.
+			verifyWindowedPost(rt, currDeadline.Challenge, sectorInfos, params.Proofs)
+		}
 
 		// Record the successful submission
 		postedPartitions := bitfield.NewFromSet(params.Partitions)
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1315,19 +1315,36 @@ func processMissingPoStFaults(rt Runtime, st *State, store adt.Store, deadlines 
 	err = st.RemoveRecoveries(failedRecoveries)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to record failed recoveries")
 
+	// Determine which faults are from the previous deadline. Faults from previous deadline are penalized
+	// the undetected fault fee. Older faults are charged the late undetected fault fee.
+	oldDetectedFaults := detectedFaults
+	newDetectedFaults := bitfield.NewFromSet(nil)
+	// New faults in previous proving period will have been handled in handleProvingPeriod
+	if beforeDeadline > 0 {
+		newDetectedFaults = deadlines.Due[beforeDeadline-1]
+		oldDetectedFaults, err = bitfield.SubtractBitField(oldDetectedFaults, newDetectedFaults)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to subtract last deadline from detected faults")
+	}
+
 	// Load info for sectors.
 	// TODO: this is potentially super expensive for a large miner failing to submit proofs.
 	// https://github.com/filecoin-project/specs-actors/issues/411
-	detectedFaultSectors, err := st.LoadSectorInfos(store, detectedFaults)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load fault sectors")
+	oldDetectedFaultSectors, err := st.LoadSectorInfos(store, oldDetectedFaults)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load old fault sectors")
+	newDetectedFaultSectors, err := st.LoadSectorInfos(store, newDetectedFaults)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load new fault sectors")
 	failedRecoverySectors, err := st.LoadSectorInfos(store, failedRecoveries)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load failed recovery sectors")
 
 	// Unlock sector penalty for all undeclared faults.
-	penalizedSectors := append(detectedFaultSectors, failedRecoverySectors...)
-	penalty, err := unlockUndeclaredFaultPenalty(st, store, info.SectorSize, currEpoch, epochReward, currentTotalPower, penalizedSectors)
+	latePenalizedSectors := append(oldDetectedFaultSectors, failedRecoverySectors...)
+	oldPenalty, err := unlockUndeclaredLateFaultPenalty(st, store, info.SectorSize, currEpoch, epochReward, currentTotalPower, latePenalizedSectors)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to charge sector penalty")
-	return detectedFaultSectors, penalty
+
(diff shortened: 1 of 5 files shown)


anorth:
@ZenGround0 could you take this review please?

actors/builtin/miner/miner_test.go
```diff
@@ -619,6 +612,17 @@ func TestWindowPost(t *testing.T) {
 		return deadline, infos, partitions
 	}
 
+	runTillFirstDeadline := func(rt *mock.Runtime) (*miner.DeadlineInfo, []*miner.SectorOnChainInfo, []uint64) {
```
ZenGround0, line 615: nice decomposition

actors/builtin/miner/miner_test.go
```diff
@@ -1830,8 +1927,21 @@ func (h *actorHarness) onProvingPeriodCron(rt *mock.Runtime, expectedEnrollment
 		},
 		exitcode.Ok)
 
-	{
-		// Detect and penalise missing faults (not yet implemented)
+	if len(undetectedFaultySectors) > 0 {
+		// Process ongoing faulty sectors (not yet implemented)
```
ZenGround0, line 1931: This phrasing is confusing, perhaps add a TODO maybe with a github issue attached.

actors/builtin/miner/miner_test.go
ZenGround0: musing: Feels like this might be at home as a method on the harness

actors/builtin/miner/miner_test.go
```diff
@@ -728,6 +732,79 @@ func TestWindowPost(t *testing.T) {
...
+			expectedRawPowerDelta: rawPower.Neg(),
+			expectedQAPowerDelta:  qaPower.Neg(),
+			expectedPenalty:       big.Add(expectedPenalty1, expectedPenalty2),
+		}
+
+		actor.submitWindowPoSt(rt, deadline, partitions, infos3, cfg)
+
+		// adds sectors as faults
+		st := getState(rt)
+		faultCount, err := st.Faults.Count()
```
ZenGround0, line 765: nit: seems about as much work and a stricter condition to check that the bitfield is set at the right sector numbers. I might be missing something though since this test is new to me.

actors/builtin/miner/miner_test.go
```diff
@@ -728,6 +732,79 @@ func TestWindowPost(t *testing.T) {
 			actor.submitWindowPoSt(rt, deadline, partitions, infos, cfg)
 		})
 	})
+
+	t.Run("detects faults from previous missed posts", func(t *testing.T) {
+		rt := builder.Build(t)
+
+		// skip two PoSts
+		_, infos1, _ := runTillFirstDeadline(rt)
```
ZenGround0, line 740: I'm mildly worried that we aren't fully testing what we want.  The reason is that I think infos1 will always be empty because `handleProvingPeriod`'s `AssignNewSectors` call always skips deadline 0 as being unassignable.  If they are empty than we always get 0 contribution to fee whether or not we implement late/non-late split correctly.

I could be reading the deadline math wrong in `advancePastProvingPeriodWithCron` but I don't think it skips deadline 0 for us.

actors/builtin/miner/miner_actor.go
```diff
@@ -1273,12 +1273,15 @@ func processSkippedFaults(rt Runtime, st *State, store adt.Store, currDeadline *
 
 	// Penalize new skipped faults and retracted recoveries
 	// This differs from declaring faults "on time", where retracting recoveries doesn't attract an extra penalty
-	penalizeFaultSectors := append(newFaultSectors, retractedRecoveryInfos...)
-	penalty, err := unlockUndeclaredFaultPenalty(st, store, minerInfo.SectorSize, currEpoch, epochReward, currentTotalPower, penalizeFaultSectors)
+	faultPenalty, err := unlockUndeclaredFaultPenalty(st, store, minerInfo.SectorSize, currEpoch, epochReward, currentTotalPower, newFaultSectors)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to charge fault fee")
+
+	// retracted recoveries are penalized at a higher rate
+	retractedPenalty, err := unlockUndeclaredLateFaultPenalty(st, store, minerInfo.SectorSize, currEpoch, epochReward, currentTotalPower, retractedRecoveryInfos)
```
anorth, line 1280: I think there's a misunderstanding here. If declared "skipped" in a PoSt submission, both new faults and retracted recoveries should have the same SP penalty. The 2*SP penalty applies if no PoSt submission is made for a sector and it was either a) not declared faulty, or b) declared recovered. But this skipped faults code path shouldn't change.

actors/builtin/miner/miner_actor.go
```diff
@@ -1303,19 +1306,36 @@ func processMissingPoStFaults(rt Runtime, st *State, store adt.Store, deadlines
 	err = st.RemoveRecoveries(failedRecoveries)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to record failed recoveries")
 
+	// Determine which faults are from the previous deadline. Faults from previous deadline are penalized
+	// the undetected fault fee. Older faults are charged the late undetected fault fee.
+	oldDetectedFaults := detectedFaults
+	newDetectedFaults := bitfield.NewFromSet(nil)
+	// New faults in previous proving period will have been handled in handleProvingPeriod
```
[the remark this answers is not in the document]
anorth: I am also pretty confused by this comment, and the code below. I expect this method to be idempotent, so having been called in PP cron not to matter. And it looks like it's marking the whole deadline faulty regardless of PoSt submissions?

actors/builtin/miner/miner_test.go
acruikshank: `runTillFirstDeadline` calls `runTillNextDeadline` at the end to skip past the zero deadline. The concern is valid though. I've added some assertions to ensure these deadlines have sectors even if the logic changes.

actors/builtin/miner/miner_test.go
acruikshank: Now doing both.

actors/builtin/miner/miner_actor.go
acruikshank: got it.

actors/builtin/miner/miner_test.go
```diff
@@ -669,6 +681,31 @@ func TestWindowPost(t *testing.T) {
...
 
+	t.Run("skipped all sectors in a deadline may be skipped", func(t *testing.T) {
+		rt := builder.Build(t)
+		deadline, infos, partitions := runTillFirstDeadline(rt)
+
+		// skip all sectors in deadline
+		st := getState(rt)
+		deadlines, err := st.LoadDeadlines(rt.AdtStore())
+		require.NoError(t, err)
+		skipped := deadlines.Due[deadline.Index]
```
anorth, line 692: assert that skipped isn't empty so we know it's doing something.

actors/builtin/miner/miner_test.go
```diff
@@ -728,6 +765,47 @@ func TestWindowPost(t *testing.T) {
...
+		// assert we have sectors in each deadline
+		assert.Greater(t, len(infos1), 0)
+		assert.Greater(t, len(infos2), 0)
+		assert.Greater(t, len(infos3), 0)
+
+		// expect power to be deducted for all sectors in first two deadlines
+		rawPower, qaPower := miner.PowerForSectors(actor.sectorSize, append(infos1, infos2...))
+
+		// expected penalty is the undeclared fault penalty for all sectors in previous deadline
+		// and the undeclared late penalty for new faults preceding that.
```
anorth, line 786: This comment is a little out of date now.

PR #586, opened 2020-07-03
Batch sector operations

Stebalien:
In the process of implementing #572, I wanted to batch sector expiration updates and got a bit carried away.

### DIFF
--- a/actors/builtin/miner/expirations.go
+++ b/actors/builtin/miner/expirations.go
@@ -1,0 +1,37 @@
+package miner
+
+import (
+	"sort"
+
+	"github.com/filecoin-project/specs-actors/actors/abi"
+)
+
+type sectorEpochSet struct {
+	epoch   abi.ChainEpoch
+	sectors []uint64
+}
+
+// Takes a slice of sector infos and returns sector info sets grouped and
+// sorted by expiration epoch.
+//
+// Note: While each sector set is sorted by epoch, the order of per-epoch sector
+// sets is maintained.
+func groupSectorsByExpiration(sectors []*SectorOnChainInfo) []sectorEpochSet {
+	sectorsByExpiration := make(map[abi.ChainEpoch][]uint64)
+
+	for _, sector := range sectors {
+		sectorsByExpiration[sector.Expiration] = append(sectorsByExpiration[sector.Expiration], uint64(sector.SectorNumber))
+	}
+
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/expirations.go
```diff
@@ -0,0 +1,29 @@
...
+
+func groupSectorsByExpiration(sectors []*SectorOnChainInfo) []sectorEpochSet {
+	sectorsByExpiration := make(map[abi.ChainEpoch][]uint64)
+
+	for _, sector := range sectors {
+		sectorsByExpiration[sector.Expiration] = append(sectorsByExpiration[sector.Expiration], uint64(sector.SectorNumber))
+	}
+
+	sectorEpochSets := make([]sectorEpochSet, 0, len(sectorsByExpiration))
+	for expiration, sectors := range sectorsByExpiration {
```
anorth, line 22: The map iteration makes me nervous, though I think this one's ok since no side-effects. We should probably document every map iteration with "this one's ok because it has no side effects or runtime interactions" to help avoid someone changing something that breaks that.

actors/builtin/miner/miner_actor.go
```diff
@@ -1441,27 +1445,23 @@ func scheduleReplaceSectorsExpiration(rt Runtime, st *State, store adt.Store, re
...
 	newExpiration := deadlineInfo.PeriodEnd()
-	for i, sector := range replaceSectors {
-		// Note: State updates could be minimized by batching these by existing expiration epoch.
-		if err := st.RemoveSectorExpirations(store, sector.Expiration, uint64(sector.SectorNumber)); err != nil {
-			return fmt.Errorf("failed to remove sector expiration %v at %v: %w",
-				sector.SectorNumber, sector.Expiration, err)
-		}
+
+	if err := st.RemoveSectorExpirations(store, replaceSectors...); err != nil {
+		return err
```
anorth, line 1452: I have moderate preference for wrapping these errors so that we or the node implementation have more information when debugging whatever corruption might cause it.

actors/builtin/miner/expirations.go
Stebalien: I've added some documentation.

actors/builtin/miner/miner_actor.go
Stebalien: Done.

PR #588, opened 2020-07-03
Multisig: Resolve address before comparing

### LINKED ISSUE #587: Multisig:  Resolve signer in passed params and the state signers before comparing
In functions like `RemoveSigner` and `SwapSigner`,  when we determine if the signer passed in the params is the same as one of the signers we have on the state, we don't account for the fact that they could have different address types.

aarshkshah1992:
Closes #587 .


wadeAlexC:
LGTM on the surface. I'll add a note for Daniel and I to investigate other areas this may occur.

PR #590, opened 2020-07-03
Market Actor: Tests for ActivateDeals and OnMinerSectorsTerminate

aarshkshah1992:
Tests for `ActivateDeals` and `OnMinerSectorsTerminate`.

### DIFF
--- a/actors/builtin/market/market_state.go
+++ b/actors/builtin/market/market_state.go
@@ -16,7 +16,7 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
-const EpochUndefined = abi.ChainEpoch(-1)
+const epochUndefined = abi.ChainEpoch(-1)
 
 // Market mutations
 // add / rm balance
@@ -87,15 +87,15 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyMSetCid cid.Cid) *State {
 func (st *State) updatePendingDealState(rt Runtime, state *DealState, deal *DealProposal, dealID abi.DealID, et, lt *adt.BalanceTable, epoch abi.ChainEpoch) (abi.TokenAmount, abi.ChainEpoch) {
 	amountSlashed := abi.NewTokenAmount(0)
 
-	everUpdated := state.LastUpdatedEpoch != EpochUndefined
-	everSlashed := state.SlashEpoch != EpochUndefined
+	everUpdated := state.LastUpdatedEpoch != epochUndefined
+	everSlashed := state.SlashEpoch != epochUndefined
 
 	Assert(!everUpdated || (state.LastUpdatedEpoch <= epoch)) // if the deal was ever updated, make sure it didn't happen in the future
 
 	// This would be the case that the first callback somehow triggers before it is scheduled to
 	// This is expected not to be able to happen
 	if deal.StartEpoch > epoch {
-		return amountSlashed, EpochUndefined
+		return amountSlashed, epochUndefined
 	}
 
 	dealEnd := deal.EndEpoch
@@ -143,12 +143,12 @@ func (st *State) updatePendingDealState(rt Runtime, state *DealState, deal *Deal
 		}
 
 		st.deleteDeal(rt, dealID)
-		return amountSlashed, EpochUndefined
+		return amountSlashed, epochUndefined
 	}
 
 	if epoch >= deal.EndEpoch {
 		st.processDealExpired(rt, deal, state, lt, dealID)
-		return amountSlashed, EpochUndefined
+		return amountSlashed, epochUndefined
 	}
 
 	next := epoch + DealUpdatesInterval
@@ -212,7 +212,7 @@ func (st *State) deleteDeal(rt Runtime, dealID abi.DealID) {
 // Delete deal, slash a portion of provider's collateral, and unlock remaining collaterals
 // for both provider and client.
 func (st *State) processDealInitTimedOut(rt Runtime, et, lt *adt.BalanceTable, dealID abi.DealID, deal *DealProposal, state *DealState) abi.TokenAmount {
-	Assert(state.SectorStartEpoch == EpochUndefined)
+	Assert(state.SectorStartEpoch == epochUndefined)
 
 	if err := st.unlockBalance(lt, deal.Client, deal.TotalStorageFee(), ClientStorageFee); err != nil {
 		rt.Abortf(exitcode.ErrIllegalState, "failure unlocking client storage fee: %s", err)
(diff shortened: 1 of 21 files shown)


actors/builtin/market/market_state.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
-const epochUndefined = abi.ChainEpoch(-1)
+const EpochUndefined = abi.ChainEpoch(-1)
```
Stebalien, line 19: I'd define this in `abi/primitives.go`, if we're going to export it.

actors/builtin/market/market_test.go
```diff
@@ -693,6 +695,363 @@ func TestPublishStorageDealsFailures(t *testing.T) {
...
+	worker := tutil.NewIDAddr(t, 103)
+	client := tutil.NewIDAddr(t, 104)
+	mAddrs := &minerAddrs{owner, worker, provider}
+
+	startEpoch := abi.ChainEpoch(10)
+	endEpoch := abi.ChainEpoch(20)
+	sectorExpiry := abi.ChainEpoch(100)
+
+	// caller is not the provider
+	{
```
Stebalien, line 753: nit: these scopes don't really add anything.

actors/builtin/market/market_test.go
```diff
@@ -693,6 +695,363 @@ func TestPublishStorageDealsFailures(t *testing.T) {
...
+
+	// deal has not been published before
+	{
+		t.Run("fail when deal has not been published before", func(t *testing.T) {
+			rt, actor := basicMarketSetup(t, marketActor, owner, provider, worker, client)
+			params := mkActivateDealParams(sectorExpiry, abi.DealID(42))
+
+			rt.ExpectValidateCallerType(builtin.StorageMinerActorCodeID)
+			rt.SetCaller(provider, builtin.StorageMinerActorCodeID)
+			rt.ExpectAbort(exitcode.ErrIllegalState, func() {
```
ZenGround0, line 794: This exitcode seems unexpected.  I think it comes from the RequireNoErr after ValidateDealsForActivation.  That should probably be changed to `ErrIllegalArgument`

ZenGround0: Thinking more it looks a little more complicated.  Some failures from that function are IllegalState, some are IllegalArgument

actors/builtin/market/market_test.go
```diff
@@ -693,6 +695,363 @@ func TestPublishStorageDealsFailures(t *testing.T) {
...
+		dealId4 := actor.generateAndPublishDeal(rt, client, maddrs2, startEpoch, endEpoch)
+		dealId5 := actor.generateAndPublishDeal(rt, client, maddrs2, startEpoch, endEpoch+1)
+		actor.activateDeals(rt, sectorExpiry, provider2, currentEpoch, dealId4, dealId5)
+
+		// provider1 terminates deal1 but that does not terminate deals2-5
+		actor.terminateDeals(rt, provider, dealId1)
+		actor.assertDealsTerminated(rt, currentEpoch, dealId1)
+		actor.assertDeaslNotTerminated(rt, dealId2, dealId3, dealId4, dealId5)
+
+		// provider2 terminates deal5 but that does not terminate del2-4
```
ZenGround0, line 917: nit: deals2-4

actors/builtin/market/market_test.go
```diff
@@ -693,6 +695,363 @@ func TestPublishStorageDealsFailures(t *testing.T) {
...
+	t.Run("fail when deal has been published but not activated", func(t *testing.T) {
+		rt, actor := basicMarketSetup(t, marketActor, owner, provider, worker, client)
+		rt.SetEpoch(currentEpoch)
+
+		dealId := actor.generateAndPublishDeal(rt, client, mAddrs, startEpoch, endEpoch)
+
+		params := mkTerminateDealParams(dealId)
+		rt.ExpectValidateCallerType(builtin.StorageMinerActorCodeID)
+		rt.SetCaller(provider, builtin.StorageMinerActorCodeID)
+		rt.ExpectAbort(exitcode.ErrIllegalState, func() {
```
ZenGround0, line 1025: Without looking at what's going on inside the market actor this seems like it should fail for IllegalArgument.  I think this is failing on the `!found` case of states.Get which would be an easy change to make if my intuition is right.

actors/builtin/market/market_state.go
anorth: Yeah, minor preference for not exporting it, but if we do, move it to `abi.`

actors/builtin/market/market_test.go
```diff
@@ -693,6 +695,363 @@ func TestPublishStorageDealsFailures(t *testing.T) {
...
+	startEpoch := abi.ChainEpoch(10)
+	endEpoch := abi.ChainEpoch(20)
+	sectorExpiry := abi.ChainEpoch(100)
+
+	// caller is not the provider
+	{
+		t.Run("fail when caller is not the provider of the deal", func(t *testing.T) {
+			rt, actor := basicMarketSetup(t, marketActor, owner, provider, worker, client)
+			provider2 := tutil.NewIDAddr(t, 201)
+			mAddrs2 := &minerAddrs{owner, worker, provider2}
```
anorth, line 757: I think it's equivalent, but this might be simpler if the deal uses the standard `mAddrs` and just set a different address at 763.

actors/builtin/market/market_test.go
anorth: This is a known problem. I've filed an issue for it. https://github.com/filecoin-project/specs-actors/issues/597

actors/builtin/market/market_test.go
```diff
@@ -933,23 +1311,44 @@ func (h *marketActorTestHarness) getDealState(rt *mock.Runtime, dealID abi.DealI
...
+	for _, d := range dealIds {
+		s := h.getDealState(rt, d)
+		require.EqualValues(h.t, epoch, s.SlashEpoch)
+	}
+}
+
+func (h *marketActorTestHarness) assertDeaslNotTerminated(rt *mock.Runtime, dealIds ...abi.DealID) {
+	for _, d := range dealIds {
+		s := h.getDealState(rt, d)
+		require.EqualValues(h.t, market.EpochUndefined, s.SlashEpoch)
```
anorth, line 1324: If this was the only reason to export EpochUndefined, just code the `-1` here instead and leave it hidden.

actors/builtin/market/market_state.go
aarshkshah1992: I don't think we need to export this. Restoring this.

actors/builtin/market/market_test.go
```diff
@@ -693,6 +695,363 @@ func TestPublishStorageDealsFailures(t *testing.T) {
 	}
 }
 
+func TestActivateDeals(t *testing.T) {
+	marketActor := tutil.NewIDAddr(t, 100)
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: Just makes the tests more readable to me.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: I agree. Have made this change.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

PR #594, opened 2020-07-03
Remove non-deterministic map traversal

### LINKED ISSUE #578: Order of removing faults is not deterministic

Stebalien:
I've looked at all instances of `range` (outside of tests & cbor_gen code) and this was the only potentially problematic case.

fixes #578


actors/builtin/market/market_actor.go
```diff
@@ -576,11 +577,18 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
...
 		}
 
-		// NB: its okay that we're doing a 'random' golang map iteration here
-		// because HAMTs and AMTs are insertion order independent, the same set of
-		// data inserted will always produce the same structure, no matter the order
-		for epoch, deals := range updatesNeeded {
-			if err := dbe.PutMany(epoch, deals); err != nil {
+		// Iterate changes in sorted order to ensure that loads/stores
+		// are deterministic. Otherwise, we could end up charging an
+		// inconsistent amount of gas.
```
ZenGround0, line 582: Out of curiousity how does total gas charged change when AMTs and HAMTs are insertion order dependent?  Is this written up anywhere?  I would love to walk through an example where things are inserted out of order, result in the same HAMT/AMT but the gas is charged differently based on the order.

Stebalien: It may change how/when intermediate nodes are created. In _this_ case, I don't think it matters because we're only inserting. It mattered in https://github.com/filecoin-project/specs-actors/issues/578 because we were deleting and inserting. E.g.,

* Insert, Insert -> Expand a new level.
* Insert, Delete, Insert -> Doesn't expand to a new level.

However, it doesn't really hurt to be safe and deterministic.

PR #595, opened 2020-07-03
trying to reduce window periods

nonsense:
An attempt to reduce the various periods in `specs-actors`, so that we can test various behaviours in Project Oni, without having to wait for thousands of epochs.

There has been some success when reducing the `WPoStProvingPeriod` to 360 and `WPoStChallengeWindow` to 10, but it is not yet clear if this introduces other issues.

In the future we will probably also want to reduce `FaultMaxAge`

---

I also considered changing `const`s to `var`s, but this is not possible, since we allocate arrays with const sizes. I am not aware of any other way to change these, but to introduce build flags, similar to what we already did on Lotus side (for example https://github.com/filecoin-project/lotus/blob/master/build/params_testground.go)

---

Related: https://github.com/filecoin-project/oni/issues/12

cc @raulk

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -3,6 +3,7 @@ package miner
 import (
 	"fmt"
 	"reflect"
+	"sort"
 
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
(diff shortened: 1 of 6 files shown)


anorth:
@nonsense could we close this PR? My impression is that there's no intention of this going into `master`, and it now seems unmaintained. You can still have the branch, of course.

nonsense:
@anorth sure, I will close it.

Depending on how well our tests work with https://github.com/filecoin-project/oni/pull/143 (mock time), we might be able to use the production windows (deadlines). For now I continue to override the deadlines, but I don't need an open PR for that!

PR #601, opened 2020-07-06
Calculate first block reward on construction

### LINKED ISSUE #374: Reward State: First block earns zero reward
The block reward is calculated based on `LastPerEpochReward` in reward state, which is zero for the first block...we can either initialise to something, or make the cron tick happen before the first block reward is awarded

ZenGround0:
Closes #374.  

FYI @magik6k @arajasek see latest comment on that issue for tips on integrating into lotus `chain/gen`.  

FYI @Kubuxu these refactoring changes should be orthogonal to logic changes in minting function etc but one of us will have to handle conflicts.

### DIFF
--- a/actors/builtin/methods.go
+++ b/actors/builtin/methods.go
@@ -29,10 +29,10 @@ var MethodsCron = struct {
 }{MethodConstructor, 2}
 
 var MethodsReward = struct {
-	Constructor        abi.MethodNum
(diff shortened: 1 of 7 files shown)


ZenGround0:
FYI merging now @anorth to keep down friction for @Kubuxu.  Please take a look though and let me know of any suggested changes so I can address in a follow on PR.

Issue #603, opened 2020-07-06
Market WithdrawBalance never slashes


ZenGround0:
There is either dead logic or a feature we have forgotten to keep track of in market actor `WithdrawBalance`.  Before making the withdraw update to the balance table we initialize a variable `amountSlashedTotal` to 0.  We then do no assignments to this variable during the state transaction making the update.  After the update we burn slashed funds if `amountSlashedTotal` is greater than 0 which is never hit.  

Either we should remove the calculations for the dead logic or we should track and do the work introducing slashing to market balance withdraw.

aarshkshah1992:
@ZenGround0 I'd discussed this with @anorth . 

1. The funds to slash are locked and the user can't withdraw such that the user's balance goes below the collateral/funds to be slashed. 
2. We slash funds in `OnCronTick`

Basically, this is dead code and we can remove it. I'll create a PR.

aarshkshah1992:
This is done.

PR #611, opened 2020-07-07
Market: Miner fixes and tests

### LINKED ISSUE #589: Market Actor: Should we be able to Terminate the same deals twice
Currently, it is possible to mark the same deal as terminated twice by calling `OnMinerSectorsTerminate` twice with the same dealID. The `SlashEpoch` on the deal will be the last `CurrentEpoch` it was called at.

Is this the expected behaviour ?

@anorth

aarshkshah1992:
Closes #589 AND #603 .

Also, as discussed offline with @anorth , we don't need to check for negative balance in the `AddBalanceCall` as the Runtime will enforce it.

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -125,19 +125,6 @@ func TestMarketActor(t *testing.T) {
 			}
 		})
 
-		t.Run("fails when called with negative value", func(t *testing.T) {
-			rt, actor := basicMarketSetup(t, owner, provider, worker, client)
-
-			rt.SetCaller(owner, builtin.AccountActorCodeID)
-			rt.SetReceived(abi.NewTokenAmount(-1))
-
-			rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
-				rt.Call(actor.AddBalance, &provider)
-			})
-
-			rt.Verify()
-		})
-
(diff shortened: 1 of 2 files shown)


actors/builtin/market/market_test.go
```diff
@@ -971,6 +958,52 @@ func TestOnMinerSectorsTerminate(t *testing.T) {
...
+		// provider1 publishes deal1 and 2 and deal3 -> deal3 has the lowest endepoch
+		dealId1 := actor.generateAndPublishDeal(rt, client, mAddrs, startEpoch, endEpoch)
+		dealId2 := actor.generateAndPublishDeal(rt, client, mAddrs, startEpoch, endEpoch+1)
+		dealId3 := actor.generateAndPublishDeal(rt, client, mAddrs, startEpoch, endEpoch-1)
+		actor.activateDeals(rt, sectorExpiry, provider, currentEpoch, dealId1, dealId2, dealId3)
+
+		// terminating the deal so slash epoch is the current epoch
+		actor.terminateDeals(rt, provider, dealId1)
+
+		// set a new epoch and terminate again -> however slash epoch will still be the old epoch.
```
ZenGround0, line 992: nit: slash epoch for dealID1 will still be the old epoch

aarshkshah1992: Will do this in a follow-up.

PR #615, opened 2020-07-07
Small Network Power Accounting

### LINKED ISSUE #266: Fix calculation of total network power in small networks

### LINKED ISSUE #565: OnConsensusFault removes power from total power without checking against the minimum power
This is a variant of #563 (which was invalid) but is also hard to exploit. If a miner _had_ enough power to win a block, then lost power, then another miner reports a consensus fault, OnConsensusFault will subtract the miner's power from the total power, even if the miner's power is below the minimum power and wasn't even included in the total power in the first place.

ZenGround0:
Closes #266 
Closes #565 

This makes total power accounting work for small networks where no miner has the minimum needed consensus power.  It also makes `OnConsensusFault` do its power changes through `AddClaim` removing the accounting error from #565.  

The approach is very simple.  It wasn't 100% clear if this simple approach meets the issue requirements, so I want feedback on this high level plan.  We keep two more top level state counters in the power actor for the sum of all qa and raw claim power.  We then modify the `CurrentTotalPower` call to check whether the total number of miners above the minimum power is above some threshold (set to 3 before this pr).  If it's above we do things as before ignoring all claims from miners below the minimum threshold.  If its below we use the total sum of claims.  

Some more things to note:
1. I have done nothing to change how we determine if a miner "meets the consensus minimum".  We still just compare QA power to some constant.  This means `minerNominalPowerMeetsConsensusMinimum` is still deadcode.  I need help understanding what we are trying to do if this work should somehow use that notion of minimum; it doesn't seem easily compatible with the issue description or what I've done here.
2. The existing solution has some weird quirks.  For example the total network power can go down if storage goes up.  Maybe we should brief cryptoecon / redteam on these changes?
3. Unrelatedly testing power updates during `OnConsensusFault` found a bug where the method panics by default which this PR should fix.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -239,9 +239,8 @@ func (a Actor) OnConsensusFault(rt Runtime, pledgeAmount *abi.TokenAmount) *adt.
 		}
 		Assert(claim.RawBytePower.GreaterThanEqual(big.Zero()))
 		Assert(claim.QualityAdjPower.GreaterThanEqual(big.Zero()))
-
-		st.TotalQualityAdjPower = big.Sub(st.TotalQualityAdjPower, claim.QualityAdjPower)
-		st.TotalRawBytePower = big.Sub(st.TotalRawBytePower, claim.RawBytePower)
+		err = st.AddToClaim(adt.AsStore(rt), minerAddr, claim.QualityAdjPower.Neg(), claim.RawBytePower.Neg())
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "could not add to claim for %s after loading existing claim for this address", minerAddr)
 
 		st.addPledgeTotal(pledgeAmount.Neg())
 		return nil
@@ -304,6 +303,13 @@ func (a Actor) CurrentTotalPower(rt Runtime, _ *adt.EmptyValue) *CurrentTotalPow
 	rt.ValidateImmediateCallerAcceptAny()
 	var st State
 	rt.State().Readonly(&st)
+	if st.NumMinersMeetingMinPower < ConsensusMinerMinMiners {
+		return &CurrentTotalPowerReturn{
+			RawBytePower:     st.TotalRawBytePowerNoMin,
+			QualityAdjPower:  st.TotalQualityAdjPowerNoMin,
+			PledgeCollateral: st.TotalPledgeCollateral,
+		}
+	}
 	return &CurrentTotalPowerReturn{
 		RawBytePower:     st.TotalRawBytePower,
 		QualityAdjPower:  st.TotalQualityAdjPower,
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -479,13 +479,16 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
 
 func (a Actor) deleteMinerActor(rt Runtime, miner addr.Address) error {
 	var st State
-	err := rt.State().Transaction(&st, func() interface{} {
-		if err := st.deleteClaim(adt.AsStore(rt), miner); err != nil {
-			return errors.Wrapf(err, "failed to delete %v from claimed power table", miner)
+	var err error
+	rt.State().Transaction(&st, func() interface{} {
+		err = st.deleteClaim(adt.AsStore(rt), miner)
+		if err != nil {
+			err = errors.Wrapf(err, "failed to delete %v from claimed power table", miner)
+			return nil
 		}
 
 		st.MinerCount -= 1
 		return nil
-	}).(error)
+	})
 	return err
 }
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -303,16 +303,11 @@ func (a Actor) CurrentTotalPower(rt Runtime, _ *adt.EmptyValue) *CurrentTotalPow
 	rt.ValidateImmediateCallerAcceptAny()
 	var st State
 	rt.State().Readonly(&st)
-	if st.NumMinersMeetingMinPower < ConsensusMinerMinMiners {
-		return &CurrentTotalPowerReturn{
-			RawBytePower:     st.TotalRawBytePowerNoMin,
-			QualityAdjPower:  st.TotalQualityAdjPowerNoMin,
-			PledgeCollateral: st.TotalPledgeCollateral,
-		}
-	}
+
+	rawBytePower, qaPower := CurrentTotalPower(&st)
 	return &CurrentTotalPowerReturn{
-		RawBytePower:     st.TotalRawBytePower,
-		QualityAdjPower:  st.TotalQualityAdjPower,
+		RawBytePower:     rawBytePower,
+		QualityAdjPower:  qaPower,
 		PledgeCollateral: st.TotalPledgeCollateral,
 	}
 }
--- a/actors/builtin/power/power_state.go
+++ b/actors/builtin/power/power_state.go
@@ -15,10 +15,12 @@ import (
 )
 
 type State struct {
-	TotalRawBytePower     abi.StoragePower
-	TotalQualityAdjPower  abi.StoragePower
-	TotalPledgeCollateral abi.TokenAmount
-	MinerCount            int64
+	TotalRawBytePower         abi.StoragePower
+	TotalRawBytePowerNoMin    abi.StoragePower
+	TotalQualityAdjPower      abi.StoragePower
+	TotalQualityAdjPowerNoMin abi.StoragePower
+	TotalPledgeCollateral     abi.TokenAmount
+	MinerCount                int64
 
 	// A queue of events to be triggered by cron, indexed by epoch.
 	CronEventQueue cid.Cid // Multimap, (HAMT[ChainEpoch]AMT[CronEvent]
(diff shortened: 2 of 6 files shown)


actors/builtin/power/power_state.go
```diff
@@ -15,10 +15,12 @@ import (
 )
 
 type State struct {
-	TotalRawBytePower     abi.StoragePower
-	TotalQualityAdjPower  abi.StoragePower
-	TotalPledgeCollateral abi.TokenAmount
-	MinerCount            int64
+	TotalRawBytePower         abi.StoragePower
+	TotalRawBytePowerNoMin    abi.StoragePower
```
ZenGround0, line 19: I don't love names, suggestions welcome

actors/builtin/power/power_test.go
```diff
@@ -145,6 +148,115 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
...
+		actor.updateClaimedPower(rt, miner4, powerUnit, mul(powerUnit, 2))
+		actor.updateClaimedPower(rt, miner5, powerUnit, mul(powerUnit, 2))		
+
+		// Below threshold small miner power is counted
+		expectedTotalBelow := big.Sum(mul(smallPowerUnit, 3), mul(powerUnit, 2))
+		actor.expectTotalPower(rt, expectedTotalBelow, mul(expectedTotalBelow, 2))
+
+		// Above threshold (power.ConsensusMinerMinMiners = 3) small miner power is ignored
+		actor.updateClaimedPower(rt, miner3, powerUnit, mul(powerUnit, 2))
+		expectedTotalAbove := big.Sum(smallPowerUnit, mul(powerUnit, 3))
```
acruikshank, line 175: I think this is miner3: `smallPower+power`, miner4: `power`, miner5: `power`, so you expect `smallPower + 3*power`. This would be a little clearer if you just added `power-smallPower` to miner3 and expected `3*power`.

actors/builtin/power/power_test.go
```diff
@@ -145,6 +148,115 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
...
+		st := getState(rt)
+		assert.Equal(t, int64(3), st.NumMinersMeetingMinPower)
+
+		// Less than 3 miners below threshold again small miner power is counted again
+		negPowerUnit := powerUnit.Neg()
+		actor.updateClaimedPower(rt, miner3, negPowerUnit, mul(negPowerUnit, 2))
+		actor.expectTotalPower(rt, expectedTotalBelow, mul(expectedTotalBelow, 2))
+	})
+
+	t.Run("all miner power dissapears when miner dips below min power threshold", func(t *testing.T) {
```
acruikshank, line 187: nit: I found this a little confusing (like total power would be zero). Maybe "removes all a miner's power when..."

actors/builtin/power/power_test.go
```diff
@@ -145,6 +148,115 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
...
+		assert.Equal(t, int64(0), st.NumMinersMeetingMinPower)
+
+		actor.updateClaimedPower(rt, miner1, big.Zero(), powerUnit)
+		actor.updateClaimedPower(rt, miner2, big.Zero(), powerUnit)
+		actor.updateClaimedPower(rt, miner3, big.Zero(), powerUnit)
+		st = getState(rt)
+		assert.Equal(t, int64(3), st.NumMinersMeetingMinPower)
+	})
+
+	t.Run("slashing miner below minimum does not impact power", func(t *testing.T) {
```
acruikshank, line 233: This is a little confusing. It's "Slashing a miner who is below minimum", not  "Slashing a miner to go below minimum".

actors/builtin/power/power_state.go
anorth: Let's reserve "power" for the thing that grants the chance to mine a block. So, how about `TotalBytesCommitted`, `TotalQABytesCommitted`. Also please add a brief note here as to why they can differ.

actors/builtin/power/power_state.go
```diff
@@ -15,10 +15,12 @@ import (
...
+	TotalQualityAdjPower      abi.StoragePower
+	TotalQualityAdjPowerNoMin abi.StoragePower
+	TotalPledgeCollateral     abi.TokenAmount
+	MinerCount                int64
```
anorth, line 23: While you're here, move `NumMinersMeetingMinPower` next to `MinerCount`. We've copped some flak for renaming things, but I'd suggest `MinerAboveMinPowerCount` or similar.

actors/builtin/power/power_state.go
```diff
@@ -52,13 +54,15 @@ type AddrKey = adt.AddrKey
...
-		NumMinersMeetingMinPower: 0,
+		TotalRawBytePower:         abi.NewStoragePower(0),
+		TotalRawBytePowerNoMin:    abi.NewStoragePower(0),
+		TotalQualityAdjPower:      abi.NewStoragePower(0),
+		TotalQualityAdjPowerNoMin: abi.NewStoragePower(0),
+		TotalPledgeCollateral:     abi.NewTokenAmount(0),
+		LastEpochTick:             -1,
+		CronEventQueue:            emptyMapCid,
+		Claims:                    emptyMapCid,
+		NumMinersMeetingMinPower:  0,
```
anorth, line 65: Nit: Initialize `MinerCount`

actors/builtin/power/power_state.go
```diff
@@ -146,6 +150,10 @@ func (st *State) AddToClaim(s adt.Store, miner addr.Address, power abi.StoragePo
 		st.TotalRawBytePower = big.Add(st.TotalRawBytePower, power)
 	}
 
+	// NoMin powers always update directly
+	st.TotalQualityAdjPowerNoMin = big.Add(st.TotalQualityAdjPowerNoMin, qapower)
+	st.TotalRawBytePowerNoMin = big.Add(st.TotalRawBytePowerNoMin, power)
```
anorth, line 155: Nit: I'd do this "simple" case first, around line 128.

actors/builtin/power/power_test.go
```diff
@@ -145,6 +148,116 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
...
+		actor.updateClaimedPower(rt, miner3, smallPowerUnit, mul(smallPowerUnit, 2))				
+
+		actor.updateClaimedPower(rt, miner4, powerUnit, mul(powerUnit, 2))
+		actor.updateClaimedPower(rt, miner5, powerUnit, mul(powerUnit, 2))		
+
+		// Below threshold small miner power is counted
+		expectedTotalBelow := big.Sum(mul(smallPowerUnit, 3), mul(powerUnit, 2))
+		actor.expectTotalPower(rt, expectedTotalBelow, mul(expectedTotalBelow, 2))
+
+		// Above threshold (power.ConsensusMinerMinMiners = 3) small miner power is ignored
```
anorth, line 173: Require `ConsensusMinerMinMiners = 3` in the test, so it breaks quickly if the number changes.

actors/builtin/power/power_test.go
```diff
@@ -145,6 +148,116 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
...
+		delta := big.Sub(powerUnit, smallPowerUnit)
+		actor.updateClaimedPower(rt, miner3, delta, mul(delta, 2))
```
anorth, line 175: The QA power is the one that counts, so this puts the power way above the threshold, not just at it. But doing the subtraction implies you meant to _just_ cross the line. The same applies to the miner4/5 set-ups.

It might be clearer to divide for the raw power than multiply for the QA power.

actors/builtin/power/power_actor.go
```diff
@@ -304,6 +303,13 @@ func (a Actor) CurrentTotalPower(rt Runtime, _ *adt.EmptyValue) *CurrentTotalPow
 	rt.ValidateImmediateCallerAcceptAny()
 	var st State
 	rt.State().Readonly(&st)
+	if st.NumMinersMeetingMinPower < ConsensusMinerMinMiners {
+		return &CurrentTotalPowerReturn{
+			RawBytePower:     st.TotalRawBytePowerNoMin,
+			QualityAdjPower:  st.TotalQualityAdjPowerNoMin,
+			PledgeCollateral: st.TotalPledgeCollateral,
+		}
+	}
```
anorth, line 312: The node's consensus system needs this same logic. Pull this out into an exported free function of state.

anorth:
@ZenGround0 this is great, `minerNominalPowerMeetsConsensusMinimum` still needs updating. Its goal is to tell whether an individual miner could possibly win a block. It should evaluate false for small miners, despite their claim being positive, if sufficient other miners exceed the threshold. This is because the small miner's claim did not form part of the total power (i.e. the sum of claims >= network power, unless you exclude the small ones).

This PR means that the iteration over claims at the end of that method is no longer necessary; you can look up the power directly. Please follow on to change `minerNominalPowerMeetsConsensusMinimum` to an exported free function of state and miner, probably moving it to `miner_actor.go`.

FYI @magik6k the actors are now doing the accounting necessary to take the miner minimum size seriously, so the node should not ignore it. Otherwise the sum of miner's power claims will exceed notional power (denominator in the election).

PR #617, opened 2020-07-07
Rework reward actor

Kubuxu:
Done, waiting for final CR.


actors/builtin/reward/expneg.go
```diff
@@ -0,0 +1,91 @@
+package reward
```
Kubuxu, line 1: Below code comes from lotus Poisson sortition

actors/builtin/reward/reward_logic.go
```diff
@@ -0,0 +1,72 @@
...
+	abi "github.com/filecoin-project/specs-actors/actors/abi"
+	big "github.com/filecoin-project/specs-actors/actors/abi/big"
+	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
+)
+
+var BaselinePowerAt = func(epoch abi.ChainEpoch) abi.StoragePower {
+	return big.NewInt(1 << 40)
+}
+
+// Computes RewardTheta, result is in Q.128 format
```
anorth, line 13: A little more explanation would be great: readers won't necessarily understand the design intent. Here, perhaps "theta is the virtual network time that elapses as the realized integral over time of committed power approaches the baseline integral"

Or something, you can probably do better 😁

actors/builtin/reward/reward_logic.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// These numbers are placeholders, but should be in units of attoFIL, 10^-18 FIL
+var SimpleTotal = big.Mul(big.NewInt(100e6), big.NewInt(1e18))   // 100M for testnet, PARAM_FINISH
+var BaselineTotal = big.Mul(big.NewInt(900e6), big.NewInt(1e18)) // 900M for testnet, PARAM_FINISH
+
+var (
+	// parameters in Q.128 format
+	// lambda = tropicalYearInSeconds/blockDelay*ln(2)
+	lambda, _ = big.FromString("186857370934482378542986172834581")
```
anorth, line 37: If possible, I would really like to compute this in code. If that's not reasonable, please add a big comment next to `builtin.EpochDurationSeconds` that points to this number needing to be recomputed if the delay changes.

actors/builtin/reward/reward_actor.go
```diff
@@ -65,7 +68,8 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
 	penalty := abi.NewTokenAmount(0)
 	var st State
 	rt.State().Readonly(&st)
-	blockReward := big.Div(st.ThisEpochReward, big.NewInt(builtin.ExpectedLeadersPerEpoch))
+
+	blockReward := st.ThisEpochReward
```
anorth, line 72: Blocking: a number of other places depend on this value being the total reward (expectation), not the per miner value. If we change this, those call sites will all have to multiply (or have the `ThisEpochReward` method do it for them).

If this was an optimization to divide once instead of 5 times, the net effect will be worse, since many more than 5 multiplies will be needed to make up for it.

actors/builtin/reward/reward_logic.go
```diff
@@ -0,0 +1,72 @@
...
+	simpleReward := big.Mul(SimpleTotal, expLamSubOne)    //Q.0 * Q.128 =>  Q.128
+	epochLam := big.Mul(big.NewInt(int64(epoch)), lambda) // Q.0 * Q.128 => Q.128
+
+	simpleReward = big.Mul(simpleReward, big.Int{Int: expneg(epochLam.Int)}) // Q.128 * Q.128 => Q.256
+	simpleReward = big.Rsh(simpleReward, precision)                          // Q.256 >> 128 => Q.128
+
+	baselineReward := big.Sub(computeBaselineSupply(prevTheta), computeBaselineSupply(currTheta)) // Q.128
+
+	reward := big.Add(simpleReward, baselineReward)                       // Q.128
+	reward = big.Div(reward, big.NewInt(builtin.ExpectedLeadersPerEpoch)) // Q.128 / Q.0  => Q.128
```
anorth, line 54: See comment in actor: don't divide here, we need the aggregate more frequently than the per-block version.

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,59 @@ package reward
...
-	CumsumBaseline       Spacetime
-	CumsumRealized       Spacetime
-	EffectiveNetworkTime NetworkTime
+	// CumsumBaseline is a target CumsumRealized needs to reach for EffectiveNetworkTime to increase
+	CumsumBaseline Spacetime
+	CumsumRealized Spacetime
 
-	SimpleSupply   abi.TokenAmount // current supply
-	BaselineSupply abi.TokenAmount // current supply
+	// EffectiveNetworkTime is ceiling of EffectiveNetworkTime based on CumsumRealized
```
anorth, line 16: This self-referential comment doesn't clarify for me.

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,59 @@ package reward
...
-	// The reward to be paid in total to block producers, if exactly the expected number of them produce a block.
+	// The reward to be paid in per WinCount to block producers.
 	// The actual reward total paid out depends on the number of winners in any round.
-	// This is computed at the end of the previous epoch, and should really be called ThisEpochReward.
 	ThisEpochReward abi.TokenAmount
-
-	// The count of epochs for which a reward has been paid.
-	// This should equal the number of non-empty tipsets after the genesis, aka "chain height".
-	RewardEpochsPaid abi.ChainEpoch
+	Epoch           abi.ChainEpoch
```
anorth, line 22: Please document, "the epoch for which ThisEpochReward and other calculations have been made".

Is "for which" rather than "in which" correct? I can't quite tell from the code. The calculations are made by cron at the end of one epoch, in order to set the reward for the subsequent one. If "for which", then this value should be zero in genesis state, if "in which" then it should be -1.

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,59 @@ package reward
...
-func ConstructState(currRealizedPower *abi.StoragePower) *State {
+func ConstructState(currRealizedPower abi.StoragePower) *State {
 	st := &State{
-		BaselinePower:        big.Zero(),
-		RealizedPower:        big.Zero(),
-		CumsumBaseline:       big.Zero(),
-		CumsumRealized:       big.Zero(),
-		EffectiveNetworkTime: big.Zero(),
+		CumsumBaseline: big.Zero(),
+		CumsumRealized: big.Zero(),
```
anorth, line 28: Init `EffectiveNetworkTime`, even if it will be overwritten.

Nit: Init `Epoch` here rather than as a mutation later?

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,59 @@ package reward
...
-//   the series term is actually multiplied by both n and the denominator of
-//   lambda.
-//
-//   When we have the numerator and denominator for a given term set up, we
-//   compute their fixed-point fraction by left-shifting the numerator before
-//   performing integer division.
-//
-//   Finally, at the end of each loop, we remove unnecessary bits of precision
-//   from both the numerator and denominator accumulators to control the
-//   computational complexity of the bigint multiplications.
```
anorth, line 214: So sad to see all this wonderful documentation go 😢

actors/builtin/reward/reward_test.go
```diff
@@ -27,6 +27,9 @@ func TestConstructor(t *testing.T) {
 			Build(t)
 		startRealizedPower := abi.NewStoragePower(0)
 		actor.constructAndVerify(rt, &startRealizedPower)
+		st := getState(rt)
+		assert.Equal(t, abi.ChainEpoch(0), st.Epoch)
+		assert.Equal(t, abi.NewStoragePower(0), st.CumsumRealized)
```
anorth, line 32: Please assert the value of the first epoch's reward. Hardcoded is probably ok, but at least check >0.

actors/builtin/reward/reward_test.go
```diff
@@ -35,8 +38,7 @@ func TestConstructor(t *testing.T) {
 		startRealizedPower := abi.NewStoragePower(1 << 25)
 		actor.constructAndVerify(rt, &startRealizedPower)
 		st := getState(rt)
-		assert.Equal(t, abi.ChainEpoch(0), st.RewardEpochsPaid) // constructor shouldn't bump count of rewards
-		assert.Equal(t, startRealizedPower, st.RealizedPower)
+		assert.Equal(t, abi.ChainEpoch(0), st.Epoch)
```
anorth, line 41: Ditto

actors/builtin/reward/expneg_test.go
```diff
@@ -0,0 +1,49 @@
+package reward
+
+import (
+	"bytes"
+	"fmt"
+	"math/big"
+	"testing"
+
+	"github.com/xorcare/golden"
```
anorth, line 9: I like the approach, but this project has a super hard line minimal dependencies policy. Even though this is only used it a test right now, if we ever accidentally reference it it will become part of the set of things for which normal Go versioning semantics are not sufficient and we'll need to vendor, etc.

It look like we only need this for something fairly simple. Could I convince you to write a few more lines of code, inline the golden data here, and be able to drop the dependency?

anorth:
@davidad could you please review the arithmetic implementation?

actors/builtin/reward/expneg_test.go
Kubuxu: @anorth I plan using more golden tests for other computations.
Inlining is not the best as I use an external program to verify the precision of calculations (not possible in Go) and I import golden files as csv.

The `github.com/xorcare/golden` is itself is very small with the only dependency being `assertify` which we are already using.
Either way, I can also just write something like `golden.Assert`.

actors/builtin/reward/reward_logic.go
Kubuxu: I was planning to add more comments and tests, my first priority was getting a working code.

actors/builtin/reward/reward_logic.go
Kubuxu: Not really possible to compute it in code (due to `ln(2)`), nor practical.
I will give exact formula.

actors/builtin/reward/reward_actor.go
Kubuxu: Mul is an order (or two) of magnitude cheaper than Div but alright.

actors/builtin/reward/reward_state.go
Kubuxu: Thanks for pointing this out, I might have an off by one in here.

Kubuxu: Genesis is block 0 AFAIK, so "in which" would be 0 and "for which" it would be 1.

actors/builtin/reward/expneg_test.go
anorth: I'll leave it to your judgement, knowing the situation. We can follow up to re-implement something like that ourselves later.

actors/builtin/reward/reward_actor.go
anorth: We could store both, but I think we'll only need 5 divs per epoch, which is hardly worth optimizing.

actors/builtin/reward/reward_state.go
anorth: The terminology is confusing, but right now, as I understand it:
- the genesis block header has epoch `0`, which means the messages it contains execute in epoch `0`
- the "genesis state" is the state _prior to_ epoch `0`
- a block header points to the parent state (usually, the state resulting from messages in the prior epoch)
- ∴ that genesis state was evaluated in epoch `-1` and the messages _in_ the genesis block in epoch `0`
- so at the end of genesis state construction this should have value `0` if this is "for which"; it's the value for the messages _in_ the genesis block, to be earned by the miner _of_ the genesis block
- for genesis state construction, that may require initializing it to `-1` before then calling the computation function to set the first epoch reward (which will bump it to `0`)

A key point of confusion is distinguishing the state prior to the genesis block from the state resulting from executing any messages contained in it.

There's a brief discussion about changing that [here](https://github.com/filecoin-project/specs/issues/874), but never actioned.

Kubuxu: Yeah, I had an off by one. Now the `Epoch` is *for which epoch*, updated comment too.

actors/abi/big/int.go
```diff
@@ -35,6 +35,16 @@ func PositiveFromUnsignedBytes(b []byte) Int {
 	return Int{i}
 }
 
+// MustFromString convers dec string into big integer and panics if conversion
+// is not sucessful.
+func MustFromString(s string) Int {
```
anorth, line 40: Nit: if this is primarily for testing, I'd prefer it live in the associated test package, so it can't be used by accident in state-affecting code.

actors/builtin/reward/reward_logic.go
```diff
@@ -0,0 +1,78 @@
...
+	} else {
+		// special case for initialization
+		rewardTheta = big.Zero()
+	}
+	return rewardTheta
+}
+
+// These numbers are placeholders, but should be in units of attoFIL, 10^-18 FIL
+var SimpleTotal = big.Mul(big.NewInt(100e6), big.NewInt(1e18))   // 100M for testnet, PARAM_FINISH
+var BaselineTotal = big.Mul(big.NewInt(900e6), big.NewInt(1e18)) // 900M for testnet, PARAM_FINISH
```
anorth, line 36: Nit: move to top next to the baseline function definition?

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,68 @@ package reward
...
 type Spacetime = big.Int
 
 type State struct {
-	BaselinePower        abi.StoragePower
-	RealizedPower        abi.StoragePower
-	CumsumBaseline       Spacetime
-	CumsumRealized       Spacetime
-	EffectiveNetworkTime NetworkTime
+	// CumsumBaseline is a target CumsumRealized needs to reach for EffectiveNetworkTime to increase
+	// CumsumBaseline and CumsumRealized and expressed in bytes
```
anorth, line 13: `and` -> `are`?

Actually they're in byte-epochs.

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,68 @@ package reward
...
-	st.BaselinePower = st.newBaselinePower()
-	st.CumsumBaseline = big.Add(st.CumsumBaseline, st.BaselinePower)
+// Takes in current realized power and updates internal state
+// Used for update of internal state during null rounds
+func (st *State) updateToNextEpoch(currRealizedPower abi.StoragePower) {
+	st.Epoch++
 
-	// Cap realized power in computing CumsumRealized so that progress is only relative to the current epoch.
-	cappedRealizedPower := big.Min(st.BaselinePower, st.RealizedPower)
+	cappedRealizedPower := big.Min(BaselinePowerAt(st.Epoch), currRealizedPower)
```
anorth, line 53: @zixuanzh I don't want to hold up the PR for this (which existed already), but this doesn't match my expectation of how this should work, so want to check whether it matches yours. As I understand this, if at any point the realized spacetime exceeds the baseline, we are forgetting about the excess. This means that if later the realized drops below the baseline, rewards are immediately reduced even if, in reality, the integral of realised power over time exceeds the integral of baseline.

This is as opposed to the approach of calculating the realized cumsum accurately, and then taking the `min` when using it to calculate reward.

anorth:
@Kubuxu please rebase and merge at your pleasure

actors/builtin/reward/reward_state.go
zixuanzh: Yes, we are explicitly capping the realized spacetime at the instantaneous baseline and it is also correct that if the realized baseline immediately drops below the baseline and the baseline share of the reward will reduce (this is part of why we have a proportion allocation to simple minting). This will happen when storage is pulled out from the network. The Min is applied at the instantaneous baseline, not at the limit. Otherwise, we will see very unstable block rewards if the min is at the cumsum. cc @davidad

actors/abi/big/int.go
Kubuxu: It is also for package variables.

actors/builtin/reward/reward_logic.go
```diff
@@ -0,0 +1,78 @@
...
+	return rewardTheta
+}
+
+var (
+	// parameters in Q.128 format
+	// lambda = tropicalYearInSeconds/blockDelay*ln(2)
+	// Precise calculation:
+	// lambda = ln(2) / (6 * 365.24219 * 24 * 60 * 60 / blockDelay(25))
+	// for Q.128: lambdaQ128 = floor(lambda * 2^128)
+	lambda = big.MustFromString("31142895155747063090497695472430")
```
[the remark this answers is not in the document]
Kubuxu: Can you share you `tropicalYear` number?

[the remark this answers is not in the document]
Kubuxu: If I'm to update this, I have to update the comment as well.

actors/builtin/reward/reward_state.go
```diff
@@ -3,305 +3,68 @@ package reward
...
-//   number, and right-shift down by MintingOutputFixedPoint. This convenience
-//   wrapper implements those conventions. However, it does NOT implement
-//   left-shifting the input by the MintingInputFixedPoint, because baseline
-//   minting will actually supply a fractional input.
-func mintingFunction(factor big.Int, t big.Int) big.Int {
-	return big.Rsh(big.Mul(factor, taylorSeriesExpansion(LambdaNum, LambdaDen, t)), MintingOutputFixedPoint)
+// Takes in a current realized power for a reward epoch and computes
+// and updates reward state to track reward for the next epoch
+func (st *State) updateToNextEpochWithReward(currRealizedPower abi.StoragePower) {
+	prevRewardTheta := computeRTheta(st.EffectiveNetworkTime, st.CumsumRealized, st.CumsumBaseline)
```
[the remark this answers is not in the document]
Kubuxu: We update the 𝜃 to be 𝜃(t-1) over here:
https://github.com/filecoin-project/specs-actors/pull/617/files#diff-2e9214f532a06a983c77ad9390267da6R119-R122
which calls https://github.com/filecoin-project/specs-actors/pull/617/files#diff-15478070063dd16268193fab3a001125R50

So no matter number of null rounds the `st.EffectiveNetworkTime, st.CumsumRealized, st.CumsumBaseline` correspond to  `𝜃(t-1)`.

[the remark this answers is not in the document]
davidad: Cool, that’s exactly the sort of thing I was missing 😄

My one remaining concern is that it looks like that catch-up loop will continue to update until `st.Epoch` == `rt.CurrEpoch`, which seems to indicate that after doing the `updateToNextEpochWithReward`, `st.Epoch` will be `rt.CurrEpoch + 1` - is that one too many updates? Or is `rt.CurrEpoch` what I think of conceptually as "t-1"? Put another way, if there aren't any null rounds, are we confident that the catch-up loop will run 0 times rather than once?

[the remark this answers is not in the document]
Kubuxu: You can think of `st.Epoch` as "for which epoch" `st.ThisEpochReward` is valid. This is the epoch after the computation is performed, as the computation is performed at the epoch end (cron tick).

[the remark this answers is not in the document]
davidad: Perfect ☺️

davidad:
My only major concern has been resolved (it was unfounded). The rest of my comments are insignificant. Great work @kubuxu!

PR #618, opened 2020-07-08
Market: Bug fixes in deal timeouts, deal deletion, payment transfers and slashing

### LINKED ISSUE #614: Market: How can a timed-out deal occur in a cron tick
We mark a deal as timeout in the `CronTick` call if the `SectorStartEpoch` of the Deal State is undefined. Take a look here:

https://github.com/filecoin-project/specs-actors/blob/a0ab84303c0ec6694001929e5442f1d7b9c2491e/actors/builtin/market/market_actor.go#L541.

However, this looks _like_ dead code. We create a state for the deal and ALSO set the `SectorStartEpoch` to the current epoch in `ActivateDeals` here:

https://github.com/filecoin-project/specs-actors/blob/a0ab84303c0ec6694001929e5442f1d7b9c2491e/actors/builtin/market/market_actor.go#L374

Note that we never mutate the `SectorStartEpoch` to set it to undefined post this. 

We also abort in `CronTick` if we are unable to find a state for the deal. This means that if a state exists, the `SectorStartEpoch` has to be defined. So, what's the point of this code ?

### LINKED ISSUE #622: Market: DealState is not deleted when deal is deleted in CronTick
We are flushing the wrong `States` variable in Market State.

### LINKED ISSUE #623: Market: Remaining payment to transfer to client after slashing the deal is not calculated correctly
The payment transfer is done as :

```
{
		// Process deal payment for the elapsed epochs.
		totalPayment := big.Mul(big.NewInt(int64(numEpochsElapsed)), deal.StoragePricePerEpoch)

		st.transferBalance(rt, deal.Client, deal.Provider, totalPayment, et, lt)
	}
```

However, to calculate to remaining amount to return to the client(as the deal has been slashed), we do:

```
func dealGetPaymentRemaining(deal *DealProposal, epoch abi.ChainEpoch) abi.TokenAmount {
	Assert(epoch <= deal.EndEpoch)

	durationRemaining := deal.EndEpoch - (epoch - 1)
	Assert(durationRemaining > 0)

	return big.Mul(big.NewInt(int64(durationRemaining)), deal.StoragePricePerEpoch)
}

```

Hence, we have an off by one error in `dealGetPaymentRemaining`. The remaining duration should simply be `durationRemaining := deal.EndEpoch - epoch`

### LINKED ISSUE #624: Market: Handle case where deal is terminated before/at it's start epoch in CronTick
In `OnCronTick`, we run into multiple errors because we always assume that the slash epoch will be greater than the start epoch.

However, a deal can very well be slashed before/at it's start epoch if the whole sector is terminated before the deal starts.

aarshkshah1992:
Closes #614 
Closes #622 
Closes #623 
Closes #624 .

* We now time out deals if they are scheduled for cron tick but they haven't been activated.
* Delete deal state correctly when deal is deleted.
* Calculate the remaining amount to return to client after slashing correctly.
* Simplify calculation of payment for storage.
* Ensure transfer amount isn't negative.
* A deal can be slashed before it's start epoch.

@whyrusleeping Tests will be in a separate PR I am creating for Cron Tick Tests.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -489,34 +489,34 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 	rt.State().Transaction(&st, func() interface{} {
 		dbe, err := AsSetMultimap(adt.AsStore(rt), st.DealOpsByEpoch)
 		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to load deal opts set: %s", err)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deal opts set")
 		}
 
 		updatesNeeded := make(map[abi.ChainEpoch][]abi.DealID)
 
 		states, err := AsDealStateArray(adt.AsStore(rt), st.States)
 		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "get state state: %v", err)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get state state")
 		}
 
 		proposals, err := AsDealProposalArray(adt.AsStore(rt), st.Proposals)
 		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "get deal proposals: %v", err)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get state proposals")
 		}
 
 		et, err := adt.AsBalanceTable(adt.AsStore(rt), st.EscrowTable)
 		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "loading escrow table: %s", err)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load escrow table")
 		}
 
 		lt, err := adt.AsBalanceTable(adt.AsStore(rt), st.LockedTable)
 		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "loading locked balance table: %s", err)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load locked balance table")
 		}
 
 		pending, err := adt.AsMap(adt.AsStore(rt), st.PendingProposals)
 		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "loading pending proposals map: %s", err)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pending proposals map")
 		}
 
 		for i := st.LastCron + 1; i <= rt.CurrEpoch(); i++ {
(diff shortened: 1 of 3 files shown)


aarshkshah1992:
@whyrusleeping Please can you give me a review ?

actors/builtin/market/market_actor.go
```diff
@@ -610,6 +620,11 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 			rt.Abortf(exitcode.ErrIllegalState, "failed to flush deal states: %s", err)
 		}
 
+		st.Proposals, err = proposals.Root()
+		if err != nil {
+			rt.Abortf(exitcode.ErrIllegalState, "failed to flush deal proposals: %s", err)
```
anorth, line 625: Use `RequireNoError` (in lots of places above where the error is unexpected).

actors/builtin/market/market_actor.go
```diff
@@ -534,21 +530,35 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
...
-					slashed := st.processDealInitTimedOut(rt, et, lt, dealID, deal, state)
+					slashed := st.processDealInitTimedOut(rt, et, lt, deal)
 					if !slashed.IsZero() {
 						amountSlashed = big.Add(amountSlashed, slashed)
 					}
 					if deal.VerifiedDeal {
 						timedOutVerifiedDeals = append(timedOutVerifiedDeals, deal)
 					}
+
+					if err := deleteDeal(dealID, states, proposals); err != nil {
```
anorth, line 550: `deleteDeal` deletes from `states` but we know it's not there. That deletion will return an error. This should fail a test.

There are many ways around this, but I think simplest is probably to have separate `deleteDealProposal` and `deleteDealState` mutators, and perhaps a `deleteDealProposalAndState` for that common case. But actually since both of those primitives are direct wrappers of the ADT operations, maybe they don't deserve to be functions: just call directly when you only want one of them.

actors/builtin/market/market_state.go
```diff
@@ -84,7 +83,8 @@ func ConstructState(emptyArrayCid, emptyMapCid, emptyMSetCid cid.Cid) *State {
 // Deal state operations
 ////////////////////////////////////////////////////////////////////////////////
 
-func (st *State) updatePendingDealState(rt Runtime, state *DealState, deal *DealProposal, dealID abi.DealID, et, lt *adt.BalanceTable, epoch abi.ChainEpoch) (abi.TokenAmount, abi.ChainEpoch) {
+func (st *State) updatePendingDealState(rt Runtime, state *DealState, deal *DealProposal, dealID abi.DealID, et, lt *adt.BalanceTable, epoch abi.ChainEpoch) (abi.TokenAmount,
+	abi.ChainEpoch, bool) {
```
anorth, line 87: Please document the return values, I have no idea what this boolean is.

aarshkshah1992: This is done.

actors/builtin/market/market_actor.go
aarshkshah1992: Oh good catch. Fixed this.

actors/builtin/market/market_actor.go
aarshkshah1992: This is done.

aarshkshah1992:
@anorth Coming up with some pretty intense tests in the next PR.

actors/builtin/market/market_actor.go
```diff
@@ -489,66 +489,77 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 	rt.State().Transaction(&st, func() interface{} {
 		dbe, err := AsSetMultimap(adt.AsStore(rt), st.DealOpsByEpoch)
 		if err != nil {
```
anorth, line 491: You don't need the `if err != nil` for `RequireNoErr`, it's handled internally.

aarshkshah1992: @anorth Will fix this next PR.

PR #625, opened 2020-07-08
unlock termination fees before burning

acruikshank:
### Motivation

Termination fees are supposed to be paid out of unlocked funds. We have been paying these fees without first unlocking them. This PR unlocks the funds before sending them to the burnt actor.

### Proposed changes

1. Unlock termination fees before burning them
2. Test that we have unlocked the fees.
3. Lint some whitespace issues from a previous commit.

### NOTES:

Since a sector's initial pledge is the undeclared fault fee (SP) and the termination fee is (at max) the late undeclared fault fee (2*SP), there is no guarantee that the miner will have enough locked funds to pay the expected termination fee. In the case where the miner does not have sufficient funds:

1. The method will not abort.
2. All locked funds will be burnt
3. Unlocked available funds will not be touched.

This is consistent with other methods that charge fees from locked funds.

Some unlinted changes appear to have been checked into master. My linter auto-corrected these and those whitespace changes account for most of the lines in this PR.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -1207,7 +1207,7 @@ func TestTerminateSectors(t *testing.T) {
 		sector := commitSector(t, rt)
 		var initialLockedFunds abi.TokenAmount
 
-		// A miner will path the minimum of termination fee and locked funds. Add some locked funds to ensure
+		// A miner will pay the minimum of termination fee and locked funds. Add some locked funds to ensure
 		// correct fee calculation is used.
 		actor.addLockedFund(rt, big.NewInt(1<<61))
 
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1205,6 +1205,11 @@ func TestTerminateSectors(t *testing.T) {
 	t.Run("removes sector with correct accounting", func(t *testing.T) {
 		rt := builder.Build(t)
 		sector := commitSector(t, rt)
+		var initialLockedFunds abi.TokenAmount
+
+		// A miner will path the minimum of termination fee and locked funds. Add some locked funds to ensure
```
anorth, line 1210: path -> pay?

Issue #416, opened 2020-06-03
Add explicit bounds checks on all slices and bitfields in method parameters


anorth:
All slices and bitfields should have explicit maximum sizes/counts that are checked before further processing.

whyrusleeping:
The goal here is to avoid serialization issues, right?

anorth:
In the case of slices, it's to bound work done iterating the slice. The message encoded size would also bound this, but we don't control it from this repo and it would be best to set some explicit bounds IMO.

For bitfields, it's also more about bounding the work done rather than encoding, although we separately also need to make sure encoding problems fail fast, rather than later like in cron (#391). An abritrarily large bitfield can fit in any message, but we should have some sane maximums on, say, faults that can be declared in one message.

anorth:
The internal call ConfirmSectorProofsValid will be in trouble if a miner proves too many sectors in one epoch. The parameter should probably be changed to a bitfield.

Stebalien:
* [ ] We don't limit the size of saved/forwarded "params". E.g., in `InitActor.Exec`, the payment channel, etc. I think this is fine.
* [ ] It's possible to mask any number of sector numbers at a single time. I  also think this is fine.
* [ ] Technically, fault, recovery, etc. declarations can include as many _declarations_ as they want, as long as they don't exceed the max sector/partition count. For example, a miner could declare multiple faults for the same partition many times.
  * Should we just say "one declaration per partition?" That's pretty easy.
  * ExtendSectorExpiration is the odd one out here. We limit the number of declarations to the "max addressed partitions" limit.
* [ ] In SubmitWindowedPoStParams, the max length of the window post proof isn't enforced. Is there a sane maximum here?

anorth:
I think the fault/recovery are ok either way, but for consistency and directness, I propose we implement the same limit that ExtendSectorExpirations has: limit declarations to max addressed partitions. This is pretty close to declaring "one declaration per partition", but since we're aggregating them anyway I don't think it's a problem to allow duplicates. The miner pays more gas for the inefficient message representation.

It's not worth replicating the aggregation behaviour in ExtendSectorExpirations.

For WPoSt, the max #proofs is the number of different seal algorithms (i.e. 1), and the expected proof length is #partitions * 192.

Stebalien:
> For WPoSt, the max #proofs is the number of different seal algorithms (i.e. 1), and the expected proof length is #partitions * 192.

[...]

PR #632, opened 2020-07-09
limit deals in sectors

acruikshank:
partially addresses #416

### Motivation
If a miner attempts to put too many (8k) deals in a sector, the `PreCommitSector` will fail because the `DealID`slice in the `SectorOnChainInfo` won't marshal into CBOR. Having this many deals in a sector would require an average deal size that is far too low even for TB sectors, anyway.

We will limit the deals in a sector so that average deal size must be above 128MiB. Specifically, this is no more than 256 deals for 32GiB sectors and 512GiB sectors. For sector sizes less than 128MiB we will allow 1 deal.

### Proposed Changes

1. Make the default StackedDRG32GiB the default seal proof type in tests with the option to change it.
2. Add `dealPerSectorLimit` function to policy.
3. Impose limit on deals per sector in `PrecommitSector`.
4. Test.

### DIFF
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -99,6 +99,9 @@ const WorkerKeyChangeDelay = ChainFinality
 // and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
 const MaxSectorExpirationExtension = builtin.EpochsInYear
 
+// Ratio of sector size to maximum deals per sector
+const DealLimitDenominator = 134217728
+
 var QualityBaseMultiplier = big.NewInt(10)         // PARAM_FINISH
 var DealWeightMultiplier = big.NewInt(11)          // PARAM_FINISH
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/policy.go
```diff
@@ -141,6 +144,15 @@ func precommitDeposit(qaSectorPower abi.StoragePower, networkQAPower abi.Storage
 	return InitialPledgeForPower(qaSectorPower, networkQAPower, networkTotalPledge, epochTargetReward, circulatingSupply)
 }
 
+// Determine maximum number of deal miner's sector can hold
+func dealPerSectorLimit(size abi.SectorSize) uint64 {
+	maxDeals := uint64(size / DealLimitDenominator)
+	if maxDeals == 0 {
+		maxDeals = 1
```
ZenGround0, line 151: fyi @whyrusleeping @magik6k we didn't hear back about this, but in case you see this we're setting max deals for 512Mib to 3 and 2kib to 1.

PR #634, opened 2020-07-10
Export min power func

ZenGround0:
Export so that node can use

### DIFF
--- a/actors/builtin/power/power_state.go
+++ b/actors/builtin/power/power_state.go
@@ -67,10 +67,11 @@ func ConstructState(emptyMapCid, emptyMMapCid cid.Cid) *State {
 	}
 }
 
-// Note: this method is currently (Feb 2020) unreferenced in the actor code, but expected to be used to validate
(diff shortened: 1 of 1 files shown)


actors/builtin/power/power_state.go
```diff
@@ -86,7 +87,7 @@ func (st *State) minerNominalPowerMeetsConsensusMinimum(s adt.Store, miner addr.
 		return true, nil
 	}
 
-	// otherwise, if ConsensusMinerMinMiners miners meets min power requirement, return false
+	// otherwise, if ConsensusMinerMinMiners miners meet min power requirement, return false
 	if st.MinerAboveMinPowerCount > ConsensusMinerMinMiners {
```
anorth, line 91: @ZenGround0 it seems to me like this should be changed to `>=`. What do you think?

ZenGround0: Yes, I'll submit a fix

PR #635, opened 2020-07-10
Lint for range map

Kubuxu:


Stebalien:
Nice!

actors/builtin/market/market_actor.go
```diff
@@ -571,10 +571,8 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 			builtin.RequireNoErr(rt, dbe.RemoveAll(i), exitcode.ErrIllegalState, "failed to delete deals from set")
 		}
 
-		// NB: its okay that we're doing a 'random' golang map iteration here
-		// because HAMTs and AMTs are insertion order independent, the same set of
-		// data inserted will always produce the same structure, no matter the order
-		for epoch, deals := range updatesNeeded {
+		// TODO FIXME: iteration over map
```
Stebalien, line 574: Fixed in https://github.com/filecoin-project/specs-actors/pull/594.

.golangci.yml
```diff
@@ -0,0 +1,18 @@
+linters-settings:
+  custom:
+    nomaprange:
+      path: .nomaprange.so
+      description: Checks for range over maps
+      original-url: github.com/Kubuxu/go-no-map-range
```
ZenGround0, line 6: can you help me find source?  If I go to https://github.com/Kubuxu/go-no-map-range I get "no branches"

Kubuxu: Sorry, I've forgot to push a branch (pushed just a tag for go.mod).

.golangci.yml
```diff
@@ -0,0 +1,18 @@
...
+linters:
+  enable:
+    - nomaprange
+
+run:
+  skip-dirs-use-default: false
+
+issues:
+  exclude-rules:
+    - path: _test\.go
```
anorth, line 17: We could probably ignore `support` directory outright too.

anorth:
This failed for me locally so I had to revert for now, but looking forward to figuring out what when wrong and bringing it back.

Issue #267, opened 2020-04-02
Delete ErrPlaceholder


anorth:
We should work out actual codes for any remaining uses.

PR #636, opened 2020-07-10
Feat/slice bound provecommits

ZenGround0:
Joint work with @acruikshank. Along with #632 this closes out the most critical non-miner slice bound work tracked in #416 

Note, given the gas charged per porep is ~130M and gas limit is somewhere in the range of 5-10B, there will only ever be ~100 poreps submitted per block (upper bound with extremely rare really big tipsets is 1000 poreps per block).  So the 8000 limit per miner will in practice never be hit.

With this change a miner who somehow was able to get more provecommits in a tipset than the serialization limit allows won't lose all gas for nothing but will fail early during prove commit before adding the porep info to power actor and avoid gas charging.  Porep infos already added will still be processed during cron.

@Kubuxu double checking in with you: is the upper bound of ~500 poreps per epoch good from a chain scalability perspective?

### DIFF
--- a/actors/builtin/power/policy.go
+++ b/actors/builtin/power/policy.go
@@ -9,3 +9,6 @@ const ConsensusMinerMinMiners = 3
 
 // Minimum power of an individual miner to meet the threshold for leader election.
 var ConsensusMinerMinPower = abi.NewStoragePower(1 << 40) // PARAM_FINISH
+
+// Maximum number of prove commits a miner can submit in one epoch
+const MaxMinerProveCommitsPerEpoch = 8000
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -23,6 +23,10 @@ type Runtime = vmr.Runtime
 type SectorTermination int64
 
 const (
+	ErrTooManyProveCommits = exitcode.FirstActorSpecificExitCode + iota
+)
+
+const (
 	SectorTerminationExpired SectorTermination = iota // Implicit termination after all deals expire
 	SectorTerminationManual                           // Unscheduled explicit termination by the miner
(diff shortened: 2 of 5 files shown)


Kubuxu:
Assuming PoRep verification takes 13ms, then ~500 PoRep verifications would take on the order of 6.5s. So yeah, this is what the network can roughly take.

actors/builtin/power/policy.go
```diff
@@ -9,3 +9,6 @@ const ConsensusMinerMinMiners = 3
 
 // Minimum power of an individual miner to meet the threshold for leader election.
 var ConsensusMinerMinPower = abi.NewStoragePower(1 << 40) // PARAM_FINISH
+
+// Maximum number of prove commits a miner can
```
anorth, line 13: Nit: unfinished comment

actors/builtin/power/power_actor.go
```diff
@@ -22,6 +22,10 @@ type Runtime = vmr.Runtime
 
 type SectorTermination int64
 
+const (
+	ErrTooManyProveCommits = exitcode.ErrPlaceholder + iota + 1
```
anorth, line 26: `ErrPlaceholder` has to go (#267). Please just hardcode a number here. I suggest `32+iota`. We could add a `FirstActorSpecificExitCode=32` if you like. (I just picked that number then based on the shared codes we have now)

support/mock/mockrt.go
```diff
@@ -64,7 +64,8 @@ type Runtime struct {
 	expectVerifyConsensusFault *expectVerifyConsensusFault
 	expectDeleteActor          *address.Address
 
-	logs []string
+	logs       []string
+	gasCharged int64
```
anorth, line 68: Add a note that this is only the gas explicitly charged through the runtime method, and that real VMs will charge most gas implicitly.

PR #638, opened 2020-07-11
Market: Tests for ComputeDataCommitment and ValidateDealsForActivation

### LINKED ISSUE #610: Market: Tests for ComputeDataCommitment and VerifyDealsForActivation
These two functions need to be unit-tested.

aarshkshah1992:
Closes #610 .


actors/builtin/market/market_test.go
```diff
@@ -1709,6 +1711,181 @@ func TestMarketActorDeals(t *testing.T) {
...
+
+		verifiedWeight := big.Add(market.DealWeight(&vd1), market.DealWeight(&vd2))
+		nvweight := big.Add(market.DealWeight(&d1), market.DealWeight(&d2))
+		require.EqualValues(t, verifiedWeight, resp.VerifiedDealWeight)
+		require.EqualValues(t, nvweight, resp.DealWeight)
+	})
+
+	t.Run("fail when caller is not a StorageMinerActor", func(t *testing.T) {
+		rt, actor := basicMarketSetup(t, owner, provider, worker, client)
+		param := &market.VerifyDealsForActivationParams{SectorStart: sectorStart, SectorExpiry: sectorExpiry}
```
anorth, line 1832: This call fails for multiple reasons, e.g. deal not found. In this case the exit code probably distinguishes them, but the test would be stronger if the caller was the _only_ thing wrong with it. In other tests below, the exit code is the same, so you can't actually tell if the condition you think is causing the test to fail is the right one (it might be masked by something else).

I expect them to all look like the basic success case above, calling `actor.verifyDealsForActivation` with a good deal

aarshkshah1992: @anorth Have fixed this.

Issue #599, opened 2020-07-06
Re-structure miner sector state as per-partition aggregates


anorth:
A collection of other issues have led us to the realization that we need to break miner sector lifecycle state (including faults, recoveries, terminations etc) into partition-sized chunks. The concept of a partition of sectors becomes first-class, and the primary structure around which per-sector information is arranged, rather than an implicit interpretation of the sectors assigned to a deadline. Each deadline will contain an explicit list of partition metadata objects.

This is a tracking issue for that work. It subsumes or is related to a bunch of issues, including:
- #411 power/penalty calculations on faults are too slow
- #483 terminating all deals is too slow
- #357 more frequent expiration buckets to avoid bitfield too large
- #519 fault epochs representation is inefficient
- #535 sector termination mutates deadlines mid-period

anorth:
Closed by #648

PR #641, opened 2020-07-13
Disable nomaprange linter plugin, failing on some systems

anorth:
When building master on my local MacBook Pro, lint failed with the new plugin.

```
[~/p/f/specs-actors](anorth/nomaprange)$ make
go build ./...
golangci-lint run ./...
ERRO Unable to load custom analyzer nomaprange:.nomaprange.so, plugin: not implemented
ERRO Running error: no such linter "nomaprange"
make: *** [lint] Error 3
```

Unfortunately I don't have time to debug this right now, will return to it after #599.

FYI @Kubuxu thank you, sorry I didn't have time to investigate right away.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,13 +1,14 @@
 linters-settings:
   custom:
-    nomaprange:
-      path: .nomaprange.so
-      description: Checks for range over maps
-      original-url: github.com/Kubuxu/go-no-map-range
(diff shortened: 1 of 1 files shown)


Kubuxu:
@anorth You have to build and run golangci-lint v1.28.2. The download version does not support plugins, IDK why.

PR #642, opened 2020-07-13
Re-run cbor-gen

### LINKED ISSUE #508: state reuse across send boundaries
We have at least one place where we load state, send a message, then reuse the state in a transaction. Unfortunately, cborgen doesn't clear struct state (yet, https://github.com/whyrusleeping/cbor-gen/issues/21).

The "extra" field could get deleted before the transaction is called (maybe?), then re-persisted in the transaction.

https://github.com/filecoin-project/specs-actors/blob/07f794c2d0eedf7ee9b8a27d53baac7d653bbd37/actors/builtin/paych/paych_actor.go#L189

The fix for this is probably just to merge https://github.com/whyrusleeping/cbor-gen/issues/21, but I wonder if we want some additional checks. We could check `reflect.ValueOf(st).IsZero()` every time (or add an `IsZero` check to state).

anorth:
I noticed deltas when re-generating locally before the update, so decided to update for regeneration.

Here is the [changelog to cbor-gen](https://github.com/whyrusleeping/cbor-gen/compare/429a0b5e922e...5379fc63235d).
- fix some null handling
- clear struct state on unmarshal
- limit max deferred length
- optimizations

Closes #508


anorth:
FYI @whyrusleeping

Issue #509, opened 2020-06-19
Write transactions may not be enough


Stebalien:
Write "transactions" are actually atomic updates. Unfortunately, there's no mirror for atomic _reads_. This can lead to code like:

https://github.com/filecoin-project/specs-actors/blob/07f794c2d0eedf7ee9b8a27d53baac7d653bbd37/actors/builtin/miner/miner_actor.go#L454-L558

This code:

1. Loads the state and a store into an `st` variable.
2. Sends messages.
3. Repeatedly _reloads_ the state into `st` in the transaction, but never reloads the `store`.
4. Uses the state and the store together.

The code may be just fine, but it's hard to reason about.

There's a TODO (#476) about performing all these updates after the loop, but it would be nice if the transaction system made it harder to write code like this. 

One solution would be to introduce read transactions. We'd still be able to copy state out of the read transaction, but we'd have to be explicit about it. Of course, this could make the code even more painful to write, making all this worse.

To avoid re-using state objects, we could use some kind of initialized flag. Or we could just store a lookaside table to record "used" state objects (`map[interface{}]struct{}`).

I'm not familiar enough with the code to know what we _should_ do here, or even if this is a real problem. It just smells a bit funny.

anorth:
The "store" is just a wrapper around the the runtime's Put/Get methods, adapting the API for use as an ipld.CborStore. If we adjusted the API appropriately the adapter wouldn't be needed. There's nothing wrong with this part of the pattern.

I'm not sure there is a real problem here. There is potential for accidentally reading stale state, if it is mutated during a `Send` and not re-loaded. An earlier version of the runtime did structure readonly state access as a "transaction" in the same way as writes, but Go type system limitations make it unreasonable to prevent access of the state object after the transaction ends anyway (it can't be lexically scoped without introducing convoluted loading incantations), so we just dropped the callback.

The linked code is not expecting the state to change during the call to external actors in the loop. Re-loading it each time is an unfortunate cost paid by the pattern, which refactoring to batch will resolve.

Stebalien:
We've fixed all the relevent cases. I'd prefer to have better transactions, but this is good enough for now.

PR #648, opened 2020-07-13
Refactor miner WindowPoSt state into per-partition aggregates and queues

### LINKED ISSUE #357: Consider more frequent sector expiration buckets, expiration registry on pre-commit
We've observed "bitfield too large" errors from `AddSectorExpirations` in `ProveCommitSector`.

This comes down to the low resolution of sector expiration boundaries: expiration is permitted only on proving period boundaries (daily).

It's particularly unfortunate that this problem is only really discoverable to the miner operator in the second stage of commitment, after their precommit deposit is posted. But they'll never be able to recover it, because the expiry epoch was locked in at pre-commit time. At a minimum we could consider registering the sector expiry at precommit, even though this may cause some (rare) junk in the expiration map from sectors that are not subsequently proven.

The below is copied from [this slack conversation](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1588966120245800).

> There is a bit of a state challenge here that might need some thought. Sector expirations are on proving period boundaries (in order that we can process them with daily cron, rather than a cron callback for each sector). The expiring sector numbers are stored in a bitfield indexed by the epoch of the end of the proving period (PP). but if the numbers expiring at the same period are not highly contiguous, then the bitfield may max out. 8k is not a very large number of sectors to commit inside a single PP, and if they all have the same expiry, they'll land in the same bucket.
>
> For context we allow 128<<10 new sectors per proving period, which bounds the size of the NewSectors bitfield when its expanded to a map. But a miner could only get close to that if the sector numbers are contiguous
>
> So you could see this as forcing the miners to help us maintain efficient chain state. Continuously committing a large number of sectors each PP but spreading their expiries out wont work for long because the expiry bitfields won't be dense. but a very large number all committed and expiring on the same boundaries, with contiguous sectornumbers, should work fine.
>
> If we need more flexibility here, we might need to adjust the expiration capability, perhaps allowing expiration on every 1hr boundary or so, rather than daily

FYI @whyrusleeping

### LINKED ISSUE #391: Bitfield too large when saving deadlines in cron
The miner's proving period cron handler can abort with "bitfield too large", which halts their progress (but also causes them to retain their power!).

We need to re-arrange logic here to force any such bitfield computation to happen *before* the cron, e.g. rejecting a new sector when it is first added if it may cause deadlines to overflow.

Generally we shouldn't have problems here if sector numbers are reasonably contiguous, but we need to force miners to meet that property.

https://ipfs.io/ipfs/QmTu2yFKvZ7vN58h25FfUyij5aqM6LZh3XRxardAHe2VAm/trace.html#bafy2bzacednhgynrysnxavsjl3speafkrjcii7dhzcwnlegr4glma2wxxzovo

```
2020-05-21T14:35:19.540-0700	ERROR	vm	vm/runtime.go:140	VM.Call failure: failed to marshal cbor object encoded bitfield was too large (8195) (RetCode=21):
    github.com/filecoin-project/lotus/chain/vm.(*Runtime).Put
        /home/why/code/lotus/chain/vm/runtime.go:127
2020-05-21T14:35:19.540-0700	WARN	vm	vm/vm.go:368	Send actor error	{"from": "t3wa5pdmlm6aewxaxu44w6heoycyfp5atstzi7d6tcdhn4wav2lca3irsndzjnqjx4f3shcwuhbjtd2sicd4ea", "to": "t01020", "nonce": 17114, "method": "7", "height": "6933", "error": "failed to marshal cbor object encoded bitfield was too large (8195) (RetCode=21):\n    github.com/filecoin-project/lotus/chain/vm.(*Runtime).Put\n        /home/why/code/lotus/chain/vm/runtime.go:127"}
2020-05-21T14:35:20.389-0700	ERROR	vm	vm/runtime.go:140	VM.Call failure: failed to marshal cbor object encoded bitfield was too large (8195) (RetCode=21):
    github.com/filecoin-project/lotus/chain/vm.(*Runtime).Put
        /home/why/code/lotus/chain/vm/runtime.go:127
2020-05-21T14:35:20.389-0700	WARN	vm	vm/vm.go:368	Send actor error	{"from": "t3wa5pdmlm6aewxaxu44w6heoycyfp5atstzi7d6tcdhn4wav2lca3irsndzjnqjx4f3shcwuhbjtd2sicd4ea", "to": "t01020", "nonce": 17114, "method": "7", "height": "6933", "error": "failed to marshal cbor object encoded bitfield was too large (8195) (RetCode=21):\n    github.com/filecoin-project/lotus/chain/vm.(*Runtime).Put\n        /home/why/code/lotus/chain/vm/runtime.go:127"}
```

cc @magik6k @whyrusleeping

### LINKED ISSUE #411: Power/penalty calculation for large fault sets or termination is too expensive

### LINKED ISSUE #418: Less strict fault/recovery declaration preconditions
At present a fault declaration including a sector already declared/detected faulty is rejected (rejecting any other declarations that are in the same message). We could easily make this less harsh by accepting but ignoring duplicate declarations. Similar for recoveries.

### LINKED ISSUE #483: Terminating all Deals/Sectors is potentially too slow
From builtin/miner/miner_actor.go: `terminateSectors` --
`// TODO: red flag that this method is potentially super expensive`

 and `requestTerminateAllDeals`:
```
// TODO: red flag this is an ~unbounded computation.
// Transform into an idempotent partial computation that can be progressed on each invocation.
```

These two TODOs stem from the root problem that "it loads all the deal IDs into a slice and makes a method call with it to terminate them."

### LINKED ISSUE #519: FaultEpochs representation is inefficient
We're using an AMT for an array that can have at most 14 entries. Ideally, we'd use a `map[int -> bitfield]`, but we can probably just serialize to an array of index/bitfield tuples (because, IIRC, IPLD doesn't support integer keys in maps).

### LINKED ISSUE #535: Sector termination mutates proving deadline mid-period
The window post mechanism maintains a property that sector deadline/partition assignments change only at the proving period boundary. This is broken by the `TerminateSectors` method, which mutates them immediately. This will cause problems for miners by re-allocating sectors to partitions mid-deadline, but also very likely messes up some accounting that is otherwise maintained only at the PP boundary.

A quick solution to this would be to change TerminateSectors to instead request sector termination at the end of the current proving period. Pay the termination fee now, and then re-register the sector's expiry for the current proving period.

An alternative, more invasive, but probably better solution is to change termination to a two-phase thing as in #551.

FYI @Stebalien

### LINKED ISSUE #552: Per-deadline miner cron
At present, the miner receives a scheduled cron call from the power actor in the last epoch of each proving period. It uses this to do a number of batched accounting things, including:
- detecting Window PoSts that did not arrive on time
- terminating faulty sectors that missed their last Window PoSt
- processing sector expirations (see also #551, #357)

This means there is a delay of up to 24hr between when a sector's final PoSt is due and when power is removed. Similarly, there is a delay before a missed PoSt is detected, and before terminating a faulty-too-long sector.

If we instead scheduled a cron call for the last epoch of every _deadline_ in the proving period, then those things could be done more eagerly, keeping the power table up to date.

Otherwise, we might have to do something awkward like require a miner to keep every sector around for Window PoSt during the subsequent proving period _after it expires_, just to prove they kept it until the proving period ended. We'd likely also need to increase the penalties charged for bad behaviour.

### LINKED ISSUE #593: Better handle DetectFaultsThisPeriod edge case
`detectFaultsThisPeriod` handles an edge case with the following comment 
```
// A cron event has not yet processed the previous proving period and established the next one.
// This is possible in the first non-empty epoch of a proving period if there was an empty tipset on the
// last epoch of the previous period.
// The period must be reset before processing missing-post faults.
```
it then aborts with `exitcode.ErrIllegalState`.  Two potential issues with this:

1.  The exitcode is inaccurate to my understanding of what `ErrIllegalState` means.  This state is legal and possible, it's just not possible to `processMissingPoStFaults` given the state.
2. It may be desirable for the message (submitPoSt, declareFault, declareRecovery) to proceed with other tasks even if it can't process missing post faults from previous deadlines.  AFAIK missing PoSt processing is convenient to add to these miner actor invocations but is separate from the main point of submitting a proof or declaring a fulat or recovery.

anorth:
For motivation, see #599.

This giant PR restructures the miner actor's state, representing partitions as first-class objects. Sectors, faults, recoveries, expiration and terminations are all tracked per-partition. A few totals of power and pledge are maintained in the partition so that power and penalty accounting for faults etc need not load the `SectorOnChainInfo`s for all the sectors (which can be a lot). The heavy per-sector information is only loaded in miner-initiated messages, never from cron.

Significant:
- Partitions are now numbered per-deadline, rather than miner-wide
- When a message references a sector, it usually needs to include information about which deadline and partition the sector is allocated to. The miner must do this search off-chain to avoid on-chain search or indices.
- New sectors are allocated to deadlines and partitions immediately, rather than waiting for cron.
- A cron callback now happens at the end of every deadline, rather than every proving period (#552)
- Penalties are now payed for the proving period in arrears, rather than in advance (in general, penalties are reduced)
- The fault and expiration queues are combined into one (per partition)
- Terminated sectors are not removed from partitions until an explicit Defrag method; they are masked over for Window PoSt like faults

TODO before merging:
- [x] Track initial pledge requirement in expiration queue, release immediately upon on-time termination
- [x] Terminate deals when sectors terminate, pay termination fee, release pledge
- [x] Deny withdrawals when there are pending termination fees
- [x] Quantize the expiration queue
- [x] At WPoSt, charge for recovering sectors (in arrears) before recovering them
- [x] Resolve all new TODOs and XXX

TODO follow-up:
- [ ] Tests!
- [ ] Implement CompactPartitions method #673
- [ ] For early termination, pay SP(t, power) immediately, pay the rest on termination processing (#674)
- [ ] Increase the CBOR-gen maximum bitfield length to accomodate 2350 sector numbers (#676)
- [ ] Fix termination of entire miner (#675)

Closes #391
Closes #357
Closes #391
Closes #411
Closes #418
Closes #483
Closes #519
Closes #535 
Closes #552
Closes #593


Stebalien:
> The heavy per-sector information is only loaded in miner-initiated messages, never from cron.

`ConfirmSectorProofsValid` still does a lot of sector info loading (called by the power actor's `OnEpochTickEnd`, which is called by cron). However, the amount of work done in `ConfirmSectorProofsValid` is proportional to the amount of work paid for by the the miner in the same epoch.

Stebalien:
>  Deny withdrawals when there are pending termination fees

Do we want to forbid sealing sectors, winning blocks, etc. as well? Or just forbid withdrawal.

Stebalien:
> Deal termination queue

It's looking like we're going to need to make this queue per-partition to deal with massive termination batches. If we do that, we're going to need:

* A bitfield at the top level indicating which deadlines have deadlines that need terminating.
* A bitfield at each deadline indicating which partitions have sectors that need terminating.

The other question here is how we're going to deal with early terminations. We don't want miner's withdrawing funds while they haven't payed per-_deal_ early termination penalties.

1. We could (a) not include early terminations in this "normal termination" queue and (b) handle early deal termination in the early termination queue. This is the simplest approach but it means we have two different ways to handle deal termination.
2. We could combine the early termination queue with this queue (two bitfields). However, because the normal termination queue is processed in the background by the chain, this might encourage miners to just let the chain deal with its early termination fees (saving on gas) instead of calling a method to deal with them up-front.

(obviously leaning towards 1 here)

Stebalien:
>    Deal termination queue

The other question is whether or not this actually needs to be a queue. Unless we expect it to fall behind regularly, I'm not sure if it does.

Stebalien:
Proposal:

1. Force the miner to handle all fees/slashing/returning funds (sectors & deals) by calling a method to process early terminations. This way, we handle all fees in one place.
2. Set a limit on how long a miner can have outstanding fees before we terminate the miner. This way, clients are guaranteed to recover locked up funds for slashed deals within a certain amount of time.
3. Handle all state cleanup on defrag. Look ma, no cron!
4. Return pledge on state cleanup to encourage miners to _actually_ defrag.

Am I missing something here?

anorth:
@Stebalien and I talked about this.
- Use the existing per partition early termination queue, and process both deal terminations and termination fee when traversing it
- Only necessary for early terminations; we don't need to tell the market actor about on-time terminations
- In cron, process a limited amount of the queue and if not yet empty, schedule again for next epoch
- Manual termination should enqueue, and then process a large amount of the queue immediately (hopefully emptying it)
    - Manual termination of zero sectors gives a manual queue-processing entry point if desired
- Release pledge requirements as soon as we can; using it to incentivise defrags will result in an inefficiently high number of defrags

Defrag of a partition will have to wait for any early terminations to be processed.

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,363 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+			} else if updated {
+				if empty, err := partition.Sectors.IsEmpty(); err != nil {
+					return err
+				} else if empty {
+					err := partitionsArr.Delete(partIdx)
+					if err != nil {
+						return err
+					}
+				} else {
+					partitionsArr.Set(partIdx, &partition)
```
magik6k, line 477: Unhandled error

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,363 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+	// 1. Inline entire deadlines into the Deadlines struct, putting PostSubmissions bitfield behind a CID.
+	//   Pro: Clean.
+	//   Con: Large Deadlines object.
+	// 2. Split deadlines into a header/details section, inlining the headers.
+	//   Pro: Small Deadlines object.
+	//   Con: Walking deadlines is now kind of funky, but not terrible.
+	// 3. Keep a separate arrays for total sectors and live sectors.
+	//   Con: Annoying to update.
+	var deadlineArr [WPoStPeriodDeadlines]*Deadline
+	deadlines.ForEach(store, func(idx uint64, dl *Deadline) error {
```
magik6k, line 670: Unhandled error

Stebalien:
So, I've hit a bit of a snag. We garbage collect deals as soon as they expire. If we process early terminations _late_, we may try to access missing deals when we compute fines. Can we keep dead deals around for a bit? (and possibly bound how late we can process early termination fees?).

Stebalien:
Answer: For manual terminations, always empty the queue before processing.

anorth:
> Can we keep dead deals around for a bit?

No. For 14-day fault terminations we'll just deal with this occasionally coming a few epochs late if we have a large queue to work through.

Stebalien:
1. How important is it to process early terminations in-order? At the moment, I'm processing all early terminations deadline by deadline, partition by partition, epoch by epoch. That means I'll completely clear a partition before I move onto the next partition. I can process early terminations in order of termination epoch, it'll just require more state. Specifically, I'd need to turn the miner and deadline level "early termination" bitfields into queues.

2. We're going to need a queue mapping epoch to target reward and network QA power at that epoch to compute termination fees. Unless we can query something for these.

I can fix 2 without fixing 1. But in that case, I won't be able to _clear_ the queue from 2 until I've processed _all_ outstanding sector expirations.

anorth:
I think out-of-order is ok for the simplicity gains of doing it by partition.

The state needed for termination fees must be written into the sector on-chain info. The BR(StartEpoch) can be computed from the initial pledge, and my impression was that that would suffice. We can store the BR explicitly if necessary.

actors/builtin/miner/deadline_state.go
```diff
@@ -440,6 +441,11 @@ func (dl *Deadline) PopEarlyTerminations(store adt.Store, max uint64) (earlyTerm
 		return nil, false, xerrors.Errorf("failed to walk early terminations bitfield for deadlines: %w", err)
 	}
 
+	// Removed finished partitions from the index.
+	for _, finished := range partitionsFinished {
+		dl.EarlyTerminations.Unset(finished)
+	}
```
Stebalien, line 447: We can do this for consistency (and I guess it's easier to reason about), but it's shouldn't be necessary (technically).

actors/builtin/miner/miner_actor.go
```diff
@@ -1642,6 +1619,9 @@ func handleProvingDeadline(rt Runtime) {
 		EventType: CronEventProvingDeadline,
 	})
 
+	// Record whether or not we _have_ early terminations now.
+	hasEarlyTerminations := havePendingEarlyTerminations(rt, &st)
```
Stebalien, line 1623: Accessing the state object from the transaction, outside of the transaction, always concerns me a bit (#509). (it's fine in this case, I just wish we had a better way to deal with this)

actors/builtin/miner/miner_actor.go
```diff
@@ -1097,315 +1371,332 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
...
-// - Skipped faults that are already declared, but not recovered will be ignored.
-// - The rest will be penalized as undeclared faults and have their power removed.
-func processSkippedFaults(rt Runtime, st *State, store adt.Store, currDeadline *DeadlineInfo, minerInfo *MinerInfo, deadlines *Deadlines, epochReward abi.TokenAmount, currentTotalPower abi.StoragePower, skipped *bitfield.BitField) ([]*SectorOnChainInfo, abi.TokenAmount) {
+// Discovers how skipped faults declared during post intersect with existing faults and recoveries, records the
+// new faults in state.
+// Returns the amount of power newly faulty, or declared recovered but faulty again.
+
+// - Skipped faults that are not in the provided partition triggers an error.
+// - Skipped faults that are already declared (but not delcared recovered) are ignored.
+func processSkippedFaults(rt Runtime, st *State, store adt.Store, faultExpiration abi.ChainEpoch, partition *Partition,
```
anorth, line 1645: Can/should we push this down to the state? We'd lose the IllegalArgument/IllegalState distinction, but then the state can maintain relations between the fields.

actors/builtin/miner/miner_actor.go
```diff
@@ -1417,146 +1708,47 @@ func validateExpiration(rt Runtime, st *State, activation, expiration abi.ChainE
...
 			expiration, expiration-activation, sealProof.SectorMaximumLifetime(), activation)
 	}
-
-	// ensure expiration is one epoch before a proving period boundary
-	periodOffset := st.ProvingPeriodStart % WPoStProvingPeriod
-	expiryOffset := (expiration + 1) % WPoStProvingPeriod
-	if expiryOffset != periodOffset {
-		rt.Abortf(exitcode.ErrIllegalArgument, "invalid expiration %d, must be immediately before proving period boundary %d mod %d",
-			expiration, periodOffset, WPoStProvingPeriod)
-	}
```
anorth, line 1427: I think we should actually add this back (clamping to epoch before deadline boundary)

actors/builtin/miner/miner_actor.go
```diff
@@ -1417,146 +1708,47 @@ func validateExpiration(rt Runtime, st *State, activation, expiration abi.ChainE
...
+			params.ReplaceSectorNumber, params.ReplaceSectorDeadline, params.ReplaceSectorPartition,
+		)
+	case SectorFaulty:
+		rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace faulty sector %d", params.ReplaceSectorNumber)
+	case SectorTerminated:
+		rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace terminated sector %d", params.ReplaceSectorNumber)
+	case SectorHealthy:
+		// pass
+	default:
+		// TODO: Just panic?
```
anorth, line 1747: Since the exits are all IllegalArgument, return an error instead of status enum?

actors/builtin/miner/miner_actor.go
```diff
@@ -1761,55 +1812,71 @@ func enrollCronEvent(rt Runtime, eventEpoch abi.ChainEpoch, callbackPayload *Cro
 	builtin.RequireSuccess(rt, code, "failed to enroll cron event")
 }
 
-func requestUpdateSectorPower(rt Runtime, sectorSize abi.SectorSize, sectorsAdded, sectorsRemoved []*SectorOnChainInfo) (rawDelta, qaDelta big.Int) {
+func requestUpdatePowerForSectors(rt Runtime, sectorSize abi.SectorSize, sectorsAdded, sectorsRemoved []*SectorOnChainInfo) (delta PowerPair) {
```
anorth, line 1815: This (called only from new sector assignment) breaks the pattern now used elsewhere of having the power returned from the partition-mutating method (which computes it for the expiration queue).

actors/builtin/miner/deadline_state.go
Stebalien: For me it's actually the opposite, but I guess that's because I've worked on the bitfield code but not the AMT code :laughing:.

I agree that maintaining the invariant could be tricky if we ever decide to perform eager updates, or anything like that.

actors/builtin/miner/deadline_assignment.go
```diff
@@ -0,0 +1,87 @@
...
+			})
+		}
+	}
+
+	// Assign sectors to deadlines.
+	for len(sectors) > 0 {
+		info := dlHeap.deadlines[0]
+
+		// Fill up any partial sectors first.
+		size := int((partitionSize - info.liveSectors%partitionSize))
```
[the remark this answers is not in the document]
Stebalien: > Whether or not the final partition is partial depends on the total sectors, not live sectors.

The goal is to fill up partitions assuming the miner defrags (where total sectors will become live sectors).

> If the partitions are exactly full, size is 0 and this step does nothing, and will loop indefinitely.

:roll_eyes: (thanks)

actors/builtin/miner/deadline_state.go
```diff
@@ -0,0 +1,503 @@
...
+			//
+			// Get/create partition to update.
+			//
+
+			partition := new(Partition)
+			if found, err := partitions.Get(partIdx, partition); err != nil {
+				return err
+			} else if !found {
+				// Calling this once per loop is fine. We're almost
+				// never going to add more than one partition per call.
```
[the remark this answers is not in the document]
Stebalien: In the common case, we won't need to create a new partition at all. Doing this here saves us a blockstore write + cid computation in that case (and constructing the empty partition where we need it seems more readable to me).

actors/builtin/miner/deadlines.go
```diff
@@ -102,183 +112,41 @@ func NewDeadlineInfo(periodStart abi.ChainEpoch, deadlineIdx uint64, currEpoch a
...
-	firstAssignableDeadline := uint64(1)
-
-	// Assigns up to `count` sectors to `deadline` and advances `nextNewSector`.
-	assignToDeadline := func(count uint64, deadline uint64) error {
-		countToAdd := min64(count, uint64(len(newSectors))-nextNewSector)
-		sectorsToAdd := newSectors[nextNewSector : nextNewSector+countToAdd]
-		err := deadlines.AddToDeadline(deadline, sectorsToAdd...)
+// FindDeadline returns the deadline index for a given sector number.
+// It returns an error if the sector number is not tracked by deadlines.
+func FindDeadline(store adt.Store, deadlines *Deadlines, sectorNum abi.SectorNumber) (uint64, error) {
```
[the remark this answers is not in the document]
Stebalien: I assume a `FindSector` that returns a deadline + partition would be more useful.

actors/builtin/miner/miner_actor.go
Stebalien: I'm returning an error when we have an illegal state. I could return two errors...

actors/builtin/miner/miner_actor.go
Stebalien: Fixed.

actors/builtin/miner/deadlines.go
```diff
@@ -60,18 +69,19 @@ func (d *DeadlineInfo) NextPeriodStart() abi.ChainEpoch {
...
-		deadlineIdx = 0
-	}
-	return NewDeadlineInfo(periodStart, deadlineIdx, currEpoch)
+	return next
+}
+
+// Returns true if the deadline is currently mutable. The open deadline and the
+// next open deadline are immutable.
+func (d *DeadlineInfo) Mutable() bool {
+	return d.CurrentEpoch < d.Open-WPoStChallengeWindow || d.Open+WPoStChallengeWindow <= d.CurrentEpoch
```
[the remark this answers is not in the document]
Stebalien: Hm. You're right, this is wrong. At the end of the current proving period, we'd mark the first deadline of the next proving period mutable when it's not.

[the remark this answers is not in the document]
Stebalien: LGTY now?

actors/builtin/market/market_actor.go
```diff
@@ -406,6 +406,7 @@ func (a Actor) ComputeDataCommitment(rt Runtime, params *ComputeDataCommitmentPa
 }
 
 type OnMinerSectorsTerminateParams struct {
+	Epoch   abi.ChainEpoch
```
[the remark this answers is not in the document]
Stebalien: Fair enough. Want me to factor this out?

actors/builtin/miner/deadline_state.go
```diff
@@ -0,0 +1,503 @@
...
+	var partitionsWithEarlyTerminations []uint64
+
+	// For each partition with an expiry, remove and collect expirations from the partition queue.
+	if err = expiredPartitions.ForEach(func(partIdx uint64) error {
+		var partition Partition
+		if found, err := partitions.Get(partIdx, &partition); err != nil {
+			return err
+		} else if !found {
+			return xerrors.Errorf("missing expected partition %d", partIdx)
+		}
```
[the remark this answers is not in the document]
Stebalien: Issue: https://github.com/filecoin-project/specs-actors/issues/677. This cleanup will likely span multiple actors.

actors/builtin/miner/miner_actor.go
Stebalien: Does that really help? We can't clamp it to the _right_ deadline.

actors/builtin/miner/expiration_queue.go
```diff
@@ -0,0 +1,636 @@
...
+	}
+	es.OnTimePledge = big.Add(es.OnTimePledge, onTimePledge)
+	es.ActivePower = es.ActivePower.Add(activePower)
+	es.FaultyPower = es.FaultyPower.Add(faultyPower)
+	return nil
+}
+
+// Removes sectors and power from the expiration set in place.
+func (es *ExpirationSet) Remove(onTimeSectors, earlySectors *abi.BitField, onTimePledge abi.TokenAmount, activePower, faultyPower PowerPair) error {
+	// Check for sector intersection. This could be cheaper with a combined intersection/difference method used below.
```
Stebalien, line 61:
```go

// Partition partitions set a with set b.
//
// Intersection = A ^ B
// Difference = A - B
func Partition(a, b *BitField) (intersection, difference BitField, err error)
```

?

actors/builtin/miner/expiration_queue.go
```diff
@@ -0,0 +1,636 @@
...
+		return xerrors.Errorf("removing early sectors %v not contained in %v", earlySectors, es.EarlySectors)
+	}
+
+	var err error
+	if es.OnTimeSectors, err = bitfield.SubtractBitField(es.OnTimeSectors, onTimeSectors); err != nil {
+		return err
+	}
+	if es.EarlySectors, err = bitfield.SubtractBitField(es.EarlySectors, earlySectors); err != nil {
+		return err
+	}
```
Stebalien, line 79: We could also make this faster counting, updating, then counting.

actors/builtin/miner/expiration_queue.go
```diff
@@ -0,0 +1,636 @@
...
+func (q ExpirationQueue) RescheduleRecovered(sectors []*SectorOnChainInfo, ssize abi.SectorSize) (PowerPair, error) {
+	remaining := make(map[abi.SectorNumber]struct{}, len(sectors))
+	for _, s := range sectors {
+		remaining[s.SectorNumber] = struct{}{}
+	}
+
+	// Traverse the expiration queue once to find each recovering sector and remove it from early/faulty there.
+	var sectorsRescheduled []*SectorOnChainInfo
+	recoveredPower := NewPowerPairZero()
+	if err := q.traverseMutate(func(epoch abi.ChainEpoch, es *ExpirationSet) (changed, keepGoing bool, err error) {
```
Stebalien, line 282: We should add a comment here explaining that it can process at most 14 epochs (14 proving periods = 14 deadlines epochs = 14 quantized epochs).

It would be nice to assert this as well (even though it should be impossible to exceed that limit). However, _not_ asserting it is actually safer (in case something goes wrong). Maybe log something?

actors/builtin/miner/expiration_queue.go
```diff
@@ -0,0 +1,636 @@
...
+			return false, false, err
+		}
+		earlySectors, err := es.EarlySectors.AllMap(SectorsMax)
+		if err != nil {
+			return false, false, err
+		}
+
+		// This loop could alternatively be done by constructing bitfields and intersecting them, but it's not
+		// clear that would be much faster (O(max(N, M)) vs O(N+M)).
+		// If faults are correlated, the first queue entry likely has them all anyway.
```
Stebalien, line 294: The first queue entry _with faults_ likely has all of them. We could still process 13 epochs with other expirations first. But I agree that optimizing this is probably not worth it.

actors/builtin/miner/miner_actor.go
anorth: Ok, after discussion I'm coming round to your way in general, but we agree this one is fine as it is.

actors/builtin/miner/deadline_assignment.go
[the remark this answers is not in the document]
Stebalien: Desired properties (in order):

1. Before compaction, avoid increasing the maximum number of partitions in any deadline, across all deadlines.
2. Before compaction, avoid increasing the number of partitions.
3. After compaction, optimize for the minimum maximum number of partitions in any deadline, across all deadlines.
4. After compaction, optimize for the minimum number of partitions.
5. Optimize for runs of sequential sector numbers.

actors/builtin/miner/miner_actor.go
anorth: Right, just panic (`Assert`) for now. The exit code error could help later.

actors/builtin/miner/policy.go
```diff
@@ -39,10 +39,19 @@ func activePartitionsMax(partitionSectorCount uint64) uint64 {
...
 
-// The maximum number of partitions that may be submitted in a single message.
-// This bounds the size of a list/set of sector numbers that might be instantiated to process a submission.
-func windowPoStMessagePartitionsMax(partitionSectorCount uint64) uint64 {
-	return 100_000 / partitionSectorCount
+// The maximum number of partitions that may be required to be loaded in a single invocation.
+// This limits the number of simultaneous fault, recovery, or sector-extension declarations.
+// With 48 deadlines (half-hour), 200 partitions per declaration permits loading a full EiB of 32GiB
+// sectors with 1 message per epoch within a single half-hour deadline. A miner can of course submit more messages.
+const AddressedPartitionsMax = 200
```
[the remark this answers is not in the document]
anorth: https://github.com/filecoin-project/specs-actors/issues/681

actors/builtin/miner/policy.go
```diff
@@ -39,10 +39,19 @@ func activePartitionsMax(partitionSectorCount uint64) uint64 {
...
+
+// The maximum number of sector infos that may be required to be loaded in a single invocation.
+const AddressedSectorsMax = 10_000
```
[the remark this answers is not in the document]
anorth: https://github.com/filecoin-project/specs-actors/issues/681

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,400 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+				if err != nil {
+					return err
+				}
 
-		bf, err = bitfield.SubtractBitField(bf, bitfield.NewFromSet(epochSet.sectors))
-		if err != nil {
-			return xerrors.Errorf("failed to update sector expirations at epoch %v: %w", epochSet.epoch, err)
-		}
+				if empty {
+					err = partitionsArr.Delete(partIdx)
```
anorth, line 475: This will never happen given the known call site. But also this is unsafe: the partitions array must be dense, we can never remove items except when compacting and re-indexing them. Just use `Set`.

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,400 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+	// Mark replaced sectors for on-time expiration at the end of the current proving period.
+	// They can't be removed right now because they may yet be challenged for Window PoSt in this period,
+	// and the deadline assignments can't be changed mid-period.
+	// If their initial pledge hasn't finished vesting yet, it just continues vesting (like other termination paths).
+	// Note that a sector's weight and power were calculated from its lifetime when the sector was first
+	// committed, but are not recalculated here. We only get away with this because we know the replaced sector
+	// has no deals and so its power does not vary with lifetime.
+	// That's a very brittle constraint, and would be much better with two-phase termination (where we could
+	// deduct power immediately).
+	// See https://github.com/filecoin-project/specs-actors/issues/535
```
anorth, line 527: This comment is basically all out of date now.

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,400 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
+func (st *State) AssignSectorsToDeadlines(
+	store adt.Store,
+	currentEpoch abi.ChainEpoch,
+	sectors []*SectorOnChainInfo,
+	partitionSize uint64,
+	sectorSize abi.SectorSize,
+	quant QuantSpec,
+) (PowerPair, error) {
+	// We should definitely sort by sector number.
+	// Should we try to find a deadline with nearby sectors? That's probably really expensive.
```
anorth, line 603: Unnecessary comments

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,400 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
-			return fmt.Errorf("too many faults %d, max %d", count, SectorsMax)
+	// TODO: We should either:
+	// 1. Inline entire deadlines into the Deadlines struct, putting PostSubmissions bitfield behind a CID.
+	//   Pro: Clean.
+	//   Con: Large Deadlines object.
+	// 2. Split deadlines into a header/details section, inlining the headers.
+	//   Pro: Small Deadlines object.
+	//   Con: Walking deadlines is now kind of funky, but not terrible.
+	// 3. Keep a separate arrays for total sectors and live sectors.
+	//   Con: Annoying to update.
```
anorth, line 622: There's a note about inlining in the Deadlines struct. Could you add any still-relevant info there (without TODO, since I don't think we intend to do any of them)

actors/builtin/miner/termination.go
```diff
@@ -0,0 +1,62 @@
...
+	PartitionsProcessed, SectorsProcessed uint64
+}
+
+func (t *TerminationResult) Add(newResult TerminationResult) error {
+	if t.Sectors == nil {
+		t.Sectors = make(map[abi.ChainEpoch]*abi.BitField, len(newResult.Sectors))
+	}
+	t.PartitionsProcessed += newResult.PartitionsProcessed
+	t.SectorsProcessed += newResult.SectorsProcessed
+	for epoch, newSectors := range newResult.Sectors {
```
anorth, line 24: Unordered map iteration. Looks like `nolint:nomaprange` is ok tho?

actors/builtin/reward/reward_test.go
```diff
@@ -43,7 +43,7 @@ func TestConstructor(t *testing.T) {
 		assert.Equal(t, startRealizedPower, st.CumsumRealized)
 
 		// Note this check is sensative to the value of startRealizedPower and the minting function
-		// so it is somewhat brittle. Values of startRealizedPower below 1<<20 mint no coins
+		// so it is somewhat brittle. OnTimeSectors of startRealizedPower below 1<<20 mint no coins
```
anorth, line 46: Oops

actors/builtin/miner/miner_actor.go
```diff
@@ -210,42 +220,51 @@ func (a Actor) ChangeMultiaddrs(rt Runtime, params *ChangeMultiaddrsParams) *adt
...
 	store := adt.AsStore(rt)
 	var st State
-	var newFaultSectors []*SectorOnChainInfo
-	var recoveredSectors []*SectorOnChainInfo
-	penalty := abi.NewTokenAmount(0)
+
+	if params.Deadline > WPoStPeriodDeadlines {
+		rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d of %d", params.Deadline, WPoStPeriodDeadlines)
+	}
+	// TODO: limit the length of proofs array https://github.com/filecoin-project/specs-actors/issues/416
```
Stebalien, line 250: @anorth, this is done, right?

https://github.com/filecoin-project/specs-actors/pull/648/files#diff-e47fb4c42d791672c9550edb1dc5f969R267

actors/builtin/miner/miner_state.go
Stebalien: Eh, all of this comment can go.

actors/builtin/miner/miner_state.go
```diff
@@ -748,49 +801,42 @@ func (st *State) LoadSectorInfos(store adt.Store, sectors *abi.BitField) ([]*Sec
...
 // If any of the sectors are declared recovered, they are returned from this method.
-func (st *State) LoadSectorInfosForProof(store adt.Store, provenSectors *abi.BitField) (sectorInfos []*SectorOnChainInfo, recoveries *bitfield.BitField, err error) {
-	// Extract a fault set relevant to the sectors being submitted, for expansion into a map.
-	declaredFaults, err := bitfield.IntersectBitField(provenSectors, st.Faults)
-	if err != nil {
-		return nil, nil, xerrors.Errorf("failed to intersect proof sectors with faults: %w", err)
-	}
-
-	recoveries, err = bitfield.IntersectBitField(declaredFaults, st.Recoveries)
+func (st *State) LoadSectorInfosForProof(store adt.Store, provenSectors, expectedFaults []*abi.BitField) ([]*SectorOnChainInfo, error) {
```
Stebalien, line 804: Any reason not to merge the bitfields outside fo this function?

actors/builtin/market/market_actor.go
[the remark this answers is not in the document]
Stebalien: Eh, :boat:.

actors/builtin/miner/miner_actor.go
anorth: We didn't make it worse, but I think still something to do here.

actors/builtin/miner/miner_state.go
anorth: No, either way is fine.

anorth:
I agree, checking more assumptions and preconditions in the partition would be worthwhile.

actors/builtin/miner/expiration_queue.go
```diff
@@ -0,0 +1,638 @@
...
+		return nil, nil, NewPowerPairZero(), big.Zero(), xerrors.Errorf("failed to add replacement sectors: %w", err)
+	}
+	return oldSnos, newSnos, newPower.Sub(oldPower), big.Sub(newPledge, oldPledge), nil
+}
+
+// Remove some sectors from the queue.
+// The sectors may be active or faulty, and scheduled either for on-time or early termination.
+// Returns the aggregate of removed sectors and power, and recovering power.
+// Fails if any sectors are not found in the queue.
+func (q ExpirationQueue) RemoveSectors(sectors []*SectorOnChainInfo, faults *abi.BitField, recovering *abi.BitField,
```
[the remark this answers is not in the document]
anorth: I see, yes. It's a bit gnarly but would be better. We could possibly change the signature to require faults and nonfault sectorinfos separately

actors/builtin/miner/miner_actor.go
```diff
@@ -1097,315 +1470,331 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
...
+			// This happens first so that any subsequent penalties are taken from locked pledge, rather than free funds.
+			newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
+		}
+
+		// Record whether or not we _had_ early terminations in the queue before this method.
+		// That way, don't re-schedule a cron callback if one is already scheduled.
+		hadEarlyTerminations = havePendingEarlyTerminations(rt, &st)
+
+		// Note: because the cron actor is not invoked on epochs with empty tipsets, the current epoch is not necessarily
```
Stebalien, line 1586: What if we're multiple deadlines ahead? I'm guessing we should skip the missed deadlines and give the miner a pass, but we should probably do something.

actors/builtin/miner/miner_actor.go
```diff
@@ -1097,315 +1470,331 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
...
+		err = st.SaveDeadlines(store, deadlines)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save deadlines")
 
-			// Set new proving period start.
-			if deadline.PeriodStarted() {
+		// Increment current deadline, and proving period if necessary.
+		if dlInfo.PeriodStarted() {
+			st.CurrentDeadline = (st.CurrentDeadline + 1) % WPoStPeriodDeadlines
+			if st.CurrentDeadline == 0 {
 				st.ProvingPeriodStart = st.ProvingPeriodStart + WPoStProvingPeriod
```
Stebalien, line 1702: Do we need to handle getting a full proving period behind?

actors/builtin/miner/miner_actor.go
```diff
@@ -1097,315 +1470,331 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
...
+		err = deadlines.UpdateDeadline(store, dlInfo.Index, deadline)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to update deadline %d", dlInfo.Index)
+
+		err = st.SaveDeadlines(store, deadlines)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save deadlines")
 
-			// Set new proving period start.
-			if deadline.PeriodStarted() {
+		// Increment current deadline, and proving period if necessary.
+		if dlInfo.PeriodStarted() {
```
Stebalien, line 1699: This can't be false, can it?

actors/builtin/miner/miner_actor.go
```diff
@@ -1097,315 +1470,331 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
...
+		quant := st.QuantEndOfDeadline()
+
+		{
+			// Detect and penalize missing proofs.
+			faultExpiration := dlInfo.Close + FaultMaxAge
+			penalizePowerTotal := big.Zero()
+
+			for i := uint64(0); i < partitions.Length(); i++ {
+				key := PartitionKey{dlInfo.Index, i}
+				proven, err := deadline.PostSubmissions.IsSet(i)
```
Stebalien, line 1610: This is potentially slow. We should probably just expand this into a map (it's not _too_ large). We can also do this with bitfield magic, but that's probably more complicated than it's worth.

actors/builtin/miner/miner_actor.go
```diff
@@ -266,79 +288,117 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+		faultExpiration := currDeadline.Close + FaultMaxAge
+
+		partitionIdxs := make([]uint64, 0, len(params.Partitions))
+		allSectors := make([]*abi.BitField, 0, len(params.Partitions))
+		allIgnored := make([]*abi.BitField, 0, len(params.Partitions))
+
+		// Accumulate sectors info for proof verification.
+		for _, post := range params.Partitions {
+			key := PartitionKey{params.Deadline, post.Index}
+			alreadyProven, err := deadline.PostSubmissions.IsSet(post.Index)
```
Stebalien, line 306: I'd expand PostSubmissions into a map.

actors/builtin/miner/miner_actor.go
```diff
@@ -1097,315 +1470,331 @@ func (a Actor) OnDeferredCronEvent(rt Runtime, payload *CronEventPayload) *adt.E
...
+				st.FaultyPower = st.FaultyPower.Add(newFaultPower)
+				powerDelta = powerDelta.Sub(newFaultPower)
+				penalizePowerTotal = big.Sum(penalizePowerTotal, newFaultPower.QA, failedRecoveryPower.QA)
+			}
+
+			// Unlock sector penalty for all undeclared faults.
+			penaltyTarget := PledgePenaltyForUndeclaredFault(epochReward, pwrTotal.QualityAdjPower, penalizePowerTotal)
+			// Subtract the "ongoing" fault fee from the amount charged now, since it will be added on just below.
+			penaltyTarget = big.Sub(penaltyTarget, PledgePenaltyForDeclaredFault(epochReward, pwrTotal.QualityAdjPower, penalizePowerTotal))
+			penalty, err := st.UnlockUnvestedFunds(store, currEpoch, penaltyTarget)
```
Stebalien, line 1640: Can we do this once at the very end?

actors/builtin/miner/deadline_assignment.go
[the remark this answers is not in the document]
Stebalien: I've broken this out into a followup issue so we can merge this PR first: https://github.com/filecoin-project/specs-actors/issues/685.

actors/builtin/miner/miner_state.go
```diff
@@ -360,367 +385,400 @@ func (st *State) ForEachSector(store adt.Store, f func(*SectorOnChainInfo)) erro
...
-	count, err := st.NewSectors.Count()
+// Walks the given sectors, deadline by deadline, partition by partition,
+// skipping missing partitions/sectors.
+func (st *State) WalkSectors(
+	store adt.Store,
+	locations []SectorLocation,
+	beforeDeadlineCb func(dlIdx uint64, dl *Deadline) (update bool, err error),
+	partitionCb func(dl *Deadline, partition *Partition, dlIdx, partIdx uint64, sectors *bitfield.BitField) (update bool, err error),
+	afterDeadlineCb func(dlIdx uint64, dl *Deadline) (update bool, err error),
+) error {
```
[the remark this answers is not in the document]
Stebalien: https://github.com/filecoin-project/specs-actors/issues/686

actors/builtin/miner/expiration_queue.go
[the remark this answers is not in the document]
Stebalien: #687.

Stebalien:
I think all the important bits have either been implemented or recorded in new issues. My only remaining concern is dealing with too many null blocks, but we can extract that into a new issue as well.

PR #651, opened 2020-07-13
Version golangci-lint and build it with make

Kubuxu:
`golangci-lint` and `no-map-range` are in a separate go.mod file so they do not increase dependency surface of specs-actors


Makefile
```diff
@@ -20,16 +18,20 @@ tidy:
...
 gen:
 	$(GO_BIN) run ./gen/gen.go
 .PHONY: gen
+
+
+# tools
+toolspath:=support/tools
+
+$(toolspath)/golangci-lint: $(toolspath)/go.mod
+	(cd $(toolspath); go build github.com/golangci/golangci-lint/cmd/golangci-lint)
```
Stebalien, line 30:
```suggestion
	(cd $(toolspath); go build -o "$@" github.com/golangci/golangci-lint/cmd/golangci-lint)
```

Stebalien: (just to be safe)

.circleci/config.yml
```diff
@@ -93,24 +93,11 @@ jobs:
...
-              cd $(mktemp -d);
-              go mod init a;
-              go get github.com/golangci/golangci-lint/cmd/golangci-lint@v<< parameters.golangci-lint-version >>
-              go build github.com/golangci/golangci-lint/cmd/golangci-lint
-              mkdir -p $HOME/.local/bin
-              mv -f golangci-lint $HOME/.local/bin/
-            )
-      - run:
-          command: make .nomaprange.so
+          command: make -j3 support/tools/golangci-lint support/tools/no-map-range.so
```
Stebalien, line 96: `-j$(nproc)`?

.gitignore
```diff
@@ -15,3 +15,6 @@
 # vendor/
 
 .idea
+
+support/tools/golangci-lint
+support/tools/no-map-range.so
```
Stebalien, line 20: Note: It's probably better to just have a `support/tools/bin` directory.

Makefile
Kubuxu: `$@` is relative, and we have to `cd` for go.mod to work

.circleci/config.yml
Kubuxu: I think circle reports way more cores than it really gives you.

Stebalien: Fair enough.

Makefile
Kubuxu: Did this: `$(@:$(toolspath)/%=%)`

PR #647, opened 2020-07-13
Market: Load Proposals AMT only once

aarshkshah1992:
Closes #608 .
Closes #609.

PR #655, opened 2020-07-13
Caching and partial possible cron tick refactoring

ZenGround0:
@anorth @acruikshank @Stebalien @wadeAlexC
Hey all, @aarshkshah1992 and I recently discussed how to handle redundant reads in adt structures. 

The current approach is to read structures once in an actor method's state transaction, pass this object around as an argument to state functions and other functions that use it, and then flush and replace pointer at the end of the state transaction.  

Another possible approach is done in this spike PR.  The state object holds onto a cache of all adt structures, initializing on the first read and from then on accessing the same adt object.  Flushes are then done on the whole state.

We wanted to see a sketch of the second approach to get a feel for how the two compare.  This PR applies this approach to the market actor state and uses it in the cron tick method.  Note that the cron tick and market state changes are not followed through to the extreme -- we'd want to make it so that none of the adts reachable from the market state's fields were passed in and only read through the cache if we ran with this.  But this PR gives the basic idea.

Cons
* This is a bit more verbose because we add all of the caching functionality to state but we don't get to get rid of all the reads at the start of the cron ticket method's state transaction because the actor needs some of these adts
* Doing this everywhere is going to be more upfront work than going down the current trajectory

Pros
* Conceptually simple. If we adopt this approach it is hard to do redundant reads.  This could make getting rid of redundant reads easier to finish.  Note that with this approach `mustGetDeal` was fixed throughout the code with a one line change in market state whereas #647 needs to change ~5 lines across state and actor.
* Noise reduced in actor.  This is especially true for flushes but also will be for a subset of reads especially if we committed to this approach and stopped passing in adt structures to state methods.

If you have time and interest I'd like some input.  If I don't hear back soon we will just scrap this and go with the current approach.

### DIFF
--- a/actors/builtin/market/market_state.go
+++ b/actors/builtin/market/market_state.go
@@ -2,6 +2,7 @@ package market
 
 import (
 	"bytes"
+	"fmt"
 
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/specs-actors/actors/builtin"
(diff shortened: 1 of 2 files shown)


actors/builtin/market/market_state.go
```diff
@@ -346,7 +475,7 @@ func (st *State) slashBalance(et, lt *adt.BalanceTable, addr addr.Address, amoun
 ////////////////////////////////////////////////////////////////////////////////
 
 func (st *State) mustGetDeal(rt Runtime, dealID abi.DealID) *DealProposal {
-	proposals, err := AsDealProposalArray(adt.AsStore(rt), st.Proposals)
+	proposals, err := st.CachedProposals(adt.AsStore(rt))
```
ZenGround0, line 478: This pattern would be repeated throughout state if we adopted this approach.  I held off from doing this with updatePendingDealState and processDealInitTimedOut because the cache read would introduce err values to those methods.

Stebalien:
I like the idea but we need to be a bit careful about not accessing this state any other way.

We might be able to introduce a caching wrapper type to deal with this:

```go
package caching

type Array struct {
  cid.Cid
  cached *adt.Array
}
//...

func (a *Array) Load(store) (*adt.Array, error) { ... }
```

Although that might be more work than it's worth.

wadeAlexC:
> I like the idea but we need to be a bit careful about not accessing this state any other way.

For this reason, I like the wrapper idea @Stebalien proposed. IMO, specs-actors doesn't have strict conventions for how/when/where the state writes happen in code. Without strict conventions, I think it'll be difficult to ensure that state is only accessed a particular way. For example:

```go
func (a Actor) ExportedMethod(rt Runtime, _ EmptyValue) EmptyValue {
  var st State
  rt.State().Transaction(&st, func() interface {} {
    st.doAThing(); // state write happens here, in another file
    return nil
  })
}
```

The convention above is used in several actors. When I want to understand a particular field in an actor's `State`, I can't just "ctrl-f" in the `x_actor.go` file - I have to search all the files in that actor's directory. When a particular field is updated by a helper method in another file, I have to add that helper method to my search criteria in order to understand which exported methods deal with which fields. This quickly racks up mental overhead for some of the more complicated actors!

anorth:
Thanks. This is very interesting. I am also worried about potential consistency problems. It makes it easier to avoid redundant reads, but a bit harder (?) to be certain that all reads are seeing the most recent state? The verbose current approach doesn't seem risky (tho we risk forgetting to flush writes instead).

PR #659, opened 2020-07-13
process power actor cron tasks set in the past

### LINKED ISSUE #413: Power actor EnrollCronEvent doesn't handle enrolments for events in the past
The `EnrollCronEvent` method does no validation of the epoch for which an event is enrolled. If it's at or prior to the recorded `LastEpochTick` then the event will never be executed and will remain in state forever. Events are stored in a HAMT keyed by epoch, which does not support ordered traversal.

A simple solution might be to require the registered epoch to be after the current epoch. However at present it is possible for a registration to be in "the past" when there is a long chain halt (greater than a proving period) from the miner actor, which steps its proving period marker forward one period at a time via cron events.

A workaround-style solution might be to change how the miner actor registers for proving period callbacks, ensuring registrations are always in the future, even when catching up from a long halt. This seems brittle, though, and easy to accidentally introduce a new case that might require an in-the-past registration (I haven't audited other cases yet).

Another possible path out is switching all such callback mechanisms to use chain height (#408), which will never have gaps or null rounds.

I think a robust solution would be to change the cron events table representation to an AMT, and always traverse it in order from beginning up to the current epoch. The event callbacks must then be robust to being called at an epoch later than they registered, but that's necessary anyway to handle empty tipsets.

acruikshank:
closes #413

### Motivation

Currently consensus processes may conspire so that a miner actor will end up scheduling a cron task in the power actor for an epoch that precedes the power actors last cron tick. If this happens the task will not be run and the data structure holding it will never be removed from power actor state.

### Proposed Changes

1. Rename `power.State.LastCronTick` to `power.State.FirstCronEpoch` and alter its value to be the first epoch _after_ the last cron (or 0).
2. Add logic to move set `FirstCronEpoch` back if a task is set to an epoch at or before the last Cron tick.
3. Test that this allows tasks run in the past to execute on the next tick and that these tasks are garbage collected.

### DIFF
--- a/actors/builtin/power/power_state.go
+++ b/actors/builtin/power/power_state.go
@@ -28,8 +28,8 @@ type State struct {
 	// A queue of events to be triggered by cron, indexed by epoch.
 	CronEventQueue cid.Cid // Multimap, (HAMT[ChainEpoch]AMT[CronEvent]
 
-	// Last chain epoch OnEpochTickEnd was called on
-	LastEpochTick abi.ChainEpoch
+	// First epoch in which a cron task may be stored.
+	FirstCronTask abi.ChainEpoch
 
 	// Claimed power for each miner.
 	Claims cid.Cid // Map, HAMT[address]Claim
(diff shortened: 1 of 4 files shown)


actors/builtin/power/power_state.go
```diff
@@ -28,8 +28,8 @@ type State struct {
 	// A queue of events to be triggered by cron, indexed by epoch.
 	CronEventQueue cid.Cid // Multimap, (HAMT[ChainEpoch]AMT[CronEvent]
 
-	// Last chain epoch OnEpochTickEnd was called on
-	LastEpochTick abi.ChainEpoch
+	// First epoch in which a cron task may be stored.
```
anorth, line 31: Please note here that this scheme means the actor will iterate every epoch from FirstCronEpoch up to the current epoch, checking the CronEventQueue.

PR #664, opened 2020-07-14
Adding a method to replace verifreg root key

mrsmkl:
Just adds a way to change the root key holder. It might be useful for initializing the root key or upgrading later on.

This change probably isn't absolutely necessary since looks like there's going to be a way to include multisig wallets in the genesis block https://github.com/filecoin-project/lotus/pulls ... So if it won't be included then this branch will be useful for testing until another solution is working.


anorth:
Thanks for your contribution. To the best of my knowledge, the root key is very much intended to be a multisig.

@whyrusleeping can you comment on the need or not for this? I'd rather avoid the extra code if the same functionality can be obtained with multisig.

whyrusleeping:
Changing the root key should necessitate a fork IMO, i'd rather not have this method here.

PR #678, opened 2020-07-15
Remove duplicate reads between AddLockedFunds and UnlockVestedFunds

### LINKED ISSUE #502: Improve performance of miner.AddLockedFund, serialize as array
Rough profiling indicates that the AddLockedFund method is the most significant sink of execution time outside of syscalls.

See this [profile snippet](https://gist.github.com/anorth/7e632d95518e9ae975f843b723380c23) from a devnet.

And from a different one
```
Average time spent in outside of syscalls per actor method:
t05-7 0.000111405
t04-11 9.31141e-05
t05-8 0.000307631
t04-13 0.000194394
t04-4 0.000269679
t01000-14 0.00131454
t02-2 0.000121549
t02-3 4.98606e-05
t01000-17 0.00367811
t02-4 0.000311329
t01000-4 0.000763825
t04-9 0.000201034
t01000-6 0.000491067
t05-5 0.000187442
t01000-7 0.000385646
t03-2 9.50445e-05
t04-10 0.000229153
```

FYI @whyrusleeping 

See also #464

ZenGround0:
Remove duplicate gets between locking and unlocking funds for #502 

We shouldn't consider #502 done until we do more profiling to be sure this is no longer a bottleneck.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -407,7 +407,9 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 			depositMinimum = replaceSector.InitialPledge
 		}
 
-		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
+		vestingFunds, err := adt.AsArray(store, st.VestingFunds)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load vested funds")
+		newlyVestedFund, err := st.UnlockVestedFunds(vestingFunds, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
 		duration := params.Expiration - rt.CurrEpoch()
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1109,8 +1130,14 @@ func handleProvingPeriod(rt Runtime) {
...
 		newlyVestedAmount := rt.State().Transaction(&st, func() interface{} {
-			newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
+			vestingFunds, err := adt.AsArray(adt.AsStore(rt), st.VestingFunds)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load vesting funds")
+			newlyVestedFund, err := st.UnlockVestedFunds(vestingFunds, rt.CurrEpoch())
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
+
+			st.VestingFunds, err = vestingFunds.Root()
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to write vesting funds")
+
```
anorth, line 1140: This is going to conflict with the #648. Would you mind either reverting this one, or hold off landing until the other does?

ZenGround0: 👍 It's a simple change I'll hold off until the big one lands

ZenGround0:
>  I suspect it's not the whole issue

Yes me too. I'm hoping the batching we did in #476 did the bulk of it but we'll see how it shakes out when we reprofile.

ZenGround0:
It looks like this is going to be made irrelevant by #843.  @aarshkshah1992 @anorth is that right?  If so I will close this down.

anorth:
Yep

PR #679, opened 2020-07-15
distinguish between randomness from drand vs ticket chain randomness

whyrusleeping:
Ref @nicola's discussion on GetRandomnessFromBeacon and GetRandomnessFromTickets

@anorth edit:
Documenting the behaviour being implemented here before landing it.
- Seal: ticket chain
- PoRep: beacon chain
- Window PoSt: beacon chain

This ties the seal to a particular chain, but not the subsequent PoRep or PoSts.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1874,7 +1874,7 @@ func getVerifyInfo(rt Runtime, params *SealVerifyStuff) *abi.SealVerifyInfo {
 	AssertNoError(err)
 
 	svInfoRandomness := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_SealRandomness, params.SealRandEpoch, buf.Bytes())
-	svInfoInteractiveRandomness := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_InteractiveSealChallengeSeed, params.InteractiveEpoch, buf.Bytes())
+	svInfoInteractiveRandomness := rt.GetRandomnessFromBeacon(crypto.DomainSeparationTag_InteractiveSealChallengeSeed, params.InteractiveEpoch, buf.Bytes())
 
 	return &abi.SealVerifyInfo{
 		SealProof: params.RegisteredSealProof,
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1814,7 +1814,7 @@ func verifyWindowedPost(rt Runtime, challengeEpoch abi.ChainEpoch, sectors []*Se
 	var addrBuf bytes.Buffer
 	err = rt.Message().Receiver().MarshalCBOR(&addrBuf)
 	AssertNoError(err)
-	postRandomness := rt.GetRandomness(crypto.DomainSeparationTag_WindowedPoStChallengeSeed, challengeEpoch, addrBuf.Bytes())
+	postRandomness := rt.GetRandomnessFromBeacon(crypto.DomainSeparationTag_WindowedPoStChallengeSeed, challengeEpoch, addrBuf.Bytes())
 
 	sectorProofInfo := make([]abi.SectorInfo, len(sectors))
 	for i, s := range sectors {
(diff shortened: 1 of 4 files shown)


whyrusleeping:
FYI @magik6k as this will need to be updated in lotus

actors/runtime/runtime.go
```diff
@@ -57,9 +57,15 @@ type Runtime interface {
 	// Look up the code ID at an actor address.
 	GetActorCodeCID(addr addr.Address) (ret cid.Cid, ok bool)
 
-	// Randomness returns a (pseudo)random byte array drawing from a
+	// GetRandomnessFromBeacon returns a (pseudo)random byte array drawing from a
 	// random beacon at a given epoch and incorporating reequisite entropy
-	GetRandomness(personalization crypto.DomainSeparationTag, randEpoch abi.ChainEpoch, entropy []byte) abi.Randomness
+	GetRandomnessFromBeacon(personalization crypto.DomainSeparationTag, randEpoch abi.ChainEpoch, entropy []byte) abi.Randomness
```
ZenGround0, line 62: These names are clear 👍

actors/builtin/miner/miner_actor.go
```diff
@@ -1875,8 +1875,8 @@ func getVerifyInfo(rt Runtime, params *SealVerifyStuff) *abi.SealVerifyInfo {
 	err = rt.Message().Receiver().MarshalCBOR(buf)
 	AssertNoError(err)
 
-	svInfoRandomness := rt.GetRandomness(crypto.DomainSeparationTag_SealRandomness, params.SealRandEpoch, buf.Bytes())
-	svInfoInteractiveRandomness := rt.GetRandomness(crypto.DomainSeparationTag_InteractiveSealChallengeSeed, params.InteractiveEpoch, buf.Bytes())
+	svInfoRandomness := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_SealRandomness, params.SealRandEpoch, buf.Bytes())
+	svInfoInteractiveRandomness := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_InteractiveSealChallengeSeed, params.InteractiveEpoch, buf.Bytes())
```
nicola, line 1879: This would be `GetRandomnessFromBeacon`

actors/runtime/runtime.go
nicola, line 62: Where do I find the implementation of this in lotus?

whyrusleeping: its not there yet, this needs to happen first

actors/builtin/miner/miner_actor.go
whyrusleeping: I thought we said all the PoRep stuff was using the vrf chain?

nicola: No, there is an analysis that shows that we could, but we should rather not do it. It reduces the PoRep security and it changes the way the VRF chain is generated.

nicola: I "be the change you want to see"d

support/mock/mockrt.go
```diff
@@ -263,14 +264,37 @@ func (rt *Runtime) GetRandomness(tag crypto.DomainSeparationTag, epoch abi.Chain
...
 		rt.failTest("unexpected get randomness\n"+
 			"         tag: %d, epoch: %d, entropy: %v\n"+
 			"expected tag: %d, epoch: %d, entropy: %v", tag, epoch, entropy, exp.tag, exp.epoch, exp.entropy)
 	}
 	defer func() {
-		rt.expectRandomness = rt.expectRandomness[1:]
+		rt.expectRandomnessBeacon = rt.expectRandomnessBeacon[1:]
+	}()
+	return exp.out
+}
```
nicola, line 277: hey @nikkolasg, can you verify that this is correct according to the spec: https://github.com/filecoin-project/specs/pull/967 ?

support/mock/mockrt.go
```diff
@@ -263,14 +264,37 @@ func (rt *Runtime) GetRandomness(tag crypto.DomainSeparationTag, epoch abi.Chain
...
+	if tag != exp.tag || epoch != exp.epoch || !bytes.Equal(entropy, exp.entropy) {
+		rt.failTest("unexpected get randomness\n"+
+			"         tag: %d, epoch: %d, entropy: %v\n"+
+			"expected tag: %d, epoch: %d, entropy: %v", tag, epoch, entropy, exp.tag, exp.epoch, exp.entropy)
+	}
+	defer func() {
+		rt.expectRandomnessTickets = rt.expectRandomnessTickets[1:]
 	}()
 	return exp.out
 }
```
nicola, line 300: hey @nikkolasg, can you verify that this is correct according to the spec: https://github.com/filecoin-project/specs/pull/967 ?

PR #680, opened 2020-07-15
Commit windowed post submissions to a particular chain

whyrusleeping:
merge after #679 

This prevents posts from being replayed on distant forks and helps to protect filecoin against long range attacks.

cc @magik6k as this implies miner changes

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -243,6 +243,10 @@ type SubmitWindowedPoStParams struct {
 	// Array of proofs, one per distinct registered proof type present in the sectors being proven.
 	// In the usual case of a single proof type, this array will always have a single element (independent of number of partitions).
 	Proofs []abi.PoStProof
+	// The epoch at which these proofs is being committed to a particular chain.
+	ChainCommitEpoch abi.ChainEpoch
+	// A signature from the miner worker over the randomness at the chain commit epoch.
+	ChainCommitSig crypto.Signature
 }
 
 // Invoked by miner's worker address to submit their fallback post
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -234,6 +242,15 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 	var recoveredSectors []*SectorOnChainInfo
 	penalty := abi.NewTokenAmount(0)
 
+	if rt.CurrEpoch()-params.ChainCommitEpoch > MaxPoStChainCommitAge {
+		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commit too far in the past")
+	}
```
anorth, line 247: Constrain `params.ChainCommitEpoch` to be in the past

actors/builtin/miner/policy.go
```diff
@@ -99,6 +99,10 @@ const WorkerKeyChangeDelay = ChainFinality
 // and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
 const MaxSectorExpirationExtension = builtin.EpochsInYear
 
+// MaxPoStChainCommitAge is the maximum distance back that a PoSt can commit to
+// a chain before being invalid
+const MaxPoStChainCommitAge = ChainFinality
```
nicola, line 104: it is better if this is smaller than finality - we can give you a better number for this.

whyrusleeping:
@nicola we don't need to actually have the extra signature here

actors/builtin/miner/miner_actor.go
```diff
@@ -234,6 +242,18 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+	if params.ChainCommitEpoch > rt.CurrEpoch() {
```
anorth, line 248: This should really be >= since I don't think randomness for the current epoch is defined until the block header's for the blocks containing this messages are accumulated into a tipset and form part of chain history.

The docs for `GetRandomness` could be more clear about that though.

nicola: any specific doc you are referring to?
if what's in here is wrong we should fix it! https://github.com/filecoin-project/specs/pull/967

nicola:
Are we saying that this signature is not needed? what randomness are you planning on using in the signature (I guess/hope the VRF, with what lookback)?

actors/builtin/miner/policy.go
nicola: We should set this to 5.

actors/builtin/miner/miner_actor.go
```diff
@@ -254,6 +258,16 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 	if params.Deadline >= WPoStPeriodDeadlines {
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d of %d", params.Deadline, WPoStPeriodDeadlines)
 	}
+	if params.ChainCommitEpoch >= currEpoch {
+		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commitment %d must be in the past", params.ChainCommitEpoch)
+	}
+	if params.ChainCommitEpoch < currEpoch - MaxPoStChainCommitAge {
+		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commitment %d too far in the past, must exceed %d", params.ChainCommitEpoch, currEpoch - MaxPoStChainCommitAge)
+	}
+	commRand := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_PoStChainCommit, params.ChainCommitEpoch, nil)
```
anorth, line 267: @nicola the first version of this PR took randomness from the beacon chain. I have changed it to the VRF ticket chain since, as I understand it, the beacon (DRAND) value wouldn't tied the proof to any particular blockchain fork, since they'd all share the same values.

Please confirm whether this change is correct.

nicola: confirm.

anorth:
I merged this by accident into the parent branch.

anorth:
I will post a new version of this based on master.

PR #684, opened 2020-07-16
Market: More tests and fixes

aarshkshah1992:
* Pending deal should be deleted on timeout.
* Test that Cron ticks for the same epoch should not change anything
* Test things work out well even if deal if processed twice in the same`CronTick` call.


actors/builtin/market/market_actor.go
```diff
@@ -462,6 +462,12 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 				state, found, err := msm.dealStates.Get(dealID)
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get deal state")
 
+				// if this is the first cron tick for the deal, it should be in the pending state.
+				if state.LastUpdatedEpoch == epochUndefined {
+					pdErr := msm.pendingDeals.Delete(adt.CidKey(dcid))
+					builtin.RequireNoErr(rt, pdErr, exitcode.ErrIllegalState, "failed to delete pending proposal")
+				}
+
```
wadeAlexC, line 470: I don't think this works. If this is the first cron tick for the deal, `state.LastUpdatedEpoch` will be 0 because it isn't set to `epochUndefined` until activation.

This should probably be included in the next block, `if !found { ... }`

actors/builtin/market/market_test.go
```diff
@@ -1302,7 +1302,34 @@ func TestCronTickTimedoutDeals(t *testing.T) {
...
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			rt.Call(actor.PublishStorageDeals, params)
+		})
+		rt.Verify()
+
+		// do a cron tick for it -> should time out and get slashed
+		rt.SetEpoch(startEpoch)
+		rt.ExpectSend(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, d.ProviderCollateral, nil, exitcode.Ok)
+		actor.cronTick(rt)
+		actor.assertDealDeleted(rt, dealId, d)
```
wadeAlexC, line 1329: The problem is that this method doesn't check `st.PendingProposals`, so this test passed even though it shouldn't have. (I think!)

actors/builtin/market/market_test.go
```diff
@@ -1342,9 +1369,12 @@ func TestCronTickTimedoutDeals(t *testing.T) {
...
 
-		actor.assertDealDeleted(rt, dealIds[0])
-		actor.assertDealDeleted(rt, dealIds[1])
-		actor.assertDealDeleted(rt, dealIds[2])
+		// a second cron tick for the same epoch should not change anything
+		actor.cronTickNoChangeBalances(rt, client, provider)
+
+		actor.assertDealDeleted(rt, dealIds[0], &deal1)
+		actor.assertDealDeleted(rt, dealIds[1], &deal2)
+		actor.assertDealDeleted(rt, dealIds[2], &deal3)
```
wadeAlexC, line 1377: I think this test checks whether 2 cron ticks in a row successfully deletes deals.

I don't think this adequately tests whether a secondary CronTick is idempotent, though. Is there a way to check the root cid of the actor's state? It should be unchanged, right?

actors/builtin/market/market_test.go
```diff
@@ -1359,6 +1389,31 @@ func TestCronTickDealExpiry(t *testing.T) {
 	endEpoch := abi.ChainEpoch(300)
 	sectorExpiry := abi.ChainEpoch(400)
 
+	t.Run("deal expiry -> deal is correctly processed twice in the same crontick", func(t *testing.T) {
+		end := abi.ChainEpoch(151) // 50 + 101
```
wadeAlexC, line 1393: Should be `startEpoch + 101`

actors/builtin/market/market_test.go
```diff
@@ -1359,6 +1389,31 @@ func TestCronTickDealExpiry(t *testing.T) {
...
+		rt.SetEpoch(current)
+		pay, slashed := actor.cronTickAndAssertBalances(rt, client, provider, current, dealId)
+		require.EqualValues(t, big.Zero(), pay)
+		require.EqualValues(t, big.Zero(), slashed)
+
+		// move the epoch to 155(anything greater than 150), so deal is first processed at 150 & then at 151 which is it's end epoch
+		// total payment = (end - start) = 151 - 50 = 101
+		current = 155
+		rt.SetEpoch(current)
+		pay, slashed = actor.cronTickAndAssertBalances(rt, client, provider, current, dealId)
```
wadeAlexC, line 1409: Is there a way to ensure this processed multiple epochs? (or maybe this exists in some other test?)

actors/builtin/market/market_test.go
```diff
@@ -1359,6 +1389,31 @@ func TestCronTickDealExpiry(t *testing.T) {
...
+		rt, actor := basicMarketSetup(t, owner, provider, worker, client)
+		dealId := actor.publishAndActivateDeal(rt, client, mAddrs, startEpoch, end, 0, sectorExpiry)
+		d := actor.getDealProposal(rt, dealId)
+
+		// move the current epoch to startEpoch so next cron epoch will be start + 100 = 150
+		current := startEpoch
+		rt.SetEpoch(current)
+		pay, slashed := actor.cronTickAndAssertBalances(rt, client, provider, current, dealId)
+		require.EqualValues(t, big.Zero(), pay)
+		require.EqualValues(t, big.Zero(), slashed)
```
wadeAlexC, line 1403: Should be a check after this cron tick to ensure the deal exists

actors/builtin/market/market_test.go
```diff
@@ -1383,9 +1438,12 @@ func TestCronTickDealExpiry(t *testing.T) {
 		current = 155
 		rt.SetEpoch(current)
 		pay, slashed = actor.cronTickAndAssertBalances(rt, client, provider, current, dealId)
-		require.EqualValues(t, pay, big.Mul(big.NewInt(100), d.StoragePricePerEpoch))
+		require.EqualValues(t, big.Mul(big.NewInt(100), d.StoragePricePerEpoch), pay)
 		require.EqualValues(t, big.Zero(), slashed)
 
+		// a second cron tick for the same epoch should not change anything
+		actor.cronTickNoChangeBalances(rt, client, provider)
+
```
wadeAlexC, line 1446: There are no checks here to ensure that a second cron tick is idempotent

actors/builtin/market/market_test.go
wadeAlexC: Maybe this should be added to `cronTickAndAssertBalances` - assert that multiple epochs were successfully processed.

actors/builtin/market/market_test.go
```diff
@@ -1619,13 +1706,19 @@ func TestCronTickDealSlashing(t *testing.T) {
 		rt.SetEpoch(current)
 		actor.cronTickNoChangeBalances(rt, client, provider)
 
+		// a second cron tick for the same epoch should not change anything
+		actor.cronTickNoChangeBalances(rt, client, provider)
```
wadeAlexC, line 1710: There are no checks here to ensure that a second cron tick is idempotent

actors/builtin/market/market_test.go
```diff
@@ -1619,13 +1706,19 @@ func TestCronTickDealSlashing(t *testing.T) {
...
+
 		//  Setting the current epoch to 155 will make another payment (5 + 100 epochs)
 		current = 155
 		rt.SetEpoch(current)
 		pay, slashed = actor.cronTickAndAssertBalances(rt, client, provider, current, dealId)
 		require.EqualValues(t, pay, big.Mul(big.NewInt(100), d.StoragePricePerEpoch))
 		require.EqualValues(t, big.Zero(), slashed)
 
+		// a second cron tick for the same epoch should not change anything
+		actor.cronTickNoChangeBalances(rt, client, provider)
```
wadeAlexC, line 1720: There are no checks here to ensure that a second cron tick is idempotent

actors/builtin/market/market_actor.go
aarshkshah1992: @wadeAlexC I've included a fix to remove this deal from pending even if it's timed out.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: @wadeAlexC I am not sure inspecting internal states like that is a good idea for tests. We have tests that ensure that deals in different epochs are process in a `CronTick`. Coupled with this,  it is proof enough.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: @wadeAlexC I've added a test to check that the `DealsByEpoch` cid does not change.

actors/builtin/market/market_test.go
aarshkshah1992: @wadeAlexC I've added a test to check that the DealsByEpoch cid does not change.

actors/builtin/market/market_test.go
aarshkshah1992: @wadeAlexC I've added a test to check that the DealsByEpoch cid does not change.

actors/builtin/market/market_test.go
aarshkshah1992: @wadeAlexC I've added a test to check that the DealsByEpoch cid does not change.

actors/builtin/market/market_test.go
aarshkshah1992: @wadeAlexC The test was passing because `msm.dealStates.Get(dealID)` in `CronTick` returns a state with `LastUpdatedEpoch = epochUndefined` if a state does not exist i.e. the deal has timed out.

However, for readability, I've fixed this by deleting pending timed-out deals as well.

wadeAlexC: I see, didn't notice this behavior in the `Get` method. Thanks for pointing this out!

wadeAlexC:
Looks great! Thanks @aarshkshah1992 !

PR #687, opened 2020-07-16
Optimize sector removal from expiration queue

Stebalien:
This version:

1. Removes on-time sectors first. These sectors can be anywhere in the queue.
2. Removes early sectors second. These sectors can only appear at the beginning (first 14 days/entries) in the queue.

Previously, when removing a bunch of on-time sectors, this function would have:

* State access linear in the length of the queue.
* A Runtime of (length of queue) * (number of sectors to remove).

### DIFF
--- a/actors/builtin/miner/expiration_queue.go
+++ b/actors/builtin/miner/expiration_queue.go
@@ -368,8 +368,35 @@ func (q ExpirationQueue) RemoveSectors(sectors []*SectorOnChainInfo, faults *abi
 		return nil, NewPowerPairZero(), xerrors.Errorf("failed to expand recoveries: %w", err)
 	}
 
+	// results
 	removed := NewExpirationSetEmpty()
 	recoveringPower := NewPowerPairZero()
+
+	// Process non-faulty sectors. These sectors can appear anywhere in the
+	// queue, but we know where to find them.
+	var (
+		activeSectors []*SectorOnChainInfo
+		faultySectors []*SectorOnChainInfo
+	)
+	for _, sector := range sectors {
+		if _, found := faultsMap[uint64(sector.SectorNumber)]; found {
+			faultySectors = append(faultySectors, sector)
+			continue
+		}
+		activeSectors = append(activeSectors, sector)
+		delete(remaining, sector.SectorNumber)
+	}
+
+	// Remove non-faulty sectors.
+	removed.OnTimeSectors, removed.ActivePower, removed.OnTimePledge, err = q.removeActiveSectors(activeSectors, ssize)
+	if err != nil {
+		return nil, NewPowerPairZero(), xerrors.Errorf("failed to remove on-time recoveries: %w", err)
+	}
+
+	// Process faulty sectors (on time and not). These sectors can only
+	// appear within the first 14 days (fault max age). Given that this
+	// queue is quantized, we should be able to stop traversing the queue
+	// after 14 entries.
 	if err = q.traverseMutate(func(epoch abi.ChainEpoch, es *ExpirationSet) (changed, keepGoing bool, err error) {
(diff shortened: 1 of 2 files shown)


Stebalien:
Fixes https://github.com/filecoin-project/specs-actors/pull/648#discussion_r455477035

actors/builtin/miner/expiration_queue.go
```diff
@@ -380,10 +403,13 @@ func (q ExpirationQueue) RemoveSectors(sectors []*SectorOnChainInfo, faults *abi
...
-		// clear that would be much faster (O(max(N, M)) vs O(N+M)).
-		// The length of sectors has a maximum of one partition size.
-		for _, sector := range sectors {
+		// NOTE: this iterates over the remaining _map_ to avoid
+		// revisiting already visited sectors. This is _ok_ because all
+		// the state updates we're doing here are idempotent.
+		//
+		// DO NOT update AMTs, HAMTs, etc. from within this loop. Only
+		// bitfields, big.Ints, etc.
+		for _, sector := range remaining { //nolint:nomaprange
```
anorth, line 412: We could alternatively accumulate a `faultySectors` slice exactly where the `activeSectors` slice is accumulated, and then traverse safely without this warning.

Stebalien: Welllll... I was trying to save a bit of time (avoid traversing sectors we've already found). But honestly, it's probably faster to create a slice and traverse it (because iterating over a slice is faster).

actors/builtin/miner/expiration_queue_test.go
```diff
@@ -597,8 +598,8 @@ func TestExpirationQueue(t *testing.T) {
 		assertBitfieldEquals(t, removed.OnTimeSectors, 1, 4, 6)
 		assertBitfieldEquals(t, removed.EarlySectors, 5)
 		assert.Equal(t, abi.NewTokenAmount(1000+1003+1005), removed.OnTimePledge) // only on-time sectors
-		assert.True(t, removed.ActivePower.Equals(PowerForSectors(sectorSize, []*SectorOnChainInfo{sectors[0], sectors[5]})))
-		assert.True(t, removed.FaultyPower.Equals(PowerForSectors(sectorSize, sectors[3:5])))
+		assert.True(t, removed.ActivePower.Equals(PowerForSectors(sectorSize, []*SectorOnChainInfo{sectors[0]})))
+		assert.True(t, removed.FaultyPower.Equals(PowerForSectors(sectorSize, sectors[3:6])))
```
Stebalien, line 602: @acruikshank did I do this right?

acruikshank: What you did makes sense to me in that a recovered sector is still a fault. I think it's right that recovered sector's power is still faulty. In that case, your changes are strictly better than what came before. 

It might be worth adding an assertion that recovering needs to be a subset of faults or at least documenting that expectation. I'm probably going to be back in here, so I can do it.

Stebalien: Makes sense. It shouldn't be too expensive and it's an invariant we want to maintain.

actors/builtin/miner/expiration_queue.go
```diff
@@ -368,8 +368,35 @@ func (q ExpirationQueue) RemoveSectors(sectors []*SectorOnChainInfo, faults *abi
 		return nil, NewPowerPairZero(), xerrors.Errorf("failed to expand recoveries: %w", err)
 	}
 
+	// results
 	removed := NewExpirationSetEmpty()
 	recoveringPower := NewPowerPairZero()
+
+	// Process non-faulty sectors. These sectors can appear anywhere in the
+	// queue, but we know where to find them.
```
anorth, line 376: This comment is a bit confusing now, given that active sectors are processed first.

PR #689, opened 2020-07-16
Feat/initial pledge refine

### LINKED ISSUE #495: Memoize total power and pledge for epoch for use in pledge/penalty calculations
Fee and pledge calculations are made in terms of the current epoch's notional block reward, as well as the total current power and (potentially) pledge. If these change during the course of a tipset's evaluation, then the order of messages which might alter these values (e.g. committing new sectors) would the calculations for subsequent sectors, which introduces message re-ordering shenanigans.

The notional block reward is already fixed for the epoch by a call from the cron actor at the end of the previous epoch. Fix an epoch's total power and pledge for use in calculations in a similar way (while also maintaining the incremental calculation).

### LINKED ISSUE #688: Latest CryptoEcon InitialPledge
Latest initial pledge calculation has changed to include an additionalIP term 
```
let lock_target = percent_lock_target * fil_circulating_supply
let storage_to_cover = max(network_qa_power, target_power)
let pledge_share = sector_qap / storage_to_cover
let additional_ip = lock_target * pledge_share
```

`IP = 20BR + additional_ip`

Additionally BR 
1. should add new sector qa power in denominator along with network qa power
2. should use consistent network qa power for entire tipset (#495)

ZenGround0:
Closes #688 
Closes #495 

New pledge calculation and stable power and pledge values per epoch

FYI @zixuanzh

### DIFF
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -27,6 +27,9 @@ var UndeclaredFaultFactorDenom = big.NewInt(1)
 // It is the expected reward this sector would pay out over a one day period.
 // BR(t) = CurrEpochReward(t) * SectorQualityAdjustedPower * EpochsInDay / TotalNetworkQualityAdjustedPower(t)
 func ExpectedDayRewardForPower(epochTargetReward abi.TokenAmount, networkQAPower abi.StoragePower, qaSectorPower abi.StoragePower) abi.TokenAmount {
+	if networkQAPower.IsZero() {
+		return epochTargetReward
+	}
 	expectedRewardForProvingPeriod := big.Mul(big.NewInt(builtin.EpochsInDay), epochTargetReward)
 	return big.Div(big.Mul(qaSectorPower, expectedRewardForProvingPeriod), networkQAPower)
 }
@@ -70,18 +73,16 @@ func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.Ch
 // newly-committed power, holding the per-epoch block reward constant (though in reality it will change over time).
 // The network total pledge and circulating supply parameters are currently unused, but may be included in a
 // future calculation.
-func InitialPledgeForPower(qaPower abi.StoragePower, networkQAPower, targetPower abi.StoragePower, networkTotalPledge abi.TokenAmount, epochTargetReward abi.TokenAmount, networkCirculatingSupply abi.TokenAmount) abi.TokenAmount {
+func InitialPledgeForPower(qaPower abi.StoragePower, networkQAPower, baselinePower abi.StoragePower, networkTotalPledge abi.TokenAmount, epochTargetReward abi.TokenAmount, networkCirculatingSupply abi.TokenAmount) abi.TokenAmount {
 	// Details here are still subject to change.
 	// PARAM_FINISH
 	// https://github.com/filecoin-project/specs-actors/issues/468
 	_ = networkTotalPledge // TODO: ce use this
 
-	if networkQAPower.IsZero() {
-		return epochTargetReward
-	}
 	ipBase := big.Mul(InitialPledgeFactor, ExpectedDayRewardForPower(epochTargetReward, networkQAPower, qaPower))
 	additionalIPNum := big.Mul(big.Mul(LockTargetFactorNum, networkCirculatingSupply), qaPower)
-	storageToCover := big.Max(networkQAPower, targetPower)
+	storageToCover := big.Max(networkQAPower, baselinePower)
+	storageToCover = big.Max(qaPower, storageToCover) // handle case where total and baseline are 0
 	additionalIPDenom := big.Mul(storageToCover, LockTargetFactorDenom)
 	additionalIP := big.Div(additionalIPNum, additionalIPDenom)
 
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -64,7 +64,7 @@ func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.Ch
 // newly-committed power, holding the per-epoch block reward constant (though in reality it will change over time).
 // The network total pledge and circulating supply parameters are currently unused, but may be included in a
 // future calculation.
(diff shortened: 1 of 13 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -215,11 +215,20 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	_, code := rt.Send(
 		builtin.RewardActorAddr,
 		builtin.MethodsReward.UpdateNetworkKPI,
-		&st.TotalRawBytePower,
+		&st.ThisEpochRawBytePower,
```
ZenGround0, line 218: Glad to see this subtle bug go -- power value unaware of small network rules

actors/builtin/power/power_actor.go
```diff
@@ -215,11 +215,20 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
 		abi.NewTokenAmount(0),
 	)
 	builtin.RequireSuccess(rt, code, "failed to update network KPI with Reward Actor")
 
+	// update next epoch's power and pledge values
```
ZenGround0, line 223: There is a question of whether we should be doing this before we send raw byte power to reward. This seems right but I wouldn't be too surprised if I was mistaken here.

actors/builtin/reward/reward_actor.go
```diff
@@ -112,6 +113,17 @@ func (a Actor) ThisEpochReward(rt vmr.Runtime, _ *adt.EmptyValue) *abi.TokenAmou
 	return &st.ThisEpochReward
 }
 
+// The baseline power the network is targeting this epoch, updated at the end of an
+// epoch through cron tick.  In the case previous epochs were null blocks this is
+// the value as calculated at the last non-null epoch
+func (a Actor) ThisEpochBaselinePower(rt vmr.Runtime, _ *adt.EmptyValue) *abi.StoragePower {
```
ZenGround0, line 119: Maybe we want to just group this return in with `ThisEpochReward`'s return and we could rename it to something more general like `ThisEpochStats`

actors/builtin/power/power_actor.go
anorth: I think we want it before the send. At the end of an epoch, we want to freeze the current power values, and then use them for computing the next epoch's block reward. Other calculations in terms of total power will then use the matching values for the block reward current during that epoch.

Describe this in a comment so that the ordering is transparent to a new reader.

actors/builtin/miner/monies.go
```diff
@@ -64,15 +70,20 @@ func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.Ch
...
 	// Details here are still subject to change.
 	// PARAM_FINISH
 	// https://github.com/filecoin-project/specs-actors/issues/468
-	_ = networkCirculatingSupply // TODO: ce use this
-	_ = networkTotalPledge       // TODO: ce use this
+	_ = networkTotalPledge // TODO: ce use this
 
 	if networkQAPower.IsZero() {
 		return epochTargetReward
 	}
```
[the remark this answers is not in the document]
ZenGround0: > I think we should push this check into ExpectedDayRewardForPower, next to the division it is protecting. Then ipBase will more correctly become 20*BR instead of, currently, 1BR.

👍 

> Logically I think we want this to come out as PledgeShare=1, which would be achieved with storageToCover=sectorQAPower. This suggests a max(sectorQAPower, networkPower, baselinePower) for storageToCover.

👍 

> This begs the question of how do we actually want to compute pledge for power in the genesis state, which I think we can address head-on separately, and then update here accordingly for networkQAPower=0.

This sounds out of scope of this PR. Should I file an issue for this here?

actors/builtin/miner/monies.go
```diff
@@ -64,15 +73,18 @@ func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.Ch
...
-	_ = networkCirculatingSupply // TODO: ce use this
-	_ = networkTotalPledge       // TODO: ce use this
+	_ = networkTotalPledge // TODO: ce use this
 
-	if networkQAPower.IsZero() {
-		return epochTargetReward
-	}
-	return big.Mul(InitialPledgeFactor, ExpectedDayRewardForPower(epochTargetReward, networkQAPower, qaPower))
+	ipBase := big.Mul(InitialPledgeFactor, ExpectedDayRewardForPower(epochTargetReward, networkQAPower, qaPower))
+	additionalIPNum := big.Mul(big.Mul(LockTargetFactorNum, networkCirculatingSupply), qaPower)
```
acruikshank, line 83: nit: In order to keep this consistent with the way the math is described, it would help to define `lockTargetNum` as in the inner multiplication here. Then `additionalIPNum` could be defined in terms of it. There's probably no need to redefine `LockTargetFactorDenom`.

actors/builtin/miner/monies.go
```diff
@@ -64,15 +73,18 @@ func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.Ch
...
+	storageToCover := big.Max(networkQAPower, baselinePower)
```
acruikshank, line 84: How about defining `pledgeShareNum := qaPower` and renaming `storageToCover` to `pledgeShareDenom`? That would map this a lot closer to how cryptoecon defines it. It would look like this:

``` golang
lockTargetNum := big.Mul(LockTargetFactorNum, networkCirculatingSupply)
lockTargetDenom := LockTargetFactorDenom
pledgeShareNum := qaPower
pledgeShareDenom := big.Max(big.Max(qaPower, storageToCover), qaPower)
additionalIPNum := big.Mul(lockTargetNum, pledgeShareNum)
additionalIPDenom := big.Mul(lockTargetDenom, pledgeShareDenom)
additionalIP := big.Div(additionalIPNum, additionalIPDenom)
```

A little more verbose, but maps cleanly to the math and has the same numerical properties.

acruikshank: You could write a little helper struct to clean this up this rational math on big integers a bit, but if you do, you have to promise you'll abbreviate its name to `BigRat`.

ZenGround0: image

ZenGround0: I like the idea but will file an issue for doing using the big rat consistently

ZenGround0: Feel free to add more details on #700

PR #691, opened 2020-07-16
add tests for expiration set and queue 

acruikshank:
### Motivation

`ExpirationSet` and `ExpirationQueue` need testing.

### Issues found

1. `PopUntil` inaccurately returned `FaultyPower` as `ActivePower`
2. `RemoveSectors` double counted (or erroneously single counted from early sectors) `OnTimePledge` when removing it from an ExpirationSet

### DIFF
--- a/actors/builtin/miner/expiration_queue_test.go
+++ b/actors/builtin/miner/expiration_queue_test.go
@@ -320,6 +320,45 @@ func TestExpirationQueue(t *testing.T) {
 		assert.True(t, activePower.Equals(set.ActivePower))
 		assert.True(t, faultyPower.Equals(set.FaultyPower))
 	})
+
+	t.Run("reschedules sectors as faults", func(t *testing.T) {
+		queue := emptyExprirationQueue(t)
+		queue.AddActiveSectors(sectors, sectorSize)
+
+		_, err := queue.Root()
+		require.NoError(t, err)
+
+		queue.RescheduleAsFaults(abi.ChainEpoch(20), sectors[:3], sectorSize)
+
+		_, err = queue.Root()
+		require.NoError(t, err)
+
+		// expect 3 rescheduled sectors to be bundled into 1 group
+		assert.Equal(t, 4, int(queue.Length()))
+
+		// pop off sectors before new expiration and expect only the rescheduled group to remain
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/expiration_queue_test.go
```diff
@@ -61,9 +663,45 @@ func TestExpirationsEmpty(t *testing.T) {
 	require.Equal(t, expected, result)
 }
 
+func testSector(expiration, number, weight, vweight, pledge int64) *SectorOnChainInfo {
```
anorth, line 666: Nit: I this sector number as first arg would be slightly better (but your call to change or not)

actors/builtin/miner/expiration_queue_test.go
```diff
@@ -9,33 +15,629 @@ import (
...
+		assert.True(t, power.Equals(PowerForSectors(sectorSize, sectors)))
+		assert.Equal(t, abi.NewTokenAmount(6015), pledge)
+
+		// default test quantizing of 1 means every sector is in its own expriation set
+		assert.Equal(t, len(sectors), int(queue.Length()))
+
+		_, err = queue.Root()
+		require.NoError(t, err)
+
+		// pop off sectors up to and including epoch 8
```
anorth, line 196: 8 -> 7

actors/builtin/miner/expiration_queue_test.go
```diff
@@ -9,33 +15,629 @@ import (
...
+		require.NoError(t, err)
+
+		assertBitfieldEquals(t, set.OnTimeSectors, 5, 6)
+		assertBitfieldEmpty(t, set.EarlySectors)
+
+		// Pledge from sector 5 is restored
+		assert.Equal(t, big.NewInt(2009), set.OnTimePledge)
+
+		assert.True(t, set.ActivePower.Equals(PowerForSectors(sectorSize, sectors[4:])))
+		assert.True(t, set.FaultyPower.Equals(NewPowerPairZero()))
```
anorth, line 512: My mental summary of this state is that it should be just like it was before the fault. Could we make that comparison more directly?

anorth:
Merging this into the integration branch so we can bring it with us to master.

Issue #278, opened 2020-04-08
ADTs should support multiple operations between flushing


anorth:
The ADT wrappers around HAMT, AMT flush on every mutation. This is much more expensive than necessary for repeated mutations.

We should change their interfaces to support batch operations, either through explicit flushing or a batching construction.

anorth:
See also #539 about using a cheaper hash function for intermediate updates.

anorth:
If taking this on, please circulate a proposal and seek some ideas before turning to code.

aarshkshah1992:
@Stebalien Would the IPFS data store batching API be an inspiration here ?

Stebalien:
We already have buffers although I guess we could have a write buffer? (not sure).

The real issue here is that we're serializing/deserializing intermediate IPLD blocks when we don't need to.

aarshkshah1992:
@ZenGround0 @acruikshank @Stebalien I'd like to this up. Any concerns ?

aarshkshah1992:
Update:

Changes have been made to the Market, Multisig, VerifReg and Power actors for this.
Other actors save the Miner actor don't have this problem.

@anorth @ZenGround0 Should we consider closing this issue ?

anorth:
Yes. There are more gains to be had, but they generally involve diving into the HAMT/AMT internals, or just using different structures or access patterns at a higher level.

PR #693, opened 2020-07-17
Power Actor: Remove duplicate reads

aarshkshah1992:
For #278 .


actors/util/adt/permission.go
```diff
@@ -0,0 +1,13 @@
...
+// MutationPermission is the mutation permission on a state field
+type MutationPermission int
+
+const (
+	// Invalid means NO permission
+	InvalidPermission MutationPermission = iota
+	// ReadOnlyPermission allows reading but not mutating the field
+	ReadOnlyPermission
+	// WritePermission allows mutating the field
+	WritePermission
```
anorth, line 12: These duplicate the same thing in market actor state. I would understand either centralizing them, or uniformly un-exporting them, but not having some private and some re-usable versions.

aarshkshah1992:
Closed in favour of #745 .

Issue #672, opened 2020-07-14
Split MaxSealDuration into MaxPreCommitRandomnessLookback and MaxProveCommitDuration


anorth:
At present, MaxSealDuration is used to bound the duration between the first randomness sample and the eventual proof. Split this into separate bounds on the delay between challenge and precommit, and the delay between precommit and provecommit. Only the latter depends on the seal type.

- Call the variable used for the first check MaxSealRandomnessLookback. This aims at representing the total maximum time that we allow for a SealRandomness to be taken from the past and it will be based on how long it takes to seal.
- Call the variable used for the second check MaxProveCommitDuration which will represent the maximum allowed time for a miner to generate a SNARK. Use the precommit epoch as the left end of this duration.

Update the precommit cron check appropriately

Issue #470, opened 2020-06-11
Raise or remove SectorsMax


ZenGround0:
From builtin/miner/policy.go:

```golang
// TODO raise this number, carefully
const SectorsMax = 32 << 20 // PARAM_FINISH
```

Unclear what carefully means or whether this TODO entails beyond the param finish but adding for completeness while flushing out todos.

ZenGround0:
From @nicola this [should be a function of RegisteredProofType](https://github.com/filecoin-project/specs-actors/pull/695/files#r456401084)

anorth:
I disagree, the purpose of `SectorsMax` is a bound on maximum computation/storage of worst-case computations. It's actually almost 100% irrelevant since #648 - all the places its used could have a _much_ tighter bound. It has nothing to do with Window PoSt.

If there are supposed to be some other limits on sector counts implied by the proof type, that is a new and distinct request that should be motivated on its own terms. @nicola I will close this issue unless you describe a new requirement (but that requirement will probably cause a new constant to be needed).

nicola:
From my PR:

> This guarantees that actors operations per miner are bounded. Also, this reflects the limit of sectors that can be proven in a day via WindowPoSt.

Let's call the two requirements SectorsMaxActors (the current SectorsMax) and SectorMaxProof[ProofType]. The number of sector for a miner should be the minimum between the two.

The reason for this need is the following: NSE won't be able to support the current SectorsMax without having more than one SNARK proof per deadline.

Stebalien:
I'm pretty sure this is the wrong variable. I believe you're looking for "partitions per deadline max". That is, the maximum number of proofs a miner would need to submit in a single deadline. That will get rid of the direct dependency on the proof type as partition size already depends on the proof type. It will also be more accurate because partitions may not be evenly spread out across all deadlines.

nicola:
This is the issue that I see: on NSE a miner cannot have 32 << 20 sectors since the total number of sectors proven per deadline is smaller - in other words, you will run out of deadlines with those many sectors.

If there is a system in place which guarantees that there is another implicit limit of maximum number of sectors per miner (e.g. if all deadlines are full, new porep is rejected), I am ok with that. I just have not found such check yet!

Let me know if this clarifies or if we need to sync on this one.

Stebalien:
This limit (SectorsMax) is a systems limit (how many sectors the chain can handle). It has nothing to do with proofs.

[...]

PR #695, opened 2020-07-17
Specs Actors Params Audit 

### LINKED ISSUE #699: Remove ElectionLookback parameter
@nicola points out [this is unused in code and used badly in test](https://github.com/filecoin-project/specs-actors/pull/695/files#r456450524).

nicola:
We are auditing the parameters, making sure that they have the correct values and the correct description. We are writing in front on all of them the `Filecoin Parameter:` prefix so that they will be easy to search in the future.

Closes #699

### Cryptoecon
Moved here: https://github.com/filecoin-project/specs-actors/issues/793

### Missing parameters
Parameters that I am not seeing:
- [ ] MaxPoStChainReference
- [ ] MinSectorExpiration

### Spec Actors (todo)
- AddressedSectorsMax
  - [ ] Checked value
  - [x] Checked usage
- WPoStPeriodDeadlines ([notes](https://hackmd.io/478poL-pQdq_qRKFEHTLTw?view))
  - [x] Value is correct/Proposed change
  - [ ] Check usage (pending confirmation of usage in [handleProvingDeadline](https://github.com/filecoin-project/specs-actors/blob/param-audit/actors/builtin/miner/miner_actor.go#L1740) and https://github.com/filecoin-project/specs-actors/issues/792) and part of [NewDeadlineInfo](https://github.com/filecoin-project/specs-actors/blob/param-audit/actors/builtin/miner/deadlines.go#L95-L106)
- ChainFinality
  - [x] Value is correct/Proposed change
  - [ ] [Used to determine SealRandomness validity](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L2335) (this is pending this change: https://github.com/filecoin-project/specs-actors/issues/672)
  - [x] [To set WorkerKeyChangeDelay](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/policy.go#L99)
- [RegisteredSealProof.SectorMaximumLifetime()](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/abi/sector.go#L206-L211) (SDR)
  - [ ] Check value (current value of 5 years seem incorrect) (TODO: @nicola, @Irenegia, @lucaniz)
- MaxSealDuration
  - [ ] Checked value (TODO: @anorth will do the proposed change or @nicola will improve current value)
  - [x] Propose that this should be changed to https://github.com/filecoin-project/specs-actors/issues/672
- ConsensusMinerMinMiners
  - [ ] Checked value (TODO: @nicola pending https://github.com/filecoin-project/research-private/issues/297)
  - [ ] [CurrentTotalPower calculation](https://github.com/filecoin-project/specs-actors/blob/param-audit/actors/builtin/power/power_state.go#L268-L273)
  - [x] [MinerNominalPowerMeetsConsensusMinimum calculation](https://github.com/filecoin-project/specs-actors/blob/param-audit/actors/builtin/power/power_state.go#L85-L108)

### Spec Actors (done)
- InitialPledgeFactor
  - [x] Checked value
  - [x] Correctly used in [TerminationFee calculation](https://github.com/filecoin-project/specs-actors/blob/365408676dbb8442a5ce35a26a36460629a6163e/actors/builtin/miner/monies.go#L57-L69) (pending resolution of https://github.com/filecoin-project/specs-actors/issues/777, https://github.com/filecoin-project/specs-actors/issues/701)
- AddressedPartitionsMax
  - [x] Checked value
  - [x] Checked usage
- MinDealDuration
  - [x] Checked Value
  - [x] Checked usage in `validateDeal`
- MaxDealDuration
  - [x] Checked Value
  - [x] Checked usage in `validateDeal`
- DealUpdatesInterval
  - [x] Checked value (value does not have audit constraints)
  - [x] Checked usage in `updatePendingDealState`
- ConsensusMinerMinPower
  - [x] Checked value
  - [x] Pending clarification on unit
  - [x] Pending minimum miner size reduction
  - [x] [MinerNominalPowerMeetsConsensusMinimum calculation](https://github.com/filecoin-project/specs-actors/blob/param-audit/actors/builtin/power/power_state.go#L85-L108)
  - [x] [AddToClaim calculation](https://github.com/filecoin-project/specs-actors/blob/365408676dbb8442a5ce35a26a36460629a6163e/actors/builtin/power/power_state.go#L110-L152)
- FaultMaxAge
  - [x] Value is correct/Proposed change
  - [x] Used for setting an termination Cron for SkippedFaults (?) (see [here](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L297) and [here](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/partition_state.go#L134))
  - [x] Used for setting termination Cron for DeclaredFaults (?) (see [here](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L1101) and [here](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L1134))
  - [x] Used for setting termination Cron for DetectedFaults (?) (see [here](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L1611))
- EpochDurationSeconds:
  - [x] Value is correct/Proposed change
  - [x] [WindowPost calculation](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/policy.go#L16)
  - [x] [Checking that if in 1 hours there is a integer number of epochs](https://github.com/filecoin-project/specs-actors/blob/927d2adfe03fc01d18898e5d70cd5dd573ca58bb/actors/builtin/network.go#L25)
- ExpectedLeadersPerEpoch
  - [x] Value is correct/Proposed change
  - [x] ~Possibly change name to ExpectedQualityPerEpoch~
  - [x] [Reward Calculation](https://github.com/filecoin-project/specs-actors/blob/e5b62bf0f506ec6ea3899fa94dcd4dd701ad51e5/actors/builtin/reward/reward_actor.go#L74)
- WPoStProvingPeriod
  - [x] Value is correct/Proposed change
  - [x] [Calculating WPoStPeriodDeadlines](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/policy.go#L19)
  - [x] [Calculating PeriodEnd and NextPeriodStart](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/deadlines.go#L64-L70)
  - [x] [Resetting the next proving period](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L1705-L1710)
- WPoStChallengeWindow
  - [x] Value is correct/Proposed change
  - [x] ~Possibly rename into: WPoStDeadlineWindow~
  - [x] [Calculate WPoStPeriodDeadlines](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/policy.go#L16)
  - [x] [Calculate Open and Close of a deadline](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/deadlines.go#L84-L92)
- WPoStChallengeLookback:
  - [x] Value is correct/Proposed change (pending resolution of: https://github.com/filecoin-project/research-private/issues/279)
- SectorsMax
  - [x] Value is correct/Proposed change
  - [x] Clarification on active or total
  - [x] Change request on sector max depending on proof type
- NewSectorsPerPeriodMax
  - [x] [Not used, it should be removed](https://github.com/filecoin-project/specs-actors/blob/param-audit/actors/builtin/miner/policy.go#L57)
- PreCommitChallengeDelay
  - [x] Value is correct/Proposed change
  - [x] [Compute the delay in the epoch in which the interactive randomness is chosen](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L571)
- ElectionLookback
  - [x] [Parameter is never used and seems to be used incorrectly in a test](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_test.go#L2236-L2240)
- FaultDeclarationCutoff
  - [x] Value is correct/Proposed change
  - [x] [Used to calculate FaultCutoff deadline](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/deadlines.go#L92)
- WorkerKeyChangeDelay
  - [x] Checked value
  - [x] [ChangeWorkerAddress](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/miner_actor.go#L155-L175) 
- QualityBaseMultiplier
  - [x] Checked value
- DealWeightMultiplier
  - [x] Checked value
- VerifiedDealWeightMultiplier
  - [x] Checked value
- DealLimitDenominator
  - [x] Checked value (clarification requested)(possible update tracked here: https://github.com/filecoin-project/research-private/issues/306)
  - [x] [DealPerSectorLimit](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/policy.go#L153-L156) function
- consensusFaultMaxReporterShare
  - [x] Checked value
  - [x] Correct usage in RewardForConsensusSlashReport
- MaxMinerProveCommitsPerEpoch
  - [x] Checked value
  - [x] [SubmitBulkPorep verify calculation](https://github.com/filecoin-project/specs-actors/blob/365408676dbb8442a5ce35a26a36460629a6163e/actors/builtin/power/power_actor.go#L279-L318)
  - [x] Correcly used in [InitialPledgeForPower](https://github.com/filecoin-project/specs-actors/blob/365408676dbb8442a5ce35a26a36460629a6163e/actors/builtin/miner/monies.go#L75-L88) (TODO: @irenegia, @lucaniz)
- SectorQualityPrecision
  - [x] Checked value
  - [x] [Used for scaling up quality](https://github.com/filecoin-project/specs-actors/blob/60a2ae96d2e656e2c4536a8f1ee905834e758912/actors/builtin/miner/policy.go#L131)
- DeclaredFaultFactorNum
  - [x] Checked value
  - [x] Usage
- DeclaredFaultFactorDenom
  - [x] Checked value
  - [x] Usage
- UndeclaredFaultFactorNum
  - [x] Checked value
  - [x] Usage
- UndeclaredFaultFactorDenom
  - [x] Checked value
  - [x] Usage

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -183,6 +183,13 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 	return info.WindowPoStProof, nil
 }
 
+
+// PARAM_SPEC
+// The maximum duration (activation to expiration) of a sector sealed with SDR.
+// Motivation: This gurantees that SDR is secure in the cost model for WindowPoSt and in the time model for WinningPoSt
+// This is based on estimation of hardware latency improvement and hardware and software cost reduction.
+const SectorMaximumLifetimeSDR = ChainEpoch(1_262_277 * 5)
+
 // SectorMaximumLifetime is the maximum duration a sector sealed with this proof may exist between activation and expiration
 func (p RegisteredSealProof) SectorMaximumLifetime() (ChainEpoch, error) {
 	info, ok := SealProofInfos[p]
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -183,10 +183,8 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 	return info.WindowPoStProof, nil
 }
 
-
-// PARAM_SPEC
-// The maximum duration (activation to expiration) of a sector sealed with SDR.
-// Motivation: This gurantees that SDR is secure in the cost model for WindowPoSt and in the time model for WinningPoSt
+// The maximum duration a sector sealed with this proof may exist between activation and expiration.
+// This ensures that SDR is secure in the cost model for Window PoSt and in the time model for Winning PoSt
 // This is based on estimation of hardware latency improvement and hardware and software cost reduction.
 const SectorMaximumLifetimeSDR = ChainEpoch(1_262_277 * 5)
 
@@ -238,6 +236,8 @@ func (p RegisteredPoStProof) WindowPoStPartitionSectors() (uint64, error) {
 	}
 	return sp.WindowPoStPartitionSectors()
 }
+
+
 
 ///
 /// Sealing
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -21,9 +21,9 @@ var InitialPledgeProjectionPeriod = abi.ChainEpoch(InitialPledgeFactor) * builti
 
 // Multiplier of share of circulating money supply for consensus pledge required to commit a sector.
 // This pledge is lost if a sector is terminated before its full committed lifetime.
-var InitialPledgeLockTarget = BigFrac{
-	numerator:   big.NewInt(3), // PARAM_SPEC PARAM_FINISH
-	denominator: big.NewInt(10),
+var InitialPledgeLockTarget = builtin.BigFrac{
+	Numerator:   big.NewInt(3), // PARAM_SPEC PARAM_FINISH
+	Denominator: big.NewInt(10),
 }
 
 // Projection period of expected daily sector block reward penalised when a fault is declared "on time".
@@ -35,23 +35,23 @@ var DeclaredFaultFactorNum = 214 // PARAM_SPEC
 var DeclaredFaultFactorDenom = 100
 var DeclaredFaultProjectionPeriod = abi.ChainEpoch((builtin.EpochsInDay * DeclaredFaultFactorNum) / DeclaredFaultFactorDenom)
 
-// PARAM_SPEC
-// Amount of fee for faults that have not been declared on time.
-// Motivation: This fee is higher than FF for two reasons:
-// (1) it guarantees that a miner is incentivized to declare a fault early
-// (2) A miner stores less than (1-spacegap) of a sector, does not not declare it as faulty and hopes to get challenged on the stored parts. SP guarantees that on expectation this strategy would not earn positive rewards.
+// Projection period of expected daily sector block reward penalised when a fault is not declared in advance.
+// This fee is higher than the declared fault fee for two reasons:
+// (1) it incentivizes a miner to declare a fault early;
+// (2) when a miner stores less than (1-spacegap) of a sector, does not not declare it as faulty,
+//     and hopes to be challenged on the stored parts, it means the miner would not be expected to earn positive rewards.
 // SP = BR(t, UndeclaredFaultProjectionPeriod)
-var UndeclaredFaultProjectionPeriod = abi.ChainEpoch(5) * builtin.EpochsInDay
+var UndeclaredFaultProjectionPeriod = abi.ChainEpoch(5) * builtin.EpochsInDay // PARAM_SPEC
 
-// Maximum number of days of BR a terminated sector can be penalized
+// Maximum number of lifetime days penalized when a sector is terminated.
 const TerminationLifetimeCap = abi.ChainEpoch(70)
 
-// Number of whole per-winner rewards covered by consensus fault penalty
+// Multiplier of whole per-winner rewards for a consensus fault penalty.
 const ConsensusFaultFactor = 5
 
-// This is the BR(t) value of the given sector for the current epoch.
-// It is the expected reward this sector would pay out over a t-day period.
-// BR(t) = CurrEpochReward(t) * SectorQualityAdjustedPower * EpochsInDay / TotalNetworkQualityAdjustedPower(t)
+// The projected block reward a sector would earn over some period.
+// Also known as "BR(t)".
+// BR(t) = ProjectedReward(t) * SectorQualityAdjustedPower / TotalNetworkQualityAdjustedPower
 func ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, qaSectorPower abi.StoragePower, projectionDuration abi.ChainEpoch) abi.TokenAmount {
 	networkQAPowerSmoothed := networkQAPowerEstimate.Estimate()
 	if networkQAPowerSmoothed.IsZero() {
@@ -63,15 +63,16 @@ func ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate *smoothing.Fi
 	return big.Max(br, big.Zero()) // negative BR is clamped at 0
 }
 
-// This is the FF(t) penalty for a sector expected to be in the fault state either because the fault was declared or because
-// it has been previously detected by the network.
-// FF(t) = DeclaredFaultFactor * BR(t)
+// The penalty for a sector declared faulty or continuing faulty for another proving period.
+// It is a projection of the expected reward earned by the sector.
+// Also known as "FF(t)"
 func PledgePenaltyForDeclaredFault(rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, qaSectorPower abi.StoragePower) abi.TokenAmount {
 	return ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate, qaSectorPower, DeclaredFaultProjectionPeriod)
 }
 
-// This is the SP(t) penalty for a newly faulty sector that has not been declared.
-// SP(t) = UndeclaredFaultFactor * BR(t)
+// The penalty for a newly faulty sector that was been declared in advance.
+// It is a projection of the expected reward earned by the sector.
+// Also known as "SP(t)"
 func PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, qaSectorPower abi.StoragePower) abi.TokenAmount {
 	return ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate, qaSectorPower, UndeclaredFaultProjectionPeriod)
 }
@@ -114,7 +115,7 @@ func PreCommitDepositForPower(rewardEstimate, networkQAPowerEstimate *smoothing.
 // network total and baseline power, per-epoch  reward, and circulating token supply.
 // The pledge comprises two parts:
 // - storage pledge, aka IP base: a multiple of the reward expected to be earned by newly-committed power
-// - pledge share, aka additional IP: a pro-rata fraction of the circulating money supply
+// - consensus pledge, aka additional IP: a pro-rata fraction of the circulating money supply
 //
 // IP = IPBase(t) + AdditionalIP(t)
 // IPBase(t) = BR(t, InitialPledgeProjectionPeriod)
@@ -124,8 +125,8 @@ func PreCommitDepositForPower(rewardEstimate, networkQAPowerEstimate *smoothing.
 func InitialPledgeForPower(qaPower, baselinePower abi.StoragePower, rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, circulatingSupply abi.TokenAmount) abi.TokenAmount {
 	ipBase := ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate, qaPower, InitialPledgeProjectionPeriod)
 
-	lockTargetNum := big.Mul(InitialPledgeLockTarget.numerator, circulatingSupply)
-	lockTargetDenom := InitialPledgeLockTarget.denominator
+	lockTargetNum := big.Mul(InitialPledgeLockTarget.Numerator, circulatingSupply)
+	lockTargetDenom := InitialPledgeLockTarget.Denominator
 	pledgeShareNum := qaPower
 	networkQAPower := networkQAPowerEstimate.Estimate()
 	pledgeShareDenom := big.Max(big.Max(networkQAPower, baselinePower), qaPower) // use qaPower in case others are 0
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -4,25 +4,42 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
 	"github.com/filecoin-project/specs-actors/actors/builtin"
-	exitcode "github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
 	"github.com/filecoin-project/specs-actors/actors/util/math"
 	"github.com/filecoin-project/specs-actors/actors/util/smoothing"
 )
 
-// PARAM_FINISH
-var PreCommitDepositFactor = 20
-var InitialPledgeFactor = 20
+// Projection period of expected sector block reward for deposit required to pre-commit a sector.
+// This deposit is lost if the pre-commitment is not timely followed up by a commitment proof.
+var PreCommitDepositFactor = 20 // PARAM_SPEC PARAM_FINISH
 var PreCommitDepositProjectionPeriod = abi.ChainEpoch(PreCommitDepositFactor) * builtin.EpochsInDay
-var InitialPledgeProjectionPeriod = abi.ChainEpoch(InitialPledgeFactor) * builtin.EpochsInDay
-var LockTargetFactorNum = big.NewInt(3)
-var LockTargetFactorDenom = big.NewInt(10)
 
+// Projection period of expected sector block rewards for storage pledge required to commit a sector.
+// This pledge is lost if a sector is terminated before its full committed lifetime.
+var InitialPledgeFactor = 20 // PARAM_SPEC PARAM_FINISH
+var InitialPledgeProjectionPeriod = abi.ChainEpoch(InitialPledgeFactor) * builtin.EpochsInDay
+
+// Multiplier of share of circulating money supply for consensus pledge required to commit a sector.
+// This pledge is lost if a sector is terminated before its full committed lifetime.
+var InitialPledgeLockTarget = BigFrac{
+	numerator:   big.NewInt(3), // PARAM_SPEC PARAM_FINISH
+	denominator: big.NewInt(10),
+}
+
+// Projection period of expected daily sector block reward penalised when a fault is declared "on time".
+// This guarantees that a miner pays back at least the expected block reward earned since the last successful PoSt.
+// The network conservatively assumes the sector was faulty since the last time it was proven.
+// This penalty is currently overly punitive for continued faults.
 // FF = BR(t, DeclaredFaultProjectionPeriod)
-// projection period of 2.14 days:  2880 * 2.14 = 6163.2.  Rounded to nearest epoch 6163
-var DeclaredFaultFactorNum = 214
+var DeclaredFaultFactorNum = 214 // PARAM_SPEC
 var DeclaredFaultFactorDenom = 100
 var DeclaredFaultProjectionPeriod = abi.ChainEpoch((builtin.EpochsInDay * DeclaredFaultFactorNum) / DeclaredFaultFactorDenom)
 
+// PARAM_SPEC
+// Amount of fee for faults that have not been declared on time.
+// Motivation: This fee is higher than FF for two reasons:
+// (1) it guarantees that a miner is incentivized to declare a fault early
+// (2) A miner stores less than (1-spacegap) of a sector, does not not declare it as faulty and hopes to get challenged on the stored parts. SP guarantees that on expectation this strategy would not earn positive rewards.
 // SP = BR(t, UndeclaredFaultProjectionPeriod)
 var UndeclaredFaultProjectionPeriod = abi.ChainEpoch(5) * builtin.EpochsInDay
 
@@ -107,8 +124,8 @@ func PreCommitDepositForPower(rewardEstimate, networkQAPowerEstimate *smoothing.
 func InitialPledgeForPower(qaPower, baselinePower abi.StoragePower, rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, circulatingSupply abi.TokenAmount) abi.TokenAmount {
 	ipBase := ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate, qaPower, InitialPledgeProjectionPeriod)
 
-	lockTargetNum := big.Mul(LockTargetFactorNum, circulatingSupply)
-	lockTargetDenom := LockTargetFactorDenom
+	lockTargetNum := big.Mul(InitialPledgeLockTarget.numerator, circulatingSupply)
+	lockTargetDenom := InitialPledgeLockTarget.denominator
 	pledgeShareNum := qaPower
 	networkQAPower := networkQAPowerEstimate.Estimate()
 	pledgeShareDenom := big.Max(big.Max(networkQAPower, baselinePower), qaPower) // use qaPower in case others are 0
--- a/actors/builtin/miner/monies.go
+++ b/actors/builtin/miner/monies.go
@@ -38,7 +38,7 @@ var DeclaredFaultProjectionPeriod = abi.ChainEpoch((builtin.EpochsInDay * Declar
 // Projection period of expected daily sector block reward penalised when a fault is not declared in advance.
 // This fee is higher than the declared fault fee for two reasons:
 // (1) it incentivizes a miner to declare a fault early;
-// (2) when a miner stores less than (1-spacegap) of a sector, does not not declare it as faulty,
+// (2) when a miner stores less than (1-spacegap) of a sector, does not declare it as faulty,
 //     and hopes to be challenged on the stored parts, it means the miner would not be expected to earn positive rewards.
 // SP = BR(t, UndeclaredFaultProjectionPeriod)
 var UndeclaredFaultProjectionPeriod = abi.ChainEpoch(5) * builtin.EpochsInDay // PARAM_SPEC
@@ -51,7 +51,9 @@ const ConsensusFaultFactor = 5
 
 // The projected block reward a sector would earn over some period.
 // Also known as "BR(t)".
-// BR(t) = ProjectedReward(t) * SectorQualityAdjustedPower / TotalNetworkQualityAdjustedPower
+// BR(t) = ProjectedRewardFraction(t) * SectorQualityAdjustedPower
+// ProjectedRewardFraction(t) is the sum of estimated reward over estimated total power
+// over all epochs in the projection period [t t+projectionDuration]
 func ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, qaSectorPower abi.StoragePower, projectionDuration abi.ChainEpoch) abi.TokenAmount {
 	networkQAPowerSmoothed := networkQAPowerEstimate.Estimate()
 	if networkQAPowerSmoothed.IsZero() {
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -245,9 +245,8 @@ var consensusFaultReporterShareGrowthRate = BigFrac{
 // PARAM_SPEC
 // Maximum share of slasher's reward
 var consensusFaultMaxReporterShare = BigFrac{
-	// PARAM_FINISH
 	numerator:   big.NewInt(1),
-	denominator: big.NewInt(4),
+	denominator: big.NewInt(20),
 }
 
 // Specification for a linear vesting schedule.
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -11,21 +11,20 @@ import (
 	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
 	. "github.com/filecoin-project/specs-actors/actors/util"
 )
-// PARAM_SPEC
+
 // The period over which a miner's active sectors are expected to be proven via WindowPoSt.
-// Motivation: This guarantees that (1) user data is proven once a day, (2) user data is stored for 24h by a rational miner (due to WindowPoSt cost assumption).
-var WPoStProvingPeriod = abi.ChainEpoch(builtin.EpochsInDay) // 24 hours
+// This guarantees that (1) user data is proven daily, (2) user data is stored for 24h by a rational miner
+// (due to Window PoSt cost assumption).
+var WPoStProvingPeriod = abi.ChainEpoch(builtin.EpochsInDay) // 24 hours PARAM_SPEC
 
-// PARAM_SPEC
-// The period between the opening and the closing of a WindowPoSt deadline in which the miner is expected to provide a WindowPoSt proof.
-// Motivation: This guarantees that a miner has enough time to propagate a WindowPoSt for the current deadline.
-// Usage: This is used to calculate the opening and close of a deadline window and to calculate the number of deadlines in a proving period.
-var WPoStChallengeWindow = abi.ChainEpoch(30 * 60 / builtin.EpochDurationSeconds) // 30 minutes (48 per day)
+// The period between the opening and the closing of a WindowPoSt deadline in which the miner is expected to
+// provide a Window PoSt proof.
+// This provides a miner enough time to compute and propagate a Window PoSt proof.
+var WPoStChallengeWindow = abi.ChainEpoch(30 * 60 / builtin.EpochDurationSeconds) // 30 minutes (48 per day) PARAM_SPEC
 
-// PARAM_SPEC
 // The number of non-overlapping PoSt deadlines in a proving period.
-// Motivation: This guarantees that a miner can spread their WindowPoSt duties across a proving period, instead of submitting a single large WindowPoSt submission.
-const WPoStPeriodDeadlines = uint64(48)
+// This spreads a miner's Window PoSt work across a proving period.
+const WPoStPeriodDeadlines = uint64(48) // PARAM_SPEC
 
 // MaxPartitionsPerDeadline is the maximum number of partitions that will be assigned to a deadline.
 // For a minimum storage of upto 1Eib, we need 300 partitions per deadline.
@@ -38,44 +37,35 @@ func init() {
 	if WPoStProvingPeriod%WPoStChallengeWindow != 0 {
 		panic(fmt.Sprintf("incompatible proving period %d and challenge window %d", WPoStProvingPeriod, WPoStChallengeWindow))
 	}
+	// Check that WPoStPeriodDeadlines is consistent with the proving period and challenge window.
 	if abi.ChainEpoch(WPoStPeriodDeadlines)*WPoStChallengeWindow != WPoStProvingPeriod {
 		panic(fmt.Sprintf("incompatible proving period %d and challenge window %d", WPoStProvingPeriod, WPoStChallengeWindow))
 	}
 }
 
-// PARAM_SPEC
-// The maximum number of sectors that a miner can have (and implicitly the maximum number of faults, recoveries, etc.)
-// Note that when sectors are terminated they count towards SectorsMax until cleanup.
-// Motivation: This guarantees that actors operations per miner are bounded.
-// TODO raise this number, carefully
-// https://github.com/filecoin-project/specs-actors/issues/470
-const SectorsMax = 32 << 20 // PARAM_FINISH
-
-// PARAM_SPEC
 // The maximum number of partitions that can be loaded in a single invocation.
 // This limits the number of simultaneous fault, recovery, or sector-extension declarations.
 // We set this to same as MaxPartitionsPerDeadline so we can process that many partitions every deadline.
 const AddressedPartitionsMax = MaxPartitionsPerDeadline
 
-// PARAM_SPEC
 // The maximum number of sector infos that can be loaded in a single invocation.
-const AddressedSectorsMax = 10_000
+// This limits the amount of state to be read in a single message execution.
+const AddressedSectorsMax = 10_000 // PARAM_SPEC
 
 // Libp2p peer info limits.
 const (
-	// MaxPeerIDLength is the maximum length allowed for any on-chain peer ID. Most
-	// Peer IDs should be less than 50 bytes.
-	MaxPeerIDLength = 128
+	// MaxPeerIDLength is the maximum length allowed for any on-chain peer ID.
+	// Most Peer IDs are expected to be less than 50 bytes.
+	MaxPeerIDLength = 128 // PARAM_SPEC
 
-	// MaxMultiaddrData is the maximum amount of data that can be stored in
-	// multiaddrs.
-	MaxMultiaddrData = 1024
+	// MaxMultiaddrData is the maximum amount of data that can be stored in multiaddrs.
+	MaxMultiaddrData = 1024 // PARAM_SPEC
 )
 
-// Maximum bytes in a single prove-commit proof.
+// Maximum size of a single prove-commit proof, in bytes.
 const MaxProveCommitSize = 1024
 
-// Maximum number of control addresses
+// Maximum number of control addresses a miner may register.
 const MaxControlAddresses = 10
 
 // The maximum number of partitions that may be required to be loaded in a single invocation,
@@ -84,14 +74,11 @@ func loadPartitionsSectorsMax(partitionSectorCount uint64) uint64 {
 	return min64(AddressedSectorsMax/partitionSectorCount, AddressedPartitionsMax)
 }
 
-// The maximum number of new sectors that may be staged by a miner during a single proving period.
-const NewSectorsPerPeriodMax = 128 << 10
-
-// PARAM_SPEC
 // Epochs after which chain state is final with overwhelming probability (hence the likelihood of two fork of this size is negligible)
-// Motivation: This is a conservative value that is chosen via simulations of all known attacks.
-const ChainFinality = abi.ChainEpoch(1400)
+// This is a conservative value that is chosen via simulations of all known attacks.
+const ChainFinality = abi.ChainEpoch(1400) // PARAM_SPEC
 
+// Prefix for sealed sector CIDs (CommR).
 var SealedCIDPrefix = cid.Prefix{
 	Version:  1,
 	Codec:    cid.FilCommitmentSealed,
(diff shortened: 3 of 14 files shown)


actors/builtin/miner/policy.go
```diff
@@ -28,8 +32,8 @@ func init() {
 	}
 }
 
-// The maximum number of sectors that a miner can have simultaneously active.
-// This also bounds the number of faults that can be declared, etc.
+// Filecoin Parameter: The maximum number of sectors that a miner can have (and implicitly the maximum number of faults, recoveries, etc.) 
+// Motivation: This guarantees that actors operations per miner are bounded. Also, this reflects the limit of sectors that can be proven in a day via WindowPoSt.
 // TODO raise this number, carefully
 // https://github.com/filecoin-project/specs-actors/issues/470
 const SectorsMax = 32 << 20 // PARAM_FINISH
```
nicola, line 39: Problems that we found:
- Previously this was referred to as "The maximum number of sectors that a miner can have simultaneously active.", we believe this is either incorrect or subject to complex edge cases (e.g. a miner that has 32<<20 sectors active and 1 sector faulty which then recovers the faulty sector, now has 32<<20+1 active sectors). Can @ZenGround0 or @anorth confirm this?
- `SectorsMax` is specific to a particular `RegisteredProof`. For example, NSE will have a much smaller `SectorsMax`. Why? One of the constrains that sets the `SectorsMax` is the number of sectors that can be proven in 24 hours via a WindowPost. If we have NSE, the number of sectors that can be proven in a WindowPoSt is reduced, hence the SectorsMax chosen for SDR will be much larger.

nicola, line 39: **Change request:** `SectorsMax` is specific to a particular `RegisteredProof`. For example, NSE will have a much smaller `SectorsMax`. Why? One of the constrains that sets the `SectorsMax` is the number of sectors that can be proven in 24 hours via a WindowPost. With NSE, the number of sectors that can be proven in a WindowPoSt is smaller than SDR, hence the SectorsMax chosen for SDR won't work for NSE WindowPoSt.

actors/builtin/miner/policy.go
```diff
@@ -71,17 +76,16 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
 }
 
-// Number of epochs between publishing the precommit and when the challenge for interactive PoRep is drawn
-// used to ensure it is not predictable by miner.
+// Filecoin Parameter: Number of epochs between publishing the precommit and when the challenge for interactive PoRep is drawn.
+// Motivation: This guarantees that (1) a miner cannot predict a future challenge (2) a miner cannot do a long fork in the past to insert a precommit after seeing a challenge.
 const PreCommitChallengeDelay = abi.ChainEpoch(150)
 
 // Lookback from the current epoch for state view for leader elections.
 const ElectionLookback = abi.ChainEpoch(1) // PARAM_FINISH
```
nicola, line 84: **Deletion request:** Parameter is never used and seems to be used incorrectly in [a test](Parameter is never used and seems to be used incorrectly in a test).

actors/builtin/miner/policy.go
```diff
@@ -52,8 +56,9 @@ func loadPartitionsSectorsMax(partitionSectorCount uint64) uint64 {
 // The maximum number of new sectors that may be staged by a miner during a single proving period.
 const NewSectorsPerPeriodMax = 128 << 10
```
nicola, line 57: **Deletion request:** Parameter is never used and should likely be removed.

actors/builtin/miner/policy.go
```diff
@@ -71,34 +76,36 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
 // it can be changed. It's the only safe value, though. We may implement a mitigation mechanism such as a second
 // key or allowing the owner account to submit PoSts while a key change is pending.
+// Motivation: This guarantees that a miner cannot choose a more favorable worker key that wins leader elections.
 const WorkerKeyChangeDelay = ChainFinality
 
-// Maximum number of epochs past the current epoch a sector may be set to expire.
+// Filecoin Parameter: Maximum number of epochs past the current epoch a sector may be set to expire.
 // The actual maximum extension will be the minimum of CurrEpoch + MaximumSectorExpirationExtension
 // and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
 const MaxSectorExpirationExtension = builtin.EpochsInYear
```
nicola, line 111: **Clarification requested:** Why is this set to one year? What is the motivation? It seems that the MaxSectorLifetime is respected in the code, can we have confirmation?

ZenGround0: No idea where it comes from.  I believe it is respected in the code when processing sector deadlines and expirations.  cc @acruikshank who knows this area better than me

actors/builtin/miner/policy.go
ZenGround0: Sounds right but you'll need @anorth or @Stebalien to confirm

ZenGround0: Now tracked on #470

actors/builtin/miner/policy.go
ZenGround0: @anorth or @Stebalien should confirm there is no intention to use a limit like this before deletion

actors/builtin/miner/policy.go
```diff
@@ -71,39 +76,41 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
 
-// Ratio of sector size to maximum deals per sector.
+// Filecoin Parameter: Ratio of sector size to maximum deals per sector.
 // The maximum number of deals is the sector size divided by this number (2^27)
 // which limits 32GiB sectors to 256 deals and 64GiB sectors to 512
 const DealLimitDenominator = 134217728
```
nicola, line 116: **Clarification requested:** Where does this come from? What is constraining this to be "256" deals for a 32GiB sector? This means an average of 128MiB per deal?

actors/builtin/miner/policy.go
ZenGround0: tracked in #699

actors/builtin/miner/policy.go
```diff
@@ -71,39 +76,41 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
+// Motivation: A miner can start the WindowPoSt computation without waiting for chain stability. This value cannot be too large since it could compromise the rationality of honest storage (due to WindowPoSt cost assumptions)
 const WPoStChallengeLookback = abi.ChainEpoch(20)
 
-// Minimum period before a deadline's challenge window opens that a fault must be declared for that deadline.
-// This lookback must not be less than WPoStChallengeLookback lest a malicious miner be able to selectively declare
-// faults after learning the challenge value.
+// Filecoin Parameter: Minimum period between fault declaration and the next deadline opening.
+// If the number of epochs between fault declaration and deadline's challenge window opening is lower than FaultDeclarationCutoff,
+// the fault declaration is considered invalid for that deadline.
+// Motivation: This guarantees that a miner is not likely to successfully fork the chain and declare a fault after seeing the challenges.
```
ZenGround0, line 94: These motivation snippets are a great idea 👏

actors/builtin/miner/policy.go
ZenGround0: This comes from @whyrusleeping, 256 deals for 32Gib, 512 deals for 64Gib. 
Resolution on relevant PR [here](https://github.com/filecoin-project/specs-actors/pull/632#pullrequestreview-446924587)

actors/builtin/miner/policy.go
```diff
@@ -52,8 +56,9 @@ func loadPartitionsSectorsMax(partitionSectorCount uint64) uint64 {
 // The maximum number of new sectors that may be staged by a miner during a single proving period.
 const NewSectorsPerPeriodMax = 128 << 10
 
-// Epochs after which chain state is final.
-const ChainFinality = abi.ChainEpoch(900)
+// Filecoin Parameter: Epochs after which chain state is final with overwhelming probability (hence the likelihood of two fork of this size is negligible)
+// Motivation: This is a conservative value that is chosen via simulations of all known attacks.
+const ChainFinality = abi.ChainEpoch(1400)
```
whyrusleeping, line 61: why is this getting bigger?

nicola:
Hey @anorth are you suggesting to keep `PARAM_FINISH` also after they are finalized? If so, I am happy to keep them in that way, if not, we need a new tag (e.g. SPEC_PARAMS).

actors/builtin/miner/policy.go
nicola: what does it mean for them to be "not yet compacted out"?
Does this imply that the maximum number of sectors can be higher than 32 << 20?

nicola: This is however a requirement from proof as we found in the Params Audit (added more motivation here and planning to add more in the PR): https://github.com/filecoin-project/specs-actors/issues/470#issuecomment-660933651

actors/builtin/miner/policy.go
nicola: Unfortunately it is not clear from the issue what is the exact motivation "After discovering discussion with @whyrusleeping elsewhere motivating this approach"

actors/builtin/miner/policy.go
```diff
@@ -71,47 +76,52 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
 
+// Filecoin Parameter: Quality multiplier for CC sectors
 var QualityBaseMultiplier = big.NewInt(10)
+// Filecoin Parameter: Quality multiplier for a standard deal in a sector
 var DealWeightMultiplier = big.NewInt(10)
+// Filecoin Parameter: Quality multiplier for verified deals
 var VerifiedDealWeightMultiplier = big.NewInt(100)
-
+// Filecoin Parameter: Precision for Quality Adjusted power
 const SectorQualityPrecision = 20
```
nicola, line 125: **Clarification Request:** Why is this set to 20?

actors/builtin/miner/policy.go
ZenGround0: > Does this imply that the maximum number of sectors can be higher than 32 << 20?

No I understand the above comment to mean that this max covers all sectors, active, faulty and terminated but not yet cleaned up

Stebalien: >  Does this imply that the maximum number of sectors can be higher than 32 << 20?

Lower.

When we terminate a sector, we just mark it as terminated and pay any fees. However, we don't actually cleanup the state (sector info) or remove it from the deadline/partition till the miner calls a "compact" method on the deadline/partition. When the miner calls "compact", re-create the deadline from all _live_ sectors in the partition, and remove any state related to old sectors.

I say "lower" because `32<<20` is the max number of sectors _including_ these terminated-but-not-yet-compacted sectors.

Stebalien: Continuing conversation on #470.

actors/builtin/miner/policy.go
```diff
@@ -28,8 +32,8 @@ func init() {
 	}
 }
 
-// The maximum number of sectors that a miner can have simultaneously active.
-// This also bounds the number of faults that can be declared, etc.
+// Filecoin Parameter: The maximum number of sectors that a miner can have (and implicitly the maximum number of faults, recoveries, etc.) 
+// Motivation: This guarantees that actors operations per miner are bounded. Also, this reflects the limit of sectors that can be proven in a day via WindowPoSt.
```
Stebalien, line 36: See https://github.com/filecoin-project/specs-actors/issues/470#issuecomment-661181105

actors/builtin/miner/monies.go
```diff
@@ -17,10 +18,18 @@ var LockTargetFactorNum = big.NewInt(3)
...
 var DeclaredFaultFactorNum = big.NewInt(214)
 var DeclaredFaultFactorDenom = big.NewInt(100)
 
 // SP = (UndeclaredFaultFactor / DeclaredFaultFactorDenom) * BR(t)
+// Filecoin parameter: Amount of fee for faults that have not been declared on time.
+// Motivation: This fee is higher than FF for two reasons:
+// (1) it guarantees that a miner is incentivized to declare a fault early
+// (2) A miner stores less than (1-spacegap) of a sector, does not not declare it as faulty and hopes to get challenged on the stored parts. SP guarantees that on expectation this strategy would not earn positive rewards.
 var UndeclaredFaultFactorNum = big.NewInt(5)
 var UndeclaredFaultFactorDenom = big.NewInt(1)
```
nicola, line 34: **Possible Change:** The fees of `FF` and `SP` are dependent on the soundness of the PoSt proof that we are using. This means that a future improvement for a PoSt for SDR (or a new PoSt for a new proof like NSE) could lead to a different soundness, which implicitly will lead to a different FF and SP.

We could have an overly punitive FF and SP and set the highest FF and SP as system wide parameters, however future optimization/proof changes could really minimize FF and SP.

The proposal here is to make FF and SP proof dependent - for example with NSE, SP could be drastically reduced (and made as close to FF) and also FF could be reduced.

anorth: This is a significant change and would need careful thought, especially with regard to potential changed proof algorithms. We probably don't have time for it.

anorth:
Yes, a tag like `SPEC_PARAMS` (or maybe `PARAM_SPEC` to ease grepping for both of them at once) should replace `PARAM_FINISH`.

actors/builtin/miner/policy.go
nicola: This is resolved, to me - feel free to re-open in case you are not happy with the wording I used.

nicola: I take this one back - this has been clarified. I will remove the second part of the sentence.

actors/builtin/miner/policy.go
```diff
@@ -171,17 +213,30 @@ type BigFrac struct {
...
 	denominator: big.NewInt(1000),
 }
+
+// PARAM_SPEC
+// Growth rate of a slasher's reward
 var consensusFaultReporterShareGrowthRate = BigFrac{
 	// PARAM_FINISH
 	numerator:   big.NewInt(101251),
 	denominator: big.NewInt(100000),
 }
```
nicola, line 230: hey @zixuanzh, @whyrusleeping, @anorth: are these numbers final? is there an analysis showing why we are using these numbers or we need to do it?

actors/builtin/miner/policy.go
```diff
@@ -171,17 +213,30 @@ type BigFrac struct {
...
 
+// PARAM_SPEC
+// Maximum share of slasher's reward
+var consensusFaultMaxReporterShare = BigFrac{
+	// PARAM_FINISH
+	numerator:   big.NewInt(1),
+	denominator: big.NewInt(2),
+}
```
nicola, line 238: **Change**: hey @anorth and @ZenGround0 we took out of the RewardForConsensusSlashReport these numbers such that they are explicitly parameters of the system

actors/builtin/miner/policy.go
```diff
@@ -81,47 +91,67 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
 const ElectionLookback = abi.ChainEpoch(1) // PARAM_FINISH
 
-// Lookback from the deadline's challenge window opening from which to sample chain randomness for the challenge seed.
+// PARAM_SPEC
+// Lookback from the deadline's challenge window opening from which to sample chain randomness for the WindowPoSt challenge seed.
 // This lookback exists so that deadline windows can be non-overlapping (which make the programming simpler)
-// but without making the miner wait for chain stability before being able to start on PoSt computation.
-// The challenge is available this many epochs before the window is actually open to receiving a PoSt.
+// Motivation: A miner can start the WindowPoSt computation without waiting for chain stability. This value cannot be too large since it could compromise the rationality of honest storage (due to WindowPoSt cost assumptions)
 const WPoStChallengeLookback = abi.ChainEpoch(20)
```
nicola, line 106: **Confirmation needed:** @irenegia to confirm that 20 + 60 epochs (assuming 30s block time) is still safe in the cost model for WindowPoSt (if I recall correctly we allocated 1h and 60+20 is about 40 mins)

actors/builtin/miner/policy.go
```diff
@@ -81,47 +91,67 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
 	abi.RegisteredSealProof_StackedDrg64GiBV1:  abi.ChainEpoch(10000),
 }
```
nicola, line 92: **Change requested**: There is an FIP where decided to split this into two variables: https://github.com/filecoin-project/specs-actors/issues/672

actors/builtin/miner/policy.go
nicola, line 238: **Clarification needed**: hey @zixuanzh, @whyrusleeping, @anorth: are these numbers final? is there an analysis showing why we are using these numbers or we need to do it?

actors/builtin/miner/policy.go
```diff
@@ -190,21 +245,28 @@ type VestSpec struct {
 	Quantization abi.ChainEpoch // Maximum precision of vesting table (limits cardinality of table).
 }
 
+// PARAM_SPEC
 var PledgeVestingSpec = VestSpec{
 	InitialDelay: abi.ChainEpoch(180 * builtin.EpochsInDay), // PARAM_FINISH
 	VestPeriod:   abi.ChainEpoch(180 * builtin.EpochsInDay), // PARAM_FINISH
 	StepDuration: abi.ChainEpoch(1 * builtin.EpochsInDay),   // PARAM_FINISH
 	Quantization: 12 * builtin.EpochsInHour,                 // PARAM_FINISH
 }
```
nicola, line 254: **Clarification request:** hey @zixuanzh, can you clarify the following?

Question:
- Miner pledges 20BR to add a sector
- Assume for a second that the miner has never mined any block (hence has no reward)
- After 180 days, the miner will get some of the pledge back (vesting period starts)
- Will the miner on day 181 have less than 20BR? In my understanding a pledge of 20BR must be constant for a miner sector to be active.

What I want to avoid:
- The pledge for a small miner that is +180 days in won't be enough to cover a minimum consensus fault fee
- The pledge is not enough to cover FF/SPs

actors/builtin/miner/policy.go
```diff
@@ -190,21 +245,28 @@ type VestSpec struct {
...
 
+// PARAM_SPEC
 var RewardVestingSpec = VestSpec{
 	InitialDelay: abi.ChainEpoch(20 * builtin.EpochsInDay),  // PARAM_FINISH
```
nicola, line 258: **Clarification request:** Hey @zixuanzh, Is this a final parameter? I recall that this used to be ~180 days. This means that a miner earns block rewards with a delay of 20 days.

actors/builtin/miner/policy.go
zixuanzh: This was introduced when max sector lifetime determined by PoRep is 3-5 times greater than the max sector lifetime that we are allowed to add at a time (1-2 year range). We can either remove it now but we may add this back later or we just remove it entirely and consider bringing it back when we switch to a different PoRep.

actors/builtin/miner/policy.go
zixuanzh: SectorQuality of a sector can be a fraction and we think a 2^20 fixed precision floating point is a good enough precision. cc @davidad

actors/builtin/miner/policy.go
zixuanzh: they are not final and we need to do more analysis

actors/builtin/miner/policy.go
zixuanzh: these are not final numbers and yes it used to be 180 days of delay

actors/builtin/miner/policy.go
irenegia: Confirming that having ~40 mins to submit the proof (i.e., 40 mins between the time the challenges are received and the deadline) is fine.

actors/builtin/miner/policy.go
nicola: Will fix: This is how much a miner is allowed to extend every time they try to extend.

actors/builtin/miner/policy.go
```diff
@@ -171,17 +213,30 @@ type BigFrac struct {
...
 	denominator: big.NewInt(1000),
 }
+
+// PARAM_SPEC
+// Growth rate of a slasher's reward
 var consensusFaultReporterShareGrowthRate = BigFrac{
 	// PARAM_FINISH
 	numerator:   big.NewInt(101251),
 	denominator: big.NewInt(100000),
 }
```
nicola, line 231: **Clarification request:** @zixuanzh will do dig why this is what it is and will update the numbers.

actors/builtin/miner/policy.go
zixuanzh: We will be updating the policy, miners need to maintain the sum of Initial Pledges for all their active sectors.

anorth: .. meaning that it won't vest.

Note that the current constructions can make no guarantee about being able to cover FF/SPs because they are calculated using a different BR as the base.

actors/builtin/miner/policy.go
```diff
@@ -81,47 +91,66 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
 const WorkerKeyChangeDelay = ChainFinality
 
-// Maximum number of epochs past the current epoch a sector may be set to expire.
-// The actual maximum extension will be the minimum of CurrEpoch + MaximumSectorExpirationExtension
-// and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
+// PARAM_SPEC
+// A sector extension can be a maximum number of epochs past the current epoch.
+// A sector may be extended multiple times, however, the maximum extension will be the minimum of
+// CurrEpoch + MaximumSectorExpirationExtension and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
 const MaxSectorExpirationExtension = builtin.EpochsInYear
```
nicola, line 132: **Clarification requested**:
Hey @zixuanzh:
- What is the value that we should choose? 3 months?
- What is motivation for choosing 3 months or 1 year as "maximum" expiration?

actors/abi/sector.go
```diff
@@ -207,12 +207,16 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 	}
 }
 
+
+// PARAM_SPEC
+// The maximum duration (activation to expiration) of a sector sealed with SDR.
+// Motivation: This gurantees that SDR is secure in the cost model for WindowPoSt and in the time model for WinningPoSt
+// This is based on estimation of hardware latency improvement and hardware and software cost reduction.
+const SectorMaximumLifetimeSDR = abi.ChainEpoch(builtin.EpochsInYear * 5)
```
nicola, line 215: **Possible Change**: we should change this according to analysis here: https://github.com/filecoin-project/research-private/issues/242

ping @irenegia, @lucaniz

actors/builtin/miner/policy.go
nicola: Ok, waiting on final param to close this.

actors/builtin/power/policy.go
```diff
@@ -4,8 +4,10 @@ import (
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
 )
 
-// Minimum number of registered miners for the minimum miner size limit to effectively limit consensus power.
-const ConsensusMinerMinMiners = 3
+// PARAM_SPEC
+// This is a switch for turning on the minimum miner size requirement.
+// When this number of miners of minimum miner size is reached, minimum miner size is enforce.
+const ConsensusMinerMinMiners = 4
```
nicola, line 10: **More analysis needed**:
- This number cannot be 3, since 3 guarantees that there is a miner with at least >33% of the power
- This number should either be 0 (meaning that only miners with minimum miner size are allowed to mine)
- Or this number should be set to >4 (e.g. 10 or 100), to ensure that the network is "enough" decentralized
  - Note that this would give the ability for small miners power to participate in the consensus and block reward earning at the beginning. You can consider this as a "threshold" of network decentralization, after which minimum miner size is enforced

Note: This check does not guarantee anything about security, since below `ConsensusMinerMinMiners` we could have a single miner with ~99% and above this threshold we can still have a single miner with ~99%.

Todo: We need to make sure that when the minimum miner size switch is turned on, we don't cutoff many small miners that contribute to 67% security.

cc @irenegia, @jzimmerman, @lucaniz, @sa8

actors/builtin/power/policy.go
```diff
@@ -4,11 +4,20 @@ import (
...
-var ConsensusMinerMinPower = abi.NewStoragePower(1 << 40) // PARAM_FINISH
+// PARAM_SPEC
+// Minimum miner size, the minimum power of an individual miner to meet the threshold for leader election (in bytes).
+// Motivation:
+// - Limits sybil generation
+// - Improves consensus fault detection
+// - Guarantees a minimum fee for consensus faults
+// - Ensures that a specific soundness for the power table
+// Future: we can consensus fault fee and sybil generation with crypto econ mechanic and we can mantain the target soundness by increasing the challenges for small miners.
+var ConsensusMinerMinPower = abi.NewStoragePower(100 << 40) // PARAM_FINISH
```
nicola, line 20: **Clarification request:**
- This was set to 1TiB (assuming that the value is in bytes), can someone confirm? @anorth 
- We set this to 100TiB, pending more research to lower it down.

actors/builtin/power/policy.go
```diff
@@ -4,11 +4,22 @@ import (
...
 
+// PARAM_SPEC
 // Maximum number of prove commits a miner can submit in one epoch
+// Motivation: This aims at limiting the number of operations on chain per miner.
 const MaxMinerProveCommitsPerEpoch = 8000
```
nicola, line 25: **Clarification request:** @ZenGround0 and @anorth can you confirm that the motivation is correct?

actors/builtin/miner/monies.go
```diff
@@ -6,21 +6,37 @@ import (
...
 // LockTarget = (LockTargetFactorNum / LockTargetFactorDenom) * FILCirculatingSupply(t)
 // PledgeShare(t) = sectorQAPower / max(BaselinePower(t), NetworkQAPower(t))
 // PARAM_FINISH
 var InitialPledgeFactor = big.NewInt(20)
+
+// PARAM_SPEC
+// Fraction of available supply that the AdditionalIP targets to lock.
+// PARAM_FINISH
 var LockTargetFactorNum = big.NewInt(3)
 var LockTargetFactorDenom = big.NewInt(10)
```
nicola, line 23: **Clarification request:** @zixuanzh will provide final parameters

actors/builtin/miner/monies.go
nicola: moved to https://github.com/filecoin-project/specs-actors/issues/743

actors/abi/sector.go
lucaniz: Considering:

1. We are in the pragmatic model
2. Our Amax is ~5 (5.14 according to this doc which is the up to date reference [checked with @irenegia ])
3. HW cost is halving every 1 to 2 years [assumption]
We have that
- A conservative approach would suggest to take
`RegisteredSealProof.SectorMaximumLifetime` = 1y
- A more aggressive approach could go to
`RegisteredSealProof.SectorMaximumLifetime` = 2 y (edited) 

Everything which is above 2 years seems insecure, under the assumption in point 3.

actors/builtin/power/policy.go
anorth: It was, because politics.

actors/builtin/power/policy.go
anorth: Not really. It's aimed at what the original comment says, because those prove commits must all be processed in end-of-epoch cron.

actors/abi/sector.go
```diff
@@ -8,6 +8,8 @@ import (
 	cid "github.com/ipfs/go-cid"
 	"github.com/pkg/errors"
 
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
```
anorth, line 12: This will introduce a circular dependency. Please revert.

actors/builtin/network.go
```diff
@@ -2,22 +2,27 @@ package builtin
...
-// This is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
+// Motivation: It guarantees that a block is propagated and WinningPoSt can be successfully done in time all supported miners.
+// Usage: It is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
 // TODO: In lieu of a real configuration mechanism for this value, we'd like to make it a var so that implementations
 // can override it at runtime. Doing so requires changing all the static references to it in this repo to go through
 // late-binding function calls, or they'll see the "wrong" value.
 // https://github.com/filecoin-project/specs-actors/issues/353
 // If EpochDurationSeconds is changed, update `lambda` and `expLamSubOne` in ./reward/reward_logic.go
-const EpochDurationSeconds = 25
+const EpochDurationSeconds = 30
```
anorth, line 14: You can't change this without doing what the comment above talks about. Please revert this. We can track changing this in a separate issue. I have not yet received a concrete request for that change.

anorth:
@nicola please rebase on master, else this will get hard to manage. I would much prefer if small parts of this were merged when ready, with your TODO list maintained in an issue or something elsewhere. This big PR ends up preventing us making improvements sooner, since they'll conflict.

actors/abi/sector.go
nicola: I did not test this code - sorry!

actors/builtin/power/policy.go
nicola: I understand it is aimed at that (I left the original comment), but what is the motivation for choosing 8000?

nicola:
Let me know how you prefer us to operate, but I believe that this issue should represent the outcome of the audit - even if this PR will never be merged and if we will apply changes and track work in new issues (like @ZenGround0 has done for some of those).

We should consider this just as place for discussion for creating issues and PRs addressed elsewhere.

actors/builtin/network.go
nicola: Will be tracked here: https://github.com/filecoin-project/lotus/issues/2547

nicola: Confirmation from @whyrusleeping here: https://github.com/filecoin-project/lotus/issues/2547#issuecomment-663127215

ZenGround0:
@nicola rounding up some untracked params from your toplevel comment checklist

[MinSectorExpiration](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/policy.go#L113)
[ProvCollateralPercentSupplyNum](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/policy.go#L14) - I'm guessing you are asking what this is for rather than pointing out it doesn't exist because this originated in code?  This comes from the cryptoecon team specifying provider deal collateral is equal to 5% of power normalized circulating supply (i.e. circ supply scaled by fraction of deal's power to total network power).
[dealDurationBounds](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/policy.go#L21) -- not sure what the issue is here.  The function in market policy has latest cryptoecon recommended params
[DealUpdatesInterval](https://github.com/filecoin-project/specs-actors/blob/968300c5249c9668da80491afd18e08967c652cd/actors/builtin/market/policy.go#L10) -- I don't know this param too well but how can I help remove confusion?
MaxPoStChainReference -- I also can't find this, what is it used for?

actors/builtin/miner/policy.go
```diff
@@ -32,19 +38,23 @@ func init() {
...
-// With 48 deadlines (half-hour), 200 partitions per declaration permits loading a full EiB of 32GiB
+// With 48 deadlines (half-hour), 238 partitions per declaration permits loading a full EiB of 32GiB
 // sectors with 1 message per epoch within a single half-hour deadline. A miner can of course submit more messages.
-const AddressedPartitionsMax = 200
+const AddressedPartitionsMax = 238
 
-// The maximum number of sector infos that may be required to be loaded in a single invocation.
+// PARAM_SPEC
+// The maximum number of sector infos that can be loaded in a single invocation.
 const AddressedSectorsMax = 10_000
```
nicola, line 58: **Clarification request**:
The calculation for `AddressedPartitionsMax` is correct (we have however updated it with the new blocktime).

However, `AddressedSectorsMax` seems incorrect. If a miner can only report a maximum of `10000` sectors per DeclareFault (for example) and send one message per epoch, it would take them ~30h to declare 1EiB faulted (calculation: 10000 sectors * 32GiB * (60hours*30*60/30).

anorth: A miner can submit many messages per epoch, limited only by competition for block gas limit.

nicola: Why is this set to 10000? The previous number had a motivation (see calculation in comments), I can’t find a valid motivation for this one.

anorth: It's a good question. It's a made up number that seems large enough but hopefully not too large. A better answer would be "empirically determined", but we haven't done the necessary benchmarking.

nicola: Ok, we can do some of that

anorth:
@nicola @ZenGround0 I have rebased and cleaned up this PR in the hope of actually landing it. Please review.

nicola:
It seems to have lost several of my updates, we may have to do this together.

anorth (approved):

actors/abi/sector.go
```diff
@@ -207,11 +207,13 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
...
-// SectorMaximumLifetime is the maximum duration a sector sealed with this proof may exist between activation and expiration
+// The maximum duration a sector sealed with this proof may exist between activation and expiration.
+// This ensures that SDR is secure in the cost model for Window PoSt and in the time model for Winning PoSt
+// This is based on estimation of hardware latency improvement and hardware and software cost reduction.
 func (p RegisteredSealProof) SectorMaximumLifetime() ChainEpoch {
 	// For all Stacked DRG sectors, the max is 5 years
-	epochsPerYear := 1_262_277
-	fiveYears := 5 * epochsPerYear
+	epochsPerYear := 1_051_200 // 365 days / 30 seconds
+	fiveYears := 5 * epochsPerYear // PARAM_SPEC
```
nicola, line 216: Ideally PARAM_SPEC is at the beginning of a comment of a variable for easy parsing for the spec.

anorth: If it's on a new line above, grep output becomes much harder to parse.

E.g. this is now almost useful:
```
[~/p/f/specs-actors](param-audit)$ rg 'PARAM_SPEC'
actors/builtin/network.go
9:// PARAM_SPEC
25:// PARAM_SPEC

actors/abi/sector.go
216:	fiveYears := 5 * epochsPerYear // PARAM_SPEC

actors/builtin/power/policy.go
10:const ConsensusMinerMinMiners = 4 // PARAM_SPEC
20:var ConsensusMinerMinPower = abi.NewStoragePower(100 << 40) // PARAM_SPEC PARAM_FINISH
26:const MaxMinerProveCommitsPerEpoch = 200 // PARAM_SPEC

actors/builtin/market/policy.go
10:const DealUpdatesInterval = builtin.EpochsInDay // PARAM_SPEC
15:	Numerator: big.NewInt(5), // PARAM_SPEC
22:var DealMinDuration = abi.ChainEpoch(180 * builtin.EpochsInDay) // PARAM_SPEC
25:var DealMaxDuration = abi.ChainEpoch(540 * builtin.EpochsInDay) // PARAM_SPEC

actors/builtin/miner/monies.go
14:var PreCommitDepositFactor = 20 // PARAM_SPEC PARAM_FINISH
19:var InitialPledgeFactor = 20 // PARAM_SPEC PARAM_FINISH
25:	Numerator:   big.NewInt(3), // PARAM_SPEC PARAM_FINISH
34:var DeclaredFaultFactorNum = 214 // PARAM_SPEC
44:var UndeclaredFaultProjectionPeriod = abi.ChainEpoch(5) * builtin.EpochsInDay // PARAM_SPEC

actors/builtin/miner/policy.go
18:var WPoStProvingPeriod = abi.ChainEpoch(builtin.EpochsInDay) // 24 hours PARAM_SPEC
23:var WPoStChallengeWindow = abi.ChainEpoch(30 * 60 / builtin.EpochDurationSeconds) // 30 minutes (48 per day) PARAM_SPEC
27:const WPoStPeriodDeadlines = uint64(48) // PARAM_SPEC
53:const AddressedSectorsMax = 10_000 // PARAM_SPEC
59:	MaxPeerIDLength = 128 // PARAM_SPEC
62:	MaxMultiaddrData = 1024 // PARAM_SPEC
79:const ChainFinality = abi.ChainEpoch(1400) // PARAM_SPEC
99:	abi.RegisteredSealProof_StackedDrg32GiBV1:  abi.ChainEpoch(10000), // PARAM_SPEC PARAM_FINISH
109:var MaxPreCommitRandomnessLookback = abi.ChainEpoch(10000) // PARAM_SPEC
114:var PreCommitChallengeDelay = abi.ChainEpoch(150) // PARAM_SPEC
120:const WPoStChallengeLookback = abi.ChainEpoch(20) // PARAM_SPEC
126:const FaultDeclarationCutoff = WPoStChallengeLookback + 50 // PARAM_SPEC
130:var FaultMaxAge = WPoStProvingPeriod * 14 // PARAM_SPEC
134:const WorkerKeyChangeDelay = ChainFinality // PARAM_SPEC
137:const MinSectorExpiration = 180 * builtin.EpochsInDay // PARAM_SPEC
142:const MaxSectorExpirationExtension = 540 * builtin.EpochsInDay // PARAM_SPEC
147:const DealLimitDenominator = 134217728 // PARAM_SPEC
204:	Numerator:   big.NewInt(1), // PARAM_FINISH PARAM_SPEC
210:	Numerator:   big.NewInt(101251), 	// PARAM_FINISH PARAM_SPEC
216:	Numerator:   big.NewInt(1), // PARAM_SPEC
229:var RewardVestingSpec = VestSpec{ // PARAM_SPEC
```

anorth:
Thanks @ZenGround0, I had not intended to leave you with that test still to fix.

PR #702, opened 2020-07-17
Increase Finality

ZenGround0:
Comes from @nicola [here](https://github.com/filecoin-project/specs-actors/pull/695/files#diff-2ecd9908929bba03f96f2c075d8d7ac5R61)

### DIFF
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -53,7 +53,7 @@ func loadPartitionsSectorsMax(partitionSectorCount uint64) uint64 {
 const NewSectorsPerPeriodMax = 128 << 10
 
 // Epochs after which chain state is final.
-const ChainFinality = abi.ChainEpoch(900)
+const ChainFinality = abi.ChainEpoch(1400)
 
 // List of proof types which can be used when creating new miner actors
 var SupportedProofTypes = map[abi.RegisteredSealProof]struct{}{


whyrusleeping:
don't merge this until we get some general agreements on this number

actors/builtin/miner/policy.go
```diff
@@ -53,7 +53,7 @@ func loadPartitionsSectorsMax(partitionSectorCount uint64) uint64 {
 const NewSectorsPerPeriodMax = 128 << 10
 
 // Epochs after which chain state is final.
-const ChainFinality = abi.ChainEpoch(900)
+const ChainFinality = abi.ChainEpoch(1400)
```
whyrusleeping, line 56:
```suggestion
const ChainFinality = abi.ChainEpoch(300)
```

whyrusleeping:
The true meaning of this number is "Forks longer than X blocks require social consensus to resolve". 1400 blocks is about 12 hours. setting this number to 1400 means that we are okay if 12 hours of the economy of filecoin get reverted.

I'm saying here that reverting 12 hours of the entirety of Filecoin is *not* okay. If it comes to a change of that magnitude, then humans need to step in and do something about it, we should not let client software do as it pleases and blindly switch views of the world.

nicola:
I propose we use two constants instead:
- ChainFinality: 1400
- ForkThreshold: x < 1400

We should keep on using ChainFinality where we have it (WorkerKeyChangeDelay and WinningPostLookback) and use ForkThreshold as. A parameter from which the nodes reject blocks from long forks.

anorth:
I'm closing this until we reach some higher level resolution.

PR #703, opened 2020-07-17
Implement smarter deadline assignment algorithm

### LINKED ISSUE #685: Improve algorithm for assigning sectors to deadlines
Broken out of #648 so we can merge it. At the moment, we'll assign sectors to deadlines such that we'll have the least number of partitions _after_ compacting partitions (dealing with terminated sectors). However, this could _force_ miners to repeatedly compact partitions whenever they almost fill any given partition, to avoid spilling over to another partition on the _same_ deadline, increasing the minimum number of GPUs required.

Desired properties (in order):

1. Before compaction, avoid increasing the maximum number of partitions in any deadline, across all deadlines.
2. Before compaction, avoid increasing the number of partitions.
3. After compaction, optimize for the minimum maximum number of partitions in any deadline, across all deadlines.
4. After compaction, optimize for the minimum number of partitions.
5. Optimize for runs of sequential sector numbers.

https://github.com/filecoin-project/specs-actors/pull/648#discussion_r455419162

Stebalien:
See the code for a description of how this works.

fixes #685

TODO

* [x] Tests

### DIFF
--- a/actors/builtin/miner/deadline_assignment_test.go
+++ b/actors/builtin/miner/deadline_assignment_test.go
@@ -78,7 +78,7 @@ func TestDeadlineAssignment(t *testing.T) {
 				deadSectors: 4,
 			},
 			1: {
-				// 2 live partitions. +1 would add another.
+				// 1 live partitions. +1 would add another.
 				liveSectors:   4,
 				expectSectors: []uint64{0},
 			},
--- a/actors/builtin/miner/deadline_assignment_test.go
+++ b/actors/builtin/miner/deadline_assignment_test.go
@@ -1,0 +1,169 @@
+package miner
+
+import (
+	"testing"
+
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/deadline_assignment_test.go
```diff
@@ -0,0 +1,169 @@
...
+		// fewer pre-compaction.
+		sectors: 1,
+		deadlines: [WPoStPeriodDeadlines]*deadline{
+			0: {
+				// one live, one dead.
+				liveSectors: 4,
+				deadSectors: 4,
+			},
+			1: {
+				// 2 live partitions. +1 would add another.
```
anorth, line 81:
```suggestion
				// 1 live partitions. +1 would add another.
```

PR #704, opened 2020-07-17
Feat/exponential baseline function

ZenGround0:
WIP exponential baseline function: x0 * e ^(k t)
- [x] implement exponential function using recurrence exp(n) = exponent * exp(n-1): `BaselinePowerNextEpoch`
- [x] track exp(ThisEpoch) and exp(EffectiveNetworkEpoch) in state
- [x] fix up calls of `BaselinePowerAt` to take in previous epoch baseline
- [x] set initial value to x0 (1 EiB)
- [x] track precision correctly within `BaselinePowerNextEpoch`
- [x] more documentation
- [x] fix tests + new tests

Exponent satisfying 200% growth rate calculated [here](https://www.wolframalpha.com/input/?i=Round%5BExp%5BLog%5B1%2B200%25%5D%2F%28%281+tropical+year%29%2F%2825+seconds%29%29%5D*2%5E128%5D) thanks to @davidad

### DIFF
--- a/actors/builtin/reward/reward_logic.go
+++ b/actors/builtin/reward/reward_logic.go
@@ -5,8 +5,30 @@ import (
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 )
 
-var BaselinePowerAt = func(epoch abi.ChainEpoch) abi.StoragePower {
-	return big.NewInt(1 << 40) // PARAM_FINISH
+const BaselineExponentString = "340282663082994238536867392845056089438"
+
+// Baseline function = BaselineInitialValue * (BaselineExponent) ^(t), t in epochs
+var BaselineExponent big.Int     // Q.128
+var BaselineInitialValue big.Int // Q.0
+
+func init() {
+	BaselineExponent = big.MustFromString(BaselineExponentString)
+	BaselineInitialValue = big.Lsh(big.NewInt(1), 60) // 1 EiB
+}
+
+// Initialize baseline power for epoch -1 so that baseline power at epoch 0 is
+// BaselineInitialValue.
+func InitBaselinePower() abi.StoragePower {
+	baselineInitialValue256 := big.Lsh(big.Lsh(BaselineInitialValue, precision), precision) // Q.0 => Q.256
+	baselineAtMinusOne := big.Div(baselineInitialValue256, BaselineExponent)                // Q.256 / Q.128 => Q.128
+	return big.Rsh(baselineAtMinusOne, precision)                                           // Q.128 => Q.0
+}
+
+// Compute BaselinePower(t) from BaselinePower(t-1) with an additional multiplication
+// of the base exponent.
+func BaselinePowerFromPrev(prevEpochBaselinePower abi.StoragePower) abi.StoragePower {
+	thisEpochBaselinePower := big.Mul(prevEpochBaselinePower, BaselineExponent) // Q.0 * Q.128 => Q.128
+	return big.Rsh(thisEpochBaselinePower, precision)                           // Q.128 => Q.0
 }
 
 // These numbers are placeholders, but should be in units of attoFIL, 10^-18 FIL
@@ -19,15 +41,15 @@ var BaselineTotal = big.Mul(big.NewInt(900e6), big.NewInt(1e18)) // 900M for tes
 // we perform linear interpolation between CumsumBaseline(⌊theta⌋) and CumsumBaseline(⌈theta⌉).
 // The effectiveNetworkTime argument is ceiling of theta.
 // The result is a fractional effectiveNetworkTime (theta) in Q.128 format.
-func computeRTheta(effectiveNetworkTime abi.ChainEpoch, cumsumRealized, cumsumBaseline big.Int) big.Int {
+func computeRTheta(effectiveNetworkTime abi.ChainEpoch, baselinePowerAtEffectiveNetworkTime, cumsumRealized, cumsumBaseline big.Int) big.Int {
 	var rewardTheta big.Int
 	if effectiveNetworkTime != 0 {
 		rewardTheta = big.NewInt(int64(effectiveNetworkTime)) // Q.0
 		rewardTheta = big.Lsh(rewardTheta, precision)         // Q.0 => Q.128
 		diff := big.Sub(cumsumBaseline, cumsumRealized)
-		diff = big.Lsh(diff, precision)                             // Q.0 => Q.128
-		diff = big.Div(diff, BaselinePowerAt(effectiveNetworkTime)) // Q.128 / Q.0 => Q.128
-		rewardTheta = big.Sub(rewardTheta, diff)                    // Q.128
+		diff = big.Lsh(diff, precision)                           // Q.0 => Q.128
+		diff = big.Div(diff, baselinePowerAtEffectiveNetworkTime) // Q.128 / Q.0 => Q.128
+		rewardTheta = big.Sub(rewardTheta, diff)                  // Q.128
 	} else {
 		// special case for initialization
 		rewardTheta = big.Zero()
--- a/actors/builtin/reward/reward_logic.go
+++ b/actors/builtin/reward/reward_logic.go
@@ -5,9 +5,15 @@ import (
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 )
 
+// Round(e^(ln[1 + 200%] / epochsInYear) * 2^128
+// Q.128 formatted number such that f(epoch) = baseExponent^epoch grows 200% in one year of epochs
+// Calculation here: https://www.wolframalpha.com/input/?i=Round%5BExp%5BLog%5B1%2B200%25%5D%2F%28%281+tropical+year%29%2F%2825+seconds%29%29%5D*2%5E128%5D
 const BaselineExponentString = "340282663082994238536867392845056089438"
 
 // Baseline function = BaselineInitialValue * (BaselineExponent) ^(t), t in epochs
+// Note: we compute exponential iteratively using recurrence e(n) = e * e(n-1).
+// Caller of baseline power function is responsible for keeping track of intermediate,
+// state e(n-1), the baseline power function just does the next multiplication
 var BaselineExponent big.Int     // Q.128
 var BaselineInitialValue big.Int // Q.0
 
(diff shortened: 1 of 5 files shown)


ZenGround0:
For reference I evaluated error rates for the implementation over 9 years.  

```
1 EiB, 1 years, diff in bytes: 13442890409, diff percent err: 3.886616262912627e-09
1 EiB, 2 years, diff in bytes: 80655046242, diff percent err: 7.773011231777352e-09
1 EiB, 3 years, diff in bytes: 362941964794, diff percent err: 1.1659332347135179e-08
1 EiB, 4 years, diff in bytes: 1451754074937, diff percent err: 1.554562885887782e-08
1 EiB, 5 years, diff in bytes: 5444044464190, diff percent err: 1.9431917152864137e-08
1 EiB, 6 years, diff in bytes: 19598477798775, diff percent err: 2.3318202696121573e-08
1 EiB, 7 years, diff in bytes: 68594464278790, diff percent err: 2.7204487312864546e-08
1 EiB, 8 years, diff in bytes: 235180483072187, diff percent err: 3.109077161079781e-08
1 EiB, 9 years, diff in bytes: 793732717286012, diff percent err: 3.497705579248354e-08
```

I also evaluated error rates if we keep all 128 bits of precision around in between computations since I was curious if this would improve things.  The difference is on the order of 10^-13 so I'm convinced keeping extra precision around is not worth it.

```
1 EiB, 1 years, diff: 13441742062, diff percent err:3.8862842521627155e-09 
1 EiB, 2 years, diff: 80650452210, diff percent err:7.77256848871292e-09 
1 EiB, 3 years, diff: 362927034238, diff percent err:1.1658852710357353e-08 
1 EiB, 4 years, diff: 1451708134129, diff percent err:1.5545136916913974e-08 
1 EiB, 5 years, diff: 5443905492402, diff percent err:1.9431421108372078e-08 
1 EiB, 6 years, diff: 19598059734565, diff percent err:2.3317705284736418e-08 
1 EiB, 7 years, diff: 68593208937690, diff percent err:2.7203989445997482e-08 
1 EiB, 8 years, diff: 235176715900810, diff percent err:3.10902735921554e-08 
1 EiB, 9 years, diff: 793721414622920, diff percent err:3.497655772321039e-08 
```

The code for this is [here](https://github.com/filecoin-project/specs-actors/tree/knowledge/exp-baseline-rounding) in case numbers look off and anyone wants to check the methodology.

actors/builtin/reward/reward_logic.go
```diff
@@ -5,8 +5,30 @@ import (
...
+var BaselineInitialValue big.Int // Q.0
+
+func init() {
+	BaselineExponent = big.MustFromString(BaselineExponentString)
+	BaselineInitialValue = big.Lsh(big.NewInt(1), 60) // 1 EiB
+}
+
+// Initialize baseline power for epoch -1 so that baseline power at epoch 0 is
+// BaselineInitialValue.
+func InitBaselinePower() abi.StoragePower {
```
ZenGround0, line 21: @anorth Perhaps it would be clearer if we special cased epoch 0 in `updateToNextEpoch`?  We'd also avoid the 1 byte difference between `BaselineInitialValue` and the constructor set `st.ThisEpochBaselinePower`.

actors/builtin/reward/reward_logic.go
```diff
@@ -5,8 +5,30 @@ import (
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 )
 
-var BaselinePowerAt = func(epoch abi.ChainEpoch) abi.StoragePower {
-	return big.NewInt(1 << 40) // PARAM_FINISH
+const BaselineExponentString = "340282663082994238536867392845056089438"
+
+// Baseline function = BaselineInitialValue * (BaselineExponent) ^(t), t in epochs
+var BaselineExponent big.Int     // Q.128
```
ZenGround0, line 11: @Kubuxu please check my `Q.x` correctness.  This is my first time doing precision tracking this way.

actors/builtin/reward/reward_logic.go
```diff
@@ -5,8 +5,30 @@ import (
...
+
+func init() {
+	BaselineExponent = big.MustFromString(BaselineExponentString)
+	BaselineInitialValue = big.Lsh(big.NewInt(1), 60) // 1 EiB
+}
+
+// Initialize baseline power for epoch -1 so that baseline power at epoch 0 is
+// BaselineInitialValue.
+func InitBaselinePower() abi.StoragePower {
+	baselineInitialValue256 := big.Lsh(big.Lsh(BaselineInitialValue, precision), precision) // Q.0 => Q.256
```
Kubuxu, line 22: Just do `2*precision`

actors/builtin/reward/reward_logic.go
```diff
@@ -5,8 +5,30 @@ import (
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 )
 
-var BaselinePowerAt = func(epoch abi.ChainEpoch) abi.StoragePower {
-	return big.NewInt(1 << 40) // PARAM_FINISH
+const BaselineExponentString = "340282663082994238536867392845056089438"
```
Kubuxu, line 8: Definition of the exponent would be great, just in case we need to recalculate it.

davidad: we in fact do, iiuc, as the block time is changing back to 30 seconds 😂 here's how I calculated it: https://www.wolframalpha.com/input/?i=Round%5BExp%5BLog%5B1%2B50%25%5D%2F%28%281+tropical+year%29%2F%2825+seconds%29%29%5D*2%5E128%5D

anorth: Yes, please specify the arithmetic here in a comment and/or link to a gist with code.

actors/builtin/reward/reward_logic.go
```diff
@@ -5,8 +5,30 @@ import (
...
+
+// Baseline function = BaselineInitialValue * (BaselineExponent) ^(t), t in epochs
+var BaselineExponent big.Int     // Q.128
+var BaselineInitialValue big.Int // Q.0
+
+func init() {
+	BaselineExponent = big.MustFromString(BaselineExponentString)
+	BaselineInitialValue = big.Lsh(big.NewInt(1), 60) // 1 EiB
```
anorth, line 16: @zixuanzh can you confirm this is the desired value?

actors/builtin/reward/reward_logic.go
anorth: Your call

actors/builtin/reward/reward_logic.go
```diff
@@ -5,8 +5,30 @@ import (
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
 )
 
-var BaselinePowerAt = func(epoch abi.ChainEpoch) abi.StoragePower {
-	return big.NewInt(1 << 40) // PARAM_FINISH
+const BaselineExponentString = "340282663082994238536867392845056089438"
+
+// Baseline function = BaselineInitialValue * (BaselineExponent) ^(t), t in epochs
```
anorth, line 10: To set up for understanding the implementation, please describe that we don't actually calculate the exponential, but do a multiply for each epoch as it happens, and store intermediate state.

actors/builtin/reward/reward_state.go
```diff
@@ -22,6 +22,9 @@ type State struct {
 	// and in advancing network time.
 	EffectiveNetworkTime abi.ChainEpoch
 
+	// EffectiveBaselinePower is the baseline power at the EffectiveNetworkTime epoch
+	EffectiveBaselinePower abi.StoragePower
```
anorth, line 26: I don't understand this well enough to approve it. I don't think this matches how I'm expecting it to behave, but I also can't quite tell, and also my expectations might be wrong. We probably need a conversation to resolve this to both our satisfaction, though it's likely that a large part of that conversation could be replaced with excessive documentation of intent and algorithm's here.

- What does `effective` mean in this name? For the network time, it's effective rather than real because it's renaming power progress as time, it's not real time. But my understanding of the baseline function is that it should be a function of real network time/age in epochs.
- I thought that the baseline (target) power was an _input_ to the calculation of effective network time. The comment here makes that very confusing. If my interpretation is correct, it would be slightly better to put this in state before the EffectiveNetworkTime, and describe the calculation of one from the other.
- I understand the need for this state as purely an optimization around computing the exponential function `BaselineExponent^st.Epoch`. Baseline is fundamentally a function of real network time.

actors/builtin/reward/reward_state.go
```diff
@@ -54,23 +58,24 @@ func ConstructState(currRealizedPower abi.StoragePower) *State {
...
-
-	cappedRealizedPower := big.Min(BaselinePowerAt(st.Epoch), currRealizedPower)
+	st.ThisEpochBaselinePower = BaselinePowerFromPrev(st.ThisEpochBaselinePower)
+	cappedRealizedPower := big.Min(st.ThisEpochBaselinePower, currRealizedPower)
 	st.CumsumRealized = big.Add(st.CumsumRealized, cappedRealizedPower)
 
 	for st.CumsumRealized.GreaterThan(st.CumsumBaseline) {
 		st.EffectiveNetworkTime++
-		st.CumsumBaseline = big.Add(st.CumsumBaseline, BaselinePowerAt(st.EffectiveNetworkTime))
+		st.EffectiveBaselinePower = BaselinePowerFromPrev(st.EffectiveBaselinePower)
```
anorth, line 67: As my comment above, I don't expect to see this being mutated here, but do expect it to be mutated once outside this loop, corresponding to one epoch elapsing.

I then expect `st.ThisEpochBaselinePower` to be set back where it was originally, as `st.ThisEpochBaselinePower = st.EffectiveBaselinePower`.

actors/builtin/reward/reward_state.go
anorth: I might be just confused. I see things kinda like this in documents elsewhere. The documentation present is just fundamentally not sufficient for me to have a clear conceptual understanding of how this is supposed to work, and so I can't verify it. I'm sure the phrase "inverse function" should appear somewhere. Maybe you could copy enough documentation from elsewhere such that this makes sense?

Kubuxu: This is BaselinePower at EffectiveNetworkTime. Previously we had straight-up formula: `BaselinePowerAt(st.EffectiveNetworkTime)` but now we have to compute it iteratively.

actors/builtin/reward/reward_logic.go
Kubuxu: Seems alright.

actors/builtin/reward/reward_state.go
Kubuxu: Maybe this will make it clearer, we need to keep track of two BaselinePowers:
 - `BaselinePower(rt.Epoch)`
 - `BaselinePower(st.EffectiveNetworkTime)`

actors/builtin/reward/reward_logic.go
zixuanzh: yes - this is our preliminary recommendation

actors/builtin/reward/reward_logic.go
zixuanzh: Note that the number here is still using a 25 seconds epoch duration, we should update this to correspond to the 30 seconds block time when the changes land [340282722315436325019587904274019633912](https://www.wolframalpha.com/input/?i=Round%5BExp%5BLog%5B1%2B200%25%5D%2F%28%281+tropical+year%29%2F%2830+seconds%29%29%5D*2%5E128%5D) related to #748

Issue #604, opened 2020-07-06
Strict parameter validation


Stebalien:
In all methods that take _slices_ of things, we should be validating that all elements in the slice are unique. Even better, we should consider validating that they're unique and sorted.

Usually duplicate elements will just end up costing the caller more, but there could be some cases where batching causes us to actually process the duplicate element multiple times, penalizing or rewarding the caller multiple times.

A simple solution here would be to have all method parameters implement a `Validate` method that must be called before passing the parameters to the actor itself.

Stebalien:
Examples:

* We don't validate that the partitions passed in SubmitWindowedPost are unique.
* We don't validate that the number of proofs in SubmitWindowedPost before passing them to the runtime.
* We don't validate that deal IDs are unique in PreCommitSector?

anorth:
Related to #416 , perhaps do them together.

Stebalien:
It would be really nice to be able to do this via codegen:

1. Codegen a `Validate` method that validates all fields.
2. Allow for tags like `validate:"ordered,nonempty"` for slices, etc.

Stebalien:
I've done a first pass, but one thing I know I've missed is non-nil validation. Especially for bitfields.

Options:

* Check everywhere. I'd consider this a non-option.
* Make bitfields nil-safe for reading. That's better than nothing, but still sketchy.
* Bitfields by value? There's no reason we _can't_ do this.
* Validate non-nil in cbor-gen. We can explicitly tag fields that _can_ be nil. This will require a signoff from @whyrusleeping. We could easily extend this to additional validation (max, etc.) but I'm not sure if this is the right place...

anorth:
> Bitfields by value

The bitfield class doesn't play nice with use as a value. We used to do that, and changed everything a while back. I would love to have an implementation that does, though.

Stebalien:
What was the problem? It looks like it should work.

anorth:
I haven't gone in depth (@whyrusleeping made the change), but from a quick glance the set/unset map overlays will shallow-copy, so mutating a copy of a bitfield will then cause an unexpected mutation in the original.

Stebalien:
> I haven't gone in depth (@whyrusleeping made the change), but from a quick glance the set/unset map overlays will shallow-copy, so mutating a copy of a bitfield will then cause an unexpected mutation in the original.

Eh, that's no different from copying a `map` or our `adt.Array` structure. I'll bug @whyrusleeping for context.

[...]

PR #705, opened 2020-07-18
validate miner params

Stebalien:
Part of https://github.com/filecoin-project/specs-actors/issues/604

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -187,6 +187,8 @@ type ChangePeerIDParams struct {
 }
 
 func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyValue {
+	// TODO: Consider limiting the maximum number of bytes used by the peer ID on-chain.
+	// https://github.com/filecoin-project/specs-actors/issues/712
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
 		info := getMinerInfo(rt, &st)
@@ -205,6 +207,8 @@ type ChangeMultiaddrsParams struct {
 }
 
 func (a Actor) ChangeMultiaddrs(rt Runtime, params *ChangeMultiaddrsParams) *adt.EmptyValue {
+	// TODO: Consider limiting the maximum number of bytes used by multiaddrs on-chain.
+	// https://github.com/filecoin-project/specs-actors/issues/712
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
 		info := getMinerInfo(rt, &st)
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -205,6 +206,7 @@ type ChangeMultiaddrsParams struct {
 }
 
 func (a Actor) ChangeMultiaddrs(rt Runtime, params *ChangeMultiaddrsParams) *adt.EmptyValue {
+	// TODO: Max number/size of multiaddrs?
```
Stebalien, line 209: ?

actors/builtin/miner/miner_actor.go
```diff
@@ -187,6 +187,7 @@ type ChangePeerIDParams struct {
 }
 
 func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyValue {
+	// TODO: Max size of peer ID?
```
Stebalien, line 190: ?

actors/builtin/miner/miner_actor.go
```diff
@@ -416,22 +418,33 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 // Proposals must be posted on chain via sma.PublishStorageDeals before PreCommitSector.
 // Optimization: PreCommitSector could contain a list of deals that are not published yet.
 func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.EmptyValue {
```
Stebalien, line 420: Most of these checks were already here, I just put them in field order,and switched to RequireParam. Please check _carefully_.

actors/builtin/miner/miner_actor.go
```diff
@@ -1328,6 +1344,8 @@ func (a Actor) CompactPartitions(rt Runtime, params *CompactPartitionsParams) *a
 
 // Locks up some amount of a the miner's unlocked balance (including any received alongside the invoking message).
 func (a Actor) AddLockedFund(rt Runtime, amountToLock *abi.TokenAmount) *adt.EmptyValue {
+	builtin.RequireParam(rt, amountToLock.Sign() >= 0, "cannot lock up a negative amount of funds")
```
Stebalien, line 1347: We do check this later... and panic. We need to audit uses of Assert.

actors/builtin/miner/miner_actor.go
anorth: If this is still TODO when you merge, please reference the issue in the comment.

Stebalien: Will do. I was going to try to resolve it now, but I'm not even sure if limits make sense here.

Issue #537, opened 2020-06-25
Consider refusing to unlock funds rather than requiring top-up


Stebalien:
Currently, each miner actor has:

* PreCommitDeposits - funds locked for future sectors
* LockedFunds - other locked funds (pledges, vesting rewards, etc.)
* VestingFunds - a vesting schedule for unlocking locked funds.

There's also a concept of initial pledges that will be enforced in https://github.com/filecoin-project/specs-actors/pull/528. If the miner's "locked funds" falls below this minimum, the miner will need to intentionally lock funds to fix the situation. Apparently miner's should receive block rewards fast enough that this shouldn't be a problem. But it would be nice if it were impossible.

I'd like to propose simplifying this a bit:

* VestingFunds - the total amount of funds currently vesting (not the schedule).
* LockedFunds (InitialPledges, PreCommits, etc...) - funds that cannot be spent for various reasons.
  * Could be stored in separate fields.
* VestingSchedule - The schedule at which funds in VestingFunds vest.

VestingFunds and LockedFunds would act as _independent_ thresholds. When attempting to withdraw funds, the total miner's funds would never be allowed to drop below either of these thresholds. But other than that, these thresholds wouldn't interact.

This way, the miner would never need to intentionally lock up funds, only to have them vest over time again. Instead, if the miner somehow drops below the minimum locked and/or vesting threshold, they'd simply need to send money to the actor to revive it.

Stebalien:
cc @acruikshank

anorth:
I don't think this works, but it's difficult to describe why. Let me try to explain the requirements and current state.

The pre-commit deposits are funds that are locked up on sector pre-commit, and released on proof (or burnt if it never comes). These don't vest. This value cannot be used as security collateral for other sectors.

The other bucket of locked funds is those that are vesting. These locked funds secure (or over-secure) the sectors the state variable called LockedFunds is just the cached sum of the unvested funds here.

The code must maintain an invariant that the miner's balance is not less than the sum of precommit deposits and locked funds.

Separately, there is a requirement that the vesting funds (excl pre-commit deposits) covers the sum of the initial pledge requirements for active sectors. Funds vest (unlock) over time so this property may cease to hold.  This is not an _invariant_, but a _requirement_ for continued participation. The initial pledge posted by the miner isn't "locked" forever. There's a distinction between the funds actually locked, and the funds required to be locked to cover initial pledges for continued participation.

So to speak to some points of your proposal:
[...]

PR #710, opened 2020-07-20
Parameter tweaking: reward vesting and deal weight multiplier

ZenGround0:
1. Change `DealWeightMultiplier` to 10
2. `AddLockedFund` defaults to using a new reward spec which is different from pledge spec.


actors/builtin/miner/miner_actor.go
```diff
@@ -1350,7 +1350,7 @@ func (a Actor) AddLockedFund(rt Runtime, amountToLock *abi.TokenAmount) *adt.Emp
 			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", availableBalance, *amountToLock)
 		}
 
-		if err := st.AddLockedFunds(store, rt.CurrEpoch(), *amountToLock, &PledgeVestingSpec); err != nil {
+		if err := st.AddLockedFunds(store, rt.CurrEpoch(), *amountToLock, &RewardVestingSpec); err != nil {
```
ZenGround0, line 1353: It seems reasonable to impose reward vesting on funds miners intentionally lock up, but perhaps there is something more sophisticated we should be doing like enforcing different specs for different callers.

anorth: I think the prior is how cryptoecon want it. The reward is considered pledged to the network, it's just not the "initial pledge" for a sector. I think there's only one PledgeVestingSpec.

anorth: It seems I was wrong. They actually would prefer something like #537, but for today, different specs are good.

anorth:
After discussion with @zixuanzh I have pushed a commit setting the initial pledge vesting spec to 180+180 days.

PR #711, opened 2020-07-20
Move queue tests to miner_test package

anorth:
This is a pure move, though required creation of a `NewQuantSpec` constructor function.


anorth:
I'm landing this pure refactor of tests before review so that I can build other tests and helpers on top of it.

PR #713, opened 2020-07-20
Recover basic Window PoSt test, fix incorrect deadline sector counts

anorth:


actors/builtin/miner/miner_test.go
```diff
@@ -290,6 +324,9 @@ func TestCommitments(t *testing.T) {
 			Build(t)
 		actor.constructAndVerify(rt)
 
+		// Move the current epoch forward so that the first deadline is a stable candidate for both sectors
+		rt.SetEpoch(periodOffset + miner.WPoStChallengeWindow)
```
anorth, line 328: This test is still skipped, but nearly working.

PR #714, opened 2020-07-20
Assert that payments are non-negative

Stebalien:
I'm not sure if anything can actually go wrong in either of these cases, but this seems like something we shouldn't allow.

Q: ~Assert non-negative instead of positive?~
A: yes (done)

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -141,6 +141,10 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	callerAddr := rt.Message().Caller()
 
+	if params.Value.Sign() < 0 {
+		rt.Abortf(exitcode.ErrIllegalArgument, "proposed value must be non-negative, was %v", params.Value)
+	}
+
 	var txnID TxnID
 	var st State
 	var txn *Transaction
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -172,6 +172,10 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 		rt.Abortf(exitcode.ErrIllegalArgument, "this voucher has expired!")
 	}
 
+	if sv.Amount.Sign() < 0 {
+		rt.Abortf(exitcode.ErrIllegalArgument, "voucher amount must be non-negative, was %v", sv.Amount)
+	}
+
 	if len(sv.SecretPreimage) > 0 {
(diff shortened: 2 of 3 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -141,6 +141,10 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	callerAddr := rt.Message().Caller()
 
+	if params.Value.Sign() <= 0 {
```
Stebalien, line 144: This appears to be checked much later, but it's easier to validate this method if we check it up-front.

actors/builtin/paych/paych_actor.go
```diff
@@ -172,6 +172,10 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 		rt.Abortf(exitcode.ErrIllegalArgument, "this voucher has expired!")
 	}
 
+	if sv.Amount.Sign() <= 0 {
+		rt.Abortf(exitcode.ErrIllegalArgument, "voucher amount not positive: %v", sv.Amount)
```
Stebalien, line 176: This is also implicitly checked later, but it's still nice to implicitly check up-front.

actors/builtin/paych/paych_actor.go
```diff
@@ -172,6 +172,10 @@ func (pca Actor) UpdateChannelState(rt vmr.Runtime, params *UpdateChannelStatePa
 		rt.Abortf(exitcode.ErrIllegalArgument, "this voucher has expired!")
 	}
 
+	if sv.Amount.Sign() <= 0 {
```
Stebalien, line 175: TODO: Do we want to assert somewhere that amount > the previous amount?

actors/builtin/multisig/multisig_actor.go
anorth: Zero must be allowed. A message send does not necessarily transfer value, it may be a method call.

actors/builtin/paych/paych_actor.go
anorth: I would think yes. If not, I think we'd need something where balanceDelta is computed in any case. @whyrusleeping could you confirm?

If correct behaviour with a smaller-than-previous amount is not already tested, could you please add tests to validate/demonstrate behaviour?

whyrusleeping: Nope, The amount can move in any direction, it just can't be negative.

actors/builtin/multisig/multisig_actor.go
Stebalien: Ah, this is a proposed invocation of a method, got it.

PR #717, opened 2020-07-20
Check input cids

magik6k:
Filecoin VM assumes that all CIDs accessible from the state tree, other than ones with the `mh.SHA2_256_TRUNC254_PADDED`/`mh.POSEIDON_BLS12_381_A1_FC1` multihashes, are part of the state tree, and that every VM has access to them.

Getting unexpected CIDs into the state tree is currently a chain-halting bug, this PR should fix all instances of this problem, and make it harder to anciently introduce again

### DIFF
--- a/support/testing/cid.go
+++ b/support/testing/cid.go
@@ -8,10 +8,25 @@ import (
 var DefaultHashFunction = uint64(mh.BLAKE2B_MIN + 31)
 var DefaultCidBuilder = cid.V1Builder{Codec: cid.DagCBOR, MhType: DefaultHashFunction}
 
-func MakeCID(input string) cid.Cid {
+func MakeCID(input string, prefix *cid.Prefix) cid.Cid {
 	c, err := DefaultCidBuilder.Sum([]byte(input))
 	if err != nil {
 		panic(err)
 	}
+
+	if prefix != nil {
+		h, err := mh.Decode(c.Hash())
+		if err != nil {
+			panic(err)
+		}
+
+		ph, err := mh.Encode(h.Digest, prefix.MhType)
+		if err != nil {
+			panic(err)
+		}
+
+		return cid.NewCidV1(prefix.Codec, ph)
+	}
+
 	return c
 }
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -230,8 +230,10 @@ type SealVerifyInfo struct {
 	Randomness            SealRandomness
 	InteractiveRandomness InteractiveSealRandomness
 	Proof                 []byte
-	SealedCID             cid.Cid // CommR
-	UnsealedCID           cid.Cid // CommD
+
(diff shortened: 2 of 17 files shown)


support/testing/cid.go
```diff
@@ -8,10 +8,25 @@ import (
 var DefaultHashFunction = uint64(mh.BLAKE2B_MIN + 31)
 var DefaultCidBuilder = cid.V1Builder{Codec: cid.DagCBOR, MhType: DefaultHashFunction}
 
-func MakeCID(input string) cid.Cid {
+func MakeCID(input string, prefix *cid.Prefix) cid.Cid {
```
Stebalien, line 11: We _really_ should be using `cid.Builder`, not `*cid.Prefix`.

Stebalien:
Wait, so the "checked" tag is just a tag to tell the runtime that we "checked" the CID in the code? That seems brittle and kind of pointless. What about adding additional type/codec information to the CIDs, either through a tag or via a newtype? That should let us validate on decode/encode and/or in the VM.

---

Note: In general, using the codec to decide whether or not we want to download something is the wrong solution. Can we not have a `remote:"true"` flag for CIDs referring to content that doesn't need to be present? Then, it would be up to the runtime to follow all links _except_ those.

support/testing/cid.go
anorth: @Stebalien could you please follow up to implement a better pattern here so that future work doesn't propagate this further?

anorth:
Thanks for jumping in to resolve this quickly so we can roll forward with devnets.

> Filecoin VM assumes that all CIDs accessible from the state tree ... are part of the state tree

In general this seems like a very unfortunate assumption to me. Is it necessary, or just convenient? The change wrought on the actors code is brittle and requires deep context to understand.

I recall discussions a long time ago where we realised that the state tree would have CIDs that point to external content, and resolution approximately of "yup, we gotta deal with it" (context was state-tree sync), possibly with some tag associated with those external CIDs. What if we wanted to reference something outside the system? This assumption certainly won't be maintainable with user-defined contracts.

If we did need to maintain it, could we enforce it at `Store.Put()` instead, and `Abort` rather than halt the chain if the state tree has a dangling CID?

whyrusleeping:
@anorth I think enforcing this on Put is the right thing to do. The enforcement needs to be such that you can only link to CIDs that are returned to you through previous puts, or were pre-existing in the current actors state.

Allowing arbitrary CIDs in the state tree makes a number of things extremely difficult, but the most pressing one is ensuring we don't fill our disks with garbage. When pulling 'the set of objects that are the state tree' we need to know the boundaries of actor state, and we can't rely on knowledge about particular actors.

support/testing/cid.go
Stebalien: After trying to fix this, I get what it's doing. We can't use a cid.Builder because this is making a fake CID.

PR #718, opened 2020-07-20
Market: Caller for withdraw must be the recipient

### LINKED ISSUE #602: Market withdraw allows anyone to trigger withdrawals
From bug report on slack

>As far as I can tell, market_actor.WithdrawBalance makes no requirements on Caller's identity when withdrawing from a non-miner actor.
Assuming Caller and params.ProviderOrClientAddress are both account actors, this function: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go#L79
returns the resolved params.ProviderOrClientAddress, and just checks that Caller is a "signable" type address. I'm pretty sure this means any account/multisig actor can withdraw unlocked, escrowed balance from any non-miner actor. (edited) 

Investigate this and if the above is true come up with a fix

aarshkshah1992:
Closes #602 .


wadeAlexC:
LGTM!

actors/builtin/market/market_actor.go
```diff
@@ -77,7 +78,11 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 		rt.Abortf(exitcode.ErrIllegalArgument, "negative amount %v", params.Amount)
 	}
 
-	nominal, recipient := escrowAddress(rt, params.ProviderOrClientAddress)
+	nominal, recipient, codeId := escrowAddress(rt, params.ProviderOrClientAddress)
+	// if caller is not a miner, caller should be the recipient of the funds
+	if !codeId.Equals(builtin.StorageMinerActorCodeID) {
```
Stebalien, line 83: Does the miner actor need to withdraw? I don't see any calls from the miner code.

actors/builtin/market/market_actor.go
```diff
@@ -5,6 +5,7 @@ import (
 	"sort"
 
 	addr "github.com/filecoin-project/go-address"
+	"github.com/ipfs/go-cid"
```
anorth, line 8: Nit: convention here is to always alias package names that don't match their final path segment.

Motivation: https://github.com/golang/go/issues/28428, https://go-review.googlesource.com/c/tools/+/152000

actors/builtin/market/market_actor.go
anorth: The escrowAddress already invokes `ValidateImmediateCallerXxx`. It's wrong to use that again here, though sometimes there is a good reason to manually check some other predicate.

The bluntest fix here is simply to change `ValidateImmediateCallerType` on line 701 to be `ValidateImmediateCallerIs` the recipient. That resolves the issue correctly, but would have the possibly-unintended consequence of also limiting AddBalance.

If we wanted to distinguish those, we would need to pull the last half of `escrowAddress` up to the callers. Perhaps it could return a tuple of `(nominal Address, approvedCallers []Address)`, and we pull the caller validation up (ignoring approved callers for AddBalance)

aarshkshah1992: This is done.

actors/builtin/market/market_actor.go
aarshkshah1992: This is done.

aarshkshah1992:
@anorth Have made the changes. Please take a look.

actors/builtin/market/market_actor.go
```diff
@@ -111,7 +113,10 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 func (a Actor) AddBalance(rt Runtime, providerOrClientAddress *addr.Address) *adt.EmptyValue {
 	msgValue := rt.Message().ValueReceived()
 
-	nominal, _ := escrowAddress(rt, *providerOrClientAddress)
+	nominal, _, codeId, approved := escrowAddress(rt, *providerOrClientAddress)
+	if codeId.Equals(builtin.StorageMinerActorCodeID) {
+		rt.ValidateImmediateCallerIs(approved...)
+	}
```
anorth, line 119: There must be an `else` here so that caller validation is always performed. [Edit] I see you left it behind in `escrowAddress` but it's too confusing to have validation in different places with different parameters.

It doesn't make sense that for clients we allow any party to deposit funds, but for providers we require it to be the worker/owner. So for AddBalance, just always use `ValidateImmediateCallerType(builtin.CallerTypesSignable...)`

There will then be no need to return the code id from `escrowAddress`.

actors/builtin/market/market_test.go
```diff
@@ -192,6 +192,32 @@ func TestMarketActor(t *testing.T) {
...
+			rt.SetCaller(tutil.NewIDAddr(t, 909), builtin.AccountActorCodeID)
+			rt.ExpectAbort(exitcode.ErrForbidden, func() {
+				rt.Call(actor.WithdrawBalance, &params)
+			})
+			rt.Verify()
+
+			// verify there was no withdrawal
+			rt.GetState(&st)
+			assert.Equal(t, abi.NewTokenAmount(20), actor.getEscrowBalance(rt, client))
+		})
```
anorth, line 219: Test also that withdraw from a miner cannot be triggered by account other than worker/owner.

aarshkshah1992: This has been added.

aarshkshah1992:
@anorth I've moved things around a bit to address your comments and make things easier to read. Please take a look.

actors/builtin/market/market_actor.go
aarshkshah1992: This is done.

PR #719, opened 2020-07-20
Market: Remove Zero balance accounts

### LINKED ISSUE #663: Market: Investigate why we never remove an entry from the Balance Table
See https://github.com/filecoin-project/specs-actors/pull/650#discussion_r453655799.

aarshkshah1992:
Closes #663 .

### DIFF
--- a/actors/builtin/market/market_balances.go
+++ b/actors/builtin/market/market_balances.go
@@ -104,6 +104,29 @@ func (m *marketStateMutation) maybeLockBalance(addr addr.Address, amount abi.Tok
 	return nil, exitcode.Ok
 }
 
+func (m *marketStateMutation) removeAccountIfNoBalance(addr addr.Address) (error, exitcode.ExitCode) {
+	bal, err, code := getBalance(m.escrowTable, addr)
+	if err != nil {
+		return fmt.Errorf("failed to get escrow balance: %w", err), code
+	}
+
+	if bal.Equals(big.Zero()) {
+		prev, err := m.escrowTable.Remove(addr)
+		if err != nil {
+			return fmt.Errorf("failed to remove account: %w", err), exitcode.ErrIllegalState
+		}
+		AssertMsg(prev.Equals(big.Zero()), "previous escrow balance should be zero")
+
+		prev, err = m.lockedTable.Remove(addr)
+		if err != nil {
+			return fmt.Errorf("failed to remove locked account: %w", err), exitcode.ErrIllegalState
+		}
+		AssertMsg(prev.Equals(big.Zero()), "previous locked balance should be zero")
+	}
+
+	return nil, exitcode.Ok
+}
+
 func getBalance(bt *adt.BalanceTable, a addr.Address) (abi.TokenAmount, error, exitcode.ExitCode) {
 	ret, found, err := bt.Get(a)
 	if err != nil {
--- a/actors/builtin/market/market_balances.go
+++ b/actors/builtin/market/market_balances.go
@@ -107,21 +107,28 @@ func (m *marketStateMutation) maybeLockBalance(addr addr.Address, amount abi.Tok
 func (m *marketStateMutation) removeAccountIfNoBalance(addr addr.Address) (error, exitcode.ExitCode) {
 	bal, err, code := getBalance(m.escrowTable, addr)
 	if err != nil {
-		return fmt.Errorf("failed to get escrow balance: %w", err), code
+		return xerrors.Errorf("failed to get escrow balance: %w", err), code
 	}
 
 	if bal.Equals(big.Zero()) {
 		prev, err := m.escrowTable.Remove(addr)
 		if err != nil {
-			return fmt.Errorf("failed to remove account: %w", err), exitcode.ErrIllegalState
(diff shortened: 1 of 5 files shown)


actors/builtin/market/market_balances.go
```diff
@@ -104,6 +104,29 @@ func (m *marketStateMutation) maybeLockBalance(addr addr.Address, amount abi.Tok
...
+		if err != nil {
+			return fmt.Errorf("failed to remove account: %w", err), exitcode.ErrIllegalState
+		}
+		AssertMsg(prev.Equals(big.Zero()), "previous escrow balance should be zero")
+
+		prev, err = m.lockedTable.Remove(addr)
+		if err != nil {
+			return fmt.Errorf("failed to remove locked account: %w", err), exitcode.ErrIllegalState
+		}
+		AssertMsg(prev.Equals(big.Zero()), "previous locked balance should be zero")
```
acruikshank, line 124: I'm wondering (but not sure myself) whether it would be better to log this than assert it. The consequences of this assertion run in a CronTick could be dire, and, if the balance had been non-zero previously, it is zero now. What do you think?

actors/builtin/market/market_balances.go
```diff
@@ -104,6 +104,29 @@ func (m *marketStateMutation) maybeLockBalance(addr addr.Address, amount abi.Tok
 	return nil, exitcode.Ok
 }
 
+func (m *marketStateMutation) removeAccountIfNoBalance(addr addr.Address) (error, exitcode.ExitCode) {
+	bal, err, code := getBalance(m.escrowTable, addr)
+	if err != nil {
+		return fmt.Errorf("failed to get escrow balance: %w", err), code
```
anorth, line 110: Nit: use xerrors

actors/builtin/market/market_balances.go
anorth: It doesn't seem wise to check the escrow table and then remove from a different table. The removal from each table should be guarded by a check on that table? If we expect them both to go at the same time, great, but we don't need to rely on it.

actors/builtin/market/market_test.go
```diff
@@ -2017,19 +2042,41 @@ func (h *marketActorTestHarness) withdrawClientBalance(rt *mock.Runtime, client
 	rt.Verify()
 }
 
-func (h *marketActorTestHarness) cronTickNoChangeBalances(rt *mock.Runtime, client, provider address.Address) {
-	// fetch current client and provider escrow balances
-	cLocked := h.getLockedBalance(rt, client)
-	cEscrow := h.getEscrowBalance(rt, client)
-	pLocked := h.getLockedBalance(rt, provider)
-	pEscrow := h.getEscrowBalance(rt, provider)
+func (h *marketActorTestHarness) cronTickNoChangeBalances(rt *mock.Runtime, client, provider address.Address, clientRemoved, providerRemoved bool) {
```
anorth, line 2045: Since this is `false, false` at all but one call site, it might be less noisy to have this method ignore removals, and check for removal explicitly at the one site where it's expected.

Change getEscrowBalance to return zero for missing accounts, if that makes it easier.

actors/builtin/market/market_balances.go
aarshkshah1992: This is done.

actors/builtin/market/market_balances.go
aarshkshah1992: This is done.

actors/util/adt/balancetable_test.go
```diff
@@ -85,6 +85,31 @@ func TestBalanceTable(t *testing.T) {
...
+	t.Run("Remove account works only if account exists", func(t *testing.T) {
+		addr := tutil.NewIDAddr(t, 100)
+		bt := buildBalanceTable()
+
+		// error if account has not been created
+		bal, err := bt.Remove(addr)
+		require.Error(t, err)
+		require.Equal(t, big.Zero(), bal)
+
+		// now create account -> remove should work
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: This is done.

aarshkshah1992:
@anorth Please take a look and approve if happy.

actors/builtin/market/market_balances.go
```diff
@@ -104,6 +104,36 @@ func (m *marketStateMutation) maybeLockBalance(addr addr.Address, amount abi.Tok
...
+
+	if bal.Equals(big.Zero()) {
+		prev, err := m.escrowTable.Remove(addr)
+		if err != nil {
+			return xerrors.Errorf("failed to remove account: %w", err), exitcode.ErrIllegalState
+		}
+		AssertMsg(prev.Equals(big.Zero()), "previous escrow balance should be zero")
+
+		// remove locked balance account
+		bal, err, code = getBalance(m.lockedTable, addr)
```
anorth, line 121: No need to nest this inside the outer `if`.

aarshkshah1992: @anorth It is possible for locked bal to be zero without the escrow bal being zero in which case we don't want to remove the locked account just yet.

PR #740, opened 2020-07-21
Partition testing

acruikshank:
### Motivation

Partition inherits some testing from miner actor but has none of its own. Quite a bit of logic happens here, so some unit tests are in order.

### Proposed Changes

1. Add tests for all exported methods of `Partition`.

### Bugs Found

1. `TerminateSectors` was adding removed power to `LivePower` rather than subtracting it. Fixed here.

Otherwise everything seems to be working.

### DIFF
--- a/actors/builtin/miner/partition_state.go
+++ b/actors/builtin/miner/partition_state.go
@@ -401,7 +401,7 @@ func (p *Partition) PopExpiredSectors(store adt.Store, until abi.ChainEpoch, qua
 	if p.Faults, err = bitfield.SubtractBitField(p.Faults, expiredSectors); err != nil {
 		return nil, err
 	}
-	p.LivePower = p.LivePower.Sub(popped.ActivePower)
+	p.LivePower = p.LivePower.Sub(popped.ActivePower.Add(popped.FaultyPower))
 	p.FaultyPower = p.FaultyPower.Sub(popped.FaultyPower)
 
 	// Record the epoch of any sectors expiring early, for termination fee calculation later.
--- a/actors/builtin/miner/partition_state.go
+++ b/actors/builtin/miner/partition_state.go
@@ -345,7 +345,7 @@ func (p *Partition) TerminateSectors(store adt.Store, epoch abi.ChainEpoch, sect
 	}
 
 	powerDelta := removed.ActivePower.Add(removed.FaultyPower)
-	p.LivePower = p.LivePower.Add(powerDelta)
+	p.LivePower = p.LivePower.Sub(powerDelta)
 	p.FaultyPower = p.FaultyPower.Sub(removed.FaultyPower)
 	p.RecoveringPower = p.RecoveringPower.Sub(removedRecovering)
 
--- a/actors/builtin/miner/partition_state.go
+++ b/actors/builtin/miner/partition_state.go
@@ -344,12 +344,12 @@ func (p *Partition) TerminateSectors(store adt.Store, epoch abi.ChainEpoch, sect
 		return NewPowerPairZero(), xerrors.Errorf("failed to add terminated sectors: %w", err)
 	}
 
-	powerDelta := removed.ActivePower.Add(removed.FaultyPower)
-	p.LivePower = p.LivePower.Sub(powerDelta)
+	powerRemoved := removed.ActivePower.Add(removed.FaultyPower)
+	p.LivePower = p.LivePower.Sub(powerRemoved)
 	p.FaultyPower = p.FaultyPower.Sub(removed.FaultyPower)
 	p.RecoveringPower = p.RecoveringPower.Sub(removedRecovering)
 
-	return powerDelta, nil
+	return powerRemoved, nil
 }
 
 // PopExpiredSectors traverses the expiration queue up to and including some epoch, and marks all expiring
--- a/actors/builtin/miner/partition_state_test.go
+++ b/actors/builtin/miner/partition_state_test.go
@@ -420,6 +420,68 @@ func TestPartitions(t *testing.T) {
 		})
 	})
 
+	t.Run("pops early terminations", func(t *testing.T) {
+		rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+		partition := emptyPartition(t, rt)
+
+		quantSpec := miner.NewQuantSpec(4, 1)
+		_, err := partition.AddSectors(adt.AsStore(rt), sectors, sectorSize, quantSpec)
+		require.NoError(t, err)
+
+		// fault sector 3, 4, 5 and 6
+		faultSet := bf(3, 4, 5, 6)
+		faultSectors := selectSectors(t, sectors, faultSet)
+		_, err = partition.AddFaults(adt.AsStore(rt), faultSet, faultSectors, abi.ChainEpoch(7), sectorSize, quantSpec)
+		require.NoError(t, err)
+
+		// mark 4and 5 as a recoveries
+		recoverSet := bf(4, 5)
+		recoverSectors := selectSectors(t, sectors, recoverSet)
+		recoveredPower := miner.PowerForSectors(sectorSize, recoverSectors)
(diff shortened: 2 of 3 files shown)


actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+func TestPartitions(t *testing.T) {
+	sectors := []*miner.SectorOnChainInfo{
+		testSector(2, 1, 50, 60, 1000),
+		testSector(3, 2, 51, 61, 1001),
+		testSector(7, 3, 52, 62, 1002),
+		testSector(8, 4, 53, 63, 1003),
+		testSector(11, 5, 54, 64, 1004),
+		testSector(13, 6, 55, 65, 1005),
+	}
+	sectorSize := abi.SectorSize(32 * 1 << 30)
```
Stebalien, line 26:
```suggestion
	sectorSize := abi.SectorSize(32 << 30)
```

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+	t.Run("adds sectors and reports sector stats", func(t *testing.T) {
+		rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+		partition := emptyPartition(t, rt)
+
+		quantSpec := miner.NewQuantSpec(4, 1)
```
Stebalien, line 32: Define this once? It's always the same.

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+		assertPartitionState(t, partition, sectors, sectorSize, bf(1, 2, 3, 4, 5, 6), bf(4, 5, 6), bf(4), bf())
+	})
+
+	t.Run("recovers faults", func(t *testing.T) {
+		rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+		partition := emptyPartition(t, rt)
+
+		quantSpec := miner.NewQuantSpec(4, 1)
+		_, err := partition.AddSectors(adt.AsStore(rt), sectors, sectorSize, quantSpec)
+		require.NoError(t, err)
```
Stebalien, line 140: Extract to a setup function? This looks repeated.

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+		// make 4, 5 and 6 faulty
+		faultSet := bitfield.NewFromSet([]uint64{4, 5, 6})
+		faultSectors := selectSectors(t, sectors, faultSet)
+		_, err = partition.AddFaults(adt.AsStore(rt), faultSet, faultSectors, abi.ChainEpoch(7), sectorSize, quantSpec)
+		require.NoError(t, err)
+
+		// add 4 and 5 as recoveries
+		recoverSet := bitfield.NewFromSet([]uint64{4, 5})
```
Stebalien, line 114: nit: use `bf` function?

Stebalien: (and below)

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+		// we pnly care whether the sectors are in the queue or not. ExpirationQueue tests can deal with early or on time.
+		allSectors, err := bitfield.MergeBitFields(set.OnTimeSectors, set.EarlySectors)
+		require.NoError(t, err)
+		assertBitfieldsEqual(t, group.sectors, allSectors)
+	}
+}
+
+func assertPartitionState(t *testing.T,
+	partition *miner.Partition,
+	sectors []*miner.SectorOnChainInfo,
```
Stebalien, line 510: nit: I'd expect this to be _all_ sectors in the partition, terminated or otherwise.

actors/builtin/miner/partition_state.go
```diff
@@ -345,7 +345,7 @@ func (p *Partition) TerminateSectors(store adt.Store, epoch abi.ChainEpoch, sect
 	}
 
 	powerDelta := removed.ActivePower.Add(removed.FaultyPower)
-	p.LivePower = p.LivePower.Add(powerDelta)
+	p.LivePower = p.LivePower.Sub(powerDelta)
```
anorth, line 348: Thanks. Conventionally I have used "delta" to signify a value that has the right sign for `Add`ing to the prior. To maintain that here, please rename `powerDelta` to `powerRemoved`, which naming would have made the original bug more obvious.

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+		testSector(3, 2, 51, 61, 1001),
+		testSector(7, 3, 52, 62, 1002),
+		testSector(8, 4, 53, 63, 1003),
+		testSector(11, 5, 54, 64, 1004),
+		testSector(13, 6, 55, 65, 1005),
+	}
+	sectorSize := abi.SectorSize(32 * 1 << 30)
+
+	t.Run("adds sectors and reports sector stats", func(t *testing.T) {
+		rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
```
anorth, line 29: The `NewBuilder` can be pulled out, only the `Build` needs to be inside each run.

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+	nonActiveIds, err := bitfield.MergeBitFields(faults, terminations)
+	require.NoError(t, err)
+	expectedActiveIds, err := bitfield.SubtractBitField(allSectorIds, nonActiveIds)
+	require.NoError(t, err)
+
+	active, err := partition.ActiveSectors()
+	require.NoError(t, err)
+	assertBitfieldsEqual(t, expectedActiveIds, active)
+}
```
anorth, line 546: There are some state invariants that should be maintained. Even though probably covered by these assertions, could we check them explicitly? Perhaps even add a "checkInvariants" method to the state object for use by tests.
- faults, recoveries, terminated are contained in sectors
- recoveries are contained in faults
- terminated is exclusive with faults
- all sectors are in expiration queue exactly once
- early terminated is subset of terminated sectors

actors/builtin/miner/partition_state_test.go
```diff
@@ -0,0 +1,591 @@
...
+
+	t.Run("reschedules expirations", func(t *testing.T) {
+		rt := mock.NewBuilder(context.Background(), address.Undef).Build(t)
+		partition := emptyPartition(t, rt)
+
+		quantSpec := miner.NewQuantSpec(4, 1)
+		_, err := partition.AddSectors(adt.AsStore(rt), sectors, sectorSize, quantSpec)
+		require.NoError(t, err)
+
+		// recovered power should equal power of recovery sectors
```
anorth, line 181: stray comment

actors/builtin/miner/partition_state_test.go
Stebalien: I created a `checkPartitionInvariants` function instead of extending the partition state. We can further split these functions up but this probably covers everything for now.

Stebalien:
Turns out we had one more bug lurking in the power calculation in PopExpiredSectors (live == active + faulty, not just active).

actors/builtin/miner/partition_state.go
```diff
@@ -401,7 +401,7 @@ func (p *Partition) PopExpiredSectors(store adt.Store, until abi.ChainEpoch, qua
 	if p.Faults, err = bitfield.SubtractBitField(p.Faults, expiredSectors); err != nil {
 		return nil, err
 	}
-	p.LivePower = p.LivePower.Sub(popped.ActivePower)
+	p.LivePower = p.LivePower.Sub(popped.ActivePower.Add(popped.FaultyPower))
```
Stebalien, line 404: @anorth double checking that this makes sense.

anorth: It does

PR #742, opened 2020-07-21
VerifReg: finish tests

### LINKED ISSUE #682: VerifReg actor: Write tests
This actor has absolutely no tests.

@anorth.

@acruikshank Would you be interested in picking this up since you are working on tests currently ?
Let me know if you are busy and I can pick this up next week.

aarshkshah1992:
Closes #682 .

Tests for `UseBytes` and `RestoreBytes`.

### DIFF
--- a/actors/builtin/verifreg/verified_registry_test.go
+++ b/actors/builtin/verifreg/verified_registry_test.go
@@ -150,13 +150,13 @@ func TestAddVerifiedClient(t *testing.T) {
 		ac.addVerifiedClient(rt, verifier2.Address, c4.Address, c4.Allowance)
 
 		// all clients should exist and verifiers should have no more allowance left
-		require.EqualValues(t, c1.Allowance, ac.getVerifiedClient(rt, c1.Address))
-		require.EqualValues(t, c2.Allowance, ac.getVerifiedClient(rt, c2.Address))
-		require.EqualValues(t, c3.Allowance, ac.getVerifiedClient(rt, c3.Address))
-		require.EqualValues(t, c4.Allowance, ac.getVerifiedClient(rt, c4.Address))
+		require.EqualValues(t, c1.Allowance, ac.getVerifiedClientCap(rt, c1.Address))
+		require.EqualValues(t, c2.Allowance, ac.getVerifiedClientCap(rt, c2.Address))
+		require.EqualValues(t, c3.Allowance, ac.getVerifiedClientCap(rt, c3.Address))
+		require.EqualValues(t, c4.Allowance, ac.getVerifiedClientCap(rt, c4.Address))
 
-		require.EqualValues(t, big.Zero(), ac.getVerifier(rt, verifierAddr))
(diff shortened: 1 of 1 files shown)


actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -272,12 +536,38 @@ func (h *verifRegActorTestHarness) constructAndVerify(rt *mock.Runtime) {
...
+	return &verifreg.AddVerifierParams{Address: a, Allowance: d}
+}
+
+func (h *verifRegActorTestHarness) generateVerifiedClient(a address.Address) *verifreg.AddVerifiedClientParams {
+	allowance := big.Add(verifreg.MinVerifiedDealSize, big.NewInt(99))
+
+	return &verifreg.AddVerifiedClientParams{a, allowance}
+}
+
 func (h *verifRegActorTestHarness) generateAndAddVerifier(rt *mock.Runtime, a address.Address) *verifreg.AddVerifierParams {
```
anorth, line 551: `addNewVerifier` ?

actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -325,7 +603,62 @@ func (h *verifRegActorTestHarness) removeVerifier(rt *mock.Runtime, verifier add
...
+	rt.ExpectValidateCallerAddr(builtin.StorageMarketActorAddr)
+	rt.SetCaller(builtin.StorageMarketActorAddr, builtin.StorageMinerActorCodeID)
+
+	// client will be removed if it's cap after the usebytes call is less than MinVerifiedDealSize
+	var removed bool
+	prev := h.getVerifiedClientCap(rt, a)
+	newCap := big.Sub(prev, dealSize)
+	if newCap.LessThan(verifreg.MinVerifiedDealSize) {
+		removed = true
+	}
```
anorth, line 616: This is ok, but borderline too much logic/smarts in the harness. It's harder to tell what the test cases are actually covering, and if this changes we might accidentally lose coverage of some case that was implicit, rather than explicit.

Alternatively this could not make this check at all (but some call sites make it explicitly), or it could take a parameter being the expected cap after the bytes are used.

actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -325,7 +603,62 @@ func (h *verifRegActorTestHarness) removeVerifier(rt *mock.Runtime, verifier add
...
+	v, err := adt.AsMap(adt.AsStore(rt), st.VerifiedClients)
+	require.NoError(h.t, err)
+	var dc verifreg.DataCap
+	found, err := v.Get(verifreg.AddrKey(a), &dc)
+	require.NoError(h.t, err)
+	if found {
+		prev = dc
+	} else {
+		prev = big.Zero()
+	}
```
anorth, line 649: Ditto, the caller probably knows what to expect and can just tell us here.

PR #746, opened 2020-07-21
dont use const for params that just introduce delays

whyrusleeping:
We need to be able to change this to run through flows in tests in a sane amount of time.


nonsense:
Thanks for that. I opened another PR that updates a few more of the consts to vars, in order to have faster tests: https://github.com/filecoin-project/specs-actors/pull/747

anorth:
Duplicate of #747

PR #747, opened 2020-07-21
decouple WPoStPeriodDeadlines calculation from WPoStProvingPeriod and WPoStChallengeWindow

nonsense:
It'd be helpful if we could decouple the WPoStChallengeWindow calculation from consts, since WPoStChallengeWindow is used to init static arrays, but we need to be able to modify WPoStProvingPeriod, WPoStChallengeWindow, PreCommitChallengeDelay and FaultMaxAge, in order to have faster tests.


actors/builtin/miner/policy_test.go
```diff
@@ -0,0 +1,13 @@
+package miner_test
+
+import (
+	"testing"
+
+	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+)
+
+func TestWPoStPeriodDeadlines(t *testing.T) {
```
nonsense, line 9: This test is a bit redundant as we already check this on `init`, but decided to add it anyways. Let me know if I should remove.

anorth: We don't need the test, but thanks.

PR #750, opened 2020-07-21
Implement partition compaction

Stebalien:
Currently, this is only implemented for partitions with no faulty sectors.

TODO:

* [x] Tests
* [x] Merge https://github.com/filecoin-project/go-bitfield/pull/38
* [x] Implement for partitions with faults?
  * https://github.com/filecoin-project/specs-actors/issues/756

### DIFF
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -1,11 +1,13 @@
 package miner
 
 import (
+	"bytes"
 	"errors"
 	"sort"
 
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/ipfs/go-cid"
+	cbg "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/specs-actors/actors/abi"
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/deadline_state.go
```diff
@@ -444,3 +446,145 @@ func (dl *Deadline) popExpiredPartitions(store adt.Store, until abi.ChainEpoch,
...
+		// Ok, actually unmarshal the partition.
+		byteReader.Reset(lazyPartition.Raw)
+		err := partition.UnmarshalCBOR(&byteReader)
+		byteReader.Reset(nil)
+		if err != nil {
+			return xerrors.Errorf("failed to decode partition %d: %w", partIdx, err)
+		}
+
+		// Don't allow compacting partitions with faulty sectors.
+		hasNoFaults, err := partition.Faults.IsEmpty()
```
Stebalien, line 504: I'm rejecting partitions with faults (for now) because handling them is non-trivial. We need to:

1. Pull out any potential early terminations from the partitions termination queue. It'll just cost more gas.
2. Record faulty/recovering status and carry this over when we re-add the sectors. Unfortunately, this means changing all of our "add new sectors" logic (or adding parallel logic for re-adding known-faulty sectors).

In practice, I don't think this limitation will be problematic.

actors/builtin/miner/partition_state.go
```diff
@@ -345,7 +345,7 @@ func (p *Partition) TerminateSectors(store adt.Store, epoch abi.ChainEpoch, sect
 	}
 
 	powerDelta := removed.ActivePower.Add(removed.FaultyPower)
-	p.LivePower = p.LivePower.Add(powerDelta)
+	p.LivePower = p.LivePower.Sub(powerDelta)
```
Stebalien, line 348: cherry-picked from #740

actors/builtin/miner/deadline_state.go
Stebalien: https://github.com/filecoin-project/specs-actors/issues/756

actors/builtin/miner/deadline_state.go
```diff
@@ -444,3 +446,161 @@ func (dl *Deadline) popExpiredPartitions(store adt.Store, until abi.ChainEpoch,
...
+		// https://github.com/filecoin-project/specs-actors/issues/597
+		return nil, nil, NewPowerPairZero(), xerrors.Errorf("failed to expand partitions into map: %w", err)
+	}
+
+	// Nothing to do.
+	if len(toCompact) == 0 {
+		return bitfield.NewFromSet(nil), bitfield.NewFromSet(nil), NewPowerPairZero(), nil
+	}
+
+	for partIdx := range toCompact { //nolint:nomaprange
```
Stebalien, line 476: I assume this is fine, right? Is it a problem if different miners fail the same transaction with different error messages?

Stebalien:
@anorth I haven't tested the actual CompactPartitions function (part of the actor), just the RemovePartitions method (part of the deadline). Should I try to add actor tests now or are you refactoring the actor tests?

actors/builtin/miner/deadline_state.go
```diff
@@ -444,3 +446,161 @@ func (dl *Deadline) popExpiredPartitions(store adt.Store, until abi.ChainEpoch,
 
 	return popped, modified, nil
 }
+
+// RemovePartitions removes the specified partitions, shifting the remaining
+// ones to the left, and returning the live and dead sectors they contained.
+//
+// Returns an error if any of the partitions contained faulty sectors or early
+// terminations.
+func (dl *Deadline) RemovePartitions(store adt.Store, partitions *bitfield.BitField, quant QuantSpec) (
```
anorth, line 455: Suggest `removePartitions` or `toRemove` for the arg, to be more clear when used.

actors/builtin/miner/miner_actor.go
```diff
@@ -1351,17 +1352,54 @@ func (a Actor) CompactPartitions(rt Runtime, params *CompactPartitionsParams) *a
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %v", params.Deadline)
 	}
 
+	store := adt.AsStore(rt)
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(info.Worker)
 
+		// TODO: enforce this for the particular deadline only?
```
anorth, line 1361: Describe why this restriction, "... so we can calculate the termination fee"

actors/builtin/miner/miner_actor.go
```diff
@@ -1351,17 +1352,54 @@ func (a Actor) CompactPartitions(rt Runtime, params *CompactPartitionsParams) *a
...
+			rt.Abortf(exitcode.ErrForbidden, "cannot compact partitions with outstanding early terminations")
+		}
+
+		if !deadlineIsMutable(st.ProvingPeriodStart, params.Deadline, rt.CurrEpoch()) {
+			rt.Abortf(exitcode.ErrForbidden,
+				"cannot compact deadline %d during its challenge window or the prior challenge window", params.Deadline)
+		}
+
+		partitionCount, err := params.Partitions.Count()
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to parse partitions bitfield")
```
anorth, line 1372: Do this outside the transaction?

actors/util/adt/array.go
```diff
@@ -81,8 +81,10 @@ func (a *Array) BatchDelete(ix []uint64) error {
 func (a *Array) ForEach(out runtime.CBORUnmarshaler, fn func(i int64) error) error {
 	return a.root.ForEach(a.store.Context(), func(k uint64, val *cbg.Deferred) error {
 		if out != nil {
-			// Why doesn't amt.ForEach() just return the value as bytes?
-			if err := out.UnmarshalCBOR(bytes.NewReader(val.Raw)); err != nil {
+			if deferred, ok := out.(*cbg.Deferred); ok {
+				// fast-path deferred -> deferred to avoid re-decoding.
+				*deferred = *val
```
anorth, line 86: Ugh, I hate this deep internal knowledge requirement. This is not the only place I wrote that "Why doesn't..." comment for the same reason. 
- Could you find and improve the others too? 
- If it makes sense to you, could you file an issue on AMT to just return bytes?

actors/builtin/miner/deadline_state.go
```diff
@@ -444,3 +446,161 @@ func (dl *Deadline) popExpiredPartitions(store adt.Store, until abi.ChainEpoch,
...
+	live, dead *abi.BitField, removedPower PowerPair, err error,
+) {
+	oldPartitions, err := dl.PartitionsArray(store)
+	if err != nil {
+		return nil, nil, NewPowerPairZero(), xerrors.Errorf("failed to load partitions: %w", err)
+	}
+
+	partitionCount := oldPartitions.Length()
+	toCompact, err := partitions.AllMap(partitionCount)
```
anorth, line 464: `toCompact` and all references to compaction are confusing - this method only does the removal part and is quite well defined as doing just that.

actors/builtin/miner/deadline_state.go
```diff
@@ -444,3 +446,161 @@ func (dl *Deadline) popExpiredPartitions(store adt.Store, until abi.ChainEpoch,
...
+		// If we're keeping the partition as-is, append it to the new partitions array.
+		if _, ok := toCompact[uint64(partIdx)]; !ok {
+			return newPartitions.AppendContinuous(&lazyPartition)
+		}
+
+		// Ok, actually unmarshal the partition.
+		byteReader.Reset(lazyPartition.Raw)
+		err := partition.UnmarshalCBOR(&byteReader)
+		byteReader.Reset(nil)
+		if err != nil {
```
anorth, line 514: Nit: fold into line 512?

actors/builtin/miner/deadline_state_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+				bf(1, 2, 3, 4),
+				bf(5, 6, 7, 8),
+				bf(9),
+			)
+		}
+
+		// Terminate sectors in partitions 0 and 1, mark sector 9 faulty.
+		// Unfortunately, we don't have per-deadline methods for stuff
+		// like this.
```
anorth, line 93: We could improve this a bit, though, right? At worst with free functions that take and mutate both deadline and partition state. Improving things like this on the way to testing is worthwhile.

actors/builtin/miner/deadline_state_test.go
```diff
@@ -0,0 +1,379 @@
...
+				bf(1, 3, 6), // terminations
+				bf(),        // posts
+
+				bf(1, 2, 3, 4),
+				bf(5, 6, 7, 8),
+				bf(9),
+			)
+		}
+
+		// Try to compact a partition with early terminations.
```
anorth, line 167: Remove "compact" from comments here

actors/builtin/miner/deadline_state_test.go
```diff
@@ -0,0 +1,379 @@
...
+			_, _, _, err := dl.RemovePartitions(store, bf(1), quantSpec)
+			require.Error(t, err, "should have failed to compact a partition with faults")
+		}
+
+		// Try to compact a missing partition.
+		{
+			_, _, _, err := dl.RemovePartitions(store, bf(2), quantSpec)
+			require.Error(t, err, "should have failed to compact missing partition")
+		}
+	})
```
anorth, line 260: This is borderline too much. We should do some work to make writing these tests more concise, because we'll need more of them. Refactoring the deadline to have a more coherent API and less direct state access may help.

actors/builtin/miner/deadline_state_test.go
```diff
@@ -0,0 +1,379 @@
...
+			for _, p := range partitions {
+				present, err := bf.IsSet(p)
+				require.NoError(t, err)
+				assert.True(t, present)
+			}
+		}
+	}
+
+	// Validate early terminations.
+	assertBitfieldEquals(t, dl.EarlyTerminations, partitionsWithEarlyTerminations...)
```
anorth, line 378: As suggested in #740, extract parts here that represent invariant checking, un-parameterized, perhaps to a method on the deadline state object. We can use that in other tests, including actor tests, to gain greater scenario coverage.

actors/builtin/miner/miner_actor.go
```diff
@@ -1022,8 +1022,9 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *Ter
 				err = partitions.Set(decl.Partition, &partition)
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to store updated partition", key)
 
-				// Record that partition now has pending early terminations.
+				// Record updates to deadline
 				deadline.EarlyTerminations.Set(decl.Partition)
+				deadline.LiveSectors -= uint64(len(sectors))
```
anorth, line 1027: Push down to a method, `deadline.RecordEarlyTerminations(partition, sectorcount)`.

anorth:
@Stebalien I have no WIP on actor tests right now, so you're clear. But you might find it difficult to get to the state where you can exercise compaction, at least via the actor's exported methods.

actors/util/adt/array.go
Stebalien: This case wasn't internal knowledge for once, just code sleuthing (ipfs doesn't use cbor-gen).

The crux of the matter is that CBOR objects are _not_ (byte) length prefixed. In order for the AMT to break the array of objects into individual objects, it needs to parse the CBOR structure of the objects (even though it just throws this structure away). This is what `Deferred` does.

If I were to guess why one _might_ want to pass in `cbg.Deferred` here, instead of a simple byte array, I'd guess:

1. It indicates that the object is a valid CBOR object.
2. In the future, we could probably memoize some of the information we learned in the initial decoding to speed up future work?

But those are just guesses. Honestly, `amt.ForEach` _should_ just take a pointer to the thing to decode and we shouldn't need this indirection. I'm not sure why it doesn't.

actors/builtin/miner/miner_actor.go
Stebalien: Actually, I have a question here: Can we narrow this restriction to partitions with pending early terminations? Currently, we prevent _all_ compactions if the miner has _any_ pending early terminations.

actors/builtin/miner/deadline_state.go
Stebalien: I wanted to reset the byte reader in all cases. It's not really an issue here, it just makes me feel a bit safer.

actors/builtin/miner/deadline_state_test.go
Stebalien: Good point. We'd have to export them but that's probably not a huge issue.

Stebalien: https://github.com/filecoin-project/specs-actors/pull/761

actors/builtin/miner/miner_actor.go
anorth: Probably. But limiting to a deadline gets a long way, so I'm not sure it's worth it. They should really just process it.

actors/util/adt/array.go
anorth: By "internal" I just meant looking at the code inside HAMT, rather than behaviour described or implied by the methods and their docs.

Stebalien: Oh, yeah, that...

actors/builtin/miner/miner_actor.go
Stebalien: Turned into an "TerminateSectors" deadline function.

actors/util/adt/array.go
Stebalien: Moved into #781

actors/builtin/miner/deadline_state_test.go
Stebalien: Done. It does take the set of all sectors in the system, but I figure that's something all tests can pass around.

actors/builtin/miner/deadline_state_test.go
Stebalien: I've introduced a state checking abstraction and deduped the tests. How does it look now?

Stebalien: I'd still like to push recording faults down into a deadline function, but that's going to need a new PR.

actors/builtin/miner/deadline_state.go
```diff
@@ -502,3 +504,161 @@ func (dl *Deadline) TerminateSectors(
...
+				return err
+			}
+			if empty, err := bf.IsEmpty(); err != nil {
+				return err
+			} else if !empty {
+				return expirationEpochs.Set(uint64(epoch), bf)
+			}
+			epochsToRemove = append(epochsToRemove, uint64(epoch))
+			return nil
+		})
```
anorth, line 649: Nit: push this down to a method on BitfieldQueue

Stebalien: Done.

Issue #556, opened 2020-06-29
Verified Registry needs to resolve addresses


arajasek:
The Verified Registry currently only works if the `Address` provided in `AddVerifierParams` is always an ID-address. If a pubkey address is sent, that address won't actually be able to add verified clients, since it won't be found in `GetVerifier`, which will be called with its ID address (since `rt.Message().Caller()` always returns an ID address).

Assuming this isn't okay, we need to do similar address resolution as the multisig's `resolve()`. It'll be a little trickier in this case, since we probably want to overwrite pubkey addresses with ID addresses in `st.Verifiers` when they first become available lest we run the risk of having two different entries in the map for the same "conceptual" address.

arajasek:
I marked this a bug, but if we're okay saying `AddVerifier` should always get ID addresses as its params, then we can just document that and close this (we shouldn't be okay with that, IMO).

anorth:
We can enforce the verifier is an ID address, rather than just document it.

@whyrusleeping can you speak to the reasonableness of requiring the verifier account to exist in the state tree?

whyrusleeping:
It might be more user friendly if we use the stable address instead of the ID address. that way things don't fail weirdly if the user hasnt sent funds to their account yet.

That said, i'm fine with either way here

anorth:
If a verifier is constrained to be an account, then we can fetch its stable address from the actor and normalize to that. If it's not so constrained (e.g. can be multisig), there's no general mechanism to get the stable address of an actor.

Schwartz10:
I think the same pattern applies to verifying clients too (not just adding verifiers) - i.e. - if you use a `t1` or `t3` address as a verified client, that address will not be able to make verified deals. Right now, you must use its ID address.

Forcing the use of an ID address for a verified client is not ideal for most users (they can't verify brand new addresses), and not ideal for most application developers either (even though v few might actually run into this exact issue).

anorth:
See #729 for an enumeration of the resolution points.

anorth:
Rather than handle unresolved addresses, an attractive alternative is to create the account actor for a public key with a zero-value send, thus ensuring it has an ID address (#730). When taking this issue, I think that should be the preferred approach (and we can do the same in the multisig).

Schwartz10:
Just so I'm clear @anorth - the creation of a new account actor would happen internally inside Lotus right? I.e. - we won't actually need to send a message with 0 value over the network to create the new account actor?

[...]

PR #758, opened 2020-07-22
VerifReg: Audit Fixes

### LINKED ISSUE #722: Verifreg: Disallow RootKey from assuming other roles
Currently, RootKey can be added as a Verified or as a VerifiedClient. In order to minimize permissions for this critical role, `AddVerifier` and `AddVerifiedClient` should fail for this case.

### LINKED ISSUE #724: Verifreg.AddVerifier: params.Allowance should be greater than or equal to MinVerifiedDealSize
Currently, Verifiers can be added whose allowance doesn't even allow them to add a VerifiedClient. By enforcing that `params.Allowance >= MinVerifiedDealSize`, this should never happen.

### LINKED ISSUE #725: Verifreg.AddVerifiedClient: Existing Verifiers should not be able to become VerifiedClients
Another role-segregation recommendation similar to #722 - `Verifiers` should not be addable in the `AddVerifiedClient` method.

### LINKED ISSUE #726: Verifreg.AddVerifiedClient: Misleading parameter "MinVerifiedDealSize"
Currently, `AddVerifiedClient` fails if the new verified client has an allowance less than or equal to `MinVerifiedDealSize`:

https://github.com/filecoin-project/specs-actors/blob/8c473f656bb07b19c61991f452d1e57d59dc5083/actors/builtin/verifreg/verified_registry_actor.go#L90-L92

This seems misleading. I recommend changing this to strictly "less than"

### LINKED ISSUE #728: Verifreg.UseBytes: Consider never deleting a VerifiedClient entry
As I understand, VerifiedClients should only ever be given one allowance, and this should never change. From code comments:

https://github.com/filecoin-project/specs-actors/blob/8c473f656bb07b19c61991f452d1e57d59dc5083/actors/builtin/verifreg/verified_registry_actor.go#L116-L119

However, deleted VerifiedClients can be re-added by a Verifier via `AddVerifiedClient`.

Additionally, `RestoreBytes` has the authority to create VerifiedClients:

https://github.com/filecoin-project/specs-actors/blob/8c473f656bb07b19c61991f452d1e57d59dc5083/actors/builtin/verifreg/verified_registry_actor.go#L192

In the event of a significant issue with StorageMarket code, this authority may provide an attacker with an additional vector into the Verifreg system.

#### Recommendation

Don't delete VerifiedClient entries.

(Related to #727)

aarshkshah1992:
Closes #722 .
Closes #724 .
Closes #725 .
Closes #726 
Closes #728 .

@wadeAlexC.


wadeAlexC:
The main concern I have so far is that `st.Rootkey` isn't explicitly required to be an ID address, and neither are any of the input `params.Address`. 

Given discussion in #556, we may be able to rely on input addresses being resolved to an ID address.

Maybe we should add an explicit check in the Constructor to ensure that `RootKey` is an ID address?

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -129,9 +148,9 @@ func (a Actor) AddVerifiedClient(rt vmr.Runtime, params *AddVerifiedClientParams
 			rt.Abortf(exitcode.ErrIllegalState, "Failed to update new verifier cap (%d) for %v", newVerifierCap, verifierAddr)
 		}
 
-		// Write-once entry and does not get changed for simplicity.
-		// If parties neeed more allowance, they can get another VerifiedClient account.
 		// This is a one-time, upfront allocation.
+		// This allowance cannot be changed by calls to AddVerifiedClient as long as the client has not been removed.
+		// If parties need more allowance, it be added via a call to RestoreBytes.
```
wadeAlexC, line 153: This isn't really accurate. RestoreBytes is only called in the event that the client already had and used this allowance, but the deal timed out

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -228,12 +247,27 @@ func (a Actor) RestoreBytes(rt vmr.Runtime, params *RestoreBytesParams) *adt.Emp
 	}
 
 	var st State
+	rt.State().Readonly(&st)
+	if st.RootKey == params.Address {
+		rt.Abortf(exitcode.ErrIllegalArgument, "Cannot restore allowance for Rootkey")
+	}
```
wadeAlexC, line 253: Will this ever catch anything?

I think that currently, the market actor sends this method a non-ID address, because it's the same address used to verify a client's signature over a DealProposal.

This may not catch anything, because (I think) RootKey is supposed to be an ID address.

wadeAlexC: On second thought, this should probably be removed. The market actor uses this during its cron tick - an abort here may allow someone to force a cron tick failure.

wadeAlexC: Same goes for the change below.

wadeAlexC: What if instead of removing these, we just removed the "RequireSuccess" condition from the market actor's cron tick?

anorth:
Yes, we can and should constrain `Rootkey` to be an ID address, since we'll know it at genesis state construction.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -50,10 +50,17 @@ type AddVerifierParams struct {
...
 func (a Actor) AddVerifier(rt vmr.Runtime, params *AddVerifierParams) *adt.EmptyValue {
+	if params.Allowance.LessThan(MinVerifiedDealSize) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "Allowance %d below MinVerifiedDealSize for add verifier %v", params.Allowance, params.Address)
+	}
+
 	var st State
 	rt.State().Readonly(&st)
 	rt.ValidateImmediateCallerIs(st.RootKey)
 
+	if params.Address == st.RootKey {
```
anorth, line 61: Please TODO and reference https://github.com/filecoin-project/specs-actors/issues/556 here and anywhere a user-provided address is inspected. Until we do that, this isn't really enforced.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -119,6 +131,13 @@ func (a Actor) AddVerifiedClient(rt vmr.Runtime, params *AddVerifiedClientParams
 			rt.Abortf(exitcode.ErrNotFound, "Invalid verifier %v", verifierAddr)
 		}
 
+		// Validate client to be added isn't a verifier
+		found, err = verifiers.Get(AddrKey(params.Address), nil)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get verifier")
+		if found {
+			rt.Abortf(exitcode.ErrIllegalArgument, "verifier %v cannot be added as a verified client", params.Address)
+		}
```
anorth, line 139: Blocking: unless we also constrain AddVerifier to prevent existing clients, we won't have an invariant.

actors/builtin/verifreg/verified_registry_actor.go
anorth: The market is supposed to resolve these, but verified registry implementation violated this for UseBytes: https://github.com/filecoin-project/specs-actors/issues/768. This case, called from cron, can rely on the ID address and so should check and enforce it.

This implicitly means that, unlike multisig parties, we do require deal participants to have instantiated account actors (fyi @arajasek)

actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -246,6 +275,31 @@ func TestAddVerifiedClient(t *testing.T) {
...
+		// add verified client with caller 1
+		verifier := ac.addNewVerifier(rt, verifierAddr, allowance)
+		client := ac.mkClientParams(verifierAddr, clientAllowance)
+		rt.SetCaller(verifier.Address, builtin.VerifiedRegistryActorCodeID)
+		rt.ExpectValidateCallerAny()
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			rt.Call(ac.AddVerifiedClient, client)
+		})
+		rt.Verify()
+	})
```
anorth, line 302: Test the opposite order too

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: This is done for all methods that need this check.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: This is done.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: This is done everywhere.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: This has been done.

actors/builtin/verifreg/verified_registry_test.go
aarshkshah1992: This is done.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: Have fixed this.

aarshkshah1992:
@anorth @wadeAlexC 

Have address all comments. We now mandate that that the `RootKey` be resolvable to an ID address and fail the construction of the actor if is not. Please approve if happy.

With regards to resolving all `Verifier` and `VerifiedClients` to ID addresses before comparing them with the `RootKey`, I think https://github.com/filecoin-project/specs-actors/issues/556#issuecomment-662848447 is decent solution.

actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -20,32 +20,74 @@ func TestExports(t *testing.T) {
...
-	actor := verifreg.Actor{}
 	receiver := tutil.NewIDAddr(t, 100)
-	builder := mock.NewBuilder(context.Background(), receiver).
-		WithCaller(builtin.SystemActorAddr, builtin.InitActorCodeID)
 
-	rt := builder.Build(t)
+	runtimeSetup := func() *mock.Runtime {
+		builder := mock.NewBuilder(context.Background(), receiver).
+			WithCaller(builtin.SystemActorAddr, builtin.InitActorCodeID)
+		rt := builder.Build(t)
```
anorth, line 28: This is incorrect: the `t` must come from the `t.Run`, not the outer scope.

You could pass it as an arg, but it's just confusing. Do make the builder outside, but then just do `builder.Build(t)` inside `t.Run`.

anorth:
Thanks. I'll try to reach a decision about auto-creating new actor accounts shortly.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -228,12 +265,29 @@ func (a Actor) RestoreBytes(rt vmr.Runtime, params *RestoreBytesParams) *adt.Emp
...
+
 	rt.State().Transaction(&st, func() interface{} {
 		verifiedClients, err := adt.AsMap(adt.AsStore(rt), st.VerifiedClients)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load verified clients")
 
+		verifiers, err := adt.AsMap(adt.AsStore(rt), st.Verifiers)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load verifiers")
+
+		// validate we are NOT attempting to do this for a verifier
+		found, err := verifiers.Get(AddrKey(params.Address), nil)
```
Stebalien, line 283: This check seems expensive (loads a HAMT) and pointless. Are we expecting this case to be possible @aarshkshah1992?

PR #761, opened 2020-07-22
Refactor to push operations into deadline object

Stebalien:
This refactor:

1. Introduces a new "sectors" type to make it easy to operate on the sectors array multiple times without reloading it. The plan is to use this elsewhere as well, possibly passing it into deadline operations that need to access the sectors array.
2. Pushes fault declaration down into a method on the deadline. If this approach is acceptable, it can be extended to other methods.

At the moment, this refactor just demonstrates some ideas. If this looks like the way we want to go, I can extend it to all deadline operations.

### DIFF
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -2,6 +2,7 @@ package miner
 
 import (
 	"errors"
+	"sort"
 
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/sectors.go
```diff
@@ -0,0 +1,89 @@
...
+	sectorsArr, err := adt.AsArray(store, root)
+	if err != nil {
+		return nil, err
+	}
+	return (*Sectors)(sectorsArr), nil
+}
+
+// Sectors is a helper type for accessing/modifying a miner's sectors. It's safe
+// to pass this object around as needed.
+type Sectors adt.Array
```
anorth, line 22: Discussion: here you opted to _newtype_ Sectors as an immediate (not pointer) Array, and then generally use pointer to sectors. In the BitfieldQueue and ExpirationQueue I opted to _extend_ Array _pointer_, and then treat the queue as an immediate/value type, generally not returned as or passed by pointer. (I subsequently added QuantSpec to it, so +1 for extensibility but -1 for being less lightweight as an immediate).

I don't think I have sufficiently deep Go experience to fully weigh these two approaches against each other, but it occurs that if one way is clearly better then we should use it everywhere, and if we're ambivalent then we should just pick one and use it consistently. 

Can you speak to the benefits of one approach over the other?

actors/builtin/miner/sectors.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (sa *Sectors) LoadSectorInfos(sectorNos *abi.BitField) ([]*SectorOnChainInfo, error) {
+	arr := sa.array()
+
+	var sectorInfos []*SectorOnChainInfo
+	if err := sectorNos.ForEach(func(i uint64) error {
+		var sectorOnChain SectorOnChainInfo
+		found, err := arr.Get(i, &sectorOnChain)
+		if err != nil {
+			return fmt.Errorf("failed to load sector %v: %w", abi.SectorNumber(i), err)
```
anorth, line 40: Use xerrors here and elsewhere.

actors/builtin/miner/sectors.go
```diff
@@ -0,0 +1,89 @@
...
+	return (*Sectors)(sectorsArr), nil
+}
+
+// Sectors is a helper type for accessing/modifying a miner's sectors. It's safe
+// to pass this object around as needed.
+type Sectors adt.Array
+
+func (sa *Sectors) array() *adt.Array {
+	return (*adt.Array)(sa)
+}
```
anorth, line 26: Nit: I prefer unexported methods last, optimising for top-down understanding by the user over the implementer.

actors/builtin/miner/deadline_state.go
```diff
@@ -444,3 +445,50 @@ func (dl *Deadline) popExpiredPartitions(store adt.Store, until abi.ChainEpoch,
 
 	return popped, modified, nil
 }
+
+func (dl *Deadline) TerminateSectors(store adt.Store, epoch abi.ChainEpoch, sectors map[uint64][]*SectorOnChainInfo, ssize abi.SectorSize, quant QuantSpec) (removedPower PowerPair, err error) {
```
anorth, line 449: `sectors` -> `partitionSectors` or similar, describing the structure.

actors/builtin/miner/sectors.go
Stebalien: I used a new type in this case because I wanted to expose functions like `Get` and `MustGet` and didn't want them to be confused with the array functions. But it doesn't really matter much either way.

In general, I try to avoid passing around a `type Foo struct { *Bar }` by value. If I then decide that I really _need_ an additional mutable field, I'll need to refactor everything to pass it around by pointer. Instead, I prefer `type Foo Bar` if I don't want to expose the underlying functions and `type Foo struct { Bar }` If I can get away with the dereference (i.e., no locks).

Otherwise, I just eat the pointer indirection and pass `type Foo struct { Bar }` by pointer.

actors/builtin/miner/sectors.go
Stebalien: I figured we don't really care about stack traces inside helper types, but that is more consistent.

actors/builtin/miner/sectors.go
Stebalien: I've switched over to `type Foo struct { *Bar }`. We can change it later if we want but I agree that consistency is king.

Stebalien:
(merging now to avoid leaving prs hanging)

actors/builtin/miner/sectors.go
anorth: Thanks, your rationale sounds good. Thanks for the consistency, I'll be happy for us to switch them all to `struct { Bar }

PR #762, opened 2020-07-22
AlphaBetaFilter

ZenGround0:
Breakdown: 
- [x] alpha beta filter type
- [x] estimation
- [x] cumsum ratio extrapolation (ln)
- [x] slight reward refactor to reuse rational poly approx
- [x] smooth power, reward, circulating supply every cron tick
- [x] pass around filter estimations in power and reward query method returns
- [x] use estimate derived values to improve BR and additional IP calculations
- [x] fix tests
- [x] test

This is ready for review.  @davidad @Kubuxu I would appreciate your help on the smoothing package in particular where the alpha beta filtering logic lives.  And @anorth I'd appreciate yours for integration into rest of the system.  

General note on precision -- the projected cumsum ratio of interest for BR can be quite low for somewhat feasible reward power combinations.  So it turns out we need Q.128 precision returned from this method.  Since all of the values we use directly from the filter as "position" estimates are inherently integers these are just returned as Q.0s.  I'm confident about this but it's tricky and new enough to me that I want y'all to double check.

Finally: the alpha beta filter is missing internal tests.  I have one set of test on the way verifying projection of cumsum ratio is the ratio of the projections.  Other ideas are welcome.


actors/builtin/power/power_state.go
```diff
@@ -41,6 +43,9 @@ type State struct {
 	// Cron will iterate every epoch between this and the current epoch inclusively to find tasks to execute.
 	FirstCronEpoch abi.ChainEpoch
 
+	// Last epoch power cron tick has been processed.
+	LastCronEpoch abi.ChainEpoch
```
anorth, line 47: These two names follow an idiomatic pattern but mean very different things (or, from another point of view, the same thing two different ways).

Change either or both names so that they don't seem so related. Maybe `FirstUnprocessCronEpoch`? `LastProcessedCronEpoch`?

actors/builtin/power/power_state.go
```diff
@@ -206,6 +212,12 @@ func (st *State) appendCronEvent(events *adt.Multimap, epoch abi.ChainEpoch, eve
 	return nil
 }
 
+func (st *State) updateSmoothedEstimate(currEpoch abi.ChainEpoch) {
+	delta := currEpoch - st.LastCronEpoch
```
anorth, line 216: I'm a bit uncertain about this using LastCronEpoch to compute a delta without then updating it to the current epoch. This could be resolved by computing the delta in the caller, which is where the update then happens.

actors/builtin/reward/reward_actor.go
```diff
@@ -139,6 +145,8 @@ func (a Actor) UpdateNetworkKPI(rt vmr.Runtime, currRealizedPower *abi.StoragePo
 		}
 
 		st.updateToNextEpochWithReward(*currRealizedPower)
+		// only update smoothed estimates after updating rewart
+		st.updateSmoothedEstimates(rt.CurrEpoch()-prev, rt.TotalFilCircSupply())
```
anorth, line 149: Blocking: This is not nice. I am guessing this request came from a side-channel from the cryptoecon team and I missed it. It _may_ be an ok way of getting what we want, but it's nasty. Lacking a written design/motivation and discussion I don't think it is.

Firstly, the node cannot compute the circulating supply without consulting the reward actors for the total mined. We have somehow got to this point with that still not being recorded in state. We need to track the total mined first.

Secondly, it's backwards and arbitrary for the reward actor to be maintaining a smoothed version of something it got from the runtime that is calculated, in part, from values the VM (should have) received from it. I think this is an oblique attempt to smooth the impact of initial pledge fluctations on the computed money supply. But we should just smooth those values as inputs.

If we really wanted to smooth the money supply calculation directly, from non-smooth inputs, we could just require the VM to do it for us and move all this computation off-chain.

For now, remove this and revert this use of smoothing; go back to consulting rt.TotalFilCircSupply() where it's needed. The rest of this PR is then clear to land. Smoothing the supply is strictly lower priority than smoothing power shocks (for which being inputs to the epoch reward, and hence money supply, is the reason we're smoothing them anyway).

@zixuanzh

zixuanzh: I agree that smoothing circulating supply is lower priority than smoothing BR. Let's land the high priority changes first and do this as a further optimization.

actors/util/smoothing/alpha_beta_filter.go
```diff
@@ -175,18 +175,21 @@ func ExtrapolatedCumSumOfRatio(delta abi.ChainEpoch, relativeStart abi.ChainEpoc
 }
 
 // The natural log of Q.128 x.
-func ln(z big.Int) big.Int {
-	// ln(z) = ln(x * 2^k) = ln(x) + k * ln2
-	intK := uint(z.BitLen() - 1 - math.Precision)
-	k := big.NewInt(int64(intK))   // Q.0
+func Ln(z big.Int) big.Int {
+	// bitlen - 1 - precision
+	k := big.Sub(big.Sub(big.NewInt(int64(z.BitLen())), big.NewInt(1)), big.NewInt(math.Precision)) // Q.0
```
Kubuxu, line 180: What is the case when you can't do this math in uint/64 precison?

ZenGround0: If you are referring to the k calculation only we should always be able to.  I switched to big int after doing this math with uints and missing the sign on k when switching k to a big.  I can switch to int64.  Is the suggestion for performance?

actors/builtin/miner/monies.go
```diff
@@ -71,19 +75,19 @@ func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.Ch
...
 }
 
 // Computes the pledge requirement for committing new quality-adjusted power to the network, given the current
 // total power, total pledge commitment, epoch block reward, and circulating token supply.
 // In plain language, the pledge requirement is a multiple of the block reward expected to be earned by the
 // newly-committed power, holding the per-epoch block reward constant (though in reality it will change over time).
-func InitialPledgeForPower(qaPower abi.StoragePower, networkQAPower, baselinePower abi.StoragePower, networkTotalPledge abi.TokenAmount, epochTargetReward abi.TokenAmount, networkCirculatingSupply abi.TokenAmount) abi.TokenAmount {
+func InitialPledgeForPower(qaPower abi.StoragePower, baselinePower abi.StoragePower, networkTotalPledge abi.TokenAmount, rewardEstimate, networkQAPowerEstimate *smoothing.FilterEstimate, networkCirculatingSupplySmoothed abi.TokenAmount) abi.TokenAmount {
+	networkQAPower := networkQAPowerEstimate.Estimate()
+	ipBase := big.Mul(InitialPledgeFactor, ExpectedDayRewardForPower(rewardEstimate, networkQAPowerEstimate, qaPower))
```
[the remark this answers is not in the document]
davidad: However, I'm not sure if what I had in mind is actually a good idea. It would certainly help reduce early pledge requirements, but would be a deviation from the idea that there is a single number `BR` which is used to compute both IP and fees (FF, SP), which is something that may be relied upon by Red Team analysis.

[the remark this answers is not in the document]
davidad: One way around this would be to compute the "daily BR" number as an extrapolation of total reward over the course of 20 days, divided by 20. That way, IP is exactly the extrapolated reward over 20 days, and it also has the previously specified relationships to SP, FF, etc. I'd like to get input from @nicola and @zixuanzh on this.

actors/builtin/miner/miner_actor.go
```diff
@@ -2218,18 +2220,15 @@ func commitWorkerKeyChange(rt Runtime) *adt.EmptyValue {
...
 
 // Requests the current epoch target block reward from the reward actor.
-func requestCurrentEpochBlockReward(rt Runtime) abi.TokenAmount {
-	_, rwd := requestCurrentEpochBaselinePowerAndReward(rt)
-	return rwd
-}
-
-func requestCurrentEpochBaselinePowerAndReward(rt Runtime) (abi.StoragePower, abi.TokenAmount) {
+// return value includes reward, smoothed estimate of reward, baseline power
+// and smoothed estimate of total circulating supply this epoch.
```
anorth, line 2224:
```suggestion
```

actors/builtin/power/power_state.go
```diff
@@ -72,6 +77,7 @@ func ConstructState(emptyMapCid, emptyMMapCid cid.Cid) *State {
 		ThisEpochRawBytePower:     abi.NewStoragePower(0),
 		ThisEpochQualityAdjPower:  abi.NewStoragePower(0),
 		ThisEpochPledgeCollateral: abi.NewTokenAmount(0),
+		ThisEpochQAPowerSmoothed:  smoothing.InitialEstimate(),
 		FirstCronEpoch:            0,
```
anorth, line 81: Init `LastProcessedCronEpoch`, probably to -1?

actors/builtin/reward/reward_state.go
```diff
@@ -79,3 +84,8 @@ func (st *State) updateToNextEpochWithReward(currRealizedPower abi.StoragePower)
 	st.ThisEpochReward = computeReward(st.Epoch, prevRewardTheta, currRewardTheta)
 
 }
+
+func (st *State) updateSmoothedEstimates(delta abi.ChainEpoch, observedFILCirculatingSupply abi.TokenAmount) {
```
anorth, line 88: Remove unused `observedFILCirculatingSupply` parameter

actors/builtin/miner/miner_actor.go
```diff
@@ -371,16 +372,16 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 		// Penalize new skipped faults and retracted recoveries as undeclared faults.
 		// These pay a higher fee than faults declared before the deadline challenge window opened.
 		undeclaredPenaltyPower := newFaultPowerTotal.Add(retractedRecoveryPowerTotal)
-		undeclaredPenaltyTarget := PledgePenaltyForUndeclaredFault(epochReward, pwrTotal.QualityAdjPower, undeclaredPenaltyPower.QA)
+		undeclaredPenaltyTarget := PledgePenaltyForUndeclaredFault(rewardStats.ThisEpochRewardSmoothed, pwrTotal.SmoothQAPowerEstimate, undeclaredPenaltyPower.QA)
```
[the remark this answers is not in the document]
ZenGround0: huh weird thought I fixed that one.  Maybe I forgot to push?  I'll get that in for sure

actors/builtin/miner/monies.go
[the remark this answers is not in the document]
ZenGround0: Great to know. I'll follow up once there is a concrete plan, likely after I merge this PR.  Tracked in #822

actors/util/smoothing/alpha_beta_filter.go
```diff
@@ -0,0 +1,221 @@
...
+	initialVelocity big.Int
+
+	DefaultAlpha                   big.Int // Q.128 value of 0.000164
+	DefaultBeta                    big.Int //  Q.128 value of 0.000115
+	ExtrapolatedCumSumRatioEpsilon big.Int // Q.128 value of 2^-50
+)
+
+func init() {
+	initialPosition = big.Zero()
+	initialVelocity = big.Zero()
```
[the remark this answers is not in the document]
ZenGround0: Awesome, I will follow up on this after this PR has merged.  #823

actors/builtin/miner/monies.go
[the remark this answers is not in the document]
nicola: I am not sure I understand the issue, if I am still needed feel free to ping me.

PR #769, opened 2020-07-23
Update data-types

### LINKED ISSUE #517: Murmur3 based HAMTs are not secure with attacker influenced keys
If an attacker can influence the keys used in a HAMT, the HAMT must use a cryptographically secure hash function. As-is, if an attacker can brute-force several (3) 64bit murmur3 hashes, they can prevent a key from being inserted into the HAMT.

Even if the key is a sha256 hash, brute-forcing 2**64 SHA256 hashes (enough to pick a specific murmur3 key) is well within the power of a bitcoin miner.

TL;DR: The hash needs to have at least 128 bits of security (but doesn't need to be safe against the birthday attack).

Stebalien:
And switch to sha256 based HAMTs to fix #517.

Requires https://github.com/filecoin-project/lotus/pull/2539 in lotus (so our HAMTs remain compatible).

* Updates the AMT to ensure that all empty AMTs are consistent. Breaking because this code will now reject some AMTs previously accepted.
* Updates the BitField library to trim trailing zeros, and reject bitfields with trailing zero bytes. Breaking because this will now reject many existing bitfields.
* Updates the HAMT library to support arbitrary hash functions. Breaking because we're switching to SHA256 (cc @whyrusleeping).


actors/util/adt/map.go
```diff
@@ -17,6 +18,15 @@ import (
 // may differ, in which case we can expose it for configuration.
 const hamtBitwidth = 5
 
+// HamtOptions specifies all the options used to construct filecoin HAMTs.
+var HamtOptions = []hamt.Option{
+	hamt.UseTreeBitWidth(hamtBitwidth),
+	hamt.UseHashFunction(func(input []byte) []byte {
+		res := sha256.Sum256(input)
```
Stebalien, line 25: @whyrusleeping said sha256 is faster on his AMD machine than murmur3 was anyways.

actors/util/adt/map.go
```diff
@@ -17,6 +18,15 @@ import (
 // may differ, in which case we can expose it for configuration.
 const hamtBitwidth = 5
 
+// HamtOptions specifies all the options used to construct filecoin HAMTs.
+var HamtOptions = []hamt.Option{
```
Stebalien, line 22: Exported in case someone really wants to construct the HAMT manually.

actors/util/adt/map.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	cid "github.com/ipfs/go-cid"
 	hamt "github.com/ipfs/go-hamt-ipld"
+	"github.com/minio/sha256-simd"
```
anorth, line 8: Hmm, I guess we have little choice but to depend on a concrete implementation. Another thing for the vendor-for-upgrades list 😢 

Note that our use of blake2b is abstracted through a runtime syscall, but that won't be practical here.

Stebalien: So, we _could_ use blake2b, but I assume that would be too slow.

PR #770, opened 2020-07-23
DRY up built-in actor CID logic

Stebalien:
(low priority drive-by change)

### DIFF
--- a/actors/builtin/codes.go
+++ b/actors/builtin/codes.go
@@ -1,6 +1,8 @@
 package builtin
 
 import (
+	"sort"
+
 	"github.com/ipfs/go-cid"
 	mh "github.com/multiformats/go-multihash"
 )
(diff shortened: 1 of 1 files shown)


actors/builtin/codes.go
```diff
@@ -21,45 +21,41 @@ var (
 	CallerTypesSignable         []cid.Cid
 )
 
+var builtinActors map[cid.Cid]string
```
anorth, line 24: Thanks. I later want to put more info in this map, like whether the actor is singleton, whether a "signer", etc.

actors/builtin/codes.go
```diff
@@ -21,45 +23,54 @@ var (
 	CallerTypesSignable         []cid.Cid
 )
 
+var builtinActors map[cid.Cid]*actorInfo
+
+type actorInfo struct {
+	name   string
+	signer bool
+}
```
Stebalien, line 31: @anorth I've added additional info.

PR #774, opened 2020-07-23
Restore CC upgrade test; fix power and pledge errors in deadline cron

anorth:
I also discovered the incorrect `LivePower` update fixed in #740.

This results on one additional but not-so-redudant write of the deadline partitions root in the deadline cron. We could resolve it by smashing together the different jobs of the cron handler, or with patterns to let us hold more in the air easily, but I'm not worried about it in the grand scheme of things right now.

I think these errors would be less likely as we push logic into the state/deadline. I opted for the direct fix here rather than larger refactor so I can focus on more tests.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1636,8 +1636,6 @@ func handleProvingDeadline(rt Runtime) {
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
 		deadline, err := deadlines.LoadDeadline(store, dlInfo.Index)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d", dlInfo.Index)
-		partitions, err := deadline.PartitionsArray(store)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load partitions for deadline %d", dlInfo.Index)
 		quant := st.QuantEndOfDeadline()
 
 		{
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1697,6 +1700,12 @@ func handleProvingDeadline(rt Runtime) {
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
 			penaltyTotal = big.Add(penaltyTotal, penalty)
 
+			// Save modified deadline state.
+			if detectedAny {
+				deadline.Partitions, err = partitions.Root()
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to store partitions")
+			}
```
anorth, line 1707: Previously, the `PopExpiredSectors` below re-loaded the partitions and missed these changes, resulting in duplicate power loss.

actors/builtin/miner/partition_state.go
```diff
@@ -537,25 +537,25 @@ func NewPowerPair(raw, qa abi.StoragePower) PowerPair {
 	return PowerPair{Raw: raw, QA: qa}
 }
 
-func (pp *PowerPair) IsZero() bool {
```
Stebalien, line 540: Any reason to change this?

anorth: Yes, when PowerPair is returned from a function Go refuses to call pointe-receiver methods on it.

`cannot call pointer method on actor.powerPairForSectors(bothSectors)`

PR #775, opened 2020-07-23
Market: Audit Fixes

### LINKED ISSUE #765: Market State.dealGetPaymentRemaining: Bad assertion allows for panic
https://github.com/filecoin-project/specs-actors/blob/25409283f9bf5d6d80b80797e7a6f8415d05d00d/actors/builtin/market/market_state.go#L225-L234

`Assert(durationRemaining > 0)` does not account for the case where `slashEpoch == deal.EndEpoch`.

(This shouldn't happen as `OnMinerSectorsTerminate` will not allow a slash epoch to be set for expired deals.)

### LINKED ISSUE #766: Market.CronTick: Branching statements imply potential abort
https://github.com/filecoin-project/specs-actors/blob/25409283f9bf5d6d80b80797e7a6f8415d05d00d/actors/builtin/market/market_actor.go#L504-L523

If `removeDeal == true`, and `nextEpoch != epochUndefined`, the deal's entry is deleted from `st.DealStates`, then will abort on `dealStates.Set`

(This shouldn't be possible due to logic in `updatePendingDealState`)

### LINKED ISSUE #767: Market.CronTick: Remove RequireSuccess for interaction with VerifiedRegistry and BurntFundsActor
The CronTick finishes by invoking the Verifreg in a loop, as well as sending burnt funds to the BurntFundsActor:

https://github.com/filecoin-project/specs-actors/blob/25409283f9bf5d6d80b80797e7a6f8415d05d00d/actors/builtin/market/market_actor.go#L567-L584

Given that this is the Market's cron tick method, these calls should not use `RequireSuccess`.

### LINKED ISSUE #768: Market: fix use of unresolved Proposal client address when checking verified deals
The market actor resolves a deal's provider and client address when the deal is published, and stores the resolved addresses in state so that they are thenceforth always ID addresses. The implementation of the verified registry actor and its interactions violated this invariant by calling to `UseBytes` before resolving the client address.

The simplest way to resolve this would be to move the calls to UseBytes below the state transaction (which will be rolled back anyway if UseBytes aborts).

FYI @wadealexc

### LINKED ISSUE #771: Market.AddBalance creates balance table entries when provided 0 value
https://github.com/filecoin-project/specs-actors/blob/25409283f9bf5d6d80b80797e7a6f8415d05d00d/actors/builtin/market/market_actor.go#L115-L116

#### Recommendation

Check that `ValueReceived` is greater than zero.

aarshkshah1992:
Closes #765 
Closes #766 .
Closes #767 .
Closes #768 .
Closes #771 .

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -512,25 +512,20 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 				}
 
 				slashAmount, nextEpoch, removeDeal := msm.updatePendingDealState(rt, state, deal, dealID, rt.CurrEpoch())
+				Assert(slashAmount.GreaterThanEqual(big.Zero()))
 				if removeDeal {
-					if err := deleteDealProposalAndState(dealID, msm.dealStates, msm.dealProposals, true, true); err != nil {
-						builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to delete deal")
-					}
-				}
-				if !slashAmount.IsZero() {
+					Assert(nextEpoch == epochUndefined)
+
 					amountSlashed = big.Add(amountSlashed, slashAmount)
-				}
+					err := deleteDealProposalAndState(dealID, msm.dealStates, msm.dealProposals, true, true)
+					builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to delete deal proposal and states")
+				} else {
+					Assert(nextEpoch > rt.CurrEpoch() && amountSlashed.IsZero())
 
-				if nextEpoch != epochUndefined && !removeDeal {
-					Assert(nextEpoch > rt.CurrEpoch())
-
-					// TODO: can we avoid having this field?
-					// https://github.com/filecoin-project/specs-actors/issues/463
+					// Update deal's LastUpdatedEpoch in DealStates
 					state.LastUpdatedEpoch = rt.CurrEpoch()
-
-					if err := msm.dealStates.Set(dealID, state); err != nil {
-						rt.Abortf(exitcode.ErrPlaceholder, "failed to set deal state state: %v", err)
-					}
+					err = msm.dealStates.Set(dealID, state)
+					builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal state")
 
 					updatesNeeded[nextEpoch] = append(updatesNeeded[nextEpoch], dealID)
 				}
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -514,13 +514,14 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 				slashAmount, nextEpoch, removeDeal := msm.updatePendingDealState(rt, state, deal, dealID, rt.CurrEpoch())
 				Assert(slashAmount.GreaterThanEqual(big.Zero()))
 				if removeDeal {
-					Assert(nextEpoch == epochUndefined)
+					AssertMsg(nextEpoch == epochUndefined, "next scheduled epoch should be undefined as deal has been removed")
 
 					amountSlashed = big.Add(amountSlashed, slashAmount)
 					err := deleteDealProposalAndState(dealID, msm.dealStates, msm.dealProposals, true, true)
 					builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to delete deal proposal and states")
 				} else {
-					Assert(nextEpoch > rt.CurrEpoch() && amountSlashed.IsZero())
+					AssertMsg(nextEpoch > rt.CurrEpoch() && slashAmount.IsZero(), "deal should not be slashed and should have a schedule for next cron tick"+
+						"as it has not been removed")
 
 					// Update deal's LastUpdatedEpoch in DealStates
 					state.LastUpdatedEpoch = rt.CurrEpoch()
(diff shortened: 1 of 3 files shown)


aarshkshah1992:
@wadeAlexC.

wadeAlexC:
From https://github.com/filecoin-project/specs-actors/issues/767#issuecomment-662811078

I know a `Send` should never fail to the BurntFunds actor, but I still think we should remove `RequireSuccess`, even if the only thing it accounts for would be a negative `amountSlashed`.

actors/builtin/market/market_state.go
```diff
@@ -231,7 +231,6 @@ func dealGetPaymentRemaining(deal *DealProposal, slashEpoch abi.ChainEpoch) abi.
 	}
 
 	durationRemaining := deal.EndEpoch - slashEpoch
-	Assert(durationRemaining > 0)
```
wadeAlexC, line 234: Because `deal.StartEpoch` is passed in from outside, I don't think this assertion should be removed. I would just change it to `durationRemaining >= 0`

actors/builtin/market/market_actor.go
```diff
@@ -525,7 +532,7 @@ func (a Actor) CronTick(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 					amountSlashed = big.Add(amountSlashed, slashAmount)
 				}
 
-				if nextEpoch != epochUndefined {
+				if nextEpoch != epochUndefined && !removeDeal {
```
wadeAlexC, line 535: My main concern with this structure is not that #766 could actually happen, it's that the branching logic does not correspond to the range of returned values from `updatePendingDealState`. Here's what I would like to see:

```go
slashAmount, nextEpoch, removeDeal := msm.updatePendingDealState(...)
// Should always hold
Assert(slashAmount >= 0)
if removeDeal {
  // If we're removing the deal, there is no "next epoch:"
  Assert(nextEpoch == epochUndefined)

  // Handle everSlashed case:
  if slashAmount != 0 {
    amountSlashed = big.Add(amountSlashed, slashAmount)
  }

  // Delete deal from Proposals and DealStates:
  if err := deleteDealProposalAndState(...); err != nil {
    builtin.RequireNoErr(rt, err, ...)
  }
} else {
  Assert(nextEpoch > rt.CurrEpoch() && amountSlashed == 0)

  // Update deal's LastUpdatedEpoch in DealStates
  state.LastUpdatedEpoch = rt.CurrEpoch()
  if err := msm.dealStates.Set(dealID, state); err != nil {
    rt.Abortf(...)
  }

  // Add dealID to updatesNeeded for re-insertion in DealOpsByEpoch
  updatesNeeded[nextEpoch] = append(updatesNeeded[nextEpoch], dealID)
}
```

I'm not crazy about adding `Asserts`, but my understanding of the market and `updatePendingDealStates` is that these should all be fine.

wadeAlexC:
Added comments. I'll leave @anorth to comment on the `removeAccountIfNoBalance` change - I'm still not sure what the intended behavior is there.

actors/builtin/market/market_actor.go
aarshkshah1992: I see what you mean now. Thanks.

aarshkshah1992:
@anorth Please take a look.

actors/builtin/market/market_actor.go
```diff
@@ -505,25 +512,20 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
+				Assert(slashAmount.GreaterThanEqual(big.Zero()))
 				if removeDeal {
-					if err := deleteDealProposalAndState(dealID, msm.dealStates, msm.dealProposals, true, true); err != nil {
-						builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to delete deal")
-					}
-				}
-				if !slashAmount.IsZero() {
-					amountSlashed = big.Add(amountSlashed, slashAmount)
-				}
+					Assert(nextEpoch == epochUndefined)
```
anorth, line 517: Prefer AssertMsg so, if it happens, we have more info about what went wrong.

actors/builtin/market/market_actor.go
```diff
@@ -221,6 +206,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			}
 			// Normalise provider and client addresses in the proposal stored on chain (after signature verification).
 			deal.Proposal.Provider = provider
+			resolvedAddrs[deal.Proposal.Client] = client
```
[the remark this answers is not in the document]
wadeAlexC: The loop iterating over verified deals happens outside of the state transition and is iterating over `params.Deals`. This mapping was needed to remember the resolution that happens within the state transition.

actors/builtin/market/market_actor.go
```diff
@@ -505,25 +512,20 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
-				}
+					Assert(nextEpoch == epochUndefined)
 
-				if nextEpoch != epochUndefined {
-					Assert(nextEpoch > rt.CurrEpoch())
+					amountSlashed = big.Add(amountSlashed, slashAmount)
+					err := deleteDealProposalAndState(dealID, msm.dealStates, msm.dealProposals, true, true)
+					builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to delete deal proposal and states")
+				} else {
+					Assert(nextEpoch > rt.CurrEpoch() && amountSlashed.IsZero())
```
wadeAlexC, line 523: Oops, I think my suggestion was wrong on this one. It's not `amountSlashed.IsZero()`, it's `slashAmount.IsZero()`.

wadeAlexC:
LGTM except for the `amountSlashed` / `slashAmount` mixup - that definitely needs to be fixed :)

I assume this slipped by because there are no tests that caught it, but there definitely should be. @aarshkshah1992 do you think you could add something that would catch this?

actors/builtin/market/market_actor.go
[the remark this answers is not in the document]
anorth: Oh `params.Deals` aren't pointers, ok.

Just for completeness, I propose putting the `deal.Proposal.Provider -> provider` mapping in here as well, then.

PR #779, opened 2020-07-23
Convenience baselineforepoch

ZenGround0:
Requested for lotus integration.

Convenience function of baseline given epoch.  It does as many multiplications as the epoch you want the baseline for so its not as fast as it could be but from tests I think its still on the order of seconds for epochs in the 9 years category so should be fine.

If we want something faster the right way is to read reward actor state to jump ahead.  Also it can be computed ahead / memoized in the consumer if it comes to that.


ZenGround0:
Also note this currently drifts from the on chain value in the case of null blocks

PR #782, opened 2020-07-23
set minimum deal duration to 180 days and fix tests

### LINKED ISSUE #735: Deal lifetime bounds
We parameterize deal lifetime bounds already we just need to change the parameters.  180 days should be the minimum and 366 should be the maximum.

acruikshank:
closes #735

### Motivation

Deal lifetime must be at least 180 days and no more than 366 days (one leap year). Set this bound and fix the tests..

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -175,7 +175,7 @@ func TestMarketActor(t *testing.T) {
 
 	t.Run("WithdrawBalance", func(t *testing.T) {
 		startEpoch := abi.ChainEpoch(10)
-		endEpoch := abi.ChainEpoch(20)
+		endEpoch := startEpoch + 200*builtin.EpochsInDay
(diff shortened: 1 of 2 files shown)


acruikshank:
cc @wadeAlexC

actors/builtin/market/market_test.go
```diff
@@ -1410,12 +1413,13 @@ func TestCronTickDealExpiry(t *testing.T) {
 		// assert deal exists
 		actor.getDealProposal(rt, dealId)
 
-		// move the epoch to 155(anything greater than 150), so deal is first processed at 150 & then at 151 which is it's end epoch
-		// total payment = (end - start) = 151 - 50 = 101
-		current = 155
+		// move the epoch to endEpoch+5(anything greater than endEpoch), so deal is processed at 150 & then at 151 which is it's end epoch
```
ZenGround0, line 1416: nit: I think 150 and 151 at the end of line are no longer correct

PR #783, opened 2020-07-24
Inline small blocks into CIDs.

Stebalien:
This should save us a blockstore access for all blocks smaller than 32 bytes (the inline limit). We could set the limit to be larger but this is a safe place to start.

WIP because this requires some changes in lotus.

### DIFF
--- a/actors/abi/cid.go
+++ b/actors/abi/cid.go
@@ -10,9 +10,29 @@ const (
 	HashLength   = 32
 )
 
-// CidBuilder is the default CID builder for Filecoin.
-var CidBuilder cid.Builder = cid.V1Builder{
-	Codec:    cid.DagCBOR,
-	MhLength: HashLength,
-	MhType:   HashFunction,
+type cidBuilder struct {
+	codec uint64
 }
+
+func (cidBuilder) WithCodec(c uint64) cid.Builder {
+	return cidBuilder{codec: c}
+}
+
+func (b cidBuilder) GetCodec() uint64 {
+	return b.codec
+}
+
+func (b cidBuilder) Sum(data []byte) (cid.Cid, error) {
+	hf := HashFunction
+	if len(data) <= HashLength {
+		hf = mh.IDENTITY
+	}
+	return cid.V1Builder{Codec: b.codec, MhType: hf}.Sum(data)
+}
+
+// CidBuilder is the default CID builder for Filecoin.
+//
+// - The default codec is CBOR. This can be changed with CidBuilder.WithCodec.
+// - The default hash function is 256bit blake2b when the data is > 32 bytes
+//   long and the identity function when the data is <= 32 bytes long.
+var CidBuilder cid.Builder = cidBuilder{codec: cid.DagCBOR}


mikeal:
I’ve said this in other threads, and @warpfork and @rvagg can follow up here, but in general we (IPLD team) don’t think inline CID’s are a good idea.

The representational pairing of a key to a value is broken when you inline the value into the key. It violates a lot of assumptions not only in **your** stack but in the stacks of everyone who has to consume the data and addresses you produce.

I’m not sure how often this would end up getting used in filecoin if this change were to be taken, so I’m not sure how often these CID’s would end up being consumed inside the data structures or outside as addresses, but I **know** there are implementations that will not handle this well and should be considered.

I don’t know what the current performance impact of not inlining small block is, but if it isn’t a substantial difference I can’t imagine the tradeoffs are worth it.

It’s also worth considering just inlining the actual data into CBOR using a union rather than linking in the places where you know you want to avoid linking to small amounts of data. This would also greatly increase the size threshold for inlining and make the data structure a lot more predictable.

Stebalien:
To be clear, the optimization we're discussing here is inlining blocks smaller than the default hash digest.

> The representational pairing of a key to a value is broken when you inline the value into the key. It violates a lot of assumptions not only in your stack but in the stacks of everyone who has to consume the data and addresses you produce.

Please specify.

>  but I know there are implementations that will not handle this well and should be considered.

There are two parts to inlining blocks into CIDs:

1. Using the identity hash function for CIDs where the size if the data is less than or equal to the cutoff (in this case, 32 bytes). Implementations of filecoin will need to consistently construct CIDs using the correct hash function/codec anyways so I can't see this being a huge issue.
2. Bypassing load/store operations for blocks inlined into CIDs. This second part is purely an optimization and should be completely optional.

> I don’t know what the current performance impact of not inlining small blocks is in this codebase, but if it isn’t a substantial difference I can’t imagine the tradeoffs are worth it.

~12% of our blocks are less than or equal to 32 bytes. I don't know how frequently these small blocks are accessed, but I assume very frequently given some of our data structures.

The problem is that we have variable size data structures. I agree that using unions would be a better approach, but defining consistent rules for this is well beyond what we can support/do right now.

mikeal:
>> The representational pairing of a key to a value is broken when you inline the value into the key. It violates a lot of assumptions not only in your stack but in the stacks of everyone who has to consume the data and addresses you produce.
> Please specify.

Sure.

Across the implementations of IPLD and its dependencies like CID there a number of abstractions for links, nodes, blocks, block storage and block services (network). The exact approach to each of these vary by implementation but at an abstract level they have a lot of similarities.

The first thing you have to do with a CID is just create a proper representation. Those vary a lot by language but none of them have an issue that I’m aware of **representing** inline CID’s, they look like any other link with a hashing function I may or may not have.

The problem is what people *do* with those links. The most obvious thing they do is resolve them, either from their block store or from a block service on the network, this is where the problems start to occur.

All code that passes a CID off to these abstractions may need to be updated to consider inline CID’s, depending on the storage/service abstraction. If they don’t consider them (they mostly don’t) then the block storage and block service abstractions need to consider them and return either block data or a Block abstraction.

If they return data this presents another problem, you need to make sure that you **also** continue to inline the data or else you’ll mutate the parent block as you roundtrip through these code paths even if you intended to make no changes. When these libraries don’t have a block abstraction they tend to rely on a static or configurable codec and assume the data and the CID for the data are separate and these libraries can’t be easily updated to handle inline cases. It’s actually a lot of work to hunt down everywhere this might exist when you don’t have a unified Block abstraction.

Block abstractions mostly don’t handle this right now either, although this problem is a bit easier to handle if you have a Block abstraction because then the representation of the key/value are unified in one place and can be handled together. These are comparatively simple to update to handle inline, but not everyone has a unified block abstraction they use everywhere.

That covers the read side, now the write side :)

Everyone who wants to write hash compatible data structures will need to write identical rules for inlining and make sure that any time they write data they follow the exact same rules. Configuring a minimum CID inline size is not supported **anywhere** I know of, yours is the first complete implementation I’ve seen. This is going to make alternative implementations much more difficult to build and, worst of all, small differences in the size estimation will lead to bugs that are quite difficult to find until they occur very occasionally. This is yet another reason we recommend that inlining be done in CBOR using unions that we can document in IPLD Schema because it makes these variations much more explicit and easier to implement and the rules for inlining can often avoid encoding to estimate the size. Sure, they’ll also need to implement data inlining rules that match the spec, but where they need to do so is predictable and it doesn’t need to be done in any of the CID abstractions.

These issues are all solvable in the code **we** write, but in an ecosystem this is just totally unmanageable. The vast majority of the time you can treat a CID as a key and pass it off to be resolved and looked up in a variety of abstractions. Adding this edge case that violates those, rather reasonable, assumptions is bound to break less sophisticated and mature libraries.

IPLD has a lot of, let’s say “casual,” implementations. These can read and write data with a very minimal level of support for various codecs and hash functions because you as long as you can create representations of most IPLD data you can do simple things without a larger stack. We just don’t have the ability to ensure that all of these handle edge cases like this and even though it’s technically allowed in the spec, that part of the spec already lives in a grey area that isn’t widely implemented enough to be relied upon. This is especially true as we migrate to `dag-cbor` because it’s actually much **more efficient** to inline the node values into CBOR so that’s the path we expect most people to take in the future.

You can see how much work this potentially is for people by just looking at what it took for you to move a bunch of code around to hit a unified code path that does the inlining and you still have a bunch of places across lotus you mention need to be updated. Expect as much or more work for anyone who writes an alternative implementation or wants to write a new library to work with the data without pulling in all of an existing Filecoin implementation.

This isn’t a problem we can solve very easily for other people, the work you’re doing to support this will end up needing to be replicated by every developer who wants to even casually work with this data because we already know from experience that the ecosystem people are building upon doesn’t support this.

> ~12% of our blocks are less than or equal to 32 bytes. I don't know how frequently these small blocks are accessed, but I assume very frequently given some of our data structures.
> The problem is that we have variable size data structures. I agree that using unions would be a better approach, but defining consistent rules for this is well beyond what we can support/do right now.

I was going to ask how predictable the locations are where this occurs but it sounds like the answer is “not very predictable” which would make the union approach more difficult than usual.

Stebalien:
Ah, sorry, I wasn't asking for something that in-depth. I just wanted to get on the same page and avoid devolving the conversation into "because reasons".

> All code that passes a CID off to these abstractions may need to be updated to consider inline CID’s, depending on the storage/service abstraction. If they don’t consider them (they mostly don’t) then the block storage and block service abstractions need to consider them and return either block data or a Block abstraction.

I think this is one of the key misunderstandings. Extracting blocks from the identity multihash is an optimization. If you can't do that, you can always get the block the normal way. Using the identity hash function just means:

1. CIDs can be _smaller_.
2. We can avoid hashing.
3. Implementations that support it can save some disk space and a few datastore reads/writes.

> If they return data this presents another problem, you need to make sure that you also continue to inline the data or else you’ll mutate the parent block as you roundtrip through these code paths even if you intended to make no changes.

This is the crux of the matter: choosing the correct hash function on write and update.

In go, we solve the write case by passing around a "CID builder" (well, we define a global one). We need to do this _anyways_ to specify the hash function and codec.

On the other hand, I agree that _maintaining_ the choice of hash function on update is tricker. That is, I can't just look at the old CID, the new object, then serialize the new object and generate a new CID based on the previous CID. Instead, I _also_ need to know the inlining rules. This isn't _too_ hard to resolve, but it's non-trivial.

Filecoin gets around this by simply requiring that all messages, blocks, state tree objects, etc. follow the same exact rules for generating CIDs to ensure that the output state is deterministic. Furthermore, I think any system that _really_ cares about consistent encoding/CIDs is going to define similar rules.

---

At the end of the day, I don't see any hard blockers to using the identity hash function for small objects as long as it's sufficiently motivated. It all comes down to a performance/complexity trade off and I don't want to just say no out the door because it slightly increases complexity (unless the performance boost isn't worth it).

mikeal:
> I think this is one of the key misunderstandings. Extracting blocks from the identity multihash is an optimization. If you can't do that, you can always get the block the normal way. Using the identity hash function just means:

What is “get the block the normal way?”

There are many block stores that are rather naive and will fail to return data from an inline CID because all they do is use the CID as a key and look up that key in a key/value store. But of course the value isn’t there, because the value is actually **in the key**.

This is why I don’t think this can be characterized as “just an optimization.” The way CID’s are used changes slightly when you consider this use case and most implementations simply can’t handle it. They can represent the CID but they fail to get the data.

> CIDs can be smaller.
> We can avoid hashing.
> Implementations that support it can save some disk space and a few datastore reads/writes.

This is all even more true of inlining the value into CBOR and avoiding creating the CID entirely. We do this pretty often in new IPLD data structures using unions.

It would be nice to know how many `Link` values in these data structures actually end up linking to blocks this small and if it’s realistic to use a union instead. 12% is a lot of blocks is quite a lot but that could potentially only be a few places in the data structure, or it could be that using a union in just a few locations cuts that percentage way down.

Stebalien:
> What is “get the block the normal way?”

Fetch it from the blockstore? A CID that uses the identity hash function is no different from any other CID.

> There are many block stores that are rather naive and will fail to return data from an inline CID because all they do is use the CID as a key and look up that key in a key/value store. But of course the value isn’t there, because the value is actually in the key.

What I'm saying is that the data _will_ be there. Either:

1. The blockstore supports extracting blocks from CIDs constructed with the identity hash function. This kind of blockstore can skip the load/store step for such CIDs.
2. The blockstore doesn't support this and will store the blocks normally. This kind of blockstore _will_ store the block, and return it later.

> This is all even more true of inlining the value into CBOR and avoiding creating the CID entirely. We do this pretty often in new IPLD data structures using unions.

We simply don't have the (stable, production-ready) go-ipld tooling to do this.

Also, if you want to talk about complexity, supporting inline CIDs is trivial next to supporting kinded unions based on size.

mikeal:
> The blockstore doesn't support this and will store the blocks normally. This kind of blockstore will store the block, and return it later.

You’re assuming a specific order of operations in which there is a block storage phase **before** this block data is ever requested, but that’s not necessarily the case. Since these only every appear as values in other blocks they **are not blocks themselves** and will likely skip any previous storage phase. For example, if you load a CAR file into a blockstore there are no block entries for the indentity CID’s. They only ever appear as values inside of other blocks, so they’ll never be stored and when they are requested from this type of block store it will simply say it doesn’t have it and when you then try to ask a network service or another node for the CID it also may not have it since these aren’t stored as unique blocks in their block storage either.

These CID’s aren’t indexed by most (maybe all) block storage systems, so if you’re unaware of Identity CID’s and can derive the block data from the key there’s actually no way it will retrieve it.

> Also, if you want to talk about complexity, supporting inline CIDs is trivial next to supporting kinded unions based on size.

You’re just moving the complexity outside of where we can reliably abstract it. There’s nothing in any IPLD library that can capture something along the lines of “inline block values under X size.” That complexity is going to exist in every application stack that works with this data.

It’s only trivial for **you** to make this change in **your** code, but communicating this to every implementation and ensuring consistency is a much harder ecosystem coordination change than the code change being proposed.

mikeal:
I think I’ve exhausted the list of reasons I don’t recommend this change. If you want to ignore the expertise of me and my team you’re welcome to.

Using identity multihashes to inline data is not a very well trafficed area of IPLD/multiformats. This **will** cause bugs in other implementations if not our own and you’re going down a very untested path of our stack at a very late stage. If we’re trying to move towards a more stable code base, this is me telling you that this is not that.

anorth:
Thank you for your extensive input, @mikeal. If we make this change, we won't be ignoring your expertise. Your expertise has been invaluable in clarifying one side of a tradeoff that we have to make. As with many other optimisations, it seems to essentially come down to an optimisation that can't be easily abstracted, and so enjoying it requires code to know about it. This is a more complex issue in an open source community, involving code and participants over which we have only indirect influence.

This represents a pretty significant optimization for Filecoin. State encoding and access is a high (highest?) impact bottleneck on chain evaluation throughput. It's not just the blocks that no longer need to be stored, although that alone would probably justify it. The Filecoin state tree is going to be huge (the largest IPLD structure in existence some day?). Avoiding the network transfer of GB's of inlined blocks is really significant. The reduced memory pressure on block store caches will further improve processing times even for access of non-inlined blocks. The reduced hashing load probably significant too. While we don't have hard data, both @Stebalien and my instinct is that many of these small blocks will be near the top of the frequently-accessed leaderboard.

_For Filecoin_, I think this is a big deal. As you point out, it's not difficult for us to make this change in Filecoin and projects in the _Filecoin_ ecosystem today. You point out that there are other implementations of IPLD elsewhere, of varying quality, and they may need change in order to correctly work with the Filecoin state tree as IPLD. I don't think that the existence of this other software, which probably doesn't care about Filecoin (yet) is a strong enough reason to impose this performance cost on Filecoin.

To remove the path dependency here, we might also consider if we had already made this change, built it into Filecoin from day 1. How strong would the case be for removing this optimization, adding 10s of GiB of extra weight to the Filecoin blockchain so that software that doesn't care about Filecoin could hypothetically work with the state tree without modification? 

Obviously this abstraction leakage can be taken too far, eventually at the cost of any benefits of a standard abstraction at all. But in order to gain wide and growing adoption, we need to accept a pragmatic balance here. Neither perfect abstraction nor optimal performance are going to be the right choice.

Stebalien:
Before we can continue discussing whether or not Filecoin should use the block inlining optimization, we need to be on the same page. Please carefully consider what I'm saying, your understanding of how "inlined" blocks interact with the rest of IPLD is incorrect.

> You’re assuming a specific order of operations in which there is a block storage phase before this block data is ever requested, but that’s not necessarily the case. Since these only every appear as values in other blocks they are not blocks themselves and will likely skip any previous storage phase. For example, if you load a CAR file into a blockstore there are no block entries for the indentity CID’s.

You sure? Looking at the go-car code, I don't see any special handling for CIDs using the identity hash function. It'll treat these CIDs like any other CID, and add the associated blocks to the CAR.

> These CID’s aren’t indexed by most (maybe all) block storage systems, so if you’re unaware of Identity CID’s and can’t derive the block data from the key there’s actually no way it will retrieve it.

A storage system is either aware of identity CIDs or not aware of identity CIDs. If it's not aware of identity CIDs, the identity hash function is _just_ another hash function.

* If I send someone a CID created with the identity hash function over bitswap, they'll send me back the associated block just like any other block.
* Walk a dag and/or create a CAR with CIDs using the identity hash function, I'll include the associated blocks as normal blocks.
* If I send someone a graphsync request for a DAG, they'll send me the blocks referenced by the identity hash function, just like any other block.
 
> You’re just moving the complexity outside of where we can reliably abstract it. There’s nothing in any IPLD library that can capture something along the lines of “inline block values under X size.” That complexity is going to exist in every application stack that works with this data.

I'm not sure what to make of this statement.

* Go's current IPLD libraries capture this with CID builders.
* go-ipld-prime captures this with the LinkBuilder type.
* rust-ipld sort-of supports it with a custom "multihasher" implementation (a bit hacky).
* js-ipld doesn't because it takes a pre-defined multihash codec. It could just as easily take a hasher function.

mikeal:
I think I see the problem Steven, you’re assuming that the “data” portion of a block for an inlined CID is the data, but that’s actually leveraging an inefficiency in these libraries. The “data” associated with an inline CID is actually *nothing*. There’s no need to actually store a value because the value is in the key and that’s how a lot of implementations will treat it.

Even your CAR file example, this is actually completely unspecified right now. You could either 

1) Not have any block entries for inline CIDs
2) Have CID entries with zero bytes of data
3) Have CID entries with the inlined data duplicated

All of these will produce a CAR file with the complete graph. You may have consistent behavior in the Go implementation here but the JS implementation is completely dependent on what the Block store does and could do either 2 or 3. The Block implementation doesn’t understand inline CID’s so it’s going to go with whatever the storage layer decides was the encoded data value.

What is the right behavior? What are you prioritizing? Bytes used? Potential bugs in naive block stores? I’m sure we’ll be discussing those tradeoffs for a while in the spec (thanks for surfacing this, I’ll followup in the specs repo) because while this is unspecified we have an issue with determinism.

Maybe you’re starting to understand why I’m concerned with relying on something underspecified and untested. We don’t have any code that tests this and we don’t know what this will do in a lot of our own libraries, much less what it’ll do in the rest of the ecosystem.

> This represents a pretty significant optimization for Filecoin. State encoding and access is a high (highest?) impact bottleneck on chain evaluation throughput. It's not just the blocks that no longer need to be stored, although that alone would probably justify it. The Filecoin state tree is going to be huge (the largest IPLD structure in existence some day?). Avoiding the network transfer of GB's of inlined blocks is really significant. The reduced memory pressure on block store caches will further improve processing times even for access of non-inlined blocks. The reduced hashing load probably significant too. While we don't have hard data, both @Stebalien and my instinct is that many of these small blocks will be near the top of the frequently-accessed leaderboard.

This is good context, thanks for this.

If it’s this important, I’d recommend taking the time to get it right and use proper, predictable, unions instead. It’ll make the data structure easier to implement and understand and definitely easier to test.

> To remove the path dependency here, we might also consider if we had already made this change, built it into Filecoin from day 1. How strong would the case be for removing this optimization, adding 10s of GiB of extra weight to the Filecoin blockchain so that software that doesn't care about Filecoin could hypothetically work with the state tree without modification?

If you had that much time with the feature already being used in the wild I’d be less concerned, a fair number of people would have already worked with data that does this and other implementations would have factored this in early.

This seems to be getting treated like a relatively minor change. It’s not. This is a huge change and alters the data structure literally everywhere you have a link. The cases in which you *may* inline or *may not* inline are everywhere you have a link which makes understanding how often this will be used and how many consumers/producers of those portions of the data structure very difficult to understand. Again, this is why we recommend using unions to inline data instead.

We don’t have a complete understanding of the bugs this can cause. It’s a very risky change, so if you do it just be aware of the risk. My frustration came from feeling as though these risks are being dismissed. These aren’t abstract problems, I’m not talking about some code that people *may* have written not playing nicely with this, I’m literally talking about code I’ve written not playing nicely with this 😁

We’ve been pointing people away from this feature for a while, including in a thread Steven was in just a few weeks ago about identity max length, so I was quite concerned to see it not only being picked up for use but being picked up during a late stage of development in which risks are trying to be mitigated.

I’ll also point out that this is a major change all implementations will need to make in order to produce consistent data and I haven’t seen anything proposed in the spec yet, although I may have missed it given how many notifications I’ve gotten lately. I’m not even sure *how* to write this into the spec since it effects everywhere there’s a link. It may also be prudent to check-in with any other implementations to understand how difficult this change will be for them to make as well.

warpfork:
I'll dip in here just briefly to try to distill a couple of things that are on my mind.  But ultimately, the choice is still filecoin's, so I'll try to keep ink spilled to a minimum.  (And you can read this all as a draft of a very "why the IPLD team is very cautious of inline CIDs"

---

>  it seems to essentially come down to an optimisation that can't be easily abstracted, and so enjoying it requires code to know about it

^ :100: 

---

I'm not really going to touch any of the earlier discussion about block storage abstractions.  (I think it's true that the CidBuilder here, and presumptively true that LinkBuilder in ipld-prime, etc, can handle this.  I also think it's probably true that there will under-exercised corners around this elsewhere.)

But there are a couple of other trains of thought I'd want to follow through this choice process too:

**Angle 1**: (reserved for the existing discussion of "under-exercised corners may abound near inline CID features")

**Angle 2**: Are there any concerns about the determinism of this choice, for tree hash consistency, and strictness of verification?

(I don't know.)

**Angle 3**: varying CID parameters, programmatically, throughout a tree, is annoying.  It means applications need to have detailed logic for fiddling with all this; and most of our other IPLD related tooling can't meaningfully _help_ you anymore (for example: schemas are purposefully generic over links -- meaning: they _do not specify_ CID parameters, either for link creation or for strict validation.  So the more complexity you have for this, the more _your application has to verify that_.)

This isn't to say you can't do varying CID parameters throughout a data forest.  Of course you can!  But it doesn't generally make anything *easier*.  As a result, it's something where we'd advise people: "default, avoid".

I think @anorth 's comment earlier about unabstractable optimisation covers this already; I just really want to double down here (for the benefit of future readers, who may also be outside of the filecoin project and looking here for patterns to copy) that in the IPLD team, we'd probably consider varying parameters throughout a data forest to be "not a best practice" for people designing new systems, because of the complexity add.

**Angle 4**: "inline" CIDs specifically are something we have very mixed feelings about in IPLD land.  _Every time we've considered doing something with an inline CID, it's turned out to be easier, lower-overhead, and usually even a more accurate description of the data semantics to **do something else**_: namely, a kinded union.

One way to probe at this is: did you *want* there to be an opportunity to switch codec in the subsequent data?  (It's not a perfect question, but it's maybe an interesting heuristic to check.)

Another thing that we've found we tend to like about the kinded union approach is that it means when we write Schemas for data, the Schema then speaks very clearly about how there's a choice between whether data is present in the same block, vs potentially being a link.  That's data we've just found is... nice to have, and nice to be able to discuss explicitly in design documents.  It communicates the complexity remarked on in Angle 3 more clearly, in other words.

I realize unions (aka sum types, aka variants, aka etc) are painful to express in golang, and involve a harrowing amount of keystrokes because (I don't know how to pussyfoot around this) Golang is a ~bad~ perfectly-nice-but-has-warts language that doesn't have a built-in concept sum types (and it should).

But I think it's really important to be conscious and cautious of how much golang brainworms might be affecting how filecoin models and designs the serial data.

If the data semantics in parts of filecoin really do desire things to be CIDs For Reasons (that aren't golang-is-bleg-at-types reasons), then by all means, move forward and disregard this consideration.  It'll just be rather a first in my database of priors (and I'll note it with interest).  And I'd be rather surprised that doing it as a blanket treatment across all CIDs built in the codebase is the way to go about it.

**Angle 5:** a union of bytes-or-CID is smaller than an inline CID.

An inline CID, in CBOR, still uses:

- a byte to say "CBOR tag",
- a byte to say 42 (it's too big of a number to bitpack into the tag-indicator byte),
- a byte to say "CBOR bytes",
- probably another byte to say how long (not many inline CIDs will be small enough to bitpack into the previous byte)
- a byte for CID version (only now did we get to CID content!),
- a byte for CID codec (you can switch codec here, after all),
- a byte for CID multihash (saying "identity"),
- a byte (or two) for CID length
- finally, the content

That's... 7 or 8 bytes of overhead, typically?  Did I count that right?

A byte segment, in CBOR, uses:

- a byte to say "CBOR bytes",
- one **or zero** or two bytes for length (*might* be bitpacked into the bytes-indicator byte, if it's small enough),
- the content

That's 1-3 bytes that aren't pure content.

Inline CIDs add significantly *more* bytes of overhead, if what you want is really just bytes.

(Note that if we're using a [kinded union](https://specs.ipld.io/schemas/representations.html#union-kinded-representation), there's no additional overhead to consider outside of this, either: whether the whole field is a CID or directly bytes is determined by that first kind-indicator byte (does it say "tag" or does it say "bytes"?).)

If you're doing more complex content than bytes, the general pattern here still holds.

I also suspect inline CIDs are probably rather more costly in execution time, if they contain anything other than raw bytes -- it'll be necessary to start a whole new unmarshal function call tree to handle them.  I don't have an estimate of how costly this is, in practice, but it's probably not zero.

---

In summary: there are a *lot* of reasons why we've become increasingly cautious of recommending inline CIDs while working on other parts of IPLD.  None of them are "hard no"'s.  But it's often turned out that when we really questioned if they were providing the value that we wanted out of them in various situations, well...

Alright, I've said my bit.  I'll probably let it rest here, and assume whatever choices follow are well-informed and well-intentioned.  Peace be with you :pray:

warpfork:
(Warpfork-uses-language-oddly note: "brainworms" isn't meant to have negative valence -- I use "brainworms" roughly substitutably with "memetic influence".)

Stebalien:
Ok, numbers:

For metadata, pretty much everything is less than 64 bytes. But metadata isn't a huge concern for performance.

For chain data, by % put/get operations for a "chain sync":

* PUT:
  * 10% <= 32 bytes
  * 51% <= 64 bytes
* GET:
  * 23% <= 32 bytes
  * 52% <= 64 bytes

So inlining up to 64 bytes (there may be a more reasonable bound but I don't currently have that data) would save us up to 50% of storage operations. Note: I say "up to" because inlining blocks > 32 bytes will change the size of other blocks.

Raw data courtesy @travisperson https://gist.github.com/travisperson/49dfc0e7f5cf786c426c81ba00d339fe.

whyrusleeping:
@Stebalien seems like we should do 64 bytes

actors/abi/cid.go
```diff
@@ -10,9 +10,29 @@ const (
...
+	if len(data) <= HashLength {
+		hf = mh.IDENTITY
+	}
+	return cid.V1Builder{Codec: b.codec, MhType: hf}.Sum(data)
+}
+
+// CidBuilder is the default CID builder for Filecoin.
+//
+// - The default codec is CBOR. This can be changed with CidBuilder.WithCodec.
+// - The default hash function is 256bit blake2b when the data is > 32 bytes
```
ribasushi, line 36: You are changing CIDs across the board with this PR. Doesn't it make sense to ditch `blake2b` as well, since it was selected before `sha2-ni` was a thing?

Stebalien: It might. Thoughts @whyrusleeping?

ribasushi: @whyrusleeping @Stebalien If nothing else this will shave 2 bytes from every CID:
[`0xA0 0xE4 0x02`](https://gchq.github.io/CyberChef/#recipe=From_Base(16)VarInt_Encode()To_Hex('Space',0)&input=MHhiMjIwCgo) vs `0x12`

Stebalien:
This is now blocked on determining the impact on downstream projects.

Stebalien:
I'm pushing through https://github.com/filecoin-project/specs-actors/pull/925 to enable further testing in lotus without having to change the defaults yet. We'll enable this if it materially impacts block validation, but any changes to the hash function will have a large impact on downstream projects as all message CIDs will change.

ZenGround0:
This PR is useful for documentation of thinking and debate around this design but it is not in a place where I expect it to be merged in the near to medium term.  If we ever decide to move forward with this change we will need to open a FIP.  If that happens I will happily reopen this PR for more visibility. Until then let's keep it closed.

ribasushi:
@ZenGround0 @anorth @Stebalien @arajasek  in light of the chain scalability conversations - should this be reconsidered?

The change is disruptive, yes, but is also extremely easy technically and rather low-risk: we now have a very long history of successful chain format changes.

ZenGround0:
Totally open to bringing this back into consideration

ZenGround0:
We should bring discussion to an issue before reopening this PR.

ribasushi:
@ZenGround0 a separate part of this discussion is [blake2 => sha256, saving 2 bytes per cid](https://github.com/filecoin-project/specs-actors/pull/783#discussion_r466755499). Not sure if worth another ticket, or too small of a saving.

PR #785, opened 2020-07-24
Balance table treats absent entries as zero, never stores zeros

anorth:
This is motivated by #784. I think this is a much more robust way of cleaning up empty accounts.

Minor conflicts with #775, but please let this land first.

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -287,8 +287,7 @@ func TestMarketActor(t *testing.T) {
 			expectedAmount := abi.NewTokenAmount(20)
 			actor.withdrawClientBalance(rt, client, withdrawAmount, expectedAmount)
 
-			rt.GetState(&st)
-			assert.Equal(t, abi.NewTokenAmount(0), actor.getEscrowBalance(rt, client))
+			actor.assertAccountZero(rt, client)
 		})
 
 		t.Run("worker withdrawing more than escrow balance limits to available funds", func(t *testing.T) {
@@ -303,8 +302,7 @@ func TestMarketActor(t *testing.T) {
 			actualWithdrawn := abi.NewTokenAmount(20)
 			actor.withdrawProviderBalance(rt, withdrawAmount, actualWithdrawn, minerAddrs)
 
-			rt.GetState(&st)
-			assert.Equal(t, abi.NewTokenAmount(0), actor.getEscrowBalance(rt, provider))
+			actor.assertAccountZero(rt, provider)
 		})
 
 		t.Run("balance after withdrawal must ALWAYS be greater than or equal to locked amount", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


actors/builtin/market/market_test.go
```diff
@@ -1612,6 +1611,21 @@ func TestCronTickDealExpiry(t *testing.T) {
...
+
+		// move the current epoch so that deal is expired
+		rt.SetEpoch(startEpoch + 1000)
+		actor.cronTick(rt)
+		require.EqualValues(t, deal.ClientCollateral, actor.getEscrowBalance(rt, client))
+
+		// client withdraws collateral -> account should be removed as it now has zero balance
+		actor.withdrawClientBalance(rt, client, deal.ClientCollateral, deal.ClientCollateral)
+		actor.assertAccountZero(rt, client)
+	})
```
anorth, line 1628: From #719

actors/builtin/market/market_test.go
```diff
@@ -1715,8 +1729,19 @@ func TestCronTickDealSlashing(t *testing.T) {
...
+						// client balances should not change
+						cLocked := actor.getLockedBalance(rt, client)
+						cEscrow := actor.getEscrowBalance(rt, client)
+						actor.cronTick(rt)
+						require.EqualValues(t, cEscrow, actor.getEscrowBalance(rt, client))
+						require.EqualValues(t, cLocked, actor.getLockedBalance(rt, client))
+					} else {
+						// running cron tick again dosen't do anything
+						actor.cronTickNoChange(rt, client, provider)
+					}
```
anorth, line 1744: From #719

anorth:
FYI @wadealexc

actors/util/adt/balancetable.go
```diff
@@ -98,17 +74,17 @@ func (t *BalanceTable) SubtractWithMinimum(key addr.Address, req abi.TokenAmount
...
 	if err != nil {
 		return err
 	}
-	if !subst.Equals(req) {
-		return errors.New("couldn't subtract the requested amount")
+	if req.GreaterThan(prev) {
+		return xerrors.New("couldn't subtract the requested amount")
 	}
-	return nil
+	return t.Add(key, req.Neg())
```
Stebalien, line 87: This can end up storing zero. Should we delete in this case?

Stebalien:
I assume you meant to store a zero balance, if previously non-zero. What was the actual problem with deleting zero balances?

actors/util/adt/balancetable.go
anorth: Yes!

anorth:
@Stebalien I did not! Good catch, fixed.

The problem previously was that various call sites required the value to be present, but since zeros were not stored, failed if they logically expected a zero to be stored (e.g. b/c deal price was zero).

Stebalien:
> The problem previously was that various call sites required the value to be present, but since zeros were not stored, failed if they logically expected a zero to be stored (e.g. b/c deal price was zero).

Ah, I see. So this is just a smaller patch that pushes that "zero" to the edges. Makes sense.

aarshkshah1992:
@anorth Can we get this in now ? Will unblock #775 .

PR #786, opened 2020-07-24
Central CID builder

Stebalien:
This is an alternative to https://github.com/filecoin-project/specs-actors/pull/783 which doesn't actually change the state structure.

Instead, this PR:

1. Adds proper support for inline CIDs (but doesn't actually crate them) so we can benchmark them without having to keep this patch lying around out-of-tree.
2. Centralizes the CID builder logic so we can change it easily.

### DIFF
--- a/actors/builtin/market/deal.go
+++ b/actors/builtin/market/deal.go
@@ -72,23 +72,11 @@ func (p *DealProposal) ProviderBalanceRequirement() abi.TokenAmount {
 }
 
 func (p *DealProposal) Cid() (cid.Cid, error) {
-	buf := new(bytes.Buffer)
-	if err := p.MarshalCBOR(buf); err != nil {
+	var buf bytes.Buffer
+	if err := p.MarshalCBOR(&buf); err != nil {
 		return cid.Undef, err
 	}
-	b := buf.Bytes()
-
-	mhType := uint64(mh.BLAKE2B_MIN + 31)
-	mhLen := -1
-
-	hash, err := mh.Sum(b, mhType, mhLen)
-	if err != nil {
-		return cid.Undef, err
-	}
-
-	c := cid.NewCidV1(cid.DagCBOR, hash)
-
-	return c, nil
+	return abi.CidBuilder.Sum(buf.Bytes())
 }
 
 type DealState struct {
--- a/actors/builtin/market/deal.go
+++ b/actors/builtin/market/deal.go
@@ -72,8 +72,8 @@ func (p *DealProposal) ProviderBalanceRequirement() abi.TokenAmount {
 }
 
 func (p *DealProposal) Cid() (cid.Cid, error) {
-	var buf bytes.Buffer
-	if err := p.MarshalCBOR(&buf); err != nil {
+	buf := new(bytes.Buffer)
+	if err := p.MarshalCBOR(buf); err != nil {
 		return cid.Undef, err
 	}
 	return abi.CidBuilder.Sum(buf.Bytes())
--- a/support/mock/mockrt.go
+++ b/support/mock/mockrt.go
@@ -12,7 +12,6 @@ import (
 	"github.com/filecoin-project/go-address"
 	addr "github.com/filecoin-project/go-address"
 	cid "github.com/ipfs/go-cid"
-	mh "github.com/multiformats/go-multihash"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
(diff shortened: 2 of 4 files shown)


Stebalien:
Basically, once I update https://github.com/filecoin-project/lotus/pull/2564 to use this instead of #783:

1. Lotus will fully support inline CIDs, if/when we choose to enable them. At the moment, lotus sort-of supports them in some places, which is not where we want to be.
2. We will be able to change the hash function and/or use inline CIDs really easily.

actors/builtin/market/deal.go
```diff
@@ -72,23 +72,11 @@ func (p *DealProposal) ProviderBalanceRequirement() abi.TokenAmount {
 }
 
 func (p *DealProposal) Cid() (cid.Cid, error) {
-	buf := new(bytes.Buffer)
-	if err := p.MarshalCBOR(buf); err != nil {
+	var buf bytes.Buffer
+	if err := p.MarshalCBOR(&buf); err != nil {
```
anorth, line 76: Aside: the miner actor alone demonstrates three different ways of doing this. We had a problem early on where I propagated a broken pattern, though I think it was yet a fourth way. If this way is good, let's use it consistently.

- `var buf bytes.Buffer`
- `buf := bytes.Buffer{}`
- `buf := new(bytes.Buffer)`

support/mock/mockrt.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/filecoin-project/go-address"
 	addr "github.com/filecoin-project/go-address"
 	cid "github.com/ipfs/go-cid"
-	mh "github.com/multiformats/go-multihash"
+	"github.com/multiformats/go-multihash"
```
anorth, line 15: Restore the alias, or alias as `multihash` explicitly since it doesn't match path.

support/mock/mockrt.go
```diff
@@ -385,15 +379,47 @@ func (rt *Runtime) checkArgument(predicate bool, msg string, args ...interface{}
...
+	prefix := c.Prefix()
+	if prefix.Codec != cid.DagCBOR {
+		rt.Abortf(exitcode.SysErrSerialization, "tried to fetch a non-cbor object: %s", c)
+	}
+
+	var data []byte
+	if prefix.MhType == multihash.IDENTITY {
+		decoded, err := multihash.Decode(c.Hash())
+		if err != nil {
+			rt.Abortf(exitcode.SysErrSerialization, "failed to parse cid %s: %s", c, err)
```
anorth, line 394:
```suggestion
			rt.Abortf(exitcode.SysErrSerialization, "failed to parse identity cid %s: %s", c, err)
```

support/mock/mockrt.go
```diff
@@ -385,15 +379,47 @@ func (rt *Runtime) checkArgument(predicate bool, msg string, args ...interface{}
...
+		data = stored
+	} else {
+		return nil, false
+	}
+	return data, true
+}
+
+// Puts raw data into the state, but only if it's not "inlined" into the CID.
+func (rt *Runtime) put(c cid.Cid, data []byte) {
+	// Store it only if we need to. Doing this in tests ensure
```
anorth, line 407: Dangling comment

support/testing/cid.go
```diff
@@ -1,32 +1,36 @@
...
+		panic(err)
 	}
 
-	return c
+	sum := sha256.Sum256(data)
+	hash, err := mh.Encode(sum[:], prefix.MhType)
+	if err != nil {
+		panic(err)
+	}
+	return cid.NewCidV1(prefix.Codec, hash)
```
anorth, line 35: FYI @magik6k

actors/builtin/market/deal.go
Stebalien: Well, they all work but I usually avoid the second case.

In this ase, I chose the `var` approach to make it clear that this doesn't escape to the heap. I guess `new` probably works just as well, but my brain always assumes "new means heap".

For now, I'll just revert this change.

support/mock/mockrt.go
Stebalien: Done. (although dropping the `go-` prefix is pretty common, IMO)

actors/builtin/market/deal.go
anorth: Me too, and the `var` approach is probably better. But the heterogeneity makes me have to think every time I see a different one.

PR #787, opened 2020-07-24
recalculate sector intial pledge

### LINKED ISSUE #522: Re-calculate sector initial pledge and defer activation until prove-commit
#516 implements the initial pledge calculation and moves it to `SectorPreCommit`, because that is the requested pre-commit deposit amount. The initial pledge requires knowing the sector epoch, so that's changed to be the pre-commit epoch. This effectively means that the sector's power is locked in at this point too. See the discussion and resolutions with a string of 👍 in #424.

However, the winds have changed and now there are requests to return sector activation epoch to prove-commit and re-calculate the power and pledge using the later activation epoch and updated block reward estimate.

From [Slack](https://filecoinproject.slack.com/archives/C015KQQLQQ1/p1592943826132100?thread_ts=1592865896.126300&cid=C015KQQLQQ1)
>  Ideally we would like to keep InitialPledge calculation in ProveCommit and PreCommitDeposit calculation a distinct concept in PreCommit. It's so happened that PreCommitDeposit uses the same formula as InitialPledge with the current PoRep. They might be different values when a new algorithm is used.
> Another complication here is that TerminationFee is also higher (since SectorAge is longer with an earlier ActivationEpoch). The main thing here is that Activation at ProveCommit is the cleanest conceptually, the sector has now been activated and its lifetime begins.

This issue tracks a decision about and implementation of re-calculating the pledge at prove-commit time.

acruikshank:
closes #522

### Motivation

Initial pledge actually does need to be calculated at prove commit time, and precommit deposit does not need to be as large as initial pledge.

### Proposed Changes

1. Introduce `PreCommitDepositForWeight` function in monies that depends on `ExpectedDayRewardForPower` but not on `InitialPledgeForPower` and delete the `precommitDeposit` function in miner_actor. Use the new function to compute a sectors `PreCommjitDeposit` in `PreCommitSector`.
2. Retrieve `baselinePower`, `epochReward`, `pwrTotal` and `circulatingSupply` at the beginning of `ConfirmSectorProofsValid` because we'll need them to compute initial pledge for sectors.
3. Use the `InitialPledgeForPower` function to compute the initial pledge for each sector. Set this in the `SectorOnChainInfo`. Also use it to compute the total pledge.
4. Sum `PreCommitDeposits` in `ConfirmSectorProofsValid` to unlock, since the `totalPledge` is no longer the correct value.
5. Change the sector's `Activation` to `rt.CurrEpoch()`. This keeps the total power calculations consistent with the total pledge calculations.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -458,10 +458,9 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	}
 
 	// gather information from other actors
-	baselinePower, epochReward := requestCurrentEpochBaselinePowerAndReward(rt)
+	_, epochReward := requestCurrentEpochBaselinePowerAndReward(rt)
 	pwrTotal := requestCurrentTotalPower(rt)
 	dealWeight := requestDealWeight(rt, params.DealIDs, rt.CurrEpoch(), params.Expiration)
-	circulatingSupply := rt.TotalFilCircSupply()
 
 	store := adt.AsStore(rt)
 	var st State
@@ -505,7 +504,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 
 		sectorWeight := QAPowerForWeight(info.SectorSize, duration, dealWeight.DealWeight, dealWeight.VerifiedDealWeight)
 		depositReq := big.Max(
-			precommitDeposit(sectorWeight, pwrTotal.QualityAdjPower, baselinePower, pwrTotal.PledgeCollateral, epochReward, circulatingSupply),
+			PreCommitDepositForPower(epochReward, pwrTotal.QualityAdjPower, sectorWeight),
 			depositMinimum,
 		)
 		if availableBalance.LessThan(depositReq) {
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -691,16 +696,21 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 
 		newSectorNos := make([]abi.SectorNumber, 0, len(preCommits))
 		for _, precommit := range preCommits {
-			// initial pledge is precommit deposit
-			initialPledge := precommit.PreCommitDeposit
+			// compute initial pledge
+			duration := precommit.Info.Expiration - rt.CurrEpoch()
```
anorth, line 700: Just so it's super-clear, I suggest setting an `ActivationEpoch := rt.CurrEpoch()` var here, using it for the calculation and setting in sector info below.

actors/builtin/miner/miner_actor.go
```diff
@@ -725,7 +736,7 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 		}
 
 		// Unlock deposit for successful proofs, make it available for lock-up as initial pledge.
-		st.AddPreCommitDeposit(totalPledge.Neg())
+		st.AddPreCommitDeposit(totalPrecommitDeposit.Neg())
```
ZenGround0, line 739: I think this is the right way to do it but flagging here that we could alternatively compute IPs in the loop above and silently drop prove commits that won't have enough funds to process.  That would get tricky because unlocking gives you some funds, but we could apply decent heuristics (PD < IP for example) to figure out if including a prove commit risks failing all others.

Sending more than enough funds is in the miner's control and acceptable way to counter this risk.

PR #789, opened 2020-07-24
Fix deadline index check

### LINKED ISSUE #753: Miner: Incorrect bounds on SubmitWindowedPoSt params.Deadline
This check is incorrect:

https://github.com/filecoin-project/specs-actors/blob/84de6c6a21972a2b03b2c6c0a80cfec0d83ed11c/actors/builtin/miner/miner_actor.go#L252-L254

Should be `>=`

Stebalien:
fixes #753

I'm punting on tests as I believe we need a framework for testing input bounds.


wadeAlexC:
LGTM!

PR #790, opened 2020-07-24
Payment: Remove actor for collect

### LINKED ISSUE #733: PayCh.Collect: Clarify whether Collect implies the channel should be terminated
Currently, `Collect` sets `st.ToSend` to 0, implying that the PaymentChannel may continue to be used. Further submission of higher-nonce vouchers via `UpdateChannelState` will continue to increment `st.ToSend` with the same behavior as before.

However, after calling `Collect`, `Settle` no longer functions properly due to `st.SettlingAt` being nonzero. Further updates to `UpdateChannelState` will allow either party to instantly `Collect` `st.ToSend`.

#### Recommendation

Clarify semantics of `Collect`: 

* If, after `Collect`, a channel should still be usable, set `st.SettlingAt` to zero.
* If `Collect` terminates the channel, consider nullifying state in some other way - like deleting the actor.

aarshkshah1992:
Closes #733 .

### DIFF
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -323,10 +323,14 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	)
 	builtin.RequireSuccess(rt, codeTo, "Failed to send funds to `To`")
 
+	// even though we will delete the actor, it dosen't hurt to set this here for sanity.
 	rt.State().Transaction(&st, func() interface{} {
 		st.ToSend = big.Zero()
 		return nil
 	})
+
+	rt.DeleteActor(builtin.BurntFundsActorAddr)
+
 	return nil
 }
 
--- a/actors/builtin/paych/paych_actor.go
+++ b/actors/builtin/paych/paych_actor.go
@@ -305,15 +305,6 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 		rt.Abortf(exitcode.ErrForbidden, "payment channel not settling or settled")
 	}
 
-	// send remaining balance to "From"
-	_, codeFrom := rt.Send(
-		st.From,
-		builtin.MethodSend,
-		nil,
-		big.Sub(rt.CurrentBalance(), st.ToSend),
-	)
-	builtin.RequireSuccess(rt, codeFrom, "Failed to send balance to `From`")
-
 	// send ToSend to "To"
 	_, codeTo := rt.Send(
 		st.To,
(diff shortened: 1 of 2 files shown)


wadeAlexC:
LGTM!

actors/builtin/paych/paych_actor.go
```diff
@@ -323,10 +323,14 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	)
 	builtin.RequireSuccess(rt, codeTo, "Failed to send funds to `To`")
 
+	// even though we will delete the actor, it dosen't hurt to set this here for sanity.
 	rt.State().Transaction(&st, func() interface{} {
 		st.ToSend = big.Zero()
 		return nil
 	})
+
+	rt.DeleteActor(builtin.BurntFundsActorAddr)
```
anorth, line 332: The send to `From` above is redundant, and leaves a possibility of actually burning some funds here if we mess something up later. 

How about replacing the Send to From with setting the From as the beneficiary on deletion here? The transaction setting balance to zero is definitely redundant then.

aarshkshah1992: @anorth Why is the send to `From` redundant ?

anorth: I mean it's redundant if we set From as the beneficiary upon deletion.

aarshkshah1992:
@anorth Please take a look.

actors/builtin/paych/paych_actor.go
aarshkshah1992: Done.

actors/builtin/paych/paych_actor.go
```diff
@@ -323,10 +314,8 @@ func (pca Actor) Collect(rt vmr.Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 	)
 	builtin.RequireSuccess(rt, codeTo, "Failed to send funds to `To`")
 
-	rt.State().Transaction(&st, func() interface{} {
-		st.ToSend = big.Zero()
-		return nil
-	})
+	rt.DeleteActor(st.From)
```
anorth, line 317: Great, just add a comment saying "return remaining balance to From upon deletion"

hannahhoward:
@anorth @aarshkshah1992 clarifying something here:

the line `rt.DeleteActor(st.From)` certainly appears to delete the actor with the address `From` which is not the payment channel actor address but rather the person sending money. I'm hoping from the comment request that some how `rt.DeleteActor(st.From)` somehow means `delete this actor that's running this code, but designate the funds to the actor at address From`.

This still seems quite unintuitive if this is the case. Is there documentation somewhere of how `DeleteActor` is supposed to work?

hannahhoward:
nm previous comment, resolved.

PR #794, opened 2020-07-24
set minimum sector expiration to 180 days

### LINKED ISSUE #734: Sector Lifetime bounds
We need to change the enforced sector lifetime bounds.  The maximum sector lifetime is already enforced but the parameter needs to change to 366 days.

We are not yet enforcing sector minimum lifetime but need to do so at precommit time.  The minimum sector lifetime parameter is 180 days.

acruikshank:
closes #734

### Motivation

We need to enforce both a minimum and a maximum sector lifetime.

### Proposed Changes

1. Introduce `MinSectorExpiration` of `180*EpochsInDay` to miner policy.
2. Add check the sector meets minimum expiration criteria in miner actor `validateExpiration`.
3. Fix miner actor tests so that sectors that are supposed to be valid expire after at least 180 days.
4. Add test for minimum sector bound.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -164,7 +164,8 @@ func TestCommitments(t *testing.T) {
 
 		// Make a good commitment for the proof to target.
 		sectorNo := abi.SectorNumber(100)
-		precommit := actor.makePreCommit(sectorNo, precommitEpoch-1, dlInfo.PeriodEnd(), nil)
+		expiration := dlInfo.PeriodEnd() + 181*miner.WPoStProvingPeriod // something on deadline boundary but > 180 days
+		precommit := actor.makePreCommit(sectorNo, precommitEpoch-1, expiration, nil)
 		actor.preCommitSector(rt, precommit)
 
 		// assert precommit exists and meets expectations
@@ -245,7 +246,8 @@ func TestCommitments(t *testing.T) {
 		assert.Equal(t, miner.NewPowerPairZero(), partition.RecoveringPower)
 
 		pQueue := actor.collectPartitionExpirations(rt, partition)
-		entry := pQueue[precommit.Expiration]
+		entry, ok := pQueue[precommit.Expiration]
+		require.True(t, ok)
 		assertBitfieldEquals(t, entry.OnTimeSectors, uint64(sectorNo))
 		assertEmptyBitfield(t, entry.EarlySectors)
 		assert.Equal(t, expectedInitialPledge, entry.OnTimePledge)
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -264,67 +266,71 @@ func TestCommitments(t *testing.T) {
...
 		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
-			actor.preCommitSector(rt, actor.makePreCommit(oldSector.SectorNumber, challengeEpoch, deadline.PeriodEnd(), nil))
+			actor.preCommitSector(rt, actor.makePreCommit(oldSector.SectorNumber, challengeEpoch, expiration, nil))
 		})
 		rt.Reset()
 
 		// Bad sealed CID
 		rt.ExpectAbortConstainsMessage(exitcode.ErrIllegalArgument, "sealed CID had wrong prefix", func() {
-			pc := actor.makePreCommit(101, challengeEpoch, deadline.PeriodEnd(), nil)
+			pc := actor.makePreCommit(102, challengeEpoch, deadline.PeriodEnd(), nil)
```
ZenGround0, line 289: Out of curiosity what motivated the change from 101 to 102?

acruikshank: A lot of these failure tests were failing because they were duplicate precommits instead of the intended failure. The change to 102 fixes that.

actors/builtin/miner/policy.go
```diff
@@ -108,10 +108,13 @@ var FaultMaxAge = WPoStProvingPeriod * 14
...
 const WorkerKeyChangeDelay = ChainFinality
 
+// Minimum number of epochs past the current epoch a sector may be set to expire.
+const MinSectorExpiration = 180 * builtin.EpochsInDay
+
 // Maximum number of epochs past the current epoch a sector may be set to expire.
 // The actual maximum extension will be the minimum of CurrEpoch + MaximumSectorExpirationExtension
 // and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
-const MaxSectorExpirationExtension = builtin.EpochsInYear
+const MaxSectorExpirationExtension = 366 * builtin.EpochsInDay
```
nicola, line 117: why was this changed?

ZenGround0: This is latest cryptoecon param recommendation.  cc @zixuanzh

acruikshank: The difference is that EpochsInYear is 365*builtin.EpochsInDay and the new spec says 366 days (leap-year upper bound). I don't know if this is a change motivated by real concerns or just an inconsistency in how various teams define a year, but I thought it best to stick to the spec.

Issue #627, opened 2020-07-08
Handle miners going bankrupt from fees


anorth:
The initial pledge locked by a miner doesn't necessarily cover the maximum fee for terminating all their sectors. Also it's possible for locked funds to decrease below the initial pledge requirement (though we could change that). We need to handle bankruptcy. In general the approach will be to claim all locked funds and de-register the miner permanently.

cc @zixuanzh do you have any detail to add?

Stebalien:
Unfortunately, we _currently_ just pay whatever fines we can and ignore the rest:

https://github.com/filecoin-project/specs-actors/blob/e5b62bf0f506ec6ea3899fa94dcd4dd701ad51e5/actors/builtin/miner/miner_actor.go#L2147-L2173

anorth:
Replaced by #888

PR #808, opened 2020-07-25
Feat/ip no vest

### LINKED ISSUE #757: Don't vest sector initial pledge, just lock it up for sector lifetime
At present, a sector's initial pledge is locked up in the vesting table, and unlocks and vests over some period that is independent of the sector's lifespan. Since it vests, there's no guarantee that the miner has locked-up value equal to the sum of their initial pledge requirements, so we have to add more checks for that explicitly anyway (#415).  This is very harsh for miners since their pledge can be locked up for longer than the sector lifetime, and the vesting configuration is coupled to our expectations of sector lifetime etc.

This is also a huge performance problem: [AddLockedFunds dominates state read/write activity](#502) on chain validation, because every sectors commitment write a bunch of vesting entries (and we expect #sectors > #blocks).

Compare with precommit deposit, which is locked up on precommit and released immediately on proof, no vesting.

Instead of vesting, we can just lock up the pledge directly, like precommit deposit: increment on provecomment and decrement on termination (we already record this number in miner state).

If not vesting, we have to determine a priority of coverage by funds, and how to process penalties, which previously always came from locked vesting funds (the precommit deposit is a separate bucket that cannot be tapped for penalties). I propose that a miner's balance is spoken for by, in order:
- pre-commit deposits, PCD
- vesting ("locked") funds, LF
- initial pledge, IP

As invariant, balance may never fall below the sum of PCD + LF w/o triggering [miner bankruptcy](#627). When vesting funds are penalised, the funds to burn are unvested first, until LF=0.

Upon penalisation [edit: see comments below with change to this]
- If balance exceeds PCD+LF+IP, burn funds until they are equal
- then while balance exceeds PCD+LF, burn funds, leaving IP un-covered
- then, while balance exceeds PCD, unlock and burn from LF (IP still un-covered)
- then when LF=0 and balance=PCD and a fee needs to be paid, bankrupt

This is in all cases penalising the funds that are most freely available, and hence valuable, to the miner.

While balance < PCD+LF+IP, a miner will be prevented from withdrawing or committing a new sector (#415). They can deposit new funds with a simple send until they cover IP again.

ZenGround0:
Closes #757 

# ⚠️  Design decisions please review ⚠️ 

- ~miner actor now tracks IP funds locked in state, InitialPledgeDeposits, as well as existing requirement number~
- ~reward (add locked funds) pays locked funds to InitialPledgeDeposits if below requirement~
- fees are paid first from unvested locked funds in the vesting table, then from IP
  - ~this PR does not draw from free balance -- but it could be modified to do so upon request~
- committed capacity upgraded sectors see their pledge released from deposit upon expiry.  Before this their pledge stayed in the vesting table until it vested 

# Progress
- [x] Track IP funds
- [x] Reward pays to pledge deposit during IP debt
- [x] Fees are paid out of IP when unvested funds exhausted
- [x] Tests fixed
- [ ] Test 
- [x] ack / nack from @anorth and maybe @zixuanzh on the above design decisions.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -385,11 +385,11 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 
 		// Note: We could delay this charge until end of deadline, but that would require more accounting state.
 		totalPenaltyTarget := big.Add(undeclaredPenaltyTarget, declaredPenaltyTarget)
-		availableBalance := st.GetUnlockedBalance(rt.CurrentBalance())
-		vestingPenaltyTotal, balancePenaltyTotal, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, totalPenaltyTarget, availableBalance)
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+		vestingPenaltyTotal, balancePenaltyTotal, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, totalPenaltyTarget, unlockedBalance)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty for %v", undeclaredPenaltyPower)
 		penaltyTotal = big.Add(vestingPenaltyTotal, balancePenaltyTotal)
-		pledgeDelta = vestingPenaltyTotal.Neg()
+		pledgeDelta = big.Sub(pledgeDelta, vestingPenaltyTotal)
 
 		// Record the successful submission
 		deadline.AddPoStSubmissions(partitionIdxs)
@@ -1415,9 +1415,12 @@ func (a Actor) AddLockedFund(rt Runtime, amountToLock *abi.TokenAmount) *adt.Emp
 		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 
-		availableBalance := st.GetUnlockedBalance(rt.CurrentBalance())
-		if availableBalance.LessThan(*amountToLock) {
-			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", availableBalance, *amountToLock)
+		// This may lock up unlocked balance that was covering InitialPledgeRequirements
+		// This ensures that the amountToLock is always locked up if the miner account
+		// can cover it.
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+		if unlockedBalance.LessThan(*amountToLock) {
+			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", unlockedBalance, *amountToLock)
 		}
 
 		if err := st.AddLockedFunds(store, rt.CurrEpoch(), *amountToLock, &RewardVestingSpec); err != nil {
@@ -1506,8 +1509,7 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 			rt.Abortf(exitcode.ErrIllegalState, "failed to vest fund: %v", err)
 		}
 
-		// Verify locked funds are are at least the sum of sector initial pledges after vesting.
-		// TODO: simplify this just to refuse to vest if pledge requirement is unmet https://github.com/filecoin-project/specs-actors/issues/537
+		// Verify InitialPledgeRequirement does not exceed unlocked funds
 		verifyPledgeMeetsInitialRequirements(rt, &st)
 
 		return newlyVestedFund
@@ -1613,8 +1615,8 @@ func processEarlyTerminations(rt Runtime) (more bool) {
 		// Unlock funds for penalties.
 		// TODO: handle bankrupt miner: https://github.com/filecoin-project/specs-actors/issues/627
 		// We're intentionally reducing the penalty paid to what we have.
-		available := st.GetUnlockedBalance(rt.CurrentBalance())
-		penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, rt.CurrEpoch(), penalty, available)
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+		penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, rt.CurrEpoch(), penalty, unlockedBalance)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock unvested funds")
 		penalty = big.Add(penaltyFromVesting, penaltyFromBalance)
 
@@ -1656,7 +1658,6 @@ func handleProvingDeadline(rt Runtime) {
 	hadEarlyTerminations := false
 
 	powerDelta := PowerPair{big.Zero(), big.Zero()}
-	newlyVested := big.Zero()
 	penaltyTotal := abi.NewTokenAmount(0)
 	pledgeDelta := abi.NewTokenAmount(0)
 
@@ -1666,8 +1667,9 @@ func handleProvingDeadline(rt Runtime) {
 		var err error
 		{
 			// Vest locked funds.
-			// This happens first so that any subsequent penalties are taken from locked pledge, rather than free funds.
-			newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+			// This happens first so that any subsequent penalties are taken
+			// from locked vesting funds before funds free this epoch.
+			newlyVested, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 			pledgeDelta = big.Add(pledgeDelta, newlyVested.Neg())
 		}
@@ -1690,7 +1692,7 @@ func handleProvingDeadline(rt Runtime) {
 		deadline, err := deadlines.LoadDeadline(store, dlInfo.Index)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d", dlInfo.Index)
 		quant := st.QuantEndOfDeadline()
-		availableBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
 
 		{
 			// Detect and penalize missing proofs.
@@ -1734,11 +1736,11 @@ func handleProvingDeadline(rt Runtime) {
 			penaltyTarget := PledgePenaltyForUndeclaredFault(epochReward, pwrTotal.QualityAdjPower, penalizePowerTotal)
 			// Subtract the "ongoing" fault fee from the amount charged now, since it will be added on just below.
 			penaltyTarget = big.Sub(penaltyTarget, PledgePenaltyForDeclaredFault(epochReward, pwrTotal.QualityAdjPower, penalizePowerTotal))
-			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, availableBalance)
+			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, unlockedBalance)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
-			availableBalance = big.Sub(availableBalance, penaltyFromBalance)
+			unlockedBalance = big.Sub(unlockedBalance, penaltyFromBalance)
 			penaltyTotal = big.Sum(penaltyTotal, penaltyFromVesting, penaltyFromBalance)
-			pledgeDelta = big.Sum(pledgeDelta, penaltyFromVesting.Neg())
+			pledgeDelta = big.Sub(pledgeDelta, penaltyFromVesting)
 
 			// Save modified deadline state.
 			if detectedAny {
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1753,8 +1755,10 @@ func handleProvingDeadline(rt Runtime) {
 			// Release pledge requirements for the sectors expiring on-time.
 			// Pledge for the sectors expiring early is retained to support the termination fee that will be assessed
 			// when the early termination is processed.
-			pledgeDelta = big.Sub(pledgeDelta, expired.OnTimePledge)
```
ZenGround0, line 1756: I believe this was just subtracting from zero before to get a negative directly, but please double check

zixuanzh:
> reward (add locked funds) pays locked funds to InitialPledgeDeposits if below requirement

AddLockedFunds should not count towards InitialPledgeRequirement since the funds are still locked and have not been earned yet. It is then consistent with the policy that LF is not counted as IPs. We have made it more lenient to allow unvested LockedFunds to be slashed first to avoid frequent top up. IP top up is required when once the LF buffer is exhausted.

zixuanzh:
> committed capacity upgraded sectors see their pledge released from deposit upon expiry. Before this their pledge stayed in the vesting table until it vested

I think the simplest for now is to require the Initial Pledge Requirement of the pre upgraded sector to be maintained until the upgraded sector expires. We can revisit this after landing the Termination Fee calculation for CC upgrades.

ZenGround0:
> I  think the simplest for now is to require the Initial Pledge Requirement of the pre upgraded sector to be maintained until the upgraded sector expires.

@anorth can you think of a way to implement this easily?  It appears to be a difficult change given my limited understanding of sector expiration.

actors/builtin/miner/miner_actor.go
anorth: Please revert. Intention is to accumulate a pledge delta over a number of different actions in this handler, in parallel with accumulating `penaltyTotal`. It just happens that this is the first and only contribution to the delta as the code stands now, but this is fragile if something affecting pledge is introduced further up.

actors/builtin/miner/miner_actor.go
```diff
@@ -1753,8 +1784,10 @@ func handleProvingDeadline(rt Runtime) {
...
-			st.AddInitialPledgeRequirement(pledgeDelta)
+			pledgeDelta = expired.OnTimePledge
+			st.AddInitialPledgeRequirement(pledgeDelta.Neg())
+			// Only record the amount we can actually unlock
+			pledgeDelta = st.UnlockInitialPledgeDeposits(pledgeDelta)
```
anorth, line 1790: Eeek this is getting hard to understand, I think due to the possible confusion between pledge requirements and coverage.

actors/builtin/miner/miner_actor.go
```diff
@@ -1798,7 +1831,7 @@ func handleProvingDeadline(rt Runtime) {
 	// Remove power for new faults, and burn penalties.
 	requestUpdatePower(rt, powerDelta)
 	burnFunds(rt, penaltyTotal)
-	notifyPledgeChanged(rt, big.Sum(newlyVested.Neg(), penaltyTotal.Neg(), pledgeDelta))
+	notifyPledgeChanged(rt, big.Sum(newlyVested.Neg(), penaltyTotal.Neg(), pledgeDelta.Neg()))
```
anorth, line 1834: This is breaking the convention of using "delta" to mean a signed quantity that is ready to be added to a total. My ability to understand what's going on is radically reduced.

anorth:
>>> committed capacity upgraded sectors see their pledge released from deposit upon expiry. Before this their pledge stayed in the vesting table until it vested

>> I think the simplest for now is to require the Initial Pledge Requirement of the pre upgraded sector to be maintained until the upgraded sector expires.

> @anorth can you think of a way to implement this easily? It appears to be a difficult change given my limited understanding of sector expiration.

No, I approve of the changed behaviour. The prior behaviour depended on how old the sector was at the time it was replaced (its pledge could have already 100% vested). Now the pledge is correctly tied to a sector's power-giving lifetime instead of being a fudge. The new sector has its own pledge, with a lower bound of the replaced sector's pledge. We should just release it.

zixuanzh:
> The new sector has its own pledge, with a lower bound of the replaced sector's pledge. 

Just to make sure we are on the same page, this is saying new sector pledge = max(replaced sector's pledge, new sector's pledge).

actors/builtin/miner/miner_actor.go
ZenGround0: done

ZenGround0:
> Just to make sure we are on the same page, this is saying new sector pledge = max(replaced sector's pledge, new sector's pledge).

Just realized we've been [doing the max all along](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L506) since @anorth built the cc feature into precommit !

FYI @zixuanzh

actors/builtin/miner/miner_actor.go
```diff
@@ -1512,6 +1515,7 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 
 	currBalance := rt.CurrentBalance()
 	amountWithdrawn := big.Min(st.GetAvailableBalance(currBalance), params.AmountRequested)
+	Assert(amountWithdrawn.GreaterThanEqual(big.Zero()))
```
anorth, line 1518: The comment at `verifyPledgeMeetsInitialRequirements` above is a bit misleading now. Also issue #537 will be redundant with this PR.

I see how this works. I think it would be a bit simpler to remove the `verifyPledgeMeetsInitialRequirements` and just explicitly abort (rather than assert) if available balance is <= zero (alt: amountWithdrawn == 0) here.

actors/builtin/miner/miner_actor.go
```diff
@@ -1654,12 +1662,14 @@ func handleProvingDeadline(rt Runtime) {
 
 	var st State
 	rt.State().Transaction(&st, func() interface{} {
+
 		var err error
 		{
 			// Vest locked funds.
 			// This happens first so that any subsequent penalties are taken from locked pledge, rather than free funds.
 			newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
```
anorth, line 1670: Remove `newlyVested` from the top-level scope so it can't be accidentally processed again.

The comment above this line is now misleading.

actors/builtin/miner/miner_actor.go
```diff
@@ -1723,9 +1734,11 @@ func handleProvingDeadline(rt Runtime) {
 			penaltyTarget := PledgePenaltyForUndeclaredFault(epochReward, pwrTotal.QualityAdjPower, penalizePowerTotal)
 			// Subtract the "ongoing" fault fee from the amount charged now, since it will be added on just below.
 			penaltyTarget = big.Sub(penaltyTarget, PledgePenaltyForDeclaredFault(epochReward, pwrTotal.QualityAdjPower, penalizePowerTotal))
-			penalty, err := st.UnlockUnvestedFunds(store, currEpoch, penaltyTarget)
+			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, availableBalance)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
-			penaltyTotal = big.Add(penaltyTotal, penalty)
+			availableBalance = big.Sub(availableBalance, penaltyFromBalance)
+			penaltyTotal = big.Sum(penaltyTotal, penaltyFromVesting, penaltyFromBalance)
+			pledgeDelta = big.Sum(pledgeDelta, penaltyFromVesting.Neg())
```
anorth, line 1741: Nit
```suggestion
			pledgeDelta = big.Sub(pledgeDelta, penaltyFromVesting)
```
?

Ok if you prefer consistently adding tho

actors/builtin/miner/miner_state.go
```diff
@@ -939,6 +941,26 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
 	return nil
 }
 
+// PenalizeFundsInPriorityOrder first unlocks unvested funds from the vesting table.
+// If the target is not yet hit it deducts funds from the (new) available balance.
+// Returns the amount unlocked from the vesting table and the amount taken from current balance.
+func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.ChainEpoch, target, availableBalance abi.TokenAmount) (fromVesting abi.TokenAmount, fromBalance abi.TokenAmount, err error) {
```
anorth, line 947: Blocking: rename `availableBalance` to `unlockedBalance`, or take `currentBalance` as parameter instead and compute unlocked internally.

Also:
```suggestion
func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.ChainEpoch, target, availableBalance abi.TokenAmount) (fromVesting, fromBalance abi.TokenAmount, err error) {
```

actors/builtin/miner/miner_actor.go
```diff
@@ -384,8 +385,11 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 
 		// Note: We could delay this charge until end of deadline, but that would require more accounting state.
 		totalPenaltyTarget := big.Add(undeclaredPenaltyTarget, declaredPenaltyTarget)
-		penaltyTotal, err = st.UnlockUnvestedFunds(store, currEpoch, totalPenaltyTarget)
+		availableBalance := st.GetUnlockedBalance(rt.CurrentBalance())
```
anorth, line 388: Blocking: don't confuse the naming of the concepts "available" and "unlocked". I had to re-read this whole PR after I got to the state file and realised these were critically distinguished.

Not blocking, but I think it would be better for `st.PenalizeFundsInPriorityOrder` to accept the `rt.CurrentBalance()` as a parameter rather than this value that is itself a function of state, which would make this whole naming confusion go away.

actors/builtin/miner/miner_actor.go
```diff
@@ -1608,11 +1613,14 @@ func processEarlyTerminations(rt Runtime) (more bool) {
 		// Unlock funds for penalties.
 		// TODO: handle bankrupt miner: https://github.com/filecoin-project/specs-actors/issues/627
 		// We're intentionally reducing the penalty paid to what we have.
-		penalty, err = st.UnlockUnvestedFunds(store, rt.CurrEpoch(), penalty)
+		available := st.GetUnlockedBalance(rt.CurrentBalance())
```
anorth, line 1616: Blocking: `available` -> `unlocked`, or change parameters for `PenalizeFundsInPriorityOrder`

actors/builtin/miner/miner_actor.go
```diff
@@ -1680,6 +1690,7 @@ func handleProvingDeadline(rt Runtime) {
 		deadline, err := deadlines.LoadDeadline(store, dlInfo.Index)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d", dlInfo.Index)
 		quant := st.QuantEndOfDeadline()
+		availableBalance := st.GetUnlockedBalance(rt.CurrentBalance())
```
anorth, line 1693: Blocking: `availableBalance` -> `unlockedBalance`.

actors/builtin/miner/miner_state.go
```diff
@@ -939,6 +941,26 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
...
+	fromVesting, err = st.UnlockUnvestedFunds(store, currEpoch, target)
+	if err != nil {
+		return abi.NewTokenAmount(0), abi.NewTokenAmount(0), err
+	}
+	if fromVesting.Equals(target) {
+		return fromVesting, abi.NewTokenAmount(0), nil
+	}
+
+	// unlocked funds were just deducted from available, so track that
+	availableBalance = big.Sub(availableBalance, fromVesting)
```
anorth, line 957: Blocking: Interpreting "available" as "unlocked", this doesn't make sense. If you consider the amount just unlocked to be unlocked, then this should be an `Add`. Alternatively, if you consider the amount just unlocked as already burnt, then this should not change at all.

actors/builtin/miner/miner_state.go
```diff
@@ -939,6 +941,26 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
...
+	remaining := big.Sub(target, fromVesting)
+
+	fromBalance = big.Min(availableBalance, remaining)
```
anorth, line 960: Call out explicitly that if the penalty exceeds the amount available, the penalty is reduced to match, and point to to issue #627 for bankruptcy.

actors/builtin/miner/miner_actor.go
```diff
@@ -1398,33 +1399,35 @@ func (a Actor) CompactPartitions(rt Runtime, params *CompactPartitionsParams) *a
...
+
 	newlyVested := rt.State().Transaction(&st, func() interface{} {
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(info.Worker, info.Owner, builtin.RewardActorAddr)
 
 		newlyVestedFund, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 
-		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
+		availableBalance := st.GetUnlockedBalance(rt.CurrentBalance())
```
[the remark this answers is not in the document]
ZenGround0: It is the available balance for the operation in question in this context.  We weren't sure about requiring understanding of the available/unlocked naming on all code readers so I didn't apply the naming convention rigorously.  I'll go back and apply this naming strictly everywhere.

actors/builtin/miner/miner_actor.go
ZenGround0: I'm confused, `verifyPledgeMeetsInitialRequirements` does just explicitly abort if available balance is <= 0.

ZenGround0: Are you saying remove the abort from the verify helper?

actors/builtin/miner/miner_actor.go
ZenGround0: If it accepts `rt.CurrentBalance()` it will also have to accept a parameter for accumulated deductions from unlocked funds because rt.CurrentBalance won't update until the funds are burnt after the state transaction.  Maybe that would be clearer but it won't get rid of the need to keep track of the unlocked funds.

actors/builtin/miner/miner_state.go
ZenGround0: Thank you, great catch

ZenGround0: This function should assume they will be burnt as it is used for Penalizing

PR #819, opened 2020-07-27
Prevent sector number reuse

### LINKED ISSUE #451: Prevent sector ID re-use
At present, a sector number (sector ID) may be re-used after a sector terminates, either due to missing ProveCommit, or termination. This could open up hard-to-spot error cases, and certainly makes sector number use from the outside very difficult. Re-use is unlikely to be efficient in bitfield-encoding, though there are contrived cases where it could be.

Consider preventing sector number re-use, perhaps by recording a sector number high-water-mark either at PreCommit or ProveCommit, and requiring new pre-commitments to have a higher number.

Stebalien:
~Depends on https://github.com/filecoin-project/go-bitfield/pull/43~

TODO

* [x] Tests

fixes #451

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -103,6 +103,9 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
 		rt.Abortf(exitcode.ErrIllegalState, "failed to construct initial state: %v", err)
 	}
 
+	emptyBitfield := bitfield.NewFromSet(nil)
+	emptyBitfieldCid := rt.Store().Put(emptyBitfield)
+
 	emptyDeadline := ConstructDeadline(emptyArray)
 	emptyDeadlineCid := rt.Store().Put(emptyDeadline)
 
@@ -119,7 +122,7 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to construct initial miner info")
 	infoCid := rt.Store().Put(info)
 
-	state, err := ConstructState(infoCid, periodStart, emptyArray, emptyMap, emptyDeadlinesCid)
+	state, err := ConstructState(infoCid, periodStart, emptyBitfieldCid, emptyArray, emptyMap, emptyDeadlinesCid)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to construct state")
 	rt.State().Create(state)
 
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -33,6 +34,9 @@ type State struct {
 	// Sectors that have been pre-committed but not yet proven.
 	PreCommittedSectors cid.Cid // Map, HAMT[SectorNumber]SectorPreCommitOnChainInfo
 
+	// Allocated sector IDs. Sector IDs can never be reused once allocated.
+	AllocatedSectors cid.Cid // BitField
```
Stebalien, line 38: I've put this behind a CID as it _could_ get large.

Q: We can save loading a few bytes every time by grouping this with `PreCommittedSectors`. That is, we can define:

```go
type PreCommitInfo {
    PreCommittedSectors cid.Cid // Map
    AllocatedSectors BitField (or behind a CID to avoid loading on prove-commit?)
}
```

And only load this when we care about precommits.

But this is probably a low priority question we can punt.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/filecoin-project/go-address v0.0.2-0.20200218010043-eb9bb40ed5be
 	github.com/filecoin-project/go-amt-ipld/v2 v2.1.0
-	github.com/filecoin-project/go-bitfield v0.1.0
+	github.com/filecoin-project/go-bitfield v0.1.2-0.20200727181427-9882314b1ea4
```
Stebalien, line 8: Need to merge https://github.com/filecoin-project/go-bitfield/pull/43 and cut a release.

actors/builtin/miner/miner_state.go
Stebalien: My plan is to punt on this for now.

actors/builtin/miner/miner_actor.go
```diff
@@ -1332,6 +1343,10 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 	return nil
 }
 
+/////////////////
+// Maintenence //
```
anorth, line 1347:
```suggestion
// Maintenance //
```

actors/builtin/miner/miner_state_test.go
```diff
@@ -543,6 +543,40 @@ func TestVestingFunds_UnvestedFunds(t *testing.T) {
...
+		assert.Error(t, harness.s.AllocateSectorNumber(harness.store, abi.MaxSectorNumber+1))
+		assert.Error(t, harness.s.MaskSectorNumbers(harness.store, bf(99, abi.MaxSectorNumber+1)))
+	})
+
+	t.Run("can allocate in range", func(t *testing.T) {
+		harness := constructStateHarness(t, abi.ChainEpoch(0))
+		assert.NoError(t, harness.s.AllocateSectorNumber(harness.store, abi.MaxSectorNumber))
+		assert.NoError(t, harness.s.MaskSectorNumbers(harness.store, bf(99, abi.MaxSectorNumber)))
+	})
+}
```
anorth, line 578: Can you test the bitfield-too-big case too?

PR #820, opened 2020-07-27
Allow up to the max sector number in precommit

### LINKED ISSUE #798: Miner.PreCommitSector: Conflicting MaxSectorNumber checks
`params.SectorNumber` may be equal to `abi.MaxSectorNumber`:

https://github.com/filecoin-project/specs-actors/blob/70a18dfa12d7a0bc7981e73c6021c11d7de14cb3/actors/builtin/miner/miner_actor.go#L426-L428

`params.ReplaceSectorNumber` may not:

https://github.com/filecoin-project/specs-actors/blob/70a18dfa12d7a0bc7981e73c6021c11d7de14cb3/actors/builtin/miner/miner_actor.go#L456-L458

Stebalien:
fixes #798

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -163,7 +163,8 @@ func TestCommitments(t *testing.T) {
 		dlInfo := actor.deadline(rt)
 
 		// Make a good commitment for the proof to target.
-		sectorNo := abi.SectorNumber(100)
+		// Use the max sector number to make sure everything works.
+		sectorNo := abi.SectorNumber(abi.MaxSectorNumber)
 		expiration := dlInfo.PeriodEnd() + 181*miner.WPoStProvingPeriod // something on deadline boundary but > 180 days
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -345,7 +346,8 @@ func TestCommitments(t *testing.T) {
 		rt.SetEpoch(periodOffset + miner.WPoStChallengeWindow)
 
 		// Commit a sector to upgrade
-		oldSector := actor.commitAndProveSectors(rt, 1, 181, nil)[0]
+		// Use the max sector number to make sure everything works.
+		oldSector := actor.commitAndProveSector(rt, abi.MaxSectorNumber, 181, nil)
```
Stebalien, line 350: @anorth would you rather I create new tests for things like this?

anorth: It's a pragmatic balance, this case is fine. What's missing, though, is a test that `abi.MaxSectorNumber+1` is rejected.

PR #821, opened 2020-07-27
Minimum Deal Provider Collateral

### LINKED ISSUE #738: Deal collateral minima
This is most parameterized, we just need to plumb circulating supply through plus tweak some functions.

Minimum client and provider deal collateral now should be 5% normalized_circulating_supply, with normalized_circulating_supply = deal_qa_power / total_qa_power * total_circulating_supply

Since qa power is only defined over sector we need some dummy sector values for calculating deal_qa_power.  We should use 64 GiB sector size + max sector lifetime for this calculation.

ZenGround0:
Closes #738  

WIP since all tests are broken.  Implementation already feels solid however.

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -2806,12 +2806,12 @@ func (h *marketActorTestHarness) generateDealAndAddFunds(rt *mock.Runtime, clien
 
 func (h *marketActorTestHarness) generateDealWithCollateralAndAddFunds(rt *mock.Runtime, client address.Address,
 	minerAddrs *minerAddrs, providerCollateral, clientCollateral abi.TokenAmount, startEpoch, endEpoch abi.ChainEpoch) market.DealProposal {
-	deal4 := generateDealProposalWithCollateral(client, minerAddrs.provider, providerCollateral, clientCollateral,
+	deal := generateDealProposalWithCollateral(client, minerAddrs.provider, providerCollateral, clientCollateral,
 		startEpoch, endEpoch)
-	h.addProviderFunds(rt, deal4.ProviderCollateral, minerAddrs)
-	h.addParticipantFunds(rt, client, deal4.ClientBalanceRequirement())
+	h.addProviderFunds(rt, deal.ProviderCollateral, minerAddrs)
+	h.addParticipantFunds(rt, client, deal.ClientBalanceRequirement())
 
-	return deal4
+	return deal
 }
 
 func generateDealProposalWithCollateral(client, provider address.Address, providerCollateral, clientCollateral abi.TokenAmount, startEpoch, endEpoch abi.ChainEpoch) market.DealProposal {
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -12,6 +12,8 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/builtin"
 	"github.com/filecoin-project/specs-actors/actors/builtin/market"
 	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/actors/builtin/reward"
 	"github.com/filecoin-project/specs-actors/actors/builtin/verifreg"
 	"github.com/filecoin-project/specs-actors/actors/crypto"
 	"github.com/filecoin-project/specs-actors/actors/runtime"
@@ -441,6 +443,7 @@ func TestPublishStorageDeals(t *testing.T) {
 			&miner.GetControlAddressesReturn{Owner: mAddr.owner, Worker: mAddr.worker},
 			exitcode.Ok,
 		)
+		expectQueryNetworkInfo(rt, actor)
 		//  create a client proposal with a valid signature
 		var params market.PublishStorageDealsParams
 		buf := bytes.Buffer{}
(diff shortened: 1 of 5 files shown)


actors/builtin/market/policy.go
```diff
@@ -20,8 +23,19 @@ func dealPricePerEpochBounds(size abi.PaddedPieceSize, duration abi.ChainEpoch)
 	return abi.NewTokenAmount(0), abi.TotalFilecoin // PARAM_FINISH
 }
 
-func dealProviderCollateralBounds(pieceSize abi.PaddedPieceSize, duration abi.ChainEpoch) (min abi.TokenAmount, max abi.TokenAmount) {
-	return abi.NewTokenAmount(0), abi.TotalFilecoin // PARAM_FINISH
+func dealProviderCollateralBounds(pieceSize abi.PaddedPieceSize, verified bool, networkQAPower, baselinePower abi.StoragePower, networkCirculatingSupply abi.TokenAmount) (min abi.TokenAmount, max abi.TokenAmount) {
```
anorth, line 26: While you're here, export all these functions so they can be used by clients & nodes for providing this collateral, pricing etc.

actors/builtin/network.go
```diff
@@ -29,3 +33,15 @@ func init() {
 		panic(fmt.Sprintf("epoch duration %d does not evenly divide one hour (%d)", EpochDurationSeconds, SecondsInHour))
 	}
 }
+
+// Quality multiplier for committed capacity (no deals) in a sector
+var QualityBaseMultiplier = big.NewInt(10)
```
anorth, line 38: Nit: @zixuanzh  can we rename this to something other than "multiplier", since it's only ever used in a denominator.

zixuanzh: any suggestion? maybe we should split this into CommittedCapacityMultiplier =10 and then QualityBase = 10? We can make this 100 instead of 10 too to have another bit of precision but we dont need it atm.

actors/builtin/market/policy.go
```diff
@@ -40,3 +54,16 @@ func DealWeight(proposal *DealProposal) abi.DealWeight {
 	dealSpaceTime := big.Mul(dealDuration, dealSize)
 	return dealSpaceTime
 }
+
+func dealQAPower(dealSize abi.PaddedPieceSize, verified bool) abi.StoragePower {
```
zixuanzh, line 58: both `dealProviderCollateralBounds` and `dealQAPower` lgtm

ZenGround0: thanks for the quick review!

actors/builtin/market/market_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/builtin"
 	"github.com/filecoin-project/specs-actors/actors/builtin/market"
 	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/actors/builtin/reward"
```
ZenGround0, line 16: Made heavy use out of the default of fil circulating supply being zero which drops the min collateral to zero and minimized breakage.

actors/builtin/market/market_test.go
```diff
@@ -2755,17 +2804,32 @@ func (h *marketActorTestHarness) generateDealAndAddFunds(rt *mock.Runtime, clien
 	return deal4
 }
 
-func generateDealProposal(client, provider address.Address, startEpoch, endEpoch abi.ChainEpoch) market.DealProposal {
+func (h *marketActorTestHarness) generateDealWithCollateralAndAddFunds(rt *mock.Runtime, client address.Address,
+	minerAddrs *minerAddrs, providerCollateral, clientCollateral abi.TokenAmount, startEpoch, endEpoch abi.ChainEpoch) market.DealProposal {
+	deal4 := generateDealProposalWithCollateral(client, minerAddrs.provider, providerCollateral, clientCollateral,
```
acruikshank, line 2809: What happened to deal1, deal2, and deal3?

actors/builtin/market/policy.go
```diff
@@ -9,6 +9,9 @@ import (
 // DealUpdatesInterval is the number of blocks between payouts for deals
 const DealUpdatesInterval = 100
 
+var ProvCollateralPercentSupplyNum = big.NewInt(5)
```
acruikshank, line 12: These should have a comment explaining the formula they are used in and meaning (if that isn't too hard to describe).

actors/builtin/market/policy.go
```diff
@@ -20,11 +23,21 @@ func dealPricePerEpochBounds(size abi.PaddedPieceSize, duration abi.ChainEpoch)
 	return abi.NewTokenAmount(0), abi.TotalFilecoin // PARAM_FINISH
 }
 
-func dealProviderCollateralBounds(pieceSize abi.PaddedPieceSize, duration abi.ChainEpoch) (min abi.TokenAmount, max abi.TokenAmount) {
-	return abi.NewTokenAmount(0), abi.TotalFilecoin // PARAM_FINISH
+func DealProviderCollateralBounds(pieceSize abi.PaddedPieceSize, verified bool, networkQAPower, baselinePower abi.StoragePower, networkCirculatingSupply abi.TokenAmount) (min abi.TokenAmount, max abi.TokenAmount) {
+	lockTargetNum := big.Mul(ProvCollateralPercentSupplyNum, networkCirculatingSupply)
```
acruikshank, line 27: It would nice to have a succinct math formula as a comment here.

actors/builtin/market/market_test.go
ZenGround0: heh yeah I had the same question.  I was propagating the existing convetion, will just change to deal.

actors/builtin/market/policy.go
ZenGround0: great idea will do

actors/builtin/market/market_test.go
```diff
@@ -647,6 +673,19 @@ func TestPublishStorageDealsFailures(t *testing.T) {
...
+					// with these two equal provider collatreal min is 5/100 * deal size
+					rt.SetCirculatingSupply(h.networkQAPower)
+					dealSize := big.NewInt(2048) // default deal size used
+					providerMin := big.Div(
+						big.Mul(dealSize, market.ProvCollateralPercentSupplyNum),
+						market.ProvCollateralPercentSupplyDenom,
+					)
+					d.ProviderCollateral = big.Sub(providerMin, big.NewInt(1))
+				},
+				exitCode: exitcode.ErrIllegalArgument,
```
anorth, line 687: Nit: `ErrForbidden` might be slightly better for "insufficient collateral"?

PR #825, opened 2020-07-28
Switch to 30 second block time

### LINKED ISSUE #748: Change epoch duration to 30s, confirm that cryptoecon params align
If we are planning to switch from 25s back to 30s for block time, some cryptoecon constants may need to be recomputed by manual intervention (if their dependence on block time isn't captured in code). These include:
* the `lambda` parameter in the minting function that computes perEpochReward
* a similar decay parameter in the baselinePower function

There may be a couple others that I'm not thinking of at the moment.

Stebalien:
* Re-define a year to be an even 365 days for portability. Matching up with real time is futile anyways as rewards are probabilistic.
* Remove exported definition of "epochs in year" to avoid abusing this variable. We don't want to start calculating sector/deal lifetimes based on this number.
* Recompute constants based on the new block time, linking to wolfram-alpha computations.
* Define computations in a python file so we don't have to rely on wolfram alpha.

fixes #748


Stebalien:
> Re-define a year to be an even 365 days for portability. Matching up with real time is futile anyways as rewards are probabilistic.

/cc @zixuanzh, @davidad 

In retrospect, we _can_ just use `365.whatever` if it really makes a difference, as long as we're consistent.

Stebalien:
Note: wolfram alpha is not a reliable tool for performing these calculations.

* 1 "tropical" year is ~31556925.11163 seconds (not ISO).
* The following two give different results:
    * https://www.wolframalpha.com/input/?i=floor%282%5E128*Log%5B2%5D%2F%286+*+31556925.11163+%2F+30%29%29
    * https://www.wolframalpha.com/input/?i=floor%282%5E128*Log%5B2%5D%2F%286+*+tropical+years+%2F+30+seconds%29%29
* The python script I wrote gives an even different result and I'm not sure where the precision is being lost.

actors/builtin/miner/policy.go
```diff
@@ -83,7 +83,7 @@ var MaxSealDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
 
 // Number of epochs between publishing the precommit and when the challenge for interactive PoRep is drawn
 // used to ensure it is not predictable by miner.
-var PreCommitChallengeDelay = abi.ChainEpoch(150)
+var PreCommitChallengeDelay = abi.ChainEpoch(125) // 1 hour
```
Stebalien, line 86: This used to be 62.5 minutes.

actors/builtin/miner/policy.go
```diff
@@ -92,15 +92,15 @@ const ElectionLookback = abi.ChainEpoch(1) // PARAM_FINISH
 // This lookback exists so that deadline windows can be non-overlapping (which make the programming simpler)
 // but without making the miner wait for chain stability before being able to start on PoSt computation.
 // The challenge is available this many epochs before the window is actually open to receiving a PoSt.
-const WPoStChallengeLookback = abi.ChainEpoch(20)
+const WPoStChallengeLookback = abi.ChainEpoch(16) // 8 minutes
```
Stebalien, line 95: This used to be 8.3333 minutes.

actors/builtin/miner/policy.go
```diff
@@ -92,15 +92,15 @@ const ElectionLookback = abi.ChainEpoch(1) // PARAM_FINISH
...
 
 // Minimum period before a deadline's challenge window opens that a fault must be declared for that deadline.
 // This lookback must not be less than WPoStChallengeLookback lest a malicious miner be able to selectively declare
 // faults after learning the challenge value.
-const FaultDeclarationCutoff = WPoStChallengeLookback + 50
+const FaultDeclarationCutoff = WPoStChallengeLookback + 42 // 29 minutes
```
Stebalien, line 100: This used to be 29.1666 minutes.

Also, this is awfully close to one full deadline. Should this just be a full deadline?

Stebalien:
Also, should finality change?

actors/builtin/miner/policy.go
nicola: yes, this should be 20.

actors/builtin/miner/policy.go
nicola: This should still be 50.

actors/builtin/miner/policy.go
nicola: This must be 150 - it has nothing to do with time, it has to do with number of blocks.

nicola:
Hey team, I flagged all the changes that are incorrect. General rule of thumb: parameters that are set in "number of blocks" (PreCommitChallengeDelay, FaultCutoff) are all defined in number of blocks and will not change with a blocktime change. This has to do with their security constraints.

Intuition: the probability of doing a selfish mining fork is 2^-30 for 250 blocks, regardless of how long a blocktime is.

actors/builtin/reward/reward_calc.py
```diff
@@ -0,0 +1,56 @@
...
+# exp(ln[1 + 200%] / epochsInYear)
+def baseline_exponent() -> Decimal: 
+    return (Decimal(1 + GROWTH_FACTOR).ln() / epochs_in_year()).exp()
+
+def baseline_exponent_q128() -> str: 
+    return q128(baseline_exponent(), decimal.ROUND_HALF_UP)
+
+# ln(2) / (6 * epochsInYear)
+def reward_lambda() -> Decimal: 
+    return Decimal(GROWTH_FACTOR).ln() / (6 * epochs_in_year())
```
davidad, line 38: This should not depend on the baseline GROWTH_FACTOR, it is just ln(2) here (as 6 years is a half-life). It’s a coincidence that GROWTH_FACTOR happens to be 2.

Stebalien: Ah, got it.

actors/builtin/miner/policy.go
Stebalien: This increases the fault declaration cutoff from 29 minutes (slightly less than a full deadline ahead) to 35 minutes (more than a deadline). Is that OK?

Stebalien:
@nicola have we considered the impact of changing the block time on these parameters?

* We now have 60 blocks/deadline to submit proofs instead of 72.
* The fault cutoff is now more than a full deadline ahead of the deadline where the sector needs to be proven.
* The pre-commit challenge delay is now 1h15m instead of 1 hour.

actors/builtin/miner/deadlines_test.go
```diff
@@ -121,15 +121,17 @@ func TestProvingPeriodDeadlines(t *testing.T) {
...
 			nxt0 := di0.NextNotElapsed()
 			assert.Equal(t, periodStart, nxt0.PeriodStart)
 			assert.Equal(t, uint64(0), nxt0.Index)
 
 			di1 := miner.NewDeadlineInfo(periodStart, 1, curr)
 			assert.False(t, di1.IsOpen())
 			assert.False(t, di1.HasElapsed())
-			assert.False(t, di1.FaultCutoffPassed())
+			// The fault cutoff is more than one deadline into the future.
+			assert.True(t, di1.FaultCutoffPassed())
```
Stebalien, line 134: @anorth is this going to be a problem? (this is due to the timing changes).

actors/util/smoothing/alpha_beta_filter_test.go
```diff
@@ -65,12 +65,16 @@ func TestCumSumRatioProjection(t *testing.T) {
 	// Note 2: since both methods are approximations with error this is not a
 	// measurement of analytic method's error, it is a sanity check that the
 	// two methods give similar results
-	errBound := big.NewInt(300)
+	errBound := big.NewInt(350)
```
Stebalien, line 68: @ZenGround0 I needed to change this. Does this means the new constants are off?

ZenGround0: no its to be expected.  thanks for checking!

actors/builtin/miner/deadlines_test.go
anorth: Slightly annoying, but not a significant problem.

anorth:
@Stebalien thanks for calling out those impacts. They are all understood by and ok by me.

PR #826, opened 2020-07-28
Store BR in sector info and use it in sector termination penalty calculation

### LINKED ISSUE #701: Calculate Termination Fee from BR(start), stored in sector info
Because of the recent change to IP the termination fee's calculation of BR(start time) is no longer correct.  We should store BR(start time) separate from IP in the sector's on chain state and then use this value directly.  @ZX and I found this out right before merging IP but decided new IP was more important than correcting termination fee.

acruikshank:
closes #701 

### Motivation

Termination fees are a function of expected day reward (BR) evaluated at activation. We had been using the simple relationship between `InitialPledge` and BR to derive BR(start). That simple relationship has gone away, so we now need to store BR in sector info and use it in the terminate sector penalty calculation.

### Proposed Changes

1. Add `ExpectedDayReward` field to `SectorOnChainInfo`.
2. Set this field in `ConfirmSectorProofsValid` when sector is activated.
3. Add `dayReward` as a parameter to `PledgePenaltyForTermination` and use it instead of dividing initial pledge by `IniitialPledgeFactor` (which is no longer correct).
4. Restore test to test TerminateSector to confirm the fee calculation is working.

### Bonus bug

Restoring the terminate sector test exposed a bug in miner.TerminateSectors. It was not adding the indices of the deadlines containing the terminated sectors to its `EarlyTerminations` bitfield. This prevented it from correctly processing the termination including charging the termination fee. This is fixed in this PR.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1029,6 +1029,8 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *Ter
 			removedPower, err := deadline.TerminateSectors(store, currEpoch, byPartition, info.SectorSize, quant)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to terminate sectors in deadline %d", dlIdx)
 
+			st.EarlyTerminations.Set(dlIdx)
+
 			powerDelta = powerDelta.Sub(removedPower)
 
 			err = deadlines.UpdateDeadline(store, dlIdx, deadline)
@@ -1053,11 +1055,6 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *Ter
 	}
 
 	requestUpdatePower(rt, powerDelta)
-
-	// Pledge requirement is not released until termination fee is paid.
-	// The termination fee is paid later, in early-termination queue processing.
-	// We could charge at least the undeclared fault fee here, which is a lower bound on the penalty.
-	// https://github.com/filecoin-project/specs-actors/issues/674
 
 	return &TerminateSectorsReturn{Done: !more}
 }
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -706,6 +706,7 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			activation := rt.CurrEpoch()
 			duration := precommit.Info.Expiration - activation
 			power := QAPowerForWeight(info.SectorSize, duration, precommit.DealWeight, precommit.VerifiedDealWeight)
+			dayReward := ExpectedDayRewardForPower(rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, power)
 			initialPledge := InitialPledgeForPower(power, rewardStats.ThisEpochBaselinePower, pwrTotal.PledgeCollateral,
 				rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, circulatingSupply)
 
@@ -721,6 +722,7 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 				DealWeight:         precommit.DealWeight,
 				VerifiedDealWeight: precommit.VerifiedDealWeight,
 				InitialPledge:      initialPledge,
+				ExpectedDayReward:  dayReward,
 			}
 			newSectors = append(newSectors, &newSectorInfo)
 			newSectorNos = append(newSectorNos, newSectorInfo.SectorNumber)
@@ -2409,7 +2411,7 @@ func terminationPenalty(sectorSize abi.SectorSize, currEpoch abi.ChainEpoch, rew
 	totalFee := big.Zero()
 	for _, s := range sectors {
 		sectorPower := QAPowerForSector(sectorSize, s)
-		fee := PledgePenaltyForTermination(s.InitialPledge, currEpoch-s.Activation, rewardEstimate, networkQAPowerEstimate, sectorPower)
+		fee := PledgePenaltyForTermination(s.InitialPledge, s.ExpectedDayReward, currEpoch-s.Activation, rewardEstimate, networkQAPowerEstimate, sectorPower)
 		totalFee = big.Add(fee, totalFee)
 	}
 	return totalFee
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -2509,26 +2514,33 @@ func makeProveCommit(sectorNo abi.SectorNumber) *miner.ProveCommitSectorParams {
...
-	//	require.NoError(t, err)
-	//	faultAtDeadline[dl] = append(faultAtDeadline[dl], uint64(sectorInfo.SectorNumber))
-	//}
-	params := &miner.DeclareFaultsParams{Faults: []miner.FaultDeclaration{}}
-	// Group together faults at the same deadline into a bitfield
-	for dl, sectorNumbers := range faultAtDeadline {
-		fault := miner.FaultDeclaration{
-			Deadline: dl,
-			Sectors:  bitfield.NewFromSet(sectorNumbers),
+	deadlines, err := st.LoadDeadlines(store)
```
acruikshank, line 2517: I thought at one point I might need to get faults working in tests. I did, but ended up not needing the faults in the end. This restoration of test code should be a leg up for getting fault tests back up.

actors/builtin/miner/miner_actor.go
```diff
@@ -704,6 +704,7 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			activation := rt.CurrEpoch()
 			duration := precommit.Info.Expiration - activation
 			power := QAPowerForWeight(info.SectorSize, duration, precommit.DealWeight, precommit.VerifiedDealWeight)
+			dayReward := ExpectedDayRewardForPower(epochReward, pwrTotal.QualityAdjPower, power)
```
anorth, line 707: Beware conceptual conflict with #762 (I think the type checker will see it).

actors/builtin/miner/miner_actor.go
```diff
@@ -1025,6 +1027,8 @@ func (a Actor) TerminateSectors(rt Runtime, params *TerminateSectorsParams) *Ter
 			removedPower, err := deadline.TerminateSectors(store, currEpoch, byPartition, info.SectorSize, quant)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to terminate sectors in deadline %d", dlIdx)
 
+			st.EarlyTerminations.Set(dlIdx)
```
anorth, line 1030: FYI @Stebalien here's another of those bugs where miner actor code broke consistency of different bits of the state. #828.

actors/builtin/miner/miner_test.go
```diff
@@ -1362,73 +1362,66 @@ func TestExtendSectorExpiration(t *testing.T) {
...
+		// correct fee calculation is used.
+		actor.addLockedFund(rt, big.Mul(big.NewInt(1e18), big.NewInt(20000)))
+		st := getState(rt)
+		initialLockedFunds := st.LockedFunds
+
+		sectorSize, err := sector.SealProof.SectorSize()
+		require.NoError(t, err)
+		sectorPower := miner.QAPowerForSector(sectorSize, sector)
+		dayReward := miner.ExpectedDayRewardForPower(actor.epochReward, actor.networkQAPower, sectorPower)
+		sectorAge := rt.Epoch() - sector.Activation
```
anorth, line 1392: It's not clear that this is definitely >0, so roll the epoch forward a bit after the `commitSector`.

actors/builtin/miner/miner_test.go
```diff
@@ -1362,73 +1362,66 @@ func TestExtendSectorExpiration(t *testing.T) {
...
+		expectedFee := miner.PledgePenaltyForTermination(sector.InitialPledge, dayReward, sectorAge, actor.epochReward, actor.networkQAPower, sectorPower)
+
+		sectors := bf(uint64(sector.SectorNumber))
+		actor.terminateSectors(rt, sectors, expectedFee)
+
+		{
+			st := getState(rt)
+
+			// expect sector to be marked as terminated but no longer early in partition
```
anorth, line 1401:
```suggestion
			// expect sector to be marked as terminated and the early termination queue to be empty (having been fully processed)
```

actors/builtin/miner/miner_test.go
```diff
@@ -2260,24 +2230,59 @@ func (h *actorHarness) terminateSectors(rt *mock.Runtime, sectors *abi.BitField,
...
+		}
+		declaration.Sectors.Set(id)
+		return nil
+	})
+	require.NoError(h.t, err)
+
+	declarations := []miner.TerminationDeclaration{}
+	for _, declaration := range declarationMap {
+		declarations = append(declarations, *declaration)
+	}
```
anorth, line 2281: The aggregation isn't strictly necessary here. It would be fine to submit a declaration per sector. Indeed that could be better to exercise the aggregation in the actor.

actors/builtin/miner/miner_test.go
```diff
@@ -2598,3 +2610,8 @@ func expectQueryNetworkInfo(rt *mock.Runtime, h *actorHarness) {
 		exitcode.Ok,
 	)
 }
+
+type decKey struct {
+	d uint64
+	p uint64
```
anorth, line 2616: See `PartitionKey` in the actor.

actors/builtin/miner/monies.go
```diff
@@ -55,18 +55,17 @@ func PledgePenaltyForUndeclaredFault(epochReward abi.TokenAmount, networkQAPower
 
 // Penalty to locked pledge collateral for the termination of a sector before scheduled expiry.
 // SectorAge is the time between the sector's activation and termination.
-func PledgePenaltyForTermination(initialPledge abi.TokenAmount, sectorAge abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower, qaSectorPower abi.StoragePower) abi.TokenAmount {
+func PledgePenaltyForTermination(initialPledge abi.TokenAmount, dayReward abi.TokenAmount, sectorAge abi.ChainEpoch, epochTargetReward abi.TokenAmount, networkQAPower, qaSectorPower abi.StoragePower) abi.TokenAmount {
```
anorth, line 58: Name this one `initialDayReward` or `dayRewardAtActivation` to distinguish from one calculated for the current epoch.

actors/builtin/miner/miner_actor.go
ZenGround0: Should just be plumbing but I can be around to reduce rebase pain.

actors/builtin/miner/miner_test.go
acruikshank: Moved forward 100 after commitSector

Issue #653, opened 2020-07-13
StoragePower: Incorrect assignment to Cid fields during construction


wadeAlexC:
The StoragePower constructor assigns the incorrect Cid root for some fields:

* `CronEventQueue` should be assigned `emptyMMapCid`, but is assigned `emptyMapCid` instead:

https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/power_state.go#L63

* `ProofValidationBatch` should be assigned `emptyMapCid`, but isn't assigned any value:

https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/power_state.go#L37

anorth:
These are lazily exploiting the fact that both the map and multimap are abstractions over a HAMT, so have the same empty state.

Stebalien:
P2 as we shouldn't leave fields uninitialized.

Stebalien:
Ah, wait, nevermind. `ProofValidationBatch` is intentionally left uninitialized (nil).

PR #830, opened 2020-07-28
Fix: StoragePower: Incorrect assignment to Cid fields during construc…

chunqizhi:
…tion


chunqizhi:
Close: https://github.com/filecoin-project/specs-actors/issues/653

PR #831, opened 2020-07-28
Power: Missing tests for Cron

### LINKED ISSUE #811: Thorough power actor testing
Some significant pieces of power actor/state code are not yet covered by tests.

aarshkshah1992:
Closes #811.


actors/builtin/power/power_actor.go
```diff
@@ -421,6 +426,8 @@ func (a Actor) processBatchProofVerifies(rt Runtime) error {
 		rt.Abortf(exitcode.ErrIllegalState, "failed to batch verify: %s", err)
 	}
 
+	confirms := make([]*minerConfirmSector, 0, len(miners))
```
anorth, line 429: I don't understand why this change is necessary. Could you please motivate it (and probably also in a comment for this structure)?

actors/builtin/power/power_test.go
```diff
@@ -752,6 +765,161 @@ func TestSubmitPoRepForBulkVerify(t *testing.T) {
...
+			miner4: []abi.SealVerifyInfo{*info7, *info8}}
+
+		ac.createMinerBasic(rt, owner, owner, miner1)
+		ac.createMinerBasic(rt, owner, owner, miner2)
+		ac.createMinerBasic(rt, owner, owner, miner3)
+		ac.createMinerBasic(rt, owner, owner, miner4)
+		ac.updateClaimedPower(rt, miner1, powerUnit, powerUnit)
+		ac.updateClaimedPower(rt, miner1, powerUnit, powerUnit)
+		ac.updateClaimedPower(rt, miner1, powerUnit, powerUnit)
+		ac.updateClaimedPower(rt, miner1, powerUnit, powerUnit)
```
anorth, line 853: I think this `UpdateNetworkKPI` call should be extracted to a different test from proof verifications.

actors/builtin/power/power_test.go
```diff
@@ -752,6 +765,161 @@ func TestSubmitPoRepForBulkVerify(t *testing.T) {
...
+		infos := map[addr.Address][]abi.SealVerifyInfo{miner1: []abi.SealVerifyInfo{*info1},
+			miner2: []abi.SealVerifyInfo{*info2}}
+
+		res := batchVerifyDefaultOutput(infos)
+		delete(res, miner1)
+		rt.ExpectBatchVerifySeals(infos, res, nil)
+		rt.ExpectValidateCallerAddr(builtin.CronActorAddr)
+
+		rt.SetEpoch(abi.ChainEpoch(0))
+		rt.SetCaller(builtin.CronActorAddr, builtin.CronActorCodeID)
```
anorth, line 890: Use the existing `onEpochTickEnd` inside the `ExpectAbort`

actors/builtin/power/power_test.go
```diff
@@ -752,6 +765,161 @@ func TestSubmitPoRepForBulkVerify(t *testing.T) {
...
+		cs := []confirmedSectorSend{{miner1, []int{1}}}
+
+		ac.onEpochTickEnd(rt, 0, big.Zero(), cs, infos)
+	})
+
+	t.Run("success with one miner and multiple confirmed sectors", func(t *testing.T) {
+		rt, ac := basicPowerSetup(t)
+		info1 := sealInfo(1)
+		info2 := sealInfo(2)
+		info3 := sealInfo(3)
```
anorth, line 794: I suggest pulling these out of the `t.Run`s for re-use and conciseness.

actors/builtin/power/power_test.go
```diff
@@ -752,6 +765,161 @@ func TestSubmitPoRepForBulkVerify(t *testing.T) {
...
+		rt.ExpectValidateCallerAddr(builtin.CronActorAddr)
+
+		rt.SetEpoch(abi.ChainEpoch(0))
+		rt.SetCaller(builtin.CronActorAddr, builtin.CronActorCodeID)
+
+		rt.ExpectAbort(exitcode.ErrIllegalState, func() {
+			rt.Call(ac.Actor.OnEpochTickEnd, nil)
+		})
+		rt.Verify()
+	})
```
anorth, line 920: Please test
- duplicate SVIs for same sector (dup ignored)
- failed verification for one SVI for a miner, with others that succeed

actors/builtin/power/power_test.go
```diff
@@ -808,6 +976,39 @@ func (h *spActorHarness) constructAndVerify(rt *mock.Runtime) {
 	verifyEmptyMap(h.t, rt, st.CronEventQueue)
 }
 
+type confirmedSectorSend struct {
+	miner      addr.Address
+	sectorNums []int
```
anorth, line 981: Just use `abi.SectorNumber`; you shouldn't need to cast when initialising as `[]abi.SectorNumber{1, 2, 3}`.

actors/builtin/power/power_test.go
aarshkshah1992: This is done.

actors/builtin/power/power_test.go
aarshkshah1992: @anorth  The existing `onEpochTickEnd` sets "correct" assumptions about the `BatchVerifySeals` call.

```
	rt.ExpectBatchVerifySeals(infos, batchVerifyDefaultOutput(infos), nil)

```

This makes it easy for the other callers to use it. Since we want to program incorrect behaviour in this method here, I'm not sure of how to use `onEpochTickEnd ` here.

actors/builtin/power/power_actor.go
aarshkshah1992: @anorth 

**Earlier:** If a call to `BatchVerifyProofs` fails, all the sends to the `ConfirmSectorProofsValid` before that are successful and we abort which means there will be no further sends to `ConfirmSectorProofsValid` even for valid sectors/miners.

**Now:** We do ALL sends to `ConfirmSectorProofsValid` ONLY after we've verified all the sectors for all miners. 

I thought this behaviour would be easy to reason about. Let me know if this does not make sense and we want to keep the original behaviour and I'll make the change.

actors/builtin/power/power_test.go
aarshkshah1992: @anorth The test for duplicates is done.

However, I'll write a test for the failure case once we agree on https://github.com/filecoin-project/specs-actors/pull/831#discussion_r462261522.

actors/builtin/power/power_actor.go
anorth: Are you talking about the `Abortf` at line 434 here, if `m` is not found in the result from `BatchVerifySeals`. If we do abort there, all state changes are rolled back anyway, including those from any calls to `ConfirmSectorProofsValid` that happened earlier. An abort means that state will roll back as if the whole method call (including any subcalls) never happened. So as far as I can tell, this change makes no difference to the semantics, except costing less gas in an error case.

But also, if we reach that abort case, something is quite broken. It's not just a proof failing to verify.

actors/builtin/power/power_test.go
anorth: Maybe I still don't understand, but as noted above I think the semantics are exactly the same either way.

I'm talking about returning a `false` value, not incorrectly failing to return a result for a miner at all.

aarshkshah1992: This is done.

PR #833, opened 2020-07-28
Allow errors to carry exit codes

### LINKED ISSUE #597: Support exit code definition in error objects
As a general design goal, use of the `Runtime` interface is mostly constrained to top-level actor code. There is a temptation to pass it all over the place for access to `rt.Abortf`, but in most cases our policy is to use standard Go error propagation rather than this exception-like escape hatch.

This introduces a problem when information about the appropriate error code is known deep in the call tree, without access to `rt.Abort`. This often results in `ErrIllegalState` being used where another code is more appropriate.

 We can resolve this by implementing a custom error type that optionally includes an exit code, and adjusting `builtin.RequireNoError` to check for that code before falling back to the one given as a parameter. E.g. see [this comment](https://github.com/filecoin-project/specs-actors/pull/530#discussion_r445724184).

Retain the use of xerrors where possible.

Stebalien:
This will make it easier to push state transformations down into the miner's state object.

I've also pushed some error messages down a layer to simplify/remove error wrapping.

fixes #597

### DIFF
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	xc "github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
@@ -67,12 +68,12 @@ func ConstructDeadlines(emptyDeadlineCid cid.Cid) *Deadlines {
 
 func (d *Deadlines) LoadDeadline(store adt.Store, dlIdx uint64) (*Deadline, error) {
 	if dlIdx >= uint64(len(d.Due)) {
-		return nil, xerrors.Errorf("invalid deadline %d", dlIdx)
+		return nil, xc.ErrIllegalArgument.Wrapf("invalid deadline %d", dlIdx)
 	}
 	deadline := new(Deadline)
 	err := store.Get(store.Context(), d.Due[dlIdx], deadline)
 	if err != nil {
-		return nil, err
+		return nil, xc.ErrIllegalState.Wrapf("failed to lookup deadline %d: %w", dlIdx, err)
 	}
 	return deadline, nil
 }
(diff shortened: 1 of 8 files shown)


Stebalien:
cc @anorth for a signoff and/or suggestions on the `code.WithContext` pattern.

actors/builtin/miner/deadline_state.go
```diff
@@ -129,10 +134,10 @@ func (d *Deadline) LoadPartition(store adt.Store, partIdx uint64) (*Partition, e
 	var partition Partition
 	found, err := partitions.Get(partIdx, &partition)
 	if err != nil {
-		return nil, err
+		return nil, exitcode.ErrIllegalState.WithContext("failed to lookup partition %d: %w", partIdx, err)
 	}
 	if !found {
-		return nil, xerrors.Errorf("no partition %d", partIdx)
+		return nil, exitcode.ErrNotFound.WithContext("no partition %d: %w", partIdx, err)
```
anorth, line 140: `err` is nil here

actors/builtin/miner/miner_state.go
```diff
@@ -415,6 +407,9 @@ func (st *State) WalkSectors(
 
 	var toVisit [WPoStPeriodDeadlines]map[uint64][]uint64
 	for _, loc := range locations {
+		if loc.Deadline >= WPoStPeriodDeadlines {
+			return xerrors.Errorf("deadline %d out of range", loc.Deadline)
```
anorth, line 411: IllegalArgument or NotFound?

actors/runtime/exitcode/exitcode.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return errors.Is(w.cause, target)
+}
+
+// Unwrap extracts an exit code from an error, defaulting to the passed default
+// exit code.
+//
+//    err := ErrIllegalState.WithContext("my description: %w", err)
+//    exitcode.Unwrap(exitcode.ErrIllegalState, err) == exitcode.ErrIllegalArgument
+func Unwrap(defaultExitCode ExitCode, err error) (code ExitCode) {
```
anorth, line 91: This would be slightly clearer to me with the arguments reversed, "unwrap this, else default to that"

Stebalien:
> A shorter name for WithContext?

`Errorf`? Alternatives: `Fmt`, `With`, `Context`, `Msg`.

PR #835, opened 2020-07-28
Update allocation params

ZenGround0:
Per request from cryptoecon FYI @nicola @zixuanzh


anorth:
LGTM when @zixuanzh ack's the numbers.

anorth:
reward_state.go has `InitialRewardPositionEstimateStr = "36266260308195979333"`. Should that be consolidated with this?

zixuanzh:
I believe "36266260308195979333" is the right number - there might be a precision issue with tests that @ZenGround0 is still planning to diagnose

PR #836, opened 2020-07-28
Fix/max sector deal duration

ZenGround0:
Per cryptoecon request: fyi @zixuanzh @nicola

### DIFF
--- a/actors/builtin/market/policy.go
+++ b/actors/builtin/market/policy.go
@@ -19,8 +19,6 @@ var ProvCollateralPercentSupplyDenom = big.NewInt(100)
 
 // Bounds (inclusive) on deal duration
 func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.ChainEpoch) {
-	// Cryptoeconomic modelling to date has used an assumption of a maximum deal duration of up to one year.
-	// It very likely can be much longer, but we're not sure yet.
 	return abi.ChainEpoch(180 * builtin.EpochsInDay), abi.ChainEpoch(540 * builtin.EpochsInDay) // PARAM_FINISH
 }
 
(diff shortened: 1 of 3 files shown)


actors/builtin/market/policy.go
```diff
@@ -21,7 +21,7 @@ var ProvCollateralPercentSupplyDenom = big.NewInt(100)
 func dealDurationBounds(size abi.PaddedPieceSize) (min abi.ChainEpoch, max abi.ChainEpoch) {
 	// Cryptoeconomic modelling to date has used an assumption of a maximum deal duration of up to one year.
 	// It very likely can be much longer, but we're not sure yet.
```
anorth, line 23: Remove the misleading comment above.

actors/builtin/miner/policy.go
```diff
@@ -115,7 +115,7 @@ const MinSectorExpiration = 180 * builtin.EpochsInDay
 // Maximum number of epochs past the current epoch a sector may be set to expire.
 // The actual maximum extension will be the minimum of CurrEpoch + MaximumSectorExpirationExtension
 // and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
-const MaxSectorExpirationExtension = 366 * builtin.EpochsInDay
+const MaxSectorExpirationExtension = 540 * builtin.EpochsInDay
```
nicola, line 118: I don't think this means what it was originally intended. This variable sets for how long more maximum we can extend a sector. I thought this to be smaller.

cc @zixuanzh

anorth: I assumed that @zixuanzh's approval here was explicitly approving a change in policy such that a sector could be immediately committed or extended to its full lifetime. If that wasn't intended, then we should revert this bit.

zixuanzh: Yes, sorry for the late response here, iiuc a sector's extension is still bounded by the maximum sector lifetime since the start epoch (of the old sector) as in the comment above this line, cc @ZenGround0 to double check. Currently, given a max sector lifetime of 540 days, there is no restriction on how long a sector can be extended for as long as that condition is still met. Now that we are looking at this again, given the minimum sector lifetime of 180 days, the effective max sector expiration extension is 360, although 540 is also fine since there is no policy restriction that a sector cannot immediately commit to its full lifetime.

ZenGround0: Tracking correction in #877

PR #838, opened 2020-07-28
Move deadline mutations into deadline methods

Stebalien:
This refactor pushes state changes down into the state & deadline objects, out of the miner actor itself.

This PR should be reviewed commit by commit.

### DIFF
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
 	xc "github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/deadline_state.go
```diff
@@ -889,3 +883,120 @@ func (dl *Deadline) ProcessPoSt(store adt.Store, quant QuantSpec, faultExpiratio
...
+func (p *PoStResult) PowerDelta() PowerPair {
+	return p.RecoveredPower.Sub(p.NewFaultyPower)
+}
+
+// PenaltyPower is the power from this PoSt that should be penalized.
+func (p *PoStResult) PenaltyPower() PowerPair {
+	return p.NewFaultyPower.Add(p.RetractedRecoveryPower)
+}
+
+func (dl *Deadline) ProcessWindowedPoSt(
```
acruikshank, line 903: Consider `ProcessSectorsToBeProven` or `ProcessProvenSectors` to clarify that this has nothing to do with proof validation.

actors/builtin/miner/miner_actor.go
```diff
@@ -298,71 +296,22 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors")
+
 		deadline, err := deadlines.LoadDeadline(store, params.Deadline)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d", params.Deadline)
 
-		partitions, err := deadline.PartitionsArray(store)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d partitions", params.Deadline)
-
 		faultExpiration := currDeadline.Last() + FaultMaxAge
+		postResult, err = deadline.ProcessWindowedPoSt(store, sectors, info.SectorSize, st.QuantEndOfDeadline(), faultExpiration, params.Partitions)
```
acruikshank, line 306: The logic of handling all deadline state changes here is sound, but I find it a little uncomfortable declaring the post successful and sectors proven before validating the proof (knowing changes will be discarded if it fails). This should be called out here in a comment in `SubmitPoSt` since the logic is now buried.

actors/builtin/miner/miner_actor.go
```diff
@@ -1126,13 +1126,20 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 		quant := st.QuantEndOfDeadline()
 
 		// Group declarations by deadline, and remember iteration order.
```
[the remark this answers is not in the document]
Stebalien: It should be fine. Honestly, I'd rather process deadlines in ascending order as well (or maybe even assert that all parameters need to be passed in in ascending order (deadline + partition)).

actors/builtin/miner/deadline_state.go
Stebalien: `RecordProvenSectors`?

PR #840, opened 2020-07-29
Test that a duplicate Window PoSt submission is ignored

anorth:


actors/builtin/miner/miner_test.go
```diff
@@ -774,9 +774,54 @@ func TestWindowPost(t *testing.T) {
...
+		actor.constructAndVerify(rt)
+		store := rt.AdtStore()
+		sector := actor.commitAndProveSectors(rt, 1, 181, nil)[0]
+
+		st := getState(rt)
+		dlIdx, pIdx, err := st.FindSector(store, sector.SectorNumber)
 		require.NoError(t, err)
-		assert.False(t, empty, "no post submission")
+
+		// Skip over deadlines until the beginning of the one with the new sector
```
acruikshank, line 793: At this point, an `advanceToDeadline(rt, actor, dlIdx)` that assumes empty cron configs would be handy.

Issue #596, opened 2020-07-05
Cover each exported actor method with at least one test


anorth:
As a basic minimum milestone to reach for testing, we should exercise each exported actor method at least once. 

Of course, many methods need much more testing in order to know they're robust, but a basic happy path still ensures that the code is executable.

- [x] system.Constructor
- [x] reward.ThisEpochReward
- [x] miner.ChangeWorkerAddress
- [x] miner.ChangePeerID
- [x] miner.ChangeMultiaddrs
- [x] miner.CheckSectorProven
- [ ] miner.CompactPartitions
- [x] miner.CompactSectorNumbers

anorth:
Some of these are covered by #479

aarshkshah1992:
@anorth Picking up `miner.ChangeWorkerAddress`.

Issue #616, opened 2020-07-07
Remove return value from State.Transaction()


anorth:
The return value from the State.Transaction callback is error prone, particularly because it requires casting. One concrete bad pattern is that a `nil` cannot be converted to an `error` interface (https://play.golang.org/p/yuTVBMCvnf2).

The return value was only helpful when there was a single piece of information to be transmitted outside the transaction. Where there's more, we already write to in-scope vars. That can have its own problems, but we should just use that pattern uniformly.

cc @ZenGround0

anorth:
I've seen some more bugs from this, elevating priority.

PR #842, opened 2020-07-29
Clean-ups while auditing code coverage.

anorth:
While auditing for #596 I fixed up easy/spurious missing coverage spots and fixed some exit codes.

- Use builtin.RequireNoErr() in some "impossible" error cases.
- Define some better exit codes using the new exitcode.Wrapf.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -5,8 +5,6 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	addr "github.com/filecoin-project/go-address"
-	errors "github.com/pkg/errors"
-	xerrors "golang.org/x/xerrors"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
 	big "github.com/filecoin-project/specs-actors/actors/abi/big"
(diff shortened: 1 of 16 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -390,15 +361,11 @@ func (a Actor) processBatchProofVerifies(rt Runtime) error {
...
 		mmap, err := adt.AsMultimap(store, *st.ProofValidationBatch)
-		if err != nil {
-			rt.Abortf(exitcode.ErrIllegalState, "failed to load proofs validation batch: %s", err)
-		}
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load proofs validation batch")
 
 		err = mmap.ForAll(func(k string, arr *adt.Array) error {
 			a, err := address.NewFromBytes([]byte(k))
-			if err != nil {
-				return xerrors.Errorf("failed to parse address key: %w", err)
```
anorth, line 400: This error was never caught, along with  other "return error" cases here. #616.

actors/builtin/power/power_actor.go
```diff
@@ -475,25 +434,20 @@ func (a Actor) processDeferredCronEvents(rt Runtime) error {
 
 		for epoch := st.FirstCronEpoch; epoch <= rtEpoch; epoch++ {
 			epochEvents, err := loadCronEvents(events, epoch)
-			if err != nil {
-				return errors.Wrapf(err, "failed to load cron events at %v", epoch)
```
anorth, line 479: This error was never caught

PR #843, opened 2020-07-29
Miner: Replace AMT with slice for vesting funds

### LINKED ISSUE #502: Improve performance of miner.AddLockedFund, serialize as array

aarshkshah1992:
Closes #502 .

@anorth @Stebalien 

Please can I get a through review ? There's a lot of potential for goof-ups/off by ones here.

@Stebalien

Don't think there's an easy to reason about way to not have to read the whole slice.


actors/builtin/miner/miner_state.go
```diff
@@ -32,7 +32,7 @@ type State struct {
 	PreCommitDeposits abi.TokenAmount // Total funds locked as PreCommitDeposits
 	LockedFunds       abi.TokenAmount // Total rewards and added funds locked in vesting table
 
-	VestingFunds cid.Cid // Array, AMT[ChainEpoch]TokenAmount
+	VestingFunds []VestingFund
```
Stebalien, line 35: This should be put behind a CID so we don't have to load it whenever we load the miner's state. See the `Deadlines` structure for how we can do that.

Stebalien: Also, instead of storing this as a single array, storing two parallel arrays (or an array and a bitfield?) would be more compact:

```go
type VestingFunds struct {
  Epochs *bitfield.BitField
  Funds []abi.TokenAmount
}
```

(not necessary, but something to consider)

actors/builtin/miner/miner_state.go
```diff
@@ -965,25 +974,24 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
...
 		}
 	}
 
-	st.VestingFunds, err = vestingFunds.Root()
-	if err != nil {
-		return err
-	}
+	// sort slice by epoch
+	sort.Slice(st.VestingFunds, func(first, second int) bool {
+		return st.VestingFunds[first].Epoch <= st.VestingFunds[second].Epoch
```
Stebalien, line 992:
```suggestion
		return st.VestingFunds[first].Epoch < st.VestingFunds[second].Epoch
```

(shouldn't matter anyways)

actors/builtin/miner/miner_state.go
```diff
@@ -1014,123 +1022,83 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
+			}
+		}
 	}
 
-	err = vestingFunds.BatchDelete(toDelete)
-	if err != nil {
-		return big.Zero(), errors.Wrapf(err, "failed to delete locked fund during slash: %v", err)
+	// remove all entries upto and including lastIndex
+	if lastIndex != -1 {
+		st.VestingFunds = st.VestingFunds[lastIndex+1:]
```
Stebalien, line 1052: This will remove the vested funds we just skipped.

anorth:
I haven't reviewed in detail yet, but in case it's approved, please hold off landing this change until we've confirmed the space race has a stable candidate network for next week.

aarshkshah1992:
@anorth Will wait for you review as well.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

aarshkshah1992:
@Stebalien Have made the changes. Please take a look.

actors/builtin/miner/miner_state.go
aarshkshah1992: @Stebalien I'm not sure what you mean. `lastIndex` will point to the last fund we zeroed. We want to remove all entries upto and including that .

If funds for an epoch are zero here, it follows that all funds before it will be zero as well. What am I missing here ?

Stebalien: We may have funds that _should_ have vested but haven't been popped from the queue yet. We'll zero out all epochs skipped by https://github.com/filecoin-project/specs-actors/pull/843/files#diff-2580e4b82238370ee854a52e8d5428e7R1055.

actors/builtin/miner/miner_actor.go
```diff
@@ -1475,9 +1491,13 @@ func (a Actor) AddLockedFund(rt Runtime, amountToLock *abi.TokenAmount) *adt.Emp
 			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", unlockedBalance, *amountToLock)
```
Stebalien, line 1491: @anorth is this still an issue?

actors/builtin/miner/miner_state.go
```diff
@@ -939,11 +960,14 @@ func (st *State) AddInitialPledgeRequirement(amount abi.TokenAmount) {
 	st.InitialPledgeRequirement = newTotal
 }
 
-func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vestingSum abi.TokenAmount, spec *VestSpec) error {
+func (st *State) AddLockedFunds(vestingFunds *VestingFunds, currEpoch abi.ChainEpoch, vestingSum abi.TokenAmount, spec *VestSpec) error {
```
Stebalien, line 963: This pattern (passing in the vesting funds object to the state like this) makes me uncomfortable.

Honestly, loading and storing vesting funds inside this function is probably not a huge issue and makes it much easier to reason about this code.

aarshkshah1992: @anorth This then ties to the issue of "batching" reads. What do we want to do here ?

anorth: I was going to say the same thing as soon as I saw the first call to `UnlockVestedFunds`.

There is only one case where this pattern makes any difference, which is this one. In the others, the funds are loaded, mutated once, then stored. The load/store should be encapsulated in the state method.

At the call site of AddLockedFunds, first the vesting is processed and then a new amount locked up. We can just combine those into this call. Break the internals of `UnlockVestedFunds` out to a private method that can be called from here, and add a return value of the amount vested. Then encapsulate the load/store with no cost.

actors/builtin/miner/miner_actor.go
```diff
@@ -475,6 +477,9 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	store := adt.AsStore(rt)
 	var st State
 	newlyVestedAmount := rt.State().Transaction(&st, func() interface{} {
+		vestingFunds, err := st.LoadVestingFunds(store)
```
anorth, line 480: As commented below, this is less convenient but no more efficient, since there is exactly one mutation between the load and store. Push this back down into `UnlockVestedFunds`.

actors/builtin/miner/miner_actor.go
anorth: Yes, this is recently added. Maybe the comment is hard to grok, but the requirement that you can't lock more than you have must be checked.

actors/builtin/miner/miner_state.go
```diff
@@ -64,6 +62,8 @@ type State struct {
 	// Faults are not subtracted from this in state, but on the fly.
 	Deadlines cid.Cid
 
+	VestingFunds cid.Cid
```
anorth, line 65: Please move this back next to the `LockedFunds` total.

actors/builtin/miner/miner_state.go
```diff
@@ -921,6 +922,26 @@ func (st *State) SaveDeadlines(store adt.Store, deadlines *Deadlines) error {
 	return nil
 }
 
+// LoadVestingFunds loads the vesting funds table from the store
+func (st *State) LoadVestingFunds(store adt.Store) (*VestingFunds, error) {
+	var funds VestingFunds
+	if err := store.Get(store.Context(), st.VestingFunds, &funds); err != nil {
+		return nil, xc.ErrIllegalState.Wrapf("failed to load vesting funds (%s): %w", st.VestingFunds, err)
```
anorth, line 929: There's no need to specify `ErrIllegalState` explicitly, that's the assumed default unless we make some giant codebase-wide change of policy. Just use `xerrors`.

actors/builtin/miner/miner_state.go
```diff
@@ -965,25 +989,24 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
...
-		err = vestingFunds.Set(key, &lockedFundEntry)
-		if err != nil {
-			return err
+		// epoch already exists. Load existing entry
+		// and update amount.
+		if index, ok := epochToIndex[vestEpoch]; ok {
+			currentAmt := vestingFunds.Funds[index].Amount
+			vestingFunds.Funds[index].Amount = big.Add(currentAmt, vestThisTime)
+		} else {
+			// create a new entry and insert at the correct position
```
anorth, line 998: This comment is a bit misleading. Instead say just blindly appended and sorted later.

actors/builtin/miner/miner_state.go
```diff
@@ -965,25 +989,24 @@ func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vesti
...
+			// st.VestingFunds is sorted by epoch
+			entry := VestingFund{Epoch: vestEpoch, Amount: vestThisTime}
+			vestingFunds.Funds = append(vestingFunds.Funds, entry)
```
anorth, line 1001: By not updating the `epochToIndex` map, this implicitly assumes that all the `vestEpoch` will be unique. Given the quantization, I don't think we can be sure of that, so update the mapping in case the same epoch appears again. Please test this with a quant spec that has a step much smaller than the quantization.

actors/builtin/miner/miner_state.go
```diff
@@ -939,11 +960,14 @@ func (st *State) AddInitialPledgeRequirement(amount abi.TokenAmount) {
...
+func (st *State) AddLockedFunds(vestingFunds *VestingFunds, currEpoch abi.ChainEpoch, vestingSum abi.TokenAmount, spec *VestSpec) error {
 	AssertMsg(vestingSum.GreaterThanEqual(big.Zero()), "negative vesting sum %s", vestingSum)
-	vestingFunds, err := adt.AsArray(store, st.VestingFunds)
-	if err != nil {
-		return err
+
+	// maps the epochs in VestingFunds to their indices in the slice
+	epochToIndex := make(map[abi.ChainEpoch]int, len(vestingFunds.Funds))
+	for i := range vestingFunds.Funds {
+		vf := vestingFunds.Funds[i]
```
anorth, line 969: `for i, vf := range vestingFunds.Funds {`

actors/builtin/miner/miner_state.go
```diff
@@ -1015,122 +1038,96 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
-						return err
-					}
-				}
-			}
-			return nil
-		} else {
-			return finished
+	lastIndex := -1
+
+	for i := range vestingFunds.Funds {
```
anorth, line 1049: Range over indices and values

actors/builtin/miner/miner_state.go
```diff
@@ -1015,122 +1038,96 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
+		if amountUnlocked.GreaterThanEqual(target) {
+			break
 		}
-	})
 
-	if err != nil && err != finished {
-		return big.Zero(), err
+		vf := vestingFunds.Funds[i]
+		if vf.Epoch >= currEpoch {
+			vestedAmt := vf.Amount
```
anorth, line 1056: Just inline this. The past-tense name is a bit misleading.

actors/builtin/miner/miner_state.go
```diff
@@ -1015,122 +1038,96 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
-				} else {
-					if err = vestingFunds.Set(uint64(k), &lockedEntry); err != nil {
-						return err
-					}
-				}
-			}
-			return nil
-		} else {
-			return finished
+	lastIndex := -1
```
anorth, line 1047: lastIndexToRemove?

actors/builtin/miner/miner_state.go
```diff
@@ -1015,122 +1038,96 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
-
-	// Iterate vestingFunds  in order of release.
-	err = vestingFunds.ForEach(&lockedEntry, func(k int64) error {
-		if k < int64(currEpoch) {
-			amountUnlocked = big.Add(amountUnlocked, lockedEntry)
-			toDelete = append(toDelete, uint64(k))
-		} else {
-			return finished // stop iterating
+
+	lastIndex := -1
```
anorth, line 1090: lastIndexToRemove?

actors/builtin/miner/vesting_state.go
```diff
@@ -0,0 +1,22 @@
+package miner
+
+import (
+	"github.com/filecoin-project/specs-actors/actors/abi"
+)
+
+type VestingFunds struct {
```
anorth, line 7: Please document this, e.g. the ordering invariant.

actors/builtin/miner/miner_state.go
aarshkshah1992: Done.

actors/builtin/miner/miner_state.go
aarshkshah1992: @anorth Is this still an issue ?

actors/builtin/miner/miner_state.go
```diff
@@ -32,7 +32,7 @@ type State struct {
 	PreCommitDeposits abi.TokenAmount // Total funds locked as PreCommitDeposits
 	LockedFunds       abi.TokenAmount // Total rewards and added funds locked in vesting table
 
-	VestingFunds cid.Cid // Array, AMT[ChainEpoch]TokenAmount
+	VestingFunds cid.Cid
```
Stebalien, line 35: Can we add a `// VestingFunds` comment (indicating the serialized type).

actors/builtin/miner/miner_state.go
```diff
@@ -875,68 +899,64 @@ func (st *State) UnlockUnvestedFunds(store adt.Store, currEpoch abi.ChainEpoch,
...
 
-	err = vestingFunds.BatchDelete(toDelete)
-	if err != nil {
-		return big.Zero(), errors.Wrapf(err, "failed to delete locked fund during vest: %v", err)
+	// remove all entries upto and including lastIndexToRemove
+	if lastIndexToRemove != -1 {
+		vestingFunds.Funds = vestingFunds.Funds[lastIndexToRemove+1:]
 	}
 
 	st.LockedFunds = big.Sub(st.LockedFunds, amountUnlocked)
```
Stebalien, line 935: This still makes me a bit uncomfortable. Can we just:

1. Add an `unlockVestedFunds` method on `vestingFunds`.
2. Call that wherever this is called, making sure to update `st.LockedFunds` on the state object at the same time.
3. Get rid of this method.

actors/builtin/miner/miner_state.go
Stebalien: If that's the case, we need to change this method to assert that there are no funds in the queue that should have already been unlocked (instead of just skipping them).

actors/builtin/miner/miner_state.go
anorth: If we do that we should move all the lock/unlock manipulations to the VestingFunds object. Those methods can't update the LockedFunds total, but by making these state methods much smaller, it would be clearer that we are getting that right. I'm ok either way.

aarshkshah1992: The whole rearrangement is done.

aarshkshah1992:
@anorth @Stebalien Have done the rearrangement and changes. Please take a look.

actors/builtin/miner/miner_state.go
```diff
@@ -745,54 +767,29 @@ func (st *State) AddInitialPledgeRequirement(amount abi.TokenAmount) {
 	st.InitialPledgeRequirement = newTotal
 }
 
-func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vestingSum abi.TokenAmount, spec *VestSpec) error {
-	AssertMsg(vestingSum.GreaterThanEqual(big.Zero()), "negative vesting sum %s", vestingSum)
-	vestingFunds, err := adt.AsArray(store, st.VestingFunds)
+func (st *State) AddLockedFunds(store adt.Store, currEpoch abi.ChainEpoch, vestingSum abi.TokenAmount, spec *VestSpec) (vested abi.TokenAmount, err error) {
```
Stebalien, line 770: Let's document this. This is really `AddAndUnlock` but I'm not sure if any name other than `AddLockedFundsAndUnlockVestedFunds` would make that clear.

actors/builtin/miner/miner_state.go
```diff
@@ -745,54 +767,29 @@ func (st *State) AddInitialPledgeRequirement(amount abi.TokenAmount) {
...
-		if elapsed < spec.VestPeriod {
-			// Linear vesting, PARAM_FINISH
-			targetVest = big.Div(big.Mul(vestingSum, big.NewInt(int64(elapsed))), vestPeriod)
-		} else {
-			targetVest = vestingSum
-		}
-
-		vestThisTime := big.Sub(targetVest, vestedSoFar)
-		vestedSoFar = targetVest
+	AssertMsg(vestingSum.GreaterThanEqual(big.Zero()), "negative vesting sum %s", vestingSum)
```
Stebalien, line 776: Put this at the top?

actors/builtin/miner/miner_state.go
```diff
@@ -875,68 +838,42 @@ func (st *State) UnlockUnvestedFunds(store adt.Store, currEpoch abi.ChainEpoch,
...
+
+		if epoch >= currEpoch {
+			break
 		}
-		return nil
-	})
-	if err != nil && err != finished {
-		return big.Zero(), err
+
+		amountUnlocked = big.Add(amountUnlocked, amount)
```
Stebalien, line 876: nit: "amount vested" because we're not actually unlocking.

actors/builtin/miner/vesting_state.go
```diff
@@ -0,0 +1,125 @@
...
+func (v *VestingFunds) unlockUnvestedFunds(currEpoch abi.ChainEpoch, target abi.TokenAmount) abi.TokenAmount {
+	amountUnlocked := abi.NewTokenAmount(0)
+	lastIndexToRemove := -1
+
+	for i, vf := range v.Funds {
+		if amountUnlocked.GreaterThanEqual(target) {
+			break
+		}
+
+		if vf.Epoch >= currEpoch {
```
Stebalien, line 93: This is still an issue. We can't just skip the funds that _should_ have already been unlocked and drop them on the floor.

We either need to:

1. Admit that this case should be impossible, and return an error if it does happen for some reason.
2. Even better, just handle it and unlock the "should have already been unlocked" funds.

actors/builtin/miner/miner_state.go
```diff
@@ -32,7 +33,7 @@ type State struct {
 	PreCommitDeposits abi.TokenAmount // Total funds locked as PreCommitDeposits
 	LockedFunds       abi.TokenAmount // Total rewards and added funds locked in vesting table
 
-	VestingFunds cid.Cid // Array, AMT[ChainEpoch]TokenAmount
+	VestingFunds cid.Cid // Vesting Funds schedule for the miner. Slice[(VestingEpoch, VestingAmount)]
```
anorth, line 36: It's not actually a Slice, it's a `VestingFunds`. I see how, given the name, this could be inferred, but please specify for consistency with other CIDs.

actors/builtin/miner/vesting_state.go
anorth: Good catch! This wasn't an issue in the original, which left the leading AMT keys in place. Could we please have a test for this explicitly, and then option 2.

anorth: On the options:
1. It seems quite possible; 2 out of 3 call paths to here have not processed vesting immediately prior
2. This will require passing out the unlocked amount and correctly using it to update pledge with the power actor

So I suggest:
3. Replicate the prior behaviour of leaving the should-have-vested entries in the table.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: Good catch ! This is done.

actors/builtin/miner/vesting_state.go
aarshkshah1992: This is done using option 3 and a test has been added for this.

aarshkshah1992:
@anorth @Stebalien Have made all the changes. Please take a look.

aarshkshah1992:
@anorth Can I merge this now ?

Issue #841, opened 2020-07-29
Incorrect faulty power in SubmitWindowPoSt


Stebalien:
This should add `newFaultPower`, not `newFaultPowerTotal`.

https://github.com/filecoin-project/specs-actors/blob/eba6261b58c671dab4efe5a7a139b59ae56b70f1/actors/builtin/miner/miner_actor.go#L342

I'm fixing this as part of #838, but want to track it so we don't forget (just in case).

Issue #829, opened 2020-07-28
Recover and extend miner actor tests


anorth:
A bunch of tests were disabled as part of the miner state refactor, and have not returned as quickly as we'd like.

anorth:
See also #828, which can push some testing down to state manipulations.

PR #844, opened 2020-07-29
recovery happy-path test

acruikshank:
work towards #829

### Motivation

Our fault recovery test had been commented out while miner actor state was being refactored. This test brings it back.

### Proposed Changes

1. Restore "successful recoveries recover power" to `TestWindowPost` test.
2. Fix bug in `SubmitWindowPoSt` logic related to #841 that caused proving period cron to continue penalizing sectors after recovery.


actors/builtin/miner/miner_test.go
```diff
@@ -782,58 +781,66 @@ func TestWindowPost(t *testing.T) {
...
-		//}
-		//
-		//actor.submitWindowPoSt(rt, deadline, partitions, infos, cfg)
+		// advance deadline and declare fault
+		advanceDeadline(rt, actor, &cronConfig{})
+		actor.declareFaults(rt, actor.declaredFaultPenalty(infos), infos...)
+
+		// advance a deadline and declare recovery
+		ongoingFee := actor.declaredFaultPenalty(infos)
+		advanceDeadline(rt, actor, &cronConfig{ongoingFaultsPenalty: ongoingFee})
```
anorth, line 804: Thanks for this. I think you've highlighted an error in implementation.

When a sector is declared faulty and then recovered all in the span of a single proving period (so it misses zero PoSts), the total penalty should be just 1*FF. It looks like we're charging two. This one is the one that should be removed, with the recovery fee paid only on successful proof of the recovered sector.

As well as fixing, could you please make a final assertion of the actor's balance at the end of the test, so the total fee is clear?

actors/builtin/miner/miner_test.go
```diff
@@ -782,58 +781,66 @@ func TestWindowPost(t *testing.T) {
...
+		// declare recovery
+		st := getState(rt)
+		dlIdx, pIdx, err := st.FindSector(rt.AdtStore(), infos[0].SectorNumber)
+		require.NoError(t, err)
+		actor.declareRecoveries(rt, dlIdx, pIdx, bf(uint64(infos[0].SectorNumber)))
+
+		// advance to epoch when submitPoSt is due
+		dlinfo := actor.deadline(rt)
+		for dlinfo.Index != dlIdx {
+			advanceDeadline(rt, actor, &cronConfig{ongoingFaultsPenalty: ongoingFee})
```
anorth, line 815: What's this fee? None of the other deadline have any sectors in them, so nothing can be faulty here...

PR #846, opened 2020-07-30
Make sure we don't generate empty expiration queue entries

Stebalien:
And test it.


Stebalien:
Based on #845.

Stebalien:
Fair enough. Also, I'm returning a nil bitfield which I definitely shouldn't be doing.

Stebalien:
Ok, all that code is wrong...

PR #847, opened 2020-07-30
Finish up deadline state testing

Stebalien:
We still need to improve our error-case testing (invalid sector numbers, invalid deadlines/partitions, etc.), but this should cover the rest of the non-error cases.

This patch covers:

* Retracted recoveries.
* Popping early terminations in multiple stages.

### DIFF
--- a/actors/builtin/miner/deadline_state_test.go
+++ b/actors/builtin/miner/deadline_state_test.go
@@ -214,6 +214,47 @@ func TestDeadlines(t *testing.T) {
 		require.Error(t, err, "should have failed to remove a partition with early terminations")
 	})
 
+	t.Run("can pop early terminations in multiple steps", func(t *testing.T) {
+		rt := builder.Build(t)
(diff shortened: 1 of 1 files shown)


Stebalien:
Based on #846.

Stebalien:
> The code uses "retracted" to refer to when a recovering sector is skipped in the PoSt, which is not tested here. Could you add a test for that case too (here or follow-up)?

I believe I test both ways of retracting recoveries:

1. Explicitly reporting a sector as faulty after reporting it as recovering: https://github.com/filecoin-project/specs-actors/pull/847/files#diff-322a2fc6c4d990d92b1126a5370c27ffR495
2. Skipping a sector that should have been recovered: https://github.com/filecoin-project/specs-actors/pull/847/files#diff-322a2fc6c4d990d92b1126a5370c27ffR448 (although the comment is outdated...)

I'll fix the comment & test description.

PR #848, opened 2020-07-30
Don't pay all fees every deadline and fix termination faulty power

anorth:
The tests in #844 revealed that we're charging the ongoing fault fee for _all_ faulty power every deadline, rather than charging for the faulty power for that deadline. Moving the `FaultyPower` memo from miner to deadline state prompted me to check that we're maintaining it on each relevant operation, and realise that sector termination was ignoring the faulty state of some power, incorrectly subtracting it again.

### DIFF
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -1039,6 +1039,14 @@ func (dl *Deadline) RecordProvenSectors(
 		return nil, xc.ErrIllegalState.Wrapf("failed to update expirations for partitions with faults: %w", err)
 	}
 
+	// Save everything back.
+	dl.FaultyPower = dl.FaultyPower.Sub(recoveredPowerTotal).Add(newFaultyPowerTotal)
+
+	dl.Partitions, err = partitions.Root()
+	if err != nil {
+		return nil, xc.ErrIllegalState.Wrapf("failed to persist partitions: %w", err)
+	}
+
 	// Collect all sectors, faults, and recoveries for proof verification.
 	allSectorNos, err := bitfield.MultiMerge(allSectors...)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


actors/builtin/miner/deadline_state.go
```diff
@@ -489,11 +497,12 @@ func (dl *Deadline) TerminateSectors(
 
 		// Record that partition now has pending early terminations.
 		dl.EarlyTerminations.Set(partIdx)
-		// Record live sector change
+		// Record change to sectors and power
 		dl.LiveSectors -= uint64(len(partSectors))
+		dl.FaultyPower = dl.FaultyPower.Sub(removed.FaultyPower)
 
-		// update power
-		removedPower = removedPower.Add(pwr)
```
anorth, line 496: This previously counted all power from terminated sectors, not only the active power.

actors/builtin/miner/miner_actor.go
```diff
@@ -1676,8 +1670,7 @@ func handleProvingDeadline(rt Runtime) {
 		{
 			// Record faulty power for penalisation of ongoing faults, before popping expirations.
 			// This includes any power that was just faulted from missing a PoSt.
-			faultyPower := st.FaultyPower.QA
```
anorth, line 1679: Oops, charging for all faulty power!

actors/builtin/miner/miner_actor.go
```diff
@@ -1124,17 +1121,15 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 			newFaultyPower, err := deadline.DeclareFaults(store, sectors, info.SectorSize, quant, faultExpirationEpoch, partitionsByDeadline[dlIdx])
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to declare faults for deadline %d", dlIdx)
 
-			newFaultPowerTotal = newFaultPowerTotal.Add(newFaultyPower)
-
 			err = deadlines.UpdateDeadline(store, dlIdx, deadline)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to store deadline %d partitions", dlIdx)
+
+			newFaultPowerTotal = newFaultPowerTotal.Add(newFaultyPower)
```
Stebalien, line 1127: Do we use this? I don't think we need to return it from `dl.DeclareFaults` either.

actors/builtin/miner/deadline_state.go
```diff
@@ -1033,17 +1049,20 @@ func (dl *Deadline) RecordProvenSectors(
...
 	}
 
-	return &PoStResult{
-		Sectors:                allSectorNos,
-		IgnoredSectors:         allIgnoredSectorNos,
-		NewFaultyPower:         newFaultyPowerTotal,
-		RecoveredPower:         recoveredPowerTotal,
-		RetractedRecoveryPower: retractedRecoveryPowerTotal,
-	}, nil
+	dl.FaultyPower = dl.FaultyPower.Sub(result.PowerDelta())
```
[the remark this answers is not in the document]
anorth: I was on the fence, will do that now.

actors/builtin/miner/miner_actor.go
anorth: Yes, it's used below for `requestUpdatePower` to inform the power actor of the change.

Issue #802, opened 2020-07-24
Miner.ChangeWorkerAddress: handled a pending change already exists


wadeAlexC:
The presence of a pending worker key change indicates that an unprocessed deferred cron event is scheduled to commit the change at a later date.

If the miner is allowed to overwrite `st.info.PendingWorkerKey`, they will cause an abort during the future deferred cron event.

Rather than allow this, simply abort if `info.PendingWorkerKey` is not `nil`

anorth:
I don't think we should black out the ability to initiate a key change because one is already in progress. But we should handle it gracefully, not abort.

I propose that we want the semantics that a new key change proposal replaces an existing one, and restarts the clock. We can achieve this simply by changing the `info.PendingWorkerKey == nil` and `info.PendingWorkerKey.EffectiveAt > rt.CurrEpoch()` cases to do nothing silently, rather than abort.

wadeAlexC:
The problem with allowing an existing proposal to replace an existing one is that in the current implementation, the cron event from the first pending change is still going to happen. Once that first cron tick executes, `commitPendingWorkerKeyChange` will abort because the new proposal has an expiration time that is in the future.

I think we should abort if a pending change already exists, so that we don't allow the miner to create cron callbacks that will definitely abort.

anorth:
That's a fair concern. As it is, the miner will only pay about half the total cost here in gas fees. I'll move that to a new issue.

Issue #574, opened 2020-07-02
Market: Tests for VerifiedDeals in PublishStorageDeals


aarshkshah1992:
In the market actor, we need to understand and test the code that talks to the `VerifiedRegistryActorAddr` in `PublishStorageDeals`.

@anorth FYI. Please give the labels.

PR #851, opened 2020-07-30
Allow owner key to act as miner

### LINKED ISSUE #547: Allow owner key to act as miner [was: while key change is in progress]
If a miner loses their worker key, they will be unable to submit Window PoSt for at least the duration of the key change delay (which is finality). This could cost a large amount of power and fees.

While an obvious remedy is "don't lose the key", in order to make this a less catastrophic scenario for a miner, we should allow their owner key to sign some messages while waiting for the change.

If a worker key change is pending, then allow the worker or owner key to sign:
- SubmitWindowPoSt
- DeclareFaults

Aside:
The key change delay is long because this same key is also used for election VRFs. The alternative of having two worker keys, with different roles and different delays, was considered and rejected as not worth the complexity.

### LINKED ISSUE #797: Reward.AwardBlockReward: Miner may be able to cause abort during block processing
`Reward.AwardBlockReward` relies on a successful Send to the Miner that produced the block:

https://github.com/filecoin-project/specs-actors/blob/70a18dfa12d7a0bc7981e73c6021c11d7de14cb3/actors/builtin/reward/reward_actor.go#L92-L93

The Miner's strategy is as follows:
* Wait until leader election allows the Miner to produce a block
* Intentionally create a consensus fault
* Create a transaction to Miner.ReportConsensusFault, and include it in the block

A valid ReportConsensusFault will end with a call to `terminateMiner`:

https://github.com/filecoin-project/specs-actors/blob/70a18dfa12d7a0bc7981e73c6021c11d7de14cb3/actors/builtin/miner/miner_actor.go#L2153-L2161

After terminating a Miner's deals, this invokes the runtime method `rt.DeleteActor`, which will remove the Miner from the state tree and cause subsequent Sends to the Miner to fail.

#### Recommendation

Remove the `RequireSuccess` condition from `Reward.AwardBlockReward`. This should never fail - and if it does, we should assume the miner is responsible.

aarshkshah1992:
Closes #547.

Tests for `miner.ChangeWorkerAddress` as part of #596 .


actors/builtin/reward/reward_actor.go
```diff
@@ -94,12 +94,12 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
 	AssertMsg(big.Add(rewardPayable, penalty).LessThanEqual(priorBalance),
 		"reward payable %v + penalty %v exceeds balance %v", rewardPayable, penalty, priorBalance)
 
-	_, code := rt.Send(minerAddr, builtin.MethodsMiner.AddLockedFund, &rewardPayable, rewardPayable)
-	builtin.RequireSuccess(rt, code, "failed to send reward to miner: %s", minerAddr)
+	// if this fails, we can assume the miner is responsible and do not to fail here
+	_, _ = rt.Send(minerAddr, builtin.MethodsMiner.AddLockedFund, &rewardPayable, rewardPayable)
```
Stebalien, line 98: Let's log this failure with `rt.Log`.

Stebalien: Also, @anorth, should we burn the funds if we fail to send them? What happens otherwise?

actors/builtin/miner/miner_actor.go
```diff
@@ -1080,7 +1086,13 @@ func (a Actor) DeclareFaults(rt Runtime, params *DeclareFaultsParams) *adt.Empty
 	newFaultPowerTotal := NewPowerPairZero()
 	rt.State().Transaction(&st, func() interface{} {
 		info := getMinerInfo(rt, &st)
-		rt.ValidateImmediateCallerIs(info.Worker)
+		// if a key change is in progress, we also allow the owner to call this
+		// otherwise, we only allow the worker.
+		if info.PendingWorkerKey == nil {
```
Stebalien, line 1091: I assume we also need to allow declaring faults recovered?

aarshkshah1992: Done.

aarshkshah1992:
@Stebalien Please take a look.

actors/builtin/reward/reward_actor.go
```diff
@@ -93,12 +93,15 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
 	AssertMsg(big.Add(rewardPayable, penalty).LessThanEqual(priorBalance),
 		"reward payable %v + penalty %v exceeds balance %v", rewardPayable, penalty, priorBalance)
 
+	// if this fails, we can assume the miner is responsible and do not to fail here
 	_, code := rt.Send(minerAddr, builtin.MethodsMiner.AddLockedFund, &rewardPayable, rewardPayable)
-	builtin.RequireSuccess(rt, code, "failed to send reward to miner: %s", minerAddr)
+	if !code.IsSuccess() {
+		rt.Log(vmr.ERROR, "failed to send AddLockedFund call to the miner actor with funds: %v, code: %v", rewardPayable, code)
```
Stebalien, line 99: @anorth should we turn move the reward to the penalty if paying the miner fails?

actors/builtin/miner/miner_actor.go
```diff
@@ -2035,12 +2053,8 @@ func commitWorkerKeyChange(rt Runtime) *adt.EmptyValue {
 	var st State
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
-		if info.PendingWorkerKey == nil {
-			rt.Abortf(exitcode.ErrIllegalState, "No pending key change.")
-		}
-
-		if info.PendingWorkerKey.EffectiveAt > rt.CurrEpoch() {
-			rt.Abortf(exitcode.ErrIllegalState, "Too early for key change. Current: %v, Change: %v)", rt.CurrEpoch(), info.PendingWorkerKey.EffectiveAt)
+		if info.PendingWorkerKey == nil || info.PendingWorkerKey.EffectiveAt > rt.CurrEpoch() {
```
Stebalien, line 2056: Can we add a comment here explaining what's going on? Specifically, the scheduled key change is no longer valid because we either aborted it or replaced it.

actors/builtin/miner/miner_actor.go
anorth: That was not omitted from the description in #547 by accident. The ability to declare recoveries is not nearly as necessary as the ability to declare faults, and thus excluded. However, given other discussion there it might make sense to include it. We'll resolve it in the issue.

actors/builtin/reward/reward_actor.go
anorth: Sorry, I can't parse your question.

anorth:
@aarshkshah1992 the reward actor change is unrelated to these others. Please submit it in a separate PR so that it can move forward independently.

actors/builtin/reward/reward_actor.go
Stebalien: (proofreading...)

If sending funds to a miner fails, what should we do with them? At the moment, it looks like the reward actor will just accrue funds. Given the fact that we're intentionally burning penalty below, I assume that any funds _not_ sent to the miner should also be burnt.

anorth: Yes, ok let's do that.

The reward actor accrues unpaid block rewards when the election winner just fails to get a block into the chain at all. But that maintains a property of the reward actor's balance decreasing exactly as expected given the blocks actually in the chain. We can retain that here by burning the reward (and also the gas reward that was sent here).

actors/builtin/reward/reward_actor.go
```diff
@@ -93,12 +93,15 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
 	AssertMsg(big.Add(rewardPayable, penalty).LessThanEqual(priorBalance),
 		"reward payable %v + penalty %v exceeds balance %v", rewardPayable, penalty, priorBalance)
 
+	// if this fails, we can assume the miner is responsible and do not to fail here
```
wadeAlexC, line 96: "If this fails, we can assume the miner is responsible and avoid failing here."

actors/builtin/miner/miner_actor.go
aarshkshah1992: @Stebalien This is done.

actors/builtin/reward/reward_actor.go
aarshkshah1992: This is done.

aarshkshah1992: This is done.

aarshkshah1992:
@anorth Moved the reward actor work to https://github.com/filecoin-project/specs-actors/pull/890.

anorth:
@aarshkshah1992 can you split out the change for #802 too? #574 is still not resolved.

aarshkshah1992:
@anorth PR for #802 in https://github.com/filecoin-project/specs-actors/pull/902.

aarshkshah1992:
Work for this PR has been subsumed by other PRs. The same privileges as the worker are now accorded to both the owner and the control addresses.

PR #857, opened 2020-07-31
Remove return value plumbing from runtime state transaction

### LINKED ISSUE #616: Remove return value from State.Transaction()

anorth:
This return value was a source of a couple of prior bugs, masked by Go's type checking semantics. I didn't find any more while making this change, so this is to prevent future ones.

Closes #616 

FYI @arajasek for integration, but I'll hold of landing this until after incentivized testnet starts.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -160,7 +160,7 @@ type ChangeWorkerAddressParams struct {
 func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams) *adt.EmptyValue {
 	var effectiveEpoch abi.ChainEpoch
 	var st State
-	rt.State().Transaction(&st, func() interface{} {
+	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
 
 		rt.ValidateImmediateCallerIs(info.Owner)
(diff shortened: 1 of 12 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1443,7 +1425,9 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 		rt.Abortf(exitcode.ErrIllegalArgument, "negative fund requested for withdrawal: %s", params.AmountRequested)
 	}
 	var info *MinerInfo
-	newlyVestedAmount := rt.State().Transaction(&st, func() interface{} {
+	newlyVested := big.Zero()
```
Stebalien, line 1428: I'd leave this (and the one above) zero initialized. That way, if we end up adding a stray colon below and fail to actually initialize this, this function will always panic.

anorth: Hmm. The reason I so conscientiously init these is that we had such panics on devnets in the past (though on an occasional code path).

Stebalien: I figured an obvious panic is better than a silent zero.

PR #859, opened 2020-07-31
Enforce BigIntMaxSerializedLen when marshalling CBOR

### LINKED ISSUE #817: Enforce `BigIntMaxSerializedLen` on `(*Int).MarshalCBOR()`
As it is enforced when encoding:

https://github.com/filecoin-project/specs-actors/blob/c47e382f64737e534890a6b286b867b0bb324cd9/actors/abi/big/int.go#L307-L309

To avoid any inconsistencies. Risk: low.

anorth:
Fixes #817

### DIFF
--- a/actors/abi/big/int.go
+++ b/actors/abi/big/int.go
@@ -55,10 +55,7 @@ func FromString(s string) (Int, error) {
 }
 
 func (bi Int) Copy() Int {
-	cpy := Int{}
-	cpy.Int = &big.Int{}
-	cpy.Int.Set(bi.Int)
-	return cpy
+	return Int{Int: new(big.Int).Set(bi.Int)}
 }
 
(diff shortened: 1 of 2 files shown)


actors/abi/big/int.go
```diff
@@ -285,7 +282,12 @@ func (bi *Int) MarshalCBOR(w io.Writer) error {
 		return err
 	}
 
-	header := cbg.CborEncodeMajorType(cbg.MajByteString, uint64(len(enc)))
+	encLen := len(enc)
+	if encLen > BigIntMaxSerializedLen {
+		return fmt.Errorf("big integer byte array too long (%d bytes)", encLen)
```
Stebalien, line 287: Could we add a test for this case?

anorth: Done

PR #861, opened 2020-07-31
Remove WalkSectors abstraction

### LINKED ISSUE #686: Remove WalkSectors abstraction in miner actor
It's not worth the complexity.

https://github.com/filecoin-project/specs-actors/pull/648#discussion_r455431725

Stebalien:
This patch introduces a new abstraction for walking sectors, but this one is
much simpler and reusable.

This patch also splits up sector expiration rescheduling into a state and a
deadline function (and pushes the partition invariants down into the partition
itself).

fixes #686

### DIFF
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -475,7 +475,7 @@ func (dl *Deadline) TerminateSectors(
 		if found, err := partitions.Get(partIdx, &partition); err != nil {
 			return xerrors.Errorf("failed to load partition %d: %w", partIdx, err)
 		} else if !found {
-			return xerrors.Errorf("failed to find partition %d", partIdx)
+			return xc.ErrNotFound.Wrapf("failed to find partition %d", partIdx)
 		}
 
 		sectorInfos, err := sectors.Load(sectorNos)
@@ -1056,9 +1056,18 @@ func (dl *Deadline) RecordProvenSectors(
 	}, nil
 }
 
+// RescheduleSectorExpirations reschedules the expirations of the given sectors
+// to the target epoch, skipping any sectors it can't find.
+//
+// The power of the rescheduled sectors is assumed to have not changed since
+// initial scheduling.
+//
+// Note: see the docs on State.RescheduleSectorExpirations for details on why we
+// skip sectors/partitions we can't find.
 func (dl *Deadline) RescheduleSectorExpirations(
-	store adt.Store, sectors Sectors, ssize abi.SectorSize, quant QuantSpec,
+	store adt.Store, sectors Sectors,
 	expiration abi.ChainEpoch, partitionSectors PartitionSectorMap,
+	ssize abi.SectorSize, quant QuantSpec,
 ) error {
 	partitions, err := dl.PartitionsArray(store)
 	if err != nil {
@@ -1069,18 +1078,20 @@ func (dl *Deadline) RescheduleSectorExpirations(
 	if err := partitionSectors.ForEach(func(partIdx uint64, sectorNos *abi.BitField) error {
 		var partition Partition
 		if found, err := partitions.Get(partIdx, &partition); err != nil {
-			return xc.ErrIllegalState.Wrapf("failed to load partition %d: %w", partIdx, err)
+			return xerrors.Errorf("failed to load partition %d: %w", partIdx, err)
 		} else if !found {
-			// We failed to find the partition. That's ok, we'll skip it.
+			// We failed to find the partition, it could have moved
+			// due to compaction. This function is only reschedules
+			// sectors it can find so we'll just skip it.
 			return nil
 		}
 
-		moved, err := partition.RescheduleExpirations(store, sectors, ssize, quant, expiration, sectorNos)
+		moved, err := partition.RescheduleExpirations(store, sectors, expiration, sectorNos, ssize, quant)
 		if err != nil {
 			return xerrors.Errorf("failed to reschedule expirations in partition %d: %w", partIdx, err)
 		}
 		if empty, err := moved.IsEmpty(); err != nil {
-			return xc.ErrIllegalState.Wrapf("failed to parse bitfield of rescheduled expirations: %w", err)
+			return xerrors.Errorf("failed to parse bitfield of rescheduled expirations: %w", err)
 		} else if empty {
 			// nothing moved.
 			return nil
@@ -1088,7 +1099,7 @@ func (dl *Deadline) RescheduleSectorExpirations(
 
 		rescheduledPartitions = append(rescheduledPartitions, partIdx)
 		if err = partitions.Set(partIdx, &partition); err != nil {
-			return xc.ErrIllegalState.Wrapf("failed to store partition %d: %w", partIdx, err)
+			return xerrors.Errorf("failed to store partition %d: %w", partIdx, err)
 		}
 		return nil
 	}); err != nil {
@@ -1098,7 +1109,7 @@ func (dl *Deadline) RescheduleSectorExpirations(
 	if len(rescheduledPartitions) > 0 {
 		dl.Partitions, err = partitions.Root()
 		if err != nil {
-			return xc.ErrIllegalState.Wrapf("failed to save partitions: %w", err)
+			return xerrors.Errorf("failed to save partitions: %w", err)
 		}
 		err := dl.AddExpirationPartitions(store, expiration, rescheduledPartitions, quant)
 		if err != nil {
--- a/actors/builtin/miner/deadline_state.go
+++ b/actors/builtin/miner/deadline_state.go
@@ -3,7 +3,6 @@ package miner
 import (
 	"bytes"
 	"errors"
-	"sort"
 
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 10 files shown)


Stebalien:
I suggest reviewing commit by commit.

actors/builtin/miner/deadline_state.go
```diff
@@ -469,40 +469,40 @@ func (dl *Deadline) TerminateSectors(
...
 	var partition Partition
-	for _, partIdx := range partitionIdxs {
-		partSectors := partitionSectors[partIdx]
+	if err := partitionSectors.ForEach(func(partIdx uint64, sectorNos *abi.BitField) error {
 		if found, err := partitions.Get(partIdx, &partition); err != nil {
-			return NewPowerPairZero(), xerrors.Errorf("failed to load partition %d: %w", partIdx, err)
+			return xerrors.Errorf("failed to load partition %d: %w", partIdx, err)
 		} else if !found {
-			return NewPowerPairZero(), xerrors.Errorf("failed to find partition %d", partIdx)
+			return xerrors.Errorf("failed to find partition %d", partIdx)
```
anorth, line 478: ErrNotFound?

actors/builtin/miner/deadline_state.go
```diff
@@ -1065,3 +1055,56 @@ func (dl *Deadline) RecordProvenSectors(
...
+		return err
+	}
+
+	var rescheduledPartitions []uint64 // track partitions with moved expirations.
+	if err := partitionSectors.ForEach(func(partIdx uint64, sectorNos *abi.BitField) error {
+		var partition Partition
+		if found, err := partitions.Get(partIdx, &partition); err != nil {
+			return xc.ErrIllegalState.Wrapf("failed to load partition %d: %w", partIdx, err)
+		} else if !found {
+			// We failed to find the partition. That's ok, we'll skip it.
```
anorth, line 1074: Why would this be ok? Could you give more context as to a state that could cause this? Is it due to compaction and the best-effort deadline-partition index queue?

actors/builtin/miner/deadline_state.go
```diff
@@ -1065,3 +1055,56 @@ func (dl *Deadline) RecordProvenSectors(
...
+			return nil
+		}
+
+		moved, err := partition.RescheduleExpirations(store, sectors, ssize, quant, expiration, sectorNos)
```
anorth, line 1078: Nit: I've been pushing the `ssize` and `qant` params to come last in every method they appear in.

actors/builtin/miner/deadline_state.go
```diff
@@ -1065,3 +1055,56 @@ func (dl *Deadline) RecordProvenSectors(
...
+	partitions, err := dl.PartitionsArray(store)
+	if err != nil {
+		return err
+	}
+
+	var rescheduledPartitions []uint64 // track partitions with moved expirations.
+	if err := partitionSectors.ForEach(func(partIdx uint64, sectorNos *abi.BitField) error {
+		var partition Partition
+		if found, err := partitions.Get(partIdx, &partition); err != nil {
+			return xc.ErrIllegalState.Wrapf("failed to load partition %d: %w", partIdx, err)
```
anorth, line 1072: Unnecessary exit code wrapping here and other error returns below.

actors/builtin/miner/miner_state.go
```diff
@@ -143,11 +143,6 @@ type SectorOnChainInfo struct {
 }
 
 // Location of a specific sector
```
anorth, line 145: Remove comment too

actors/builtin/miner/miner_state.go
```diff
@@ -418,190 +413,47 @@ func (st *State) FindSector(store adt.Store, sno abi.SectorNumber) (uint64, uint
...
 
-// Walks the given sectors, deadline by deadline, partition by partition,
-// skipping missing partitions/sectors.
-func (st *State) WalkSectors(
-	store adt.Store,
-	locations []SectorLocation,
-	beforeDeadlineCb func(dlIdx uint64, dl *Deadline) (update bool, err error),
-	partitionCb func(dl *Deadline, partition *Partition, dlIdx, partIdx uint64, sectors *bitfield.BitField) (update bool, err error),
-	afterDeadlineCb func(dlIdx uint64, dl *Deadline) (update bool, err error),
+// Schedule's each sector to expire at its next deadline end.
```
anorth, line 416:
```suggestion
// Schedules each sector to expire at its next deadline end.
```

actors/builtin/miner/miner_state.go
```diff
@@ -418,190 +413,47 @@ func (st *State) FindSector(store adt.Store, sno abi.SectorNumber) (uint64, uint
...
-		var deadlineUpdated bool
-		dl, err := deadlines.LoadDeadline(store, uint64(dlIdx))
+		dl, err := deadlines.LoadDeadline(store, dlIdx)
 		if err != nil {
 			return err
 		}
 
-		deadlineUpdated, err = beforeDeadlineCb(uint64(dlIdx), dl)
-		if err != nil {
+		if err := dl.RescheduleSectorExpirations(store, sectors, ssize, quant, newExpiration, pm); err != nil {
```
anorth, line 444: Nit: `ssize` and `quant` last.

actors/builtin/miner/partition_state.go
```diff
@@ -238,17 +238,46 @@ func (p *Partition) RemoveRecoveries(sectorNos *abi.BitField, power PowerPair) (
...
+		return nil, err
+	}
+
+	// Filter out terminated sectors.
+	live, err := bitfield.SubtractBitField(present, p.Terminated)
+	if err != nil {
+		return nil, err
+	}
+
+	// And faulty sectors.
```
anorth, line 257:
```suggestion
	// Filter out faulty sectors.
```

("and" is very context sensitive, and I misread it as "add")

actors/builtin/miner/sector_map.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Count counts the number of partitions & sectors within the map.
+func (dm DeadlineSectorMap) Count() (partitions, sectors uint64, err error) {
+	for dlIdx, pm := range dm { //nolint:nomaprange
+		partCount, sectorCount, err := pm.Count()
+		if err != nil {
+			return 0, 0, xerrors.Errorf("when counting deadline %d: %w", dlIdx, err)
+		}
+		if partCount > math.MaxUint64-partitions {
```
anorth, line 44: Go doesn't have overflow-safe arithmetic functions?

It's a great insight to add these checks. We should write our own arithmetic wrappers and do this everywhere.

actors/builtin/miner/sector_map.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Add records the given sector bitfield at the given deadline/partition index.
+func (dm DeadlineSectorMap) Add(dlIdx, partIdx uint64, sectorNos *bitfield.BitField) error {
+	if dlIdx >= WPoStPeriodDeadlines {
+		return xerrors.Errorf("invalid deadline %d", dlIdx)
+	}
+	dl, ok := dm[dlIdx]
+	if !ok {
+		dl = make(PartitionSectorMap, 1)
```
anorth, line 64: The capacity of `1` here isn't informed by knowledge of the eventual size, and may lead to more allocations (but only sometimes less memory). It's probably better to just leave it unspecified.

actors/builtin/miner/sector_map_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/filecoin-project/go-bitfield"
+	rlepluslazy "github.com/filecoin-project/go-bitfield/rle"
+	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDeadlineSectorMap(t *testing.T) {
+	dm := make(miner.DeadlineSectorMap)
```
anorth, line 16: Please also tests counts, iteration etc on an empty map, and map with an empty partition.

actors/builtin/miner/deadline_state.go
Stebalien: See the new comments.

actors/builtin/miner/deadline_state.go
Stebalien: Ah, I've been doing the opposite (common args first, important args later). But either way is fine as long as we're consistent.

actors/builtin/miner/sector_map.go
Stebalien: Unfortunately, no, they don't. It's not an issue in 99% of cases, but this is one of the few where it is (because someone can pass an arbitrarily large bitfield).

Of course, now that we're restricting bitfield runs to `2**63-1` instead of `2**64`, this may become a non-issue.

actors/builtin/miner/sector_map.go
Stebalien: How so? I know I'm going to add _at least_ one element in `dl.Add`, that's why I'm pre-allocating for it.

Looking at the source code, it looks like go always allocates at least one bucket (i.e., if the hint is 0-8, one bucket is allocated regardless).

anorth: You're probably more familiar with internals (tho they might change). But specifying 1 could cause the allocation of fewer buckets than specifying nothing, which would then cause a re-hashing later when it expands.

Stebalien: Well, there are two possible assumptions:

1. Go will always allocate enough space for at least one element.
2. Providing an allocation hint will never cause go to allocate less than it would otherwise.

Either way, we're making an assumption about how go works (and will work) internally. In this case, I'm making the second assumption. I had actually assumed that maps with no hint started at size 0 (which isn't the case), hence the hint of 1.

Stebalien: Note: I'm happy to change this, I just feel like it's clearer to tell the compiler "please make room for at least one element" when we know we'll need that space.

anorth: I had the opposite assumption for 2, but I'm happy to leave it as is.

Stebalien: Eh, you're right. It's better to avoid specifying things we don't actually know: https://github.com/filecoin-project/specs-actors/pull/898.

PR #863, opened 2020-07-31
Add some more reward constants

Stebalien:
TODO: Replace THING with what the thing actually is.

Actually, it would be really nice to have descriptions of all of these things.

----

#


actors/builtin/reward/reward_calc.py
```diff
@@ -15,6 +15,9 @@
 # Precision factor.
 Q128 = 2**128
 
+# TODO: what is this?
+THING=330000000
```
Stebalien, line 19: Ah, is this the supply total?

anorth: The simple supply total

actors/builtin/reward/reward_calc.py
```diff
@@ -26,6 +30,9 @@ def epochs_in_year() -> Decimal:
 def q128(val, round) -> str:
     return str((Q128 * val).to_integral_value(round))
 
+def atto(val) -> str:
+    return str((10**18 * val).to_integral_value(decimal.ROUND_DOWN))
```
Stebalien, line 34: @davidad I'm not finding any rhyme or reason to our rounding. Can we just safely round down (i.e., truncate)?

davidad: Fine with me. All our fixed-point arithmetic does truncation too, we just aim to have enough precision that it doesn't matter.

PR #870, opened 2020-07-31
recover skipped fault tests

acruikshank:
work towards #829

### Motivation

We had skipped many miner tests to simplify the miner state refactor. This PR brings some of them back

### Proposed Changes

1. Modify to `actorHarness.submitPoSt` so that it chooses sectors for validation from sectors that are not skipped and not previously faulty.
2. Restore "skipped faults are penalized and adjust power adjusted"
3. Restore "skipped all sectors in a deadline may be skipped"
4. Restore "skipped recoveries are penalized and do not recover power"
5. Restore "skipping a fault from the wrong partition is an error"
6. Delete "detects faults from previous missed posts" from `TestSubmitPoSt` because that behavior now occurs in cron.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -1221,7 +1221,7 @@ func TestProvingPeriodCron(t *testing.T) {
 		ongoingPenalty := miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, pwr.QA)
 
 		actor.onDeadlineCron(rt, &cronConfig{
-			expectedEntrollment:      dlinfo.NextPeriodStart(),
+			expectedEnrollment:       dlinfo.NextPeriodStart(),
 			detectedFaultsPenalty:    undetectedPenalty,
 			detectedFaultsPowerDelta: &powerDeltaClaim,
 			ongoingFaultsPenalty:     ongoingPenalty,
@@ -1252,7 +1252,7 @@ func TestProvingPeriodCron(t *testing.T) {
 		ongoingPenalty = miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, faultPwr.QA)
 
 		actor.onDeadlineCron(rt, &cronConfig{
-			expectedEntrollment:   dlinfo.Close,
+			expectedEnrollment:    dlinfo.Close,
 			detectedFaultsPenalty: retractedPenalty,
 			ongoingFaultsPenalty:  ongoingPenalty,
 		})
@@ -2066,7 +2066,7 @@ func (h *actorHarness) advancePastProvingPeriodWithCron(rt *mock.Runtime) {
 	rt.SetEpoch(deadline.PeriodEnd())
 	nextCron := deadline.NextPeriodStart() + miner.WPoStProvingPeriod - 1
 	h.onDeadlineCron(rt, &cronConfig{
-		expectedEntrollment: nextCron,
+		expectedEnrollment: nextCron,
 	})
 	rt.SetEpoch(deadline.NextPeriodStart())
 }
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1114,146 +1165,98 @@ func TestProvingPeriodCron(t *testing.T) {
...
+			dlinfo = advanceDeadline(rt, actor, &cronConfig{})
+		}
+
 		// The proving period start isn't changed, because the period hadn't started yet.
 		st = getState(rt)
 		assert.Equal(t, periodOffset, st.ProvingPeriodStart)
 
-		rt.SetEpoch(secondCronEpoch)
-		actor.onDeadlineCron(rt, &cronConfig{
-			expectedEntrollment: periodOffset + 2*miner.WPoStProvingPeriod - 1,
```
ZenGround0, line 1135: hmm enrollment still happens right?  If so it looks like we are no longer testing that it does.

actors/builtin/miner/miner_test.go
```diff
@@ -1114,146 +1165,98 @@ func TestProvingPeriodCron(t *testing.T) {
...
-		})
-		// Proving period moves forward
+		// next cron moves proving period forward
+		advanceDeadline(rt, actor, &cronConfig{})
 		st = getState(rt)
 		assert.Equal(t, periodOffset+miner.WPoStProvingPeriod, st.ProvingPeriodStart)
 	})
 
-	t.Run("first period gets randomness from previous epoch", func(t *testing.T) {
```
ZenGround0, line 1142: Is this functionality completely removed after the refactor?

actors/builtin/miner/miner_test.go
acruikshank: Enrollment still happens. The expectation is in `advanceDeadline`, so it is under test. The fact that you called it out, suggests that it should be more visible in the test. I'll use the more explicit version.

actors/builtin/miner/miner_test.go
acruikshank: It is very unclear what this is supposed to be testing. The code itself just appears to be going though a couple proving periods and occasionally setting the now non-existant `NewSectors`. The test description doesn't really make literal sense even when "first period" is well defined. I confirmed that every deadline in a proving period uses the same formula for `deadlineInfo.Challenge`, so maybe whatever behavior this is testing has gone away.

actors/builtin/miner/miner_test.go
```diff
@@ -2513,12 +2481,13 @@ func completeProvingPeriod(rt *mock.Runtime, h *actorHarness, config *cronConfig
 
 // Completes a deadline by moving the epoch forward to the penultimate one, calling the deadline cron handler,
 // and then advancing to the first epoch in the new deadline.
-func advanceDeadline(rt *mock.Runtime, h *actorHarness, config *cronConfig) {
+func advanceDeadline(rt *mock.Runtime, h *actorHarness, config *cronConfig) *miner.DeadlineInfo {
 	deadline := h.deadline(rt)
 	rt.SetEpoch(deadline.Last())
 	config.expectedEntrollment = deadline.Last() + miner.WPoStChallengeWindow
```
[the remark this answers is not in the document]
acruikshank: You gotta be entrolling me. Don't know how we missed this before.

actors/builtin/miner/miner_test.go
```diff
@@ -883,158 +883,212 @@ func TestWindowPost(t *testing.T) {
...
+		// Submit first PoSt to ensure we are sufficiently early to add a fault
+		// advance to next proving period
+		advanceAndSubmitPoSt(rt, actor, infos[0])
+
+		// advance deadline and declare fault
+		advanceDeadline(rt, actor, &cronConfig{})
+		actor.declareFaults(rt, infos...)
+
+		// advance a deadline and declare recovery
+		advanceDeadline(rt, actor, &cronConfig{})
```
anorth, line 1020: This shouldn't be necessary

PR #872, opened 2020-08-01
Reverse the penalty to sectorAge

### LINKED ISSUE #871: PledgePenaltyForTermination should be negatively correlative to SectorAge
See the implementation of `PledgePenaltyForTermination` from [here](https://github.com/filecoin-project/specs-actors/blob/866b3a6274eff7679d8de6d97d1846eec3fe509a/actors/builtin/miner/monies.go#L62-L75)

Sadly, the penalty is the most at the last day of a sector life, while the least at the first day. It looks the SectorRestOfLife should be used instead of SectorAge in this function.

steven004:
close #871


anorth:
I'm marking this as draft until discussion in #871 resolves.

anorth:
We won't be making this change, sorry.

PR #878, opened 2020-08-03
Split seal delays into randomness lookback and max porep duration

### LINKED ISSUE #672: Split MaxSealDuration into MaxPreCommitRandomnessLookback and MaxProveCommitDuration

ZenGround0:
Closes #672

fyi @nicola not changing any numbers yet but splitting out so we can.  Should we be changing either of these from 10,000?  If so this PR would be a natural place.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -472,7 +472,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		rt.Abortf(exitcode.ErrIllegalArgument, "seal challenge epoch %v must be before now %v", params.SealRandEpoch, rt.CurrEpoch())
 	}
 
-	challengeEarliest := sealChallengeEarliest(rt.CurrEpoch(), params.SealProof)
+	challengeEarliest := rt.CurrEpoch() - ChainFinality - MaxPreCommitRandomnessLookback
 	if params.SealRandEpoch < challengeEarliest {
 		rt.Abortf(exitcode.ErrIllegalArgument, "seal challenge epoch %v too old, must be after %v", params.SealRandEpoch, challengeEarliest)
 	}
@@ -1975,12 +1975,6 @@ func getVerifyInfo(rt Runtime, params *SealVerifyStuff) *abi.SealVerifyInfo {
 		rt.Abortf(exitcode.ErrForbidden, "too early to prove sector")
 	}
 
-	// Check randomness.
-	challengeEarliest := sealChallengeEarliest(rt.CurrEpoch(), params.RegisteredSealProof)
-	if params.SealRandEpoch < challengeEarliest {
-		rt.Abortf(exitcode.ErrIllegalArgument, "seal epoch %v too old, expected >= %v", params.SealRandEpoch, challengeEarliest)
-	}
-
 	commD := requestUnsealedSectorCID(rt, params.RegisteredSealProof, params.DealIDs)
 
 	minerActorID, err := addr.IDFromAddress(rt.Message().Receiver())
@@ -2271,11 +2265,6 @@ func PowerForSectors(ssize abi.SectorSize, sectors []*SectorOnChainInfo) PowerPa
 		Raw: big.Mul(big.NewIntUnsigned(uint64(ssize)), big.NewIntUnsigned(uint64(len(sectors)))),
 		QA:  qa,
 	}
-}
-
-// The oldest seal challenge epoch that will be accepted in the current epoch.
-func sealChallengeEarliest(currEpoch abi.ChainEpoch, proof abi.RegisteredSealProof) abi.ChainEpoch {
-	return currEpoch - ChainFinality - MaxPreCommitRandomnessLookback
 }
 
 func getMinerInfo(rt Runtime, st *State) *MinerInfo {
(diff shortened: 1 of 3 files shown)


anorth:
We should use this to decouple the usages of the two constants in `sealChallengeEarliest`; ambiguity over the word "seal" in the issue obscures this.

The `MaxPreCommitRandomnessLookback` should be the maximum lookback _from pre-commit_, evaluated at pre-commit time, and never used again. Then `MaxProveCommitDuration` should be the maximum delay between pre-commit and proof. There should be no checking of the seal randomness in ProveCommit. The comments around links 403-405 should disappear because the check is simple and no longer speculative.

nicola:
Correct, the numbers should be different - we will try to give you a better number soon.

ZenGround0:
> There should be no checking of the seal randomness age in ProveCommit.

I believe this is the current state.  We only check prove commit duration during prove commit .

> The comments around links 403-405 should disappear because the check is simple and no longer speculative.

Done.

> Correct, the numbers should be different - we will try to give you a better number soon.

Great.  I'll wait to land this until we have better numbers

actors/builtin/miner/miner_actor.go
```diff
@@ -400,11 +400,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		rt.Abortf(exitcode.ErrIllegalArgument, "seal challenge epoch %v must be before now %v", params.SealRandEpoch, rt.CurrEpoch())
 	}
 
-	challengeEarliest := sealChallengeEarliest(rt.CurrEpoch(), params.SealProof)
+	challengeEarliest := rt.CurrEpoch() - ChainFinality - MaxPreCommitRandomnessLookback
```
anorth, line 403: I don't think `ChainFinality` should be part of this. I'm guessing you preserved it from the prior state. @nicola could you please confirm?

nicola: It will not, we will set `MaxPreCommitRandomnessLookback` in the right way (if it will need finality, we will do so in the param)

actors/builtin/miner/miner_actor.go
```diff
@@ -1869,12 +1866,6 @@ func getVerifyInfo(rt Runtime, params *SealVerifyStuff) *abi.SealVerifyInfo {
 		rt.Abortf(exitcode.ErrForbidden, "too early to prove sector")
 	}
 
-	// Check randomness.
-	challengeEarliest := sealChallengeEarliest(rt.CurrEpoch(), params.RegisteredSealProof)
-	if params.SealRandEpoch < challengeEarliest {
-		rt.Abortf(exitcode.ErrIllegalArgument, "seal epoch %v too old, expected >= %v", params.SealRandEpoch, challengeEarliest)
-	}
-
```
nicola, line 1877: Why are we not checking this anymore?

ZenGround0: This check is incorrect.  This function is only called during prove commit.  The intention behind the check is to ensure the precommit seal rand epoch is not too far back from the precommit epoch.  However by using current epoch it is checking that the difference between prove commit time and precommit seal rand epoch is less than max duration between precommit and precommit seal rand epoch.  So this is overly punitive.

On top of the incorrectness the `params.SealRandEpoch` value comes from an onchain value already correctly checked [here](https://github.com/filecoin-project/specs-actors/pull/878/files#diff-e47fb4c42d791672c9550edb1dc5f969R404), so it is fine to remove this check.

nicola:
Safe numbers:
- MaxPreCommitRandomnessLookback: 12h worth of epochs + finality
- MaxProveCommitDuration: 5h worth of epochs

We will change these numbers to be tighter

anorth:
@nicola re `MaxProveCommitDuration`, `PreCommitChallengeDelay` is 150 (1.5h), so a miner must
- wait 1.5 hr + settle time
- take the challenge and begin proof
- submit before a further 3.5hr has passed

Did you want to add `PreCommitChallengeDelay` to that time, or is it already incorporated? Certainly it would have to be added if it were larger, like the Finality value proposed some time ago.

anorth:
Also `MaxProveCommitDuration` depends on sector size. I interpret your number for 32GiB sectors. Is it the same for 64GiB?

(I'll use the same for the smaller sector sizes anyway).

anorth:
@magik6k @arajasek this max prove commit delay looks like it might be a moderately tight bound on proof times. Is Lotus prepared for this?

anorth:
@ZenGround0 I have added numbers and rebased on master. Need @nicola's ack on the implications and Lotus ok. If Lotus don't want it right now, we could land it with the 10000 numbers instead and follow-up later.

nicola:
hey @anorth,

1. I had a misunderstanding on the usage of MaxProveCommitDuration. In my understanding it should be from the time the challenge is available. We can solve this in two ways: either set `MaxProveCommitDuration=5h + PreCommitChallengeDelay` or explicitly add `PreCommitChallengeDelay` where needed in the code.
2. MaxPreCommitRandomnessLookback should be 24h + finality for 64GiB (and we can re-use the same for 32GiB)
3. I am using these numbers based on my assumption that it takes maximum ~24h to PreCommit 64GiB, ~12h to PreCommit 32GiB and maximum 5h to generate a valid ProveCommit. If this is not the case, we should change these numbers accordingly. Maybe @porcuquine, @magik6k  can chip in with better timing that they have witnessed

anorth:
Ok. We'll check the numbers match impl reality, and we can use different values for different sector size.

PR #876, opened 2020-08-03
Move fault handling logic into partition methods

Stebalien:
Move fault handling logic into partition methods. This way, partition invariants can be enforced and tested on partition objects themselves.

Based on #875.
fixes #773.

PR #879, opened 2020-08-03
Remove runtime from partition/deadline state assertions

Stebalien:
This way, we can use these assertions from the miner state tests.

Based on #876. This is an entirely mechanical refactor paving the way for some state tests.

### DIFF
--- a/actors/builtin/miner/deadline_state_test.go
+++ b/actors/builtin/miner/deadline_state_test.go
@@ -55,8 +55,7 @@ func TestDeadlines(t *testing.T) {
 	// Partition 1: sectors 1, 2, 3, 4
 	// Partition 2: sectors 5, 6, 7, 8
 	// Partition 3: sectors 9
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/deadline_state_test.go
```diff
@@ -162,64 +157,61 @@ func TestDeadlines(t *testing.T) {
...
+			).assert(t, store, dl)
 	}
 
 	// Test the basic scenarios (technically, we could just run the final one).
 
 	t.Run("adds sectors", func(t *testing.T) {
-		rt := builder.Build(t)
-		dl := emptyDeadline(t, rt)
-		addSectors(t, rt, dl)
+		store := adt.AsStore(builder.Build(t))
```
anorth, line 166: It would be even better if we could avoid instantiating the mock runtime at all!

Try `NewADTStore` in `support/ipld.go`

PR #880, opened 2020-08-03
Quantize sector expiration queues per-deadline

### LINKED ISSUE #739: Switch to deadline specific quantization for expiration queues
Currently, we round up the expiration of a sector to the next deadline. However, we really should be rounding up to the next instance of the deadline to which the sector has been assigned as we won't actually expire it until we hit that deadline.

Stebalien:
This also adds a test for sector assignment which should test this quantization.

fixes #739

### DIFF
--- a/actors/builtin/miner/miner_state_test.go
+++ b/actors/builtin/miner/miner_state_test.go
@@ -546,6 +546,68 @@ func TestVestingFunds_UnvestedFunds(t *testing.T) {
 
 }
 
+func TestSectorAssignment(t *testing.T) {
+	partitionSectors, err := abi.RegisteredSealProof_StackedDrg32GiBV1.WindowPoStPartitionSectors()
+	require.NoError(t, err)
+	sectorSize, err := abi.RegisteredSealProof_StackedDrg32GiBV1.SectorSize()
+	require.NoError(t, err)
+
+	openDeadlines := miner.WPoStPeriodDeadlines - 2
+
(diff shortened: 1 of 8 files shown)


actors/builtin/miner/miner_state_test.go
```diff
@@ -546,6 +546,68 @@ func TestVestingFunds_UnvestedFunds(t *testing.T) {
...
+
+			var partitions []*bitfield.BitField
+			for i := uint64(0); i < uint64(partitionsPerDeadline); i++ {
+				var runs []rlepluslazy.Run
+				start := ((i * openDeadlines) + (dlIdx - 2)) * partitionSectors
+				if start > 0 {
+					runs = append(runs, rlepluslazy.Run{Val: false, Len: start})
+				}
+				runs = append(runs, rlepluslazy.Run{Val: true, Len: partitionSectors})
+				bf, err := bitfield.NewFromIter(&rlepluslazy.RunSliceIterator{Runs: runs})
```
anorth, line 598: This is a bit clever. I infer the goal is just to allocate sequential sector numbers. I think a simple loop over `bf.Set` or a `[]uint64` would be much clearer. There used to be a `seq` helper function just for that.

Stebalien: I've added a `seq` helper and used it in several places. I kept the run slice iterator trick as some cases needed really long sequences (~max uint64).

PR #885, opened 2020-08-04
Market actor cleanups

anorth:
Via @austinabell


aarshkshah1992:
Thanks !

actors/builtin/market/market_actor.go
```diff
@@ -350,12 +350,12 @@ type ComputeDataCommitmentParams struct {
 func (a Actor) ComputeDataCommitment(rt Runtime, params *ComputeDataCommitmentParams) *cbg.CborCid {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 
-	pieces := make([]abi.PieceInfo, 0)
 	var st State
 	rt.State().Readonly(&st)
 	proposals, err := AsDealProposalArray(adt.AsStore(rt), st.Proposals)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deal dealProposals")
 
+	pieces := make([]abi.PieceInfo, 0)
```
austinabell, line 358: The point was actually to set the capacity on this, with adding `, len(params.DealIDs)`, just to avoid reallocations, but this doesn't matter

PR #889, opened 2020-08-04
Market: Randomize first cron tick post the startepoch for deal processing

### LINKED ISSUE #813: Randomise deal state processing epochs to avoid load imbalance
The epochs at which deal states are processed by the market actor cron are distributed as their start epochs. These start epochs are controlled from the outside, so there's an avenue for causing a large chain load imbalance by posting many deals with matching start epoch modulo `DealUpdatesInterval`.

We should randomise the first epoch after StartEpoch at which a deal is scheduled for regular payment processing. Use chain randomness to seed this.

FYI @whyrusleeping @zixuanzh

aarshkshah1992:
Closes #813.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -1,9 +1,12 @@
 package market
 
 import (
+	"bytes"
+	"encoding/binary"
 	"sort"
 
 	addr "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/specs-actors/actors/crypto"
 	cbg "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
 
@@ -491,6 +494,7 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 
 				slashAmount, nextEpoch, removeDeal := msm.updatePendingDealState(rt, state, deal, rt.CurrEpoch())
 				Assert(slashAmount.GreaterThanEqual(big.Zero()))
+
 				if removeDeal {
 					AssertMsg(nextEpoch == epochUndefined, "next scheduled epoch should be undefined as deal has been removed")
 
@@ -500,6 +504,22 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 				} else {
 					AssertMsg(nextEpoch > rt.CurrEpoch() && slashAmount.IsZero(), "deal should not be slashed and should have a schedule for next cron tick"+
 						" as it has not been removed")
+
+					// If this is the first cron tick for the deal, we should  randomize the next epoch
+					// for when it will be processed so an attacker isn't able to schedule too many deals for the same tick.
+					if state.LastUpdatedEpoch == epochUndefined {
+						buf := bytes.Buffer{}
+						err = deal.MarshalCBOR(&buf)
+						builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to marshal proposal")
+
+						rb := rt.GetRandomness(crypto.DomainSeparationTag_MarketNextCronSchedule, rt.CurrEpoch()-1, buf.Bytes())
+
+						// generate a random epoch in [currentEpoch + 1, currentEpoch + DealUpdatesInterval]
+						var offset uint64
+						err = binary.Read(bytes.NewBuffer(rb), binary.BigEndian, &offset)
+						builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to generate offset")
+						nextEpoch = rt.CurrEpoch() + abi.ChainEpoch(offset%uint64(DealUpdatesInterval)) + 1
+					}
 
 					// Update deal's LastUpdatedEpoch in DealStates
 					state.LastUpdatedEpoch = rt.CurrEpoch()
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -219,7 +219,12 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			err = msm.dealProposals.Set(id, &deal.Proposal)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal")
 
-			err = msm.dealsByEpoch.Put(deal.Proposal.StartEpoch, id)
+			// We should randomize the first epoch for when the deal will be processed so an attacker isn't able to
+			// schedule too many deals for the same tick.
+			processEpoch, err := genRandNextEpoch(deal.Proposal.StartEpoch, rt.CurrEpoch(), &deal.Proposal, rt.GetRandomness)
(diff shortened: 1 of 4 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -501,6 +505,22 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
 					AssertMsg(nextEpoch > rt.CurrEpoch() && slashAmount.IsZero(), "deal should not be slashed and should have a schedule for next cron tick"+
 						" as it has not been removed")
 
+					// If this is the first cron tick for the deal, we should  randomize the next epoch
+					// for when it will be processed so an attacker isn't able to schedule too many deals for the same tick.
+					if state.LastUpdatedEpoch == epochUndefined {
```
anorth, line 510: This should be done in PublishStorageDeals instead, thus randomising _all_ processing by cron. As it stands, an attacker can still stack a single cron epoch with a flood of new deals (but then they're distributed after that).

Also please break the logic below out to a free function so it doesn't clutter up whichever method it's called from.

actors/builtin/market/market_state.go
```diff
@@ -149,9 +149,6 @@ func (m *marketStateMutation) updatePendingDealState(rt Runtime, state *DealStat
 	}
 
 	nextEpoch = epoch + DealUpdatesInterval
-	if nextEpoch > deal.EndEpoch {
-		nextEpoch = deal.EndEpoch
-	}
```
anorth, line 154: Thanks. Please leave a comment here stating that we're explicitly not inspecting the end epoch and may process a deal's expiration late, in order to avoid outside influence over cron loading.

actors/crypto/randomness.go
```diff
@@ -11,4 +11,5 @@ const (
 	DomainSeparationTag_SealRandomness
 	DomainSeparationTag_InteractiveSealChallengeSeed
 	DomainSeparationTag_WindowedPoStDeadlineAssignment
+	DomainSeparationTag_MarketNextCronSchedule
```
anorth, line 14:
```suggestion
	DomainSeparationTag_MarketDealCronSeed
```

actors/builtin/market/market_test.go
```diff
@@ -1506,7 +1582,7 @@ func TestLockedFundTrackingStates(t *testing.T) {
 
 	// deal1 and deal2 will now be charged at epoch curr + market.DealUpdatesInterval, so nothing changes before that.
 	rt.SetEpoch(curr + market.DealUpdatesInterval - 1)
-	actor.cronTick(rt)
+	actor.cronTick(rt, curr+market.DealUpdatesInterval-1)
```
anorth, line 1585: I haven't gone into the tests in detail since I expect them to change when the randomization moves to deal publishing, but I'm hoping that the testing cron calls will not need crafted arguments since there's nothing nondeterministic going on.

actors/builtin/market/market_test.go
```diff
@@ -1434,11 +1440,80 @@ func TestCronTick(t *testing.T) {
...
+		require.NotNil(t, actor.getDealProposal(rt, dealId))
+
+		// cron tick at nextEpoch expires the deal -> payment is ONLY for one epoch
+		rt.SetEpoch(nextEpoch)
+		pay, _ = actor.cronTickAndAssertBalances(rt, client, provider, nextEpoch, dealId, undefinedEpoch)
+		require.EqualValues(t, d.StoragePricePerEpoch, pay)
+		actor.assertDealDeleted(rt, dealId, d)
+	})
+}
+
```
anorth, line 1516: Since deal states are no longer definitely processed on their start or end epoch, we need higher level tests for new boundary conditions:
- activation after deal start epoch, but before it was processed (should fail)
- cron processing of deal after missed activation ↑
- termination of deal after it already expired, but before expiration was processed (should not slash)

actors/builtin/market/market_actor.go
aarshkshah1992: This is done.

actors/builtin/market/market_test.go
aarshkshah1992: All three have been written. Please take a look at `TestRandomCronEpochDuringPublish`.

actors/builtin/market/market_actor.go
```diff
@@ -216,7 +219,12 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			err = msm.dealProposals.Set(id, &deal.Proposal)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal")
 
-			err = msm.dealsByEpoch.Put(deal.Proposal.StartEpoch, id)
+			// We should randomize the first epoch for when the deal will be processed so an attacker isn't able to
+			// schedule too many deals for the same tick.
+			processEpoch, err := genRandNextEpoch(deal.Proposal.StartEpoch, rt.CurrEpoch(), &deal.Proposal, rt.GetRandomness)
```
anorth, line 224: Since you pass the proposal anyway, don't bother passing the start epoch separately.

aarshkshah1992: This is done.

actors/builtin/market/market_actor.go
```diff
@@ -563,13 +572,29 @@ func (a Actor) CronTick(rt Runtime, _ *adt.EmptyValue) *adt.EmptyValue {
...
+	buf := bytes.Buffer{}
+	if err := deal.MarshalCBOR(&buf); err != nil {
+		return epochUndefined, xerrors.Errorf("failed to marshal proposal: %w", err)
+	}
+
+	rb := rbF(crypto.DomainSeparationTag_MarketDealCronSeed, currEpoch-1, buf.Bytes())
+
+	// generate a random epoch in [baseEpoch, baseEpoch + DealUpdatesInterval)
+	var offset uint64
+	if err := binary.Read(bytes.NewBuffer(rb), binary.BigEndian, &offset); err != nil {
```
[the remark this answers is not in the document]
Stebalien: Ok, we're guaranteed 32 bytes of randomness.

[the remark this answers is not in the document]
aarshkshah1992: This is done.

actors/builtin/market/market_actor.go
```diff
@@ -216,7 +219,12 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 			err = msm.dealProposals.Set(id, &deal.Proposal)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal")
 
-			err = msm.dealsByEpoch.Put(deal.Proposal.StartEpoch, id)
+			// We should randomize the first epoch for when the deal will be processed so an attacker isn't able to
+			// schedule too many deals for the same tick.
+			processEpoch, err := genRandNextEpoch(rt.CurrEpoch(), &deal.Proposal, rt.GetRandomness)
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

PR #890, opened 2020-08-04
Reward Actor: Do not abort on sending funds to the miner

### LINKED ISSUE #797: Reward.AwardBlockReward: Miner may be able to cause abort during block processing

aarshkshah1992:
Closes #797.

### DIFF
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -93,8 +93,15 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams) 
 	AssertMsg(big.Add(rewardPayable, penalty).LessThanEqual(priorBalance),
 		"reward payable %v + penalty %v exceeds balance %v", rewardPayable, penalty, priorBalance)
 
+	// if this fails, we can assume the miner is responsible and avoid failing here.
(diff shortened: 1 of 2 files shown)


wadeAlexC:
Could you add a test that checks the failure case mentioned in #797 ?

PR #891, opened 2020-08-04
proposal requester must be ID addresses err msg

exp0nge:
It's not clear that `requester` must be an ID address where as `params.to` can be either, so let's at least make the error clear


anorth:
Thanks!

PR #892, opened 2020-08-04
restore report consensus fault test, and fix terminate sectors bug

acruikshank:
work towards #829

### Motivation

The ReportConsensusFaults tests had been skipped for the miner state refactor. This PR restores that test and fixes a bug

### Proposed changes

1. Fix bug in `miner.requestTerminateDeals` that puts it in an infinite send loop if deals are present.
2. Remove skipped "faulty committed capacity sector not replaced". The case is covered in the test before it.
3. Restore report consensus fault test.
4. Add report consensus fault test for the case where sectors containing more than 8k deals are being terminating (resulting in multiple sends to the market actor's `OnMinerSectorsTerminate`.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1775,7 +1775,7 @@ func requestTerminateDeals(rt Runtime, epoch abi.ChainEpoch, dealIDs []abi.DealI
 			abi.NewTokenAmount(0),
 		)
 		builtin.RequireSuccess(rt, code, "failed to terminate deals, exit code %v", code)
-		dealIDs = dealIDs[:size]
+		dealIDs = dealIDs[size:]
 	}
 }
 
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -559,60 +559,6 @@ func TestCommitments(t *testing.T) {
 		rt.Verify()
 	})
 
-	t.Run("faulty committed capacity sector not replaced", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1774,7 +1774,7 @@ func requestTerminateDeals(rt Runtime, epoch abi.ChainEpoch, dealIDs []abi.DealI
 			abi.NewTokenAmount(0),
 		)
 		builtin.RequireSuccess(rt, code, "failed to terminate deals, exit code %v", code)
-		dealIDs = dealIDs[:size]
```
Stebalien, line 1777: Great catch!

actors/builtin/miner/miner_test.go
```diff
@@ -1649,32 +1595,66 @@ func TestWithdrawBalance(t *testing.T) {
...
-	dealIDs := [][]abi.DealID{{1, 2}, {3, 4}}
-	sectorInfo := actor.commitAndProveSectors(rt, 2, 10, dealIDs)
-	_ = sectorInfo
-
-	params := &miner.ReportConsensusFaultParams{
-		BlockHeader1:     nil,
-		BlockHeader2:     nil,
-		BlockHeaderExtra: nil,
-	}
+	t.Run("Report consensus fault terminates deals when multipler sectors have multiple deals", func(t *testing.T) {
```
Stebalien, line 1603:
```suggestion
	t.Run("Report consensus fault terminates deals when multiple sectors have multiple deals", func(t *testing.T) {
```

PR #896, opened 2020-08-05
Update ReportConsensusFault Parameters

zixuanzh:
To balance incentives for prompt consensus fault reporting and self slashing to unvest rewards, maxReporterShare is set to 1/20. 

Previously, reporterShare starts from 1/1000, growing exponentially over finality (500 epochs), and capped at 1/2. This is now starting from 1/1000, growing exponentially over finality (900 epochs), and capped at 1/20.

Note that when improving consensus fault slashing penalty, the remaining of the penalty is burned (currently done by `terminateMiner`).

cc @nicola @sa8 @davidad @anorth


actors/builtin/miner/policy.go
```diff
@@ -171,8 +171,8 @@ var consensusFaultReporterInitialShare = BigFrac{
 }
 var consensusFaultReporterShareGrowthRate = BigFrac{
 	// PARAM_FINISH
-	numerator:   big.NewInt(101251),
-	denominator: big.NewInt(100000),
+	numerator:   big.NewInt(1004356),
+	denominator: big.NewInt(1000000),
```
anorth, line 175: I gather that this ratio is derived from finality. Please document the arithmetic involved here or, even better, do the calculation of numerator in code if possible.

nicola:
1. Why are we maxing this out at finality? The earlier is reported the better no?
2. Since we are abandoning miner termination and since the collateral lost is now much smaller, it may make very little sense to start from `1/1000`. Depending on the consensus fault fee value, 1/1000 may not be enough to gain from transaction fees.

zixuanzh:
@nicola There are a few subtle issues here

1. The earlier it is reported the better and how much there is to gain after subtracting the transaction fee.

This is handled by the mechanism being strategically equivalent to a Dutch auction. Whichever slasher that is willing to do it for the lowest reporter reward (minus transaction fee) will get the job.

2. Why are we maxing this out at finality?

There is no requirement from economics to max out at finality that I know of. It was taken as a consensus-related requirement. Are you or @sa8 aware of any requirement on the max out duration? 

3. Since the collateral lost is now much smaller, it may make very little sense to start from 1/1000

It depends on the size of the penalty. We can revisit when we have more clarity on  its magnitude.

davidad:
My sense is that, from a consensus PoV, it’s important for the share to max out before finality (so that potential reporters aren’t waiting until after finality to get a bigger share), but indeed, earlier is better. The only reason I know of to make it grow slower rather than faster is to improve the effective resolution of the Dutch auction’s price discovery mechanism (basically, the auction's minimum bid increment). With less than, say, 100 epochs before maxing out, the auction starts to look pretty coarse. But 100 should be fine, and it wouldn't be terrible to go as low as 40.

nicola:
Ok my worry: it takes X blocks for the transaction fees to make it worth to report the fault. We want this X to be as small as possible.

zixuanzh:
ConsensusFaultReporterShare grows from 1/1000 to 1/20 over 500 epoch.

anorth (approved):

Issue #895, opened 2020-08-05
Switch to bitfields by value


Stebalien:
Currently, we're passing around bitfields by reference. Unfortunately, we're not carefully checking for null pointers (see #604).

Plan:

* [ ] Switch all bitfields to by-value, both in specs actors and in lotus
* [ ] If reasonable, make bitfields immutable.

Pull requests:

* [x] Bitfield PR: https://github.com/filecoin-project/go-bitfield/pull/44
* [x] Specs actors PR: https://github.com/filecoin-project/specs-actors/pull/900
* [ ] Lotus PR: https://github.com/filecoin-project/lotus/pull/2839

anorth:
@magik6k @arajasek we'll work to make this as easy for you to consume as possible. Please ping us on a sync channel when you eventually consume this if you encounter any friction or would like to recruit our help. 

We expect this change to eliminate a class of potential errors introduced by the possibility of nil pointers.

anorth:
@Stebalien Done?

Stebalien:
Ah, yes.

PR #900, opened 2020-08-05
Switch to by-value bitfields

Stebalien:
Unfortunately, by-pointer bitfields meant we needed to check for nil bitfields in function parameters. Given the fact that we rarely do this, it's safer to just use bitfields by-value.

Note: "zero"/uninitialized bitfields are still valid for reading. However, updating a zero-value/uninitialized bitfield will cause a panic.

Part of https://github.com/filecoin-project/specs-actors/issues/895
Depends on https://github.com/filecoin-project/go-bitfield/pull/44

### DIFF
--- a/actors/abi/bitfield.go
+++ b/actors/abi/bitfield.go
@@ -1,21 +1,11 @@
 package abi
 
 import (
-	"fmt"
-
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-bitfield/rle"
 )
 
 type BitField = bitfield.BitField
-
-func NewBitField() *BitField {
-	bf, err := bitfield.NewFromBytes([]byte{})
-	if err != nil {
-		panic(fmt.Sprintf("creating empty rle: %+v", err))
-	}
-	return &bf
-}
 
 func isEmpty(iter rlepluslazy.RunIterator) (bool, error) {
 	// Look for the first non-zero bit.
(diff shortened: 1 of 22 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -38,7 +38,7 @@ const (
 
 type CronEventPayload struct {
 	EventType CronEventType
-	Sectors   *abi.BitField
+	Sectors   bitfield.BitField
```
Stebalien, line 41: Should this one still be by-pointer? Technically, an uninitialized bitfield will correctly serialize to an empty bitfield, so we don't _have_ to make this nullable.

actors/builtin/miner/miner_actor.go
```diff
@@ -1139,7 +1139,7 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 
 type CompactPartitionsParams struct {
 	Deadline   uint64
-	Partitions *abi.BitField
+	Partitions bitfield.BitField
```
Stebalien, line 1142: note: I switched from `abi.BitField` to `bitfield.BitField` to be consistent. We can also go the other way if we want to alias everything inside `abi`.

actors/abi/bitfield.go
```diff
@@ -9,12 +9,12 @@ import (
 
 type BitField = bitfield.BitField
 
-func NewBitField() *BitField {
+func NewBitField() BitField {
```
anorth, line 12: Remove this if we no longer need it

actors/builtin/miner/miner_actor.go
anorth: This is fine. I think we'll be able to delete this soon anywa.

actors/builtin/miner/miner_actor.go
anorth: The name stuttering is unfortunate, but generally 👍 ; aliasing was an interim thing when the bitfield package didn't have everything we needed.

actors/abi/bitfield.go
Stebalien: Lotus uses it, but I can fix that pretty easily.

PR #901, opened 2020-08-05
Miner: Efficiently expire pre-committed sectors

### LINKED ISSUE #417: Quantize sector pre-commit expiry epochs to avoid per-sector cron registration
When each sector is pre-committed, a cron event is registered for the expiry of that precommit, which penalises a miner for failing to follow through. This creates a huge load of cron registrations in state, and processing in the tipset cron handler. For high-scale miners, many events per epoch. Most of the events are for naught, but can't be unregistered when the proof arrives on time.

This load could reduced by quantizing the epochs at which expiries are processed, and using a single cron event per miner per quanta which checks all the pre-commits due for proof before the event. The existing proving-period event might be a suitable entry point. It doesn't matter much if it's quite infrequent: a late-arriving proof is already rejected. This is just cleaning up state.

The precommit sector IDs for processing could be stored either in a queue-like `AMT[epoch]Bitfield`, perhaps quantized (like `FaultEpochs`)

aarshkshah1992:
Closes #417.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -37,7 +37,6 @@ const (
 
 type CronEventPayload struct {
 	EventType CronEventType
-	Sectors   *abi.BitField
 }
 
 // Identifier for a single partition within a miner.
@@ -493,35 +492,12 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		}); err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "failed to write pre-committed sector %v: %v", params.SectorNumber, err)
 		}
+		// add precommit expiry to the queue
+		err = st.AddPreCommitExpiry(store, rt.CurrEpoch(), params.SealProof, params.SectorNumber)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add pre-commit expiry to queue")
 	})
 
 	notifyPledgeChanged(rt, newlyVested.Neg())
-
-	// add precommit expiry to the queue
-	msd, ok := MaxSealDuration[params.SealProof]
-	if !ok {
-		rt.Abortf(exitcode.ErrIllegalArgument, "no max seal duration set for proof type: %d", params.SealProof)
-	}
-	// The +1 here is critical for the batch verification of proofs. Without it, if a proof arrived exactly on the
-	// due epoch, ProveCommitSector would accept it, then the expiry event would remove it, and then
-	// ConfirmSectorProofsValid would fail to find it.
-	expiryBound := rt.CurrEpoch() + msd + 1
-
-	// create BitField for this sector
-	bf := abi.NewBitField()
-	bf.Set(uint64(params.SectorNumber))
-
-	// Load BitField Queue for sector expiry
-	quant := st.QuantEndOfDeadline()
-	queue, err := LoadBitfieldQueue(store, st.PreCommittedSectorsExpiry, quant)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-commit expiry queue")
-
-	// add entry for this sector to the queue
-	err = queue.AddToQueue(expiryBound, bf)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add pre-commit sector expiry to queue")
-
-	st.PreCommittedSectorsExpiry, err = queue.Root()
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save pre-commit sector queue")
 
 	return nil
 }
@@ -1514,6 +1490,25 @@ func handleProvingDeadline(rt Runtime) {
 			pledgeDelta = big.Add(pledgeDelta, newlyVested.Neg())
 		}
 
+		{
+			// expire pre-committed sectors
+			expiryQ, err := LoadBitfieldQueue(store, st.PreCommittedSectorsExpiry, NewQuantSpec(WPoStChallengeWindow, st.ProvingPeriodStart))
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sector expiry queue")
+
+			bf, modified, err := expiryQ.PopUntil(currEpoch)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to pop expired sectors")
+			AssertMsg(bf != nil, "sectors to check expiry for at epoch %d were nil", currEpoch)
+
+			if modified {
+				st.PreCommittedSectorsExpiry, err = expiryQ.Root()
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save expiry queue")
+			}
+
+			depositToBurn, err := st.checkPrecommitExpiry(store, bf)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expire pre-committed sectors")
+			penaltyTotal = big.Add(penaltyTotal, depositToBurn)
+		}
+
 		// Record whether or not we _had_ early terminations in the queue before this method.
 		// That way, don't re-schedule a cron callback if one is already scheduled.
 		hadEarlyTerminations = havePendingEarlyTerminations(rt, &st)
@@ -1611,24 +1606,6 @@ func handleProvingDeadline(rt Runtime) {
 				st.ProvingPeriodStart = st.ProvingPeriodStart + WPoStProvingPeriod
 			}
 		}
-
-		// expire pre-committed sectors
-		expiryQ, err := LoadBitfieldQueue(store, st.PreCommittedSectorsExpiry, quant)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sector expiry queue")
-
-		bf, modified, err := expiryQ.PopUntil(currEpoch)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to pop expired sectors")
-
-		if modified {
-			st.PreCommittedSectorsExpiry, err = expiryQ.Root()
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save expiry queue")
-		}
-
-		if bf != nil {
-			depositToBurn, err := st.checkPrecommitExpiry(store, bf)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expire pre-committed sectors")
-			burnFunds(rt, depositToBurn)
-		}
 	})
 
 	// Remove power for new faults, and burn penalties.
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -498,25 +497,31 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
 	// ConfirmSectorProofsValid would fail to find it.
 	expiryBound := rt.CurrEpoch() + msd + 1
-	enrollCronEvent(rt, expiryBound, &cronPayload)
+
+	// create BitField for this sector
+	bf := abi.NewBitField()
+	bf.Set(uint64(params.SectorNumber))
+
+	// Load BitField Queue for sector expiry
+	quant := st.QuantEndOfDeadline()
```
Stebalien, line 515: This has gone away. We'll need to quantize for every deadline, at the end of every deadline.

actors/builtin/miner/miner_actor.go
```diff
@@ -498,25 +497,31 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
+	queue, err := LoadBitfieldQueue(store, st.PreCommittedSectorsExpiry, quant)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-commit expiry queue")
+
+	// add entry for this sector to the queue
+	err = queue.AddToQueue(expiryBound, bf)
```
Stebalien, line 520: nit:

```suggestion
	err = queue.AddToQueueValues(expiryBound, uint64(params.SectorNumber))
```

actors/builtin/miner/miner_actor.go
```diff
@@ -1610,6 +1611,24 @@ func handleProvingDeadline(rt Runtime) {
...
+
+		bf, modified, err := expiryQ.PopUntil(currEpoch)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to pop expired sectors")
+
+		if modified {
+			st.PreCommittedSectorsExpiry, err = expiryQ.Root()
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save expiry queue")
+		}
+
+		if bf != nil {
```
Stebalien, line 1627: this should never return nil, unless it returns an error.

actors/builtin/miner/miner_actor.go
```diff
@@ -1610,6 +1611,24 @@ func handleProvingDeadline(rt Runtime) {
...
+			depositToBurn, err := st.checkPrecommitExpiry(store, bf)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expire pre-committed sectors")
+			burnFunds(rt, depositToBurn)
```
Stebalien, line 1630: Can't call this in a transaction.

Stebalien: This should probably just be added to `penaltyTotal` and burnt at the end.

actors/builtin/miner/miner_actor.go
```diff
@@ -498,25 +497,31 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 
 	notifyPledgeChanged(rt, newlyVested.Neg())
 
-	bf := abi.NewBitField()
-	bf.Set(uint64(params.SectorNumber))
-
-	// Request deferred Cron check for PreCommit expiry check.
-	cronPayload := CronEventPayload{
-		EventType: CronEventPreCommitExpiry,
-		Sectors:   bf,
```
Stebalien, line 507: We should be able to remove `Sectors` from the cron payload.

actors/builtin/miner/miner_actor.go
anorth: `NewQuantSpec(WPoStChallengeWindow, st.ProvingPeriodStart)`

actors/builtin/miner/miner_actor.go
```diff
@@ -498,25 +497,31 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
+	quant := st.QuantEndOfDeadline()
+	queue, err := LoadBitfieldQueue(store, st.PreCommittedSectorsExpiry, quant)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-commit expiry queue")
+
+	// add entry for this sector to the queue
+	err = queue.AddToQueue(expiryBound, bf)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add pre-commit sector expiry to queue")
+
+	st.PreCommittedSectorsExpiry, err = queue.Root()
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save pre-commit sector queue")
```
anorth, line 524: Please push this whole delta down to a method on miner_state, and test it directly.

actors/builtin/miner/miner_actor.go
```diff
@@ -1610,6 +1611,24 @@ func handleProvingDeadline(rt Runtime) {
...
+		if modified {
+			st.PreCommittedSectorsExpiry, err = expiryQ.Root()
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save expiry queue")
+		}
+
+		if bf != nil {
+			depositToBurn, err := st.checkPrecommitExpiry(store, bf)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expire pre-committed sectors")
+			burnFunds(rt, depositToBurn)
+		}
```
anorth, line 1631: Let's move this whole lot to the top, immediately after vesting funds.

Push it down into a method on miner.State that returns the total penalty.

actors/builtin/miner/miner_actor.go
```diff
@@ -1686,46 +1705,44 @@ func validateReplaceSector(rt Runtime, st *State, store adt.Store, params *Secto
 	return replaceSector
 }
 
-func checkPrecommitExpiry(rt Runtime, sectors *abi.BitField) {
-	store := adt.AsStore(rt)
-	var st State
-
+func (st *State) checkPrecommitExpiry(store adt.Store, sectors *abi.BitField) (depositToBurn abi.TokenAmount, err error) {
```
anorth, line 1708: Move to miner_state.go

actors/builtin/miner/miner_actor.go
```diff
@@ -1686,46 +1705,44 @@ func validateReplaceSector(rt Runtime, st *State, store adt.Store, params *Secto
...
-				return nil
-			}
+	depositToBurn = abi.NewTokenAmount(0)
 
-			// mark it for deletion
-			sectorNos = append(sectorNos, sectorNo)
-
-			// increment deposit to burn
-			depositToBurn = big.Add(depositToBurn, sector.PreCommitDeposit)
+	var sectorNos []abi.SectorNumber
```
anorth, line 1712:
```suggestion
	var precommitsToDelete []abi.SectorNumber
```

actors/builtin/miner/miner_actor.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_actor.go
aarshkshah1992: Have removed this check and added an Assert for this.

actors/builtin/miner/miner_actor.go
aarshkshah1992: @anorth Not sure what you mean here. Do you mean push the pre commit expiry part in a separate func ?

actors/builtin/miner/miner_actor.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
```diff
@@ -970,6 +975,81 @@ func (st *State) MeetsInitialPledgeCondition(balance abi.TokenAmount) bool {
...
+// pre-commit expiry
+func (st *State) AddPreCommitExpiry(store adt.Store, currEpoch abi.ChainEpoch, sealProof abi.RegisteredSealProof, sectorNum abi.SectorNumber) error {
+	msd, ok := MaxSealDuration[sealProof]
+	if !ok {
+		return xerrors.Errorf("no max seal duration set for proof type: %d", sealProof)
+	}
+	// The +1 here is critical for the batch verification of proofs. Without it, if a proof arrived exactly on the
+	// due epoch, ProveCommitSector would accept it, then the expiry event would remove it, and then
+	// ConfirmSectorProofsValid would fail to find it.
+	expiryBound := currEpoch + msd + 1
```
anorth, line 987: Nit: pull this lookup and addition back out to the actor, separating the policy (in the actor) from the mechanics here.

actors/builtin/miner/miner_state.go
```diff
@@ -970,6 +975,81 @@ func (st *State) MeetsInitialPledgeCondition(balance abi.TokenAmount) bool {
...
+
+	// create BitField for this sector
+	bf := abi.NewBitField()
+	bf.Set(uint64(sectorNum))
```
anorth, line 991: Delete this bf, unused.

actors/builtin/miner/miner_state.go
```diff
@@ -970,6 +975,81 @@ func (st *State) MeetsInitialPledgeCondition(balance abi.TokenAmount) bool {
...
+
+	// Load BitField Queue for sector expiry
+	quant := NewQuantSpec(WPoStChallengeWindow, st.ProvingPeriodStart)
```
anorth, line 994: We create this twice, so extract `st.QuantSpecEveryDeadline()`.

actors/builtin/miner/miner_state.go
```diff
@@ -970,6 +975,81 @@ func (st *State) MeetsInitialPledgeCondition(balance abi.TokenAmount) bool {
...
+	}
+
+	// add entry for this sector to the queue
+	if err := queue.AddToQueueValues(expiryBound, uint64(sectorNum)); err != nil {
+		return xerrors.Errorf("failed to add pre-commit sector expiry to queue: %w", err)
+	}
+
+	st.PreCommittedSectorsExpiry, err = queue.Root()
+	if err != nil {
+		return xerrors.Errorf("failed to save pre-commit sector queue")
```
anorth, line 1007: Wrap the error

actors/builtin/miner/miner_state.go
```diff
@@ -970,6 +975,81 @@ func (st *State) MeetsInitialPledgeCondition(balance abi.TokenAmount) bool {
...
+	}
+
+	return nil
+}
+
+func (st *State) checkPrecommitExpiry(store adt.Store, sectors *abi.BitField) (depositToBurn abi.TokenAmount, err error) {
+	// initialize here to add together for all sectors and minimize calls across actors
```
anorth, line 1014: This comment is out of place now, there are no actor calls

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_state.go
aarshkshah1992: Removed.

actors/builtin/miner/miner_state.go
aarshkshah1992: This is done.

PR #908, opened 2020-08-05
Commit windowed post submissions to a particular chain.

anorth:
This prevents posts from being replayed on distant forks and helps to protect filecoin against long range attacks.

Replaces #680, which I accidentally merged into the wrong branch.

cc @magik6k @whyrusleeping

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -244,8 +244,8 @@ type SubmitWindowedPoStParams struct {
 	Proofs []abi.PoStProof
 	// The epoch at which these proofs is being committed to a particular chain.
 	ChainCommitEpoch abi.ChainEpoch
-	// A signature from the miner worker over the randomness at the chain commit epoch.
-	ChainCommitSig crypto.Signature
+	// The ticket randomness on the chain at the ChainCommitEpoch on the chain this post is committed to
+	ChainCommitRand abi.Randomness
 }
 
 // Invoked by miner's worker address to submit their fallback post
@@ -266,8 +266,9 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commitment %d too far in the past, must exceed %d", params.ChainCommitEpoch, currEpoch - MaxPoStChainCommitAge)
 	}
 	commRand := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_PoStChainCommit, params.ChainCommitEpoch, nil)
-	err := rt.Syscalls().VerifySignature(params.ChainCommitSig, rt.Message().Caller(), commRand)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "chain commit signature invalid")
+	if !bytes.Equal(commRand, params.ChainCommitRand) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "post commit randomness mismatched")
+	}
 
 	// Get the total power/reward. We need these to compute penalties.
 	rewardStats := requestCurrentEpochBlockReward(rt)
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/policy.go
```diff
@@ -117,6 +117,10 @@ const MinSectorExpiration = 180 * builtin.EpochsInDay
 // and sector.ActivationEpoch+sealProof.SectorMaximumLifetime()
 const MaxSectorExpirationExtension = 540 * builtin.EpochsInDay
 
+// MaxPoStChainCommitAge is the maximum distance back that a PoSt can commit to
+// a chain before being invalid
+const MaxPoStChainCommitAge = 5
```
anorth, line 122: @nicola has indicated this isn't the right number, so we won't merge until we have a better one.

nicola:
@irenegia and @sa8 to confirm that we can set a larger MaxPoStChainCommitAge or if there are better mechanics that can prevent short replay periods.

actors/builtin/miner/miner_actor.go
```diff
@@ -255,6 +259,16 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 	if params.Deadline >= WPoStPeriodDeadlines {
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d of %d", params.Deadline, WPoStPeriodDeadlines)
 	}
+	if params.ChainCommitEpoch >= currEpoch {
+		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commitment %d must be in the past", params.ChainCommitEpoch)
+	}
+	if params.ChainCommitEpoch < currEpoch-MaxPoStChainCommitAge {
+		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commitment %d too far in the past, must exceed %d", params.ChainCommitEpoch, currEpoch-MaxPoStChainCommitAge)
+	}
+	commRand := rt.GetRandomnessFromTickets(crypto.DomainSeparationTag_PoStChainCommit, params.ChainCommitEpoch, nil)
```
Stebalien, line 268: Where do we tie this into the proof? We'll need to pass this into `verifyWindowPost` and check it in the runtime. That or the runtime will have to return the randomness.

nicola: It is not an input to the proof, it's just a value the miner must sign

nicola:
Set MaxPoStChainCommitAge = WPoStChallengeWindow

actors/builtin/miner/miner_actor.go
Stebalien: Ah, I see.

actors/builtin/miner/miner_actor.go
```diff
@@ -255,6 +259,16 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 	if params.Deadline >= WPoStPeriodDeadlines {
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid deadline %d of %d", params.Deadline, WPoStPeriodDeadlines)
 	}
+	if params.ChainCommitEpoch >= currEpoch {
+		rt.Abortf(exitcode.ErrIllegalArgument, "PoSt chain commitment %d must be in the past", params.ChainCommitEpoch)
+	}
+	if params.ChainCommitEpoch < currEpoch-MaxPoStChainCommitAge {
```
Stebalien, line 265: What happens for epochs <= MaxPoStChainCommitAge?

anorth: I don't think I understand the question. The chain commit must be sufficiently recent, or the proof will be rejected.

Stebalien: I mean, what happens when the current epoch is less than MaxPoStChainCommitAge? I.e., less than a day old. It looks like this will wrap around and "just work", but that seems funky.

Stebalien: Ah, nevermind. I keep forgetting that we allow negative chain epochs.

PR #911, opened 2020-08-06
Clarifications to runtime interface documentation

anorth:
These clarifications mostly inspired by @wadealexc's questions.

I also filed a number of [chain validation issues](https://github.com/filecoin-project/chain-validation/issues) for new tests.

### DIFF
--- a/actors/crypto/randomness.go
+++ b/actors/crypto/randomness.go
@@ -1,7 +1,7 @@
 package crypto
 
 // Specifies a domain for randomness generation.
-type DomainSeparationTag int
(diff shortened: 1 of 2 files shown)


actors/crypto/randomness.go
```diff
@@ -1,7 +1,7 @@
 package crypto
 
 // Specifies a domain for randomness generation.
-type DomainSeparationTag int
+type DomainSeparationTag int64
```
anorth, line 4: Lotus already casts to int64 when serializing this value  to mix with entropy.

Issue #730, opened 2020-07-20
Multisig, verifreg, paych actors create accounts that don't already exist


wadeAlexC:
From discussion here #325:

> I don't think we can resolve to pubkey/stable addresses because there is no mapping from ID->stable address for actors. The reverse exists in the init actor, and account actors have a workaround of storing their pubkey in state, but other actors don't.

If it's not too much work, I propose that the Multisig actor creates an account in state if it doesn't already exist. The goal of this change is to make it easier to reason about the properties of the Multisig by enforcing that all addresses may be referenced by an ID address.

Edit: To be more clear, the proposed solution is that the Multisig (and verified registry, and payment channel) only store ID addresses in state. At construction (or relevant methods where addresses are accepted from outside), if provided with an address that does not resolve to an ID address, the actor should make a 0-value Send to the address, triggering its creation in state. Then resolve the address to an ID address again, which we can expect to succeed.

anorth:
This is a great solution, for both multisig and verified registry actor. I expect it to simplify the code, and certainly the tests.

anorth:
And the payment channel actor too!

aarshkshah1992:
The PRs for MultiSig, Verifreg and Payment Channel actors have been merged,

PR #912, opened 2020-08-06
Multisig: State should ONLY have ID address -> create accounts if they do NOT exist

aarshkshah1992:
For #730.

@anorth We should move the `resolveToIdAddr` func to a common location so the `Verifreg` and `PaymentChannel` actors can use it too.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -78,7 +78,7 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 	}
 
 	// resolve signer addresses and do not allow duplicate signers
-	resolvedAddrs := make([]addr.Address, 0, len(params.Signers))
+	resolvedSigners := make([]addr.Address, 0, len(params.Signers))
 	deDupSigners := make(map[addr.Address]struct{}, len(params.Signers))
 	for _, signer := range params.Signers {
 		resolved, err := builtin.ResolveToIDAddr(rt, signer)
@@ -88,7 +88,7 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 			rt.Abortf(exitcode.ErrIllegalArgument, "duplicate signer not allowed: %s", signer)
 		}
 
-		resolvedAddrs = append(resolvedAddrs, resolved)
+		resolvedSigners = append(resolvedSigners, resolved)
 		deDupSigners[resolved] = struct{}{}
 	}
 
@@ -110,7 +110,7 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 	}
 
 	var st State
-	st.Signers = resolvedAddrs
+	st.Signers = resolvedSigners
 	st.NumApprovalsThreshold = params.NumApprovalsThreshold
 	st.PendingTxns = pending
 	st.InitialBalance = abi.NewTokenAmount(0)
@@ -144,7 +144,7 @@ type ProposeReturn struct {
 
 func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
-	callerAddr := rt.Message().Caller()
+	proposer := rt.Message().Caller()
 
 	if params.Value.Sign() < 0 {
 		rt.Abortf(exitcode.ErrIllegalArgument, "proposed value must be non-negative, was %v", params.Value)
@@ -154,9 +154,8 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	var st State
 	var txn *Transaction
 	rt.State().Transaction(&st, func() {
-		proposerIsSigner := isSigner(callerAddr, st.Signers)
-		if !proposerIsSigner {
-			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", callerAddr)
+		if !isSigner(proposer, st.Signers) {
+			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", proposer)
 		}
 
 		ptx, err := adt.AsMap(adt.AsStore(rt), st.PendingTxns)
@@ -312,14 +311,14 @@ type RemoveSignerParams struct {
 func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.EmptyValue {
 	// Can only be called by the multisig wallet itself.
 	rt.ValidateImmediateCallerIs(rt.Message().Receiver())
-	resolvedSigner, err := builtin.ResolveToIDAddr(rt, params.Signer)
+	resolvedOldSigner, err := builtin.ResolveToIDAddr(rt, params.Signer)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve address %v", params.Signer)
 
 	var st State
 	rt.State().Transaction(&st, func() {
-		isSigner := isSigner(resolvedSigner, st.Signers)
+		isSigner := isSigner(resolvedOldSigner, st.Signers)
 		if !isSigner {
-			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", resolvedSigner)
+			rt.Abortf(exitcode.ErrForbidden, "%s is not a signer", resolvedOldSigner)
 		}
 
 		if len(st.Signers) == 1 {
@@ -329,7 +328,7 @@ func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.Emp
 		newSigners := make([]addr.Address, 0, len(st.Signers))
 		// signers have already been resolved
 		for _, s := range st.Signers {
-			if resolvedSigner != s {
+			if resolvedOldSigner != s {
 				newSigners = append(newSigners, s)
 			}
 		}
@@ -410,7 +409,6 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
 }
 
 func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn *Transaction) (bool, []byte, exitcode.ExitCode) {
-
 	caller := rt.Message().Caller()
 
 	var st State
@@ -502,9 +500,6 @@ func executeTransactionIfApproved(rt vmr.Runtime, st State, txnID TxnID, txn *Tr
 
 	return applied, out, code
 }
-
-type AddressResolveFunc func(address addr.Address) (resolved addr.Address, found bool)
-type SendFunc func(toAddr addr.Address, methodNum abi.MethodNum, params vmr.CBORMarshaler, value abi.TokenAmount) (vmr.SendReturn, exitcode.ExitCode)
 
 func isSigner(address addr.Address, signers []addr.Address) bool {
 	AssertMsg(address.Protocol() == addr.ID, "address %v passed to isSigner must be a resolved address", address)
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -81,7 +81,7 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 	resolvedAddrs := make([]addr.Address, 0, len(params.Signers))
 	deDupSigners := make(map[addr.Address]struct{}, len(params.Signers))
 	for _, signer := range params.Signers {
-		resolved, err := resolveToIDAddr(signer, rt.ResolveAddress, rt.Send)
+		resolved, err := builtin.ResolveToIDAddr(rt, signer)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve addr %v to ID addr", signer)
 
 		if _, ok := deDupSigners[resolved]; ok {
@@ -146,9 +146,6 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	callerAddr := rt.Message().Caller()
 
-	resolvedCaller, err := resolveToIDAddr(callerAddr, rt.ResolveAddress, rt.Send)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolved address %v", callerAddr)
-
 	if params.Value.Sign() < 0 {
 		rt.Abortf(exitcode.ErrIllegalArgument, "proposed value must be non-negative, was %v", params.Value)
 	}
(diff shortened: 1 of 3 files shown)


aarshkshah1992:
@wadeAlexC.

wadeAlexC:
These changes make me think the runtime should expose a "ForceResolve" method that automatically creates an account if it doesn't exist.

Either that, or the concept of non-existent accounts could just be abstracted away entirely, and accounts are always created if they don't already exist. i.e. things like `rt.DeleteActor(beneficiary)` would create `beneficiary` if it doesn't exist.

wadeAlexC:
What would be really cool is if this were abstracted in a way s.t. ALL addresses were always ID-addresses within code, and a reverse ID -> BLS/SECP/ACTOR lookup existed.

wadeAlexC:
I would like to see these changes decouple ID address resolution from Multisig-specific functionality.

`isSigner` should assert that it receives an ID address; resolution should be an individual, explicit action taken as soon as an address is referenced for the first time.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -302,17 +320,20 @@ func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.Emp
...
 			rt.Abortf(exitcode.ErrForbidden, "cannot remove only signer")
 		}
 
 		newSigners := make([]addr.Address, 0, len(st.Signers))
+		// signers have already been resolved
 		for _, s := range st.Signers {
-			if !isAddressEqual(rt.ResolveAddress, s, params.Signer) {
+			if resolved != s {
 				newSigners = append(newSigners, s)
 			}
```
wadeAlexC, line 338: Do we still need to iterate over st.Signers? The !isSigner check should be functionally identical.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -344,21 +365,25 @@ func (a Actor) SwapSigner(rt vmr.Runtime, params *SwapSignerParams) *adt.EmptyVa
...
 		newSigners := make([]addr.Address, 0, len(st.Signers))
 		for _, s := range st.Signers {
-			if !isAddressEqual(rt.ResolveAddress, s, params.From) {
+			if s != fromResolved {
 				newSigners = append(newSigners, s)
 			}
 		}
-		newSigners = append(newSigners, params.To)
+		newSigners = append(newSigners, toResolved)
 		st.Signers = newSigners
```
wadeAlexC, line 387: Same comment here. I dont think we need to iterate over `st.Signers` anymore.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -384,7 +409,7 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
 	return nil
 }
 
-func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn *Transaction) (bool, []byte, exitcode.ExitCode) {
+func (a Actor) approveTransaction(rt vmr.Runtime, resolvedApprover addr.Address, txnID TxnID, txn *Transaction) (bool, []byte, exitcode.ExitCode) {
```
wadeAlexC, line 412: Should AssertMsg that `resolvedApprover` is an ID address

actors/builtin/multisig/multisig_actor.go
```diff
@@ -399,9 +424,9 @@ func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn *Transaction)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pending transactions")
 
 		// update approved on the transaction
-		txn.Approved = append(txn.Approved, rt.Message().Caller())
-		 err = ptx.Put(txnID, txn)
-		 builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to put transaction %v for approval", txnID)
+		txn.Approved = append(txn.Approved, resolvedApprover)
+		err = ptx.Put(txnID, txn)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to put transaction %v for approval", txnID)
```
wadeAlexC, line 429: Should the "abort duplicate approval" check above use `resolveApprover` as well?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -476,33 +501,50 @@ func executeTransactionIfApproved(rt vmr.Runtime, st State, txnID TxnID, txn *Tr
...
 
-	for _, ap := range st.Signers {
-		signerResolved := resolve(resolveFunc, ap)
-		if signerResolved == candidateResolved {
-			return true
+	// signer addresses have already been resolved
+	for _, ap := range signers {
+		if ap == candidateResolved {
+			return true, candidateResolved, nil
 		}
```
wadeAlexC, line 517: `ap`? Not sure what this stands for. Maybe use `signer` instead?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -476,33 +501,50 @@ func executeTransactionIfApproved(rt vmr.Runtime, st State, txnID TxnID, txn *Tr
...
-	if resolved.Protocol() != addr.ID {
-		idAddr, found := resolveFunc(resolved)
-		if found {
-			resolved = idAddr
-		}
+func resolveToIDAddr(address addr.Address, resolveFunc AddressResolveFunc, sendFunc SendFunc) (addr.Address, error) {
+	// if it's already an ID address, we can return immediately
+	if address.Protocol() == addr.ID {
+		return address, nil
 	}
```
wadeAlexC, line 527: This short-circuit doesn't guarantee that `address` exists if it's already an ID-address.

wadeAlexC: Plus, `resolveFunc` performs the same short-circuit internally, so I think this branch is just redundant.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: This is done.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: I agree. This has been removed.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: This is done.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: This is done.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: I am not sure what you mean here, This iteration is to get the Signers slice without the signer we need to remove.
Can you elaborate on what you mean here ?

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: Same comment as above.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: Also added a test to catch this.

aarshkshah1992:
@wadeAlexC 

> I would like to see these changes decouple ID address resolution from Multisig-specific functionality.
> 
> isSigner should assert that it receives an ID address; resolution should be an individual, explicit action taken as soon as an address is referenced for the first time.

This is done.

> 
> These changes make me think the runtime should expose a "ForceResolve" method that automatically creates an account if it doesn't exist.
> 
> Either that, or the concept of non-existent accounts could just be abstracted away entirely, and accounts are always created if they don't already exist. i.e. things like rt.DeleteActor(beneficiary) would create beneficiary if it doesn't exist.

While I think it's a bit too late to make any change to the `Runtime` Impl, I'll leave it to @anorth to comment on this.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -141,6 +146,9 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	callerAddr := rt.Message().Caller()
 
+	resolvedCaller, err := resolveToIDAddr(callerAddr, rt.ResolveAddress, rt.Send)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolved address %v", callerAddr)
```
anorth, line 150: Unnecessary, the caller is always an ID address.

actors/builtin/multisig/multisig_actor.go
anorth: It was unnecessary to resolve the caller in the first place, so also unnecessary to pass it as a parameter here or assert anything.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -476,33 +510,40 @@ func executeTransactionIfApproved(rt vmr.Runtime, st State, txnID TxnID, txn *Tr
...
 }
 
-func resolve(resolveFunc AddressResolveFunc, address addr.Address) addr.Address {
-	resolved := address
-	if resolved.Protocol() != addr.ID {
-		idAddr, found := resolveFunc(resolved)
-		if found {
-			resolved = idAddr
-		}
+func resolveToIDAddr(address addr.Address, resolveFunc AddressResolveFunc, sendFunc SendFunc) (addr.Address, error) {
```
anorth, line 527: Move this to `builtin/shared.go`. Also it's ok for it to just take Runtime as a parameter, since it' always called at the top level and obviously needs it.

actors/builtin/multisig/multisig_test.go
```diff
@@ -31,15 +31,15 @@ func TestConstruction(t *testing.T) {
...
 
 	bob := tutil.NewIDAddr(t, 102)
 	bobNonId := tutil.NewBLSAddr(t, 2)
 
 	charlie := tutil.NewIDAddr(t, 103)
 
-	builder := mock.NewBuilder(context.Background(), receiver).WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
-
 	t.Run("simple construction", func(t *testing.T) {
+		builder := mock.NewBuilder(context.Background(), receiver).WithCaller(builtin.InitActorAddr, builtin.InitActorCodeID)
```
anorth, line 42: Why did you move and duplicate this line in all the cases? They all need it, and if something needs something special, you can shadow it or use another name.

actors/builtin/multisig/multisig_actor.go
wadeAlexC: I guess my question is - can you just do `st.Signers = append(st.Signers, toResolved)`?

wadeAlexC: Iterating over happens twice in this method - once in isSigner, and once here. We only need to iterate a single time, I think!

wadeAlexC: Nevermind, this doesn't work for SwapSigner. But I think it does work for AddSigner!

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: This is fixed everywhere.

actors/builtin/multisig/multisig_actor.go
aarshkshah1992: This is done.

actors/builtin/multisig/multisig_test.go
aarshkshah1992: Apologies, this is fixed.

aarshkshah1992:
@anorth Have made all the changes. Please take a look.

wadeAlexC:
LGTM

actors/builtin/multisig/multisig_actor.go
```diff
@@ -105,7 +110,7 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 	}
 
 	var st State
-	st.Signers = params.Signers
+	st.Signers = resolvedAddrs
```
anorth, line 113: Nit: resolvedSigners?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -149,7 +154,8 @@ func (a Actor) Propose(rt vmr.Runtime, params *ProposeParams) *ProposeReturn {
 	var st State
 	var txn *Transaction
 	rt.State().Transaction(&st, func() {
-		if !isSigner(rt.ResolveAddress, &st, callerAddr) {
+		proposerIsSigner := isSigner(callerAddr, st.Signers)
+		if !proposerIsSigner {
```
anorth, line 158: Nit: I would fold these back to one line; the predicate code is already pretty clear. Possibly rename `callerAddr` for even more clarity.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -299,20 +312,24 @@ type RemoveSignerParams struct {
 func (a Actor) RemoveSigner(rt vmr.Runtime, params *RemoveSignerParams) *adt.EmptyValue {
 	// Can only be called by the multisig wallet itself.
 	rt.ValidateImmediateCallerIs(rt.Message().Receiver())
+	resolvedSigner, err := builtin.ResolveToIDAddr(rt, params.Signer)
```
anorth, line 315: resolvedOldSigner?

actors/builtin/multisig/multisig_actor.go
```diff
@@ -385,10 +410,13 @@ func (a Actor) ChangeNumApprovalsThreshold(rt vmr.Runtime, params *ChangeNumAppr
 }
 
 func (a Actor) approveTransaction(rt vmr.Runtime, txnID TxnID, txn *Transaction) (bool, []byte, exitcode.ExitCode) {
+
```
anorth, line 413: Nit: unnecessary blank line

actors/builtin/multisig/multisig_actor.go
```diff
@@ -476,35 +504,20 @@ func executeTransactionIfApproved(rt vmr.Runtime, st State, txnID TxnID, txn *Tr
 }
 
 type AddressResolveFunc func(address addr.Address) (resolved addr.Address, found bool)
+type SendFunc func(toAddr addr.Address, methodNum abi.MethodNum, params vmr.CBORMarshaler, value abi.TokenAmount) (vmr.SendReturn, exitcode.ExitCode)
```
anorth, line 507: I think this is redundant now

actors/builtin/shared.go
```diff
@@ -54,3 +56,27 @@ type MinerAddrs struct {
...
+	// if we are able to resolve it to an ID address, return the resolved address
+	idAddr, found := rt.ResolveAddress(address)
+	if found {
+		return idAddr, nil
+	}
+
+	// send 0 balance to the account so an ID address for it is created and then try to resolve
+	_, code := rt.Send(address, MethodSend, nil, abi.NewTokenAmount(0))
+	if !code.IsSuccess() {
+		return address, fmt.Errorf("failed to send zero balance to account %v, got code %v", address, code)
```
anorth, line 72: `code` here is an `ExitCode` which implements error, so we should return a thing that will result in propagating that exit code. I think you want something like `code.Wrapf(msg, args)`.

Issue #621, opened 2020-07-08
Implement a simple/reference VM for scenario testing


anorth:
The current mock runtime supports detailed unit testing of individual actors, but not the coordinated actions of multiple actors. This "scenario testing" is an important responsibility that should be carried out within this project.

The [chain validation](https://github.com/filecoin-project/chain-validation) project does some scenario testing as a means of exercising the VMs that it is targeting. But that project should really focus on VM edge cases rather than actor correctness.

To support multi-actor tests, we need a runtime that can process invocations between actors and maintain a ~full state tree. We don't need a _full_ VM coupled to a blockchain, though - we still want to be able to fake out system calls like proof and signature verification (note: fake/stub, rather than mock). This could also function as a reference for the behaviour (but not architecture/optimizations) for a real VM. We should be able to have the chain validation tests target it directly.

A promising path here would be to copy the VM from go-filecoin, and dramatically trim it down to the bare necessities.

anorth:
Closed by #914.

See #810 for more tests using this.

PR #914, opened 2020-08-06
scenario testing with create miner test

acruikshank:
work towards #621

### Motivation

We have fairly thorough unit testing of actor code, but these test stub out all interactions that happen between actors and provide relatively low confidence that top-level messages that trigger multiple actor actions will work. This PR introduces a VM for scenario testing and some infrastructure for writing those tests. The VM is a simplified and stripped-down version of go-filecoin's VM with some setup code to initialize singleton actors.

The VM is missing important functionality like proof syscalls, nonce checking and gas. This functionality belongs to the implementations, and should be validated there. The code introduced here should really be the least that is need to initiate all actor processes without stubbing out actor code or erroring out from otherwise working code.

### Proposed Changes

1. Introduce `VM` and `invocationContext` to apply messages and sub messages.
2. Introduce `setup` to create singleton and account actors (i.e. genesis).
3. Add proof of concept `TestCreateMiner` test that hits Init, StoragePower, and Miner actors with a single message.


support/testing/scenarios/power_test.go
```diff
@@ -0,0 +1,84 @@
...
+	vm, addrs := scenarios.NewVMWithSingletons(ctx, t, 1)
+
+	params := power.CreateMinerParams{
+		Owner:         addrs[0],
+		Worker:        addrs[0],
+		SealProofType: abi.RegisteredSealProof_StackedDrg32GiBV1,
+		Peer:          abi.PeerID("not really a peer id"),
+	}
+	ret, code := vm.ApplyMessage(addrs[0], builtin.StoragePowerActorAddr, big.NewInt(1e10), builtin.MethodsPower.CreateMiner, &params)
+	assert.Equal(t, exitcode.Ok, code)
```
anorth, line 32: We can do it in follow-up as we write more tests, but I suggest we make an API as ergonomic for test writing as we can. So, for example:
- add ApplyMessageOk that checks the exit code
- add GetState helper that checks errors already, and maybe returns values (that need casting) to save pre-allocation
- store methods that use a context the VM already has

Maybe we want a wrapper that encapsulates the `testing.T` for making all these checks (without pushing the T down into the VM proper)?

support/testing/scenarios/power_test.go
```diff
@@ -0,0 +1,84 @@
...
+	var ev power.CronEvent
+	err = queue.ForAll(func(k string, arr *adt.Array) error {
+		return arr.ForEach(&ev, func(i int64) error {
+			assert.Equal(t, createMinerRet.IDAddress, ev.MinerAddr)
+			cronCount++
+			return nil
+		})
+	})
+	require.NoError(t, err)
+	assert.Equal(t, uint64(1), cronCount)
```
anorth, line 83: I think we should avoid diving too deeply into the actor state internals in scenario tests. This is fine now because it's a simple test, but we don't want a lot of friction to scripting up complicated interactions. We can probably thin it down as other tests arrive (which will also create miners and run their crons).

support/testing/scenarios/setup.go
```diff
@@ -0,0 +1,150 @@
...
+	actor_testing "github.com/filecoin-project/specs-actors/support/testing"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/specs-actors/actors/runtime"
+	"github.com/ipfs/go-cid"
+	"github.com/stretchr/testify/require"
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+func NewVMWithSingletons(ctx context.Context, t *testing.T, additionalAccounts int) (*VM, []address.Address) {
```
anorth, line 30: Please document

support/testing/scenarios/setup.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// burnt funds
+	initializeActor(ctx, t, vm, &account.State{Address: builtin.BurntFundsActorAddr}, builtin.AccountActorCodeID, builtin.BurntFundsActorAddr, big.Zero())
+
+	// verified root
+	rootVerifierState := &account.State{Address: rootVerifier.pubAddr}
+	initializeActor(ctx, t, vm, rootVerifierState, builtin.AccountActorCodeID, rootVerifier.idAddr, big.Zero())
+
+	// additional accounts
+	initialAccountBalance := big.Mul(big.NewInt(10_000), FIL)
```
anorth, line 81: Please make this a parameter. I'd rather separate the creation of accounts like this from the singletons. A few ways this could go:
- pass a slice of AccountSpec objects to this method, each spec having a pubkey and initial balance
- don't init accounts here at all, but export a new method for adding an account to state directly

The accounts should come after the other actors, so create the init and other singletons first, and _then_ add the accounts (if any)

support/testing/scenarios/setup.go
```diff
@@ -0,0 +1,150 @@
...
+func initAddressTree(t *testing.T, vm *VM, rootVerifier addrPair, additionalAccounts int) (cid.Cid, []addrPair) {
+	m := adt.MakeEmptyMap(vm.store)
+	err := m.Put(adt.AddrKey(rootVerifier.pubAddr), cborAddrID(t, rootVerifier.idAddr))
+	require.NoError(t, err)
+
+	newAccounts := make([]addrPair, additionalAccounts)
+	for i := 0; i < additionalAccounts; i++ {
+		addr := actor_testing.NewBLSAddr(t, 93837778)
+		require.NoError(t, err)
+		idAddr, err := address.NewIDAddress(uint64(builtin.FirstNonSingletonActorId + i))
```
anorth, line 121: It would be much better if this just used `init.State.MapAddressToNewID`. This seems to be reaching around the existing code unnecessarily.

support/testing/scenarios/setup.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+type addrPair struct {
+	pubAddr address.Address
+	idAddr  address.Address
+}
+
+func initAddressTree(t *testing.T, vm *VM, rootVerifier addrPair, additionalAccounts int) (cid.Cid, []addrPair) {
+	m := adt.MakeEmptyMap(vm.store)
+	err := m.Put(adt.AddrKey(rootVerifier.pubAddr), cborAddrID(t, rootVerifier.idAddr))
```
anorth, line 114: I don't understand the restrictions of the verified registry actor enough to know why this is here, but neither will ~anyone else. Could you please add some more commentary about what and why is happening here?

support/testing/scenarios/vm.go
```diff
@@ -0,0 +1,364 @@
...
+
+// VM holds the state and executes messages over the state.
+type VM struct {
+	ctx   context.Context
+	store adt.Store
+
+	currentEpoch abi.ChainEpoch
+
+	actorImpls  ActorImplLookup
+	actorRoot   cid.Cid  // The last committed root.
```
anorth, line 30: Is this just the state root? The name suggests it's something different, but if it's not, please use the name we're all used to.

support/testing/scenarios/vm.go
```diff
@@ -0,0 +1,364 @@
...
+	actorsDirty bool
+
+	emptyObject cid.Cid
+
+	logs []string
+}
+
+// VM types
+
+type TestActor struct {
```
anorth, line 41: Please document, and note how it differs from the real version

support/testing/scenarios/vm.go
```diff
@@ -0,0 +1,364 @@
...
+func (vm *VM) deleteActor(ctx context.Context, key address.Address) error {
+	err := vm.actors.Delete(adt.AddrKey(key))
+	vm.actorsDirty = true
+	if err == hamt.ErrNotFound {
+		return nil
+	}
+	return err
+}
+
+func (vm *VM) checkpoint() (cid.Cid, error) {
```
anorth, line 128: We can probably remove this, or rename `commit()` to `checkpoint()`

support/testing/scenarios/vm.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	// get a view into the actor state
+	var state init_.State
+	if err := vm.store.Get(vm.ctx, initActorEntry.Head, &state); err != nil {
+		panic(err)
+	}
+
+	idAddr, err := state.ResolveAddress(vm.store, addr)
+	if err == init_.ErrAddressNotFound {
```
anorth, line 166: FYI https://github.com/filecoin-project/specs-actors/pull/910

support/testing/scenarios/power_test.go
acruikshank: I went ahead and implemented a higher level approach that tracks invocations in the VM and provides some infrastructure for making assertions against them. I think this flow through actors is what we should really focus on testing with scenarios.

support/testing/scenarios/setup.go
acruikshank: Added. The simple version is that verified registry references a special account actor which needs to be initialized prior to initializing the registry. We now handle this one-off separately from initializing accounts used for testing.

actors/test/power_scenario_test.go
```diff
@@ -0,0 +1,143 @@
...
+	minerAddrs, ok := ret.(*power.CreateMinerReturn)
+	require.True(t, ok)
+
+	// find epoch of miner's next cron task (4 levels deep, first message each level)
+	cronParams := vm.ParamsForInvocation(t, v, 0, 0, 0, 0)
+	cronConfig, ok := cronParams.(*power.EnrollCronEventParams)
+	require.True(t, ok)
+
+	// create new vm at epoch 1 less than epoch requested by miner
+	v, err := v.NewVMAtEpoch(cronConfig.EventEpoch - 1)
```
anorth, line 94: Hmm. `v.WithEpoch`? Or just have the epoch and other context be mutable?

actors/test/power_scenario_test.go
```diff
@@ -0,0 +1,143 @@
...
+	// expect miner call to be missing
+	vm.ExpectInvocation{
+		// Original send to storage power actor
+		To:     builtin.StoragePowerActorAddr,
+		Method: builtin.MethodsPower.OnEpochTickEnd,
+		SubInvocations: []vm.ExpectInvocation{{
+			// expect call to reward to update kpi
+			To:     builtin.RewardActorAddr,
+			Method: builtin.MethodsReward.UpdateNetworkKPI,
+			From:   vm.ExpectAddress(builtin.StoragePowerActorAddr),
```
anorth, line 110: Factor this repeated thing out?

support/vm/testing.go
```diff
@@ -0,0 +1,276 @@
...
+	require.NoError(t, err)
+
+	initializeActor(ctx, t, vm, &system.State{}, builtin.SystemActorCodeID, builtin.SystemActorAddr, big.Zero())
+
+	// Verified registry account needs to be instantiated and initialized in init actor prior to construction
+	// of verified registry actor.
+	rootVerifier := verifregRootAddresses(t)
+	addrTreeRoot := initAddressTree(t, vm, rootVerifier)
+	initState := initactor.ConstructState(addrTreeRoot, "scenarios")
+	initState.NextID = abi.ActorID(builtin.FirstNonSingletonActorId)
```
anorth, line 65: `ConstructState` already does this

support/vm/testing.go
```diff
@@ -0,0 +1,276 @@
...
+	initializeActor(ctx, t, vm, initState, builtin.InitActorCodeID, builtin.InitActorAddr, big.Zero())
+
+	rewardState := reward.ConstructState(big.Mul(big.NewInt(100), FIL))
```
anorth, line 68: The argument is a power value, so multiplying by FIL is very weird. Add a new constant for `Petabyte` or similar.

support/vm/testing.go
```diff
@@ -0,0 +1,276 @@
...
+
+// Creates n account actors in the VM with the given balance
+func CreateAccounts(ctx context.Context, t *testing.T, vm *VM, n int, balance abi.TokenAmount) []address.Address {
+	var initState initactor.State
+	err := vm.GetState(builtin.InitActorAddr, &initState)
+	require.NoError(t, err)
+
+	addrPairs := make([]addrPair, n)
+	for i := range addrPairs {
+		addr := actor_testing.NewBLSAddr(t, 93837778)
```
anorth, line 104: These addresses are all the same! That's probably going to result in only one account actually in state.

support/vm/testing.go
```diff
@@ -0,0 +1,276 @@
...
+	if marshaller != nil {
+		marshaller.MarshalCBOR(paramBuf2) // nolint: errcheck
+	}
+	return bytes.Equal(paramBuf1.Bytes(), paramBuf2.Bytes())
+}
+
+var okExitCode = exitcode.Ok
+var ExpectOK = &okExitCode
+
+type ExpectInvocation struct {
```
anorth, line 165: Please document this, e.g. the meaning of zero/nil values.

support/vm/testing.go
```diff
@@ -0,0 +1,276 @@
...
+	require.NoError(t, err)
+	cbgId := cbg.CborInt(id)
+	return &cbgId
+}
+
+func verifregRootAddresses(t *testing.T) addrPair {
+	rootVerifierAddr, err := address.NewFromString("t3qfoulel6fy6gn3hjmbhpdpf6fs5aqjb5fkurhtwvgssizq4jey5nw4ptq5up6h7jk7frdvvobv52qzmgjinq")
+	require.NoError(t, err)
+
+	rootVerifierID, err := address.NewFromString("t080")
```
anorth, line 273: Are we considering this a singleton actor? If so, we should declare it in the main actor code as a singleton. Please file an issue to follow up with that.

support/vm/testing.go
acruikshank: ✅

support/vm/testing.go
acruikshank: Looking into it a bit, this actually should be a multisig actor, which means multiple account actors and an multisig actor need to be added to properly simulate adding verifiers. I don't think the multisig is a proper singleton in the sense that it needs a well known address, since it will only be addressed by its signers.

Since the VM doesn't check message signatures, it's currently possible to simply use this t080 address as the root verifier in the test VM. I'm dropping the root verifier account for now.

PR #915, opened 2020-08-06
Don't allow omitting the proposal hash.

Stebalien:


Stebalien:
According to @whyrusleeping, the proposal hash is an optional additional security check to make sure the approval applies to the right proposal. If that's the case, we need to document that and reason about the security trade-offs.

As written, there is no security issue here. The caller would have to intentionally drop the proposal hash to hit this. However, if the feature is actually important, we might want to require it.

Stebalien:
Ok, digging into this, I think the issue is "finality". If I can propose something on one chain, get someone to approve it before the chain stabilizes, then replace the proposal with a new proposal (on a fork), I can re-play the approval on the new fork.

A more general fix would be the ability to declare a dependency between two messages or on a state. But that's much harder.

Stebalien:
Ok, closing as there's a bigger question here around finality. This is working as intended.

droark:
> the proposal hash is an optional additional security check to make sure the approval applies to the right proposal.

Hi. Just to be clear, the hash originated from a need for offline signing to have something more than an integer to look at when signing a transaction. In and of itself, the integer means nothing to an offline signer. If you pass the proposal to the offline signer, they can generate a hash based off of that data, with a reasonable (IMO) degree of certainty that the network will reject the TX if any of the hashed data isn't what was in the original proposal. It's not perfect but it is the best compromise that didn't require a significant refactor of the codebase.

PR #919, opened 2020-08-06
avoid prove commit aborts when sector lifetime <= 0

### LINKED ISSUE #795: Miner: Sector activation at ProveCommit may allow proving expired/expiring sectors, leading to panic during cron tick
#787 re-calculates the sector initial pledge, and moves sector activation to the epoch in which a successful ProveCommit is made.

This change may allow `precommit.Info.Expiration - rt.CurrEpoch()` to have a negative result: https://github.com/filecoin-project/specs-actors/blob/0dcd04ea4dca004872a4984d40179a67765f2544/actors/builtin/miner/miner_actor.go#L700-L701

This was not possible before, as duration was calculated in `PreCommitSector`, where a bounds check implies that a sector's activation epoch cannot come before its expiration: https://github.com/filecoin-project/specs-actors/blob/0dcd04ea4dca004872a4984d40179a67765f2544/actors/builtin/miner/miner_actor.go#L447-L449

#### Recommendation

Because `ConfirmSectorProofsValid` needs to consider null rounds, I think the best way to handle this would be to simply add a check here: https://github.com/filecoin-project/specs-actors/blob/0dcd04ea4dca004872a4984d40179a67765f2544/actors/builtin/miner/miner_actor.go#L701

If the result is zero, or negative, this function should probably `Abortf`

acruikshank:
closes #795

### Motivation

It's probably not possible for prove commit to occur past a valid sector's lifetime, but if it does, it will trigger panics in the power math, causing all the other sectors to fail. This PR check that the sector lifetime is less than equal to zero and skips the sector if so.

### Proposed Changes

1. Use current epoch + MaxSealDuration for activation time when validating in precommit.
2. Add check that sector duration meets or exceeds minimum in prove commit, and skip sector if not.
3. Test

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -475,6 +475,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		}
 
 		// Require sector lifetime meets minimum by assuming activation happens at last epoch permitted for seal proof.
+		// This could make sector maximum lifetime validation more lenient if the maximum sector limit isn't hit first.
 		maxActivation := rt.CurrEpoch() + MaxSealDuration[params.SealProof]
 		validateExpiration(rt, maxActivation, params.Expiration, params.SealProof)
 
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -676,6 +676,13 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			// compute initial pledge
 			activation := rt.CurrEpoch()
 			duration := precommit.Info.Expiration - activation
+
+			// Avoid math errors if duration is too low somehow.
+			if duration <= 0 {
+				rt.Log(vmr.WARN, "precommit %d has non-positive lifetime %d. ignoring", precommit.Info.SectorNumber, duration)
+				continue
+			}
+
 			power := QAPowerForWeight(info.SectorSize, duration, precommit.DealWeight, precommit.VerifiedDealWeight)
 			dayReward := ExpectedRewardForPower(rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, power, builtin.EpochsInDay)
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -669,6 +669,13 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			// compute initial pledge
 			activation := rt.CurrEpoch()
 			duration := precommit.Info.Expiration - activation
+
+			// Avoid math errors if duration is too low somehow.
+			if duration <= 0 {
```
Stebalien, line 674: I wonder if we should be checking min sector lifetime instead?

anorth: Yes.
1. PreCommitSector should add the max seal duration to current epoch when invoking `validateExpiration()` ([from the issue](https://github.com/filecoin-project/specs-actors/issues/795#issuecomment-664100263))
2. Here we should check again that the min lifetime meets the bound (which we can assert is positive)

actors/builtin/miner/miner_actor.go
```diff
@@ -457,7 +457,9 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 			rt.Abortf(exitcode.ErrIllegalState, "sector %v already committed", params.SectorNumber)
 		}
 
-		validateExpiration(rt, rt.CurrEpoch(), params.Expiration, params.SealProof)
+		// Require sector lifetime meets minimum by assuming activation happens at last epoch permitted for seal proof.
+		maxActivation := rt.CurrEpoch() + MaxSealDuration[params.SealProof]
```
Stebalien, line 461: So, this will allow the sector to potentially exceed the maximum lifetime (if it activates early). Is that OK? If so, we should document it somewhere.

anorth: I'm fairly confident this is the calculation that @nicola verbally expressed to me. Yes, an extra comment sentence here would be good.

nicola: If I am not wrong, I will check this with @irenegia, the alternative is having some PreCommits that pass, but their ProveCommit fails (if we check expiration at ProveCommit).

Issue #897, opened 2020-08-05
Bound length of strings and arrays in state


anorth:
A few items in state are strings or byte arrays that are provided from off chain. We should bound the length of these strings to a tight reasonable bound, else we could incur unwanted de/serialization cost as these things are read and written (potentially by parties other than the party controlling the value).

- [ ] DealProposal.Label: 100s of bytes?
- [x] MinerInfo.PeerID: 100s of bytes?
- [x] MinerInfo.Multiaddrs: array length of 10s? each value 100s of bytes?
- [ ] Multisig.Signers: array length of 100s?

The Address type is limited on CBOR decoding to 64 bytes, though it [should](https://github.com/filecoin-project/go-address/issues/9) also check on encoding.

BigIntegers are limited to 128 bytes, which is probably ok from a serde point of view (though maybe not for arithmetic, #866).

Related to #604

Stebalien:
#920 currently deals with MinerInfo.PeerID and MinerInfo.Multiaddrs.

anorth:
@Stebalien done?

Stebalien:
Not quite. Still need deal label, and signers.

I'm picking 256 for the deal label, same for max signers.

Issue #727, opened 2020-07-20
Verifreg.UseBytes: newVcCap is permanently lost if it's smaller than MinVerifiedDealSize


wadeAlexC:
From the code: https://github.com/filecoin-project/specs-actors/blob/8c473f656bb07b19c61991f452d1e57d59dc5083/actors/builtin/verifreg/verified_registry_actor.go#L165-L172

The comment implies that this should somehow be restored via `RestoreBytes`, but I don't think that happens. Rather than deleting the verified client, consider keeping them active so that `newVcCap` is still usable if `RestoreBytes` is called.

aarshkshah1992:
@whyrusleeping @anorth I agree with @wadeAlexC here.

Do you guys have any objections before I make this change ?

whyrusleeping:
This was an explicit design choice. We are prioritizing easy cleanup of the structures involved.

> The comment implies that this should somehow be restored via RestoreBytes, but I don't think that happens.

It does, see the `RestoreBytes` method, its called by the storage market actor near the end of its cron tick if a verified deal does not get activated in time.

wadeAlexC:
To be clear, I'm referring to the edge case where `newVcCap < MinVerifiedDealSize` - in this case, the entry is deleted, and `newVcCap` is permanently lost - it won't be recovered in `RestoreBytes`. 

I can understand why it would be a design decision to just drop this remainder, though.

aarshkshah1992:
@whyrusleeping Question is, do we want to drop the remainder or keep the verified client around in case it's balance is replenished later via `RestoreBytes` ?

whyrusleeping:
Yeah, dropping the remainder is fine. Its an edge case thats probably not worth handling. Getting more verified datacap (especially at those low of sizes) should be relatively easy if you need it.

PR #920, opened 2020-08-06
Check parameters

Stebalien:
* Limit the size of the peer ID to 128 bytes.
* Limit the total size of all multiaddrs combined to 1024 bytes, and forbid empty multiaddrs.
* Require at most one window post and require that it (if present) match the miner's declared proof type. This should be fine (?) given that we don't currently support changing proof types at runtime anyways.
* Cleanup expiration validation.
* Validate "max sector number" in a couple of places. Honestly, this isn't strictly speaking necessary, but it's nice to do.
* Log if we somehow end up trying to confirm more sector proofs than we should.
* Verify that we don't allow extending sector expirations after their expiration.
* Limit payment channel secrets to 256 bytes. Really, they shouldn't ever be more than 32.
* Limit seal proofs to 1KiB.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -86,6 +86,8 @@ type ConstructorParams = power.MinerConstructorParams
 func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.InitActorAddr)
 
+	checkPeerInfo(rt, params.PeerId, params.Multiaddrs)
+
 	_, ok := SupportedProofTypes[params.SealProofType]
 	if !ok {
 		rt.Abortf(exitcode.ErrIllegalArgument, "proof type %d not allowed for new miner actors", params.SealProofType)
@@ -230,8 +232,8 @@ type ChangePeerIDParams struct {
 }
 
 func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyValue {
-	// TODO: Consider limiting the maximum number of bytes used by the peer ID on-chain.
-	// https://github.com/filecoin-project/specs-actors/issues/712
+	checkPeerInfo(rt, params.NewID, nil)
+
 	var st State
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
@@ -250,8 +252,8 @@ type ChangeMultiaddrsParams struct {
 }
 
 func (a Actor) ChangeMultiaddrs(rt Runtime, params *ChangeMultiaddrsParams) *adt.EmptyValue {
-	// TODO: Consider limiting the maximum number of bytes used by multiaddrs on-chain.
-	// https://github.com/filecoin-project/specs-actors/issues/712
+	checkPeerInfo(rt, nil, params.NewMultiaddrs)
+
 	var st State
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
@@ -310,7 +312,6 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 	if !bytes.Equal(commRand, params.ChainCommitRand) {
 		rt.Abortf(exitcode.ErrIllegalArgument, "post commit randomness mismatched")
 	}
-	// TODO: limit the length of proofs array https://github.com/filecoin-project/specs-actors/issues/416
 
 	// Get the total power/reward. We need these to compute penalties.
 	rewardStats := requestCurrentEpochBlockReward(rt)
@@ -326,17 +327,26 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 
 		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
 
+		// Verify that the miner has passed 0 or 1 proofs. If they've
+		// passed 1, verify that it's a good proof.
+		//
+		// This can be 0 if the miner isn't actually proving anything,
+		// just skipping all sectors.
+		windowPoStProofType, err := info.SealProofType.RegisteredWindowPoStProof()
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to determine window PoSt type")
+		if len(params.Proofs) > 1 {
+			rt.Abortf(exitcode.ErrIllegalArgument, "expected at most one proof, got %d", len(params.Proofs))
+		} else if len(params.Proofs) == 1 && params.Proofs[0].PoStProof != windowPoStProofType {
+			rt.Abortf(exitcode.ErrIllegalArgument, "expected proof of type %s, got proof of type %s", params.Proofs[0], windowPoStProofType)
+		}
+
 		// Validate that the miner didn't try to prove too many partitions at once.
 		submissionPartitionLimit := loadPartitionsSectorsMax(info.WindowPoStPartitionSectors)
 		if uint64(len(params.Partitions)) > submissionPartitionLimit {
 			rt.Abortf(exitcode.ErrIllegalArgument, "too many partitions %d, limit %d", len(params.Partitions), submissionPartitionLimit)
 		}
 
-		// Load and check deadline.
 		currDeadline := st.DeadlineInfo(currEpoch)
-		deadlines, err := st.LoadDeadlines(adt.AsStore(rt))
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
-
 		// Check that the miner state indicates that the current proving deadline has started.
 		// This should only fail if the cron actor wasn't invoked, and matters only in case that it hasn't been
 		// invoked for a whole proving period, and hence the missed PoSt submissions from the prior occurrence
@@ -353,6 +363,9 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 
 		sectors, err := LoadSectors(store, st.Sectors)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors")
+
+		deadlines, err := st.LoadDeadlines(adt.AsStore(rt))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
 
 		deadline, err := deadlines.LoadDeadline(store, params.Deadline)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d", params.Deadline)
@@ -380,6 +393,10 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		// Skip verification if all sectors are faults.
 		// We still need to allow this call to succeed so the miner can declare a whole partition as skipped.
 		if len(sectorInfos) > 0 {
+			if len(params.Proofs) == 0 {
+				// The miner _was_ supposed to prove something, but didn't.
+				rt.Abortf(exitcode.ErrIllegalArgument, "no proofs submitted in window PoSt for %d sectors", len(sectorInfos))
+			}
 			// Verify the proof.
 			// A failed verification doesn't immediately cause a penalty; the miner can try again.
 			//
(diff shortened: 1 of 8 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -132,7 +132,7 @@ func (a Actor) AddBalance(rt Runtime, providerOrClientAddress *addr.Address) *ad
 }
 
 type PublishStorageDealsParams struct {
```
Stebalien, line 134: Do we want to specify a "max deals per call" limit? It doesn't seem necessary (gas and message sizes should handle this) but we don't have any explicit limits here.

actors/builtin/miner/miner_actor.go
```diff
@@ -242,7 +242,7 @@ type SubmitWindowedPoStParams struct {
 	Partitions []PoStPartition
 	// Array of proofs, one per distinct registered proof type present in the sectors being proven.
 	// In the usual case of a single proof type, this array will always have a single element (independent of number of partitions).
-	Proofs []abi.PoStProof
+	Proofs []abi.PoStProof // PARCHECK: arbitrary size, proof type not registered, proofs can have arbitrary sizes.
```
Stebalien, line 245: These will be checked by the runtime, but I'm not sure what assumptions the runtime makes.

actors/builtin/miner/miner_actor.go
```diff
@@ -275,11 +275,7 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 			rt.Abortf(exitcode.ErrIllegalArgument, "too many partitions %d, limit %d", len(params.Partitions), submissionPartitionLimit)
 		}
 
-		// Load and check deadline.
 		currDeadline := st.DeadlineInfo(currEpoch)
-		deadlines, err := st.LoadDeadlines(adt.AsStore(rt))
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
```
Stebalien, line 281: Small change to perform all checks before loading state.

actors/builtin/miner/miner_actor.go
```diff
@@ -407,9 +406,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		rt.Abortf(exitcode.ErrIllegalArgument, "seal challenge epoch %v too old, must be after %v", params.SealRandEpoch, challengeEarliest)
 	}
 
-	if params.Expiration <= rt.CurrEpoch() {
-		rt.Abortf(exitcode.ErrIllegalArgument, "sector expiration %v must be after now (%v)", params.Expiration, rt.CurrEpoch())
-	}
+	validateExpiration(rt, rt.CurrEpoch(), params.Expiration, params.SealProof)
```
Stebalien, line 409: Small change to consolidate expiration validation, and to validate expiration as soon as possible.

actors/builtin/miner/miner_actor.go
```diff
@@ -582,6 +585,13 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSectorProofsParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.StoragePowerActorAddr)
 
+	if len(params.Sectors) > power.MaxMinerProveCommitsPerEpoch {
+		// XXX: log only?
```
Stebalien, line 589: This shouldn't be possible. Personally, I'd remove this or log, but I wanted to bring it up.

actors/builtin/miner/miner_actor.go
Stebalien: TL;DR: is it safe to trust the runtime to verify proofs and handle arbitrarily sized proofs?

actors/builtin/miner/miner_actor.go
```diff
@@ -1282,9 +1308,9 @@ func (a Actor) AddLockedFund(rt Runtime, amountToLock *abi.TokenAmount) *adt.Emp
 }
 
 type ReportConsensusFaultParams struct {
-	BlockHeader1     []byte
-	BlockHeader2     []byte
-	BlockHeaderExtra []byte
+	BlockHeader1     []byte // PARCHECK: arbitrary size
+	BlockHeader2     []byte // PARCHECK: arbitrary size
+	BlockHeaderExtra []byte // PARCHECK: arbitrary size
```
Stebalien, line 1313: Again, this is another "can I just trust the runtime" question.

actors/builtin/paych/paych_actor.go
```diff
@@ -86,7 +86,7 @@ func (pca *Actor) resolveAccount(rt vmr.Runtime, raw addr.Address) (addr.Address
 
 type UpdateChannelStateParams struct {
 	Sv     SignedVoucher
-	Secret []byte
+	Secret []byte // PARCHECK: hash of unbounded byte string
```
Stebalien, line 89: It doesn't look like we even use this in lotus.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -123,7 +123,7 @@ type ProposeParams struct {
 	To     addr.Address
 	Value  abi.TokenAmount
 	Method abi.MethodNum
-	Params []byte
+	Params []byte // PARCHECK: unbounded
```
Stebalien, line 126: This almost certainly isn't an issue, I just wanted to check.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -64,9 +64,9 @@ func (a Actor) Exports() []interface{} {
 var _ abi.Invokee = Actor{}
 
 type ConstructorParams struct {
-	Signers               []addr.Address
+	Signers               []addr.Address // PARCHECK unbounded
 	NumApprovalsThreshold uint64
-	UnlockDuration        abi.ChainEpoch
+	UnlockDuration        abi.ChainEpoch // PARCHECK unbounded
```
Stebalien, line 69: I assume these are fine, but I wanted to check to see if there are any constraints in the crypto on the max number of signers.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -218,6 +220,7 @@ func (a Actor) UseBytes(rt vmr.Runtime, params *UseBytesParams) *adt.EmptyValue
 		if newVcCap.LessThan(MinVerifiedDealSize) {
 			// Delete entry if remaining DataCap is less than MinVerifiedDealSize.
 			// Will be restored later if the deal did not get activated with a ProvenSector.
+			// XXX: This isn't quite true. The client can lose up to MinVerifiedDealSize.
```
Stebalien, line 223: Do we care about this?

actors/builtin/market/market_actor.go
anorth: Message size limit is probably enough.

 I'm torn about whether as a matter of policy we should set a limit on all such slices anyway, though.

actors/builtin/market/market_actor.go
```diff
@@ -251,7 +251,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 }
 
 type VerifyDealsForActivationParams struct {
-	DealIDs      []abi.DealID
+	DealIDs      []abi.DealID // PARCHECK: unbounded
```
anorth, line 254: These ones are called by actor code, so we need to check the call sites will respect the bound.

actors/builtin/miner/miner_actor.go
anorth: Probably safe to rely on message size limits, but an explicit bound would still be even safer.

actors/builtin/miner/miner_actor.go
```diff
@@ -843,12 +859,22 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
...
 				for i, sector := range oldSectors {
+					// XXX: < or <=? Should we even have this?
+					// This can happen if the sector should
+					// have already expired, but hasn't
+					// because the end of its deadline
+					// hasn't passed yet.
+					if sector.Expiration <= currEpoch {
+						rt.Abortf(exitcode.ErrIllegalArgument, "cannot extend expiration for expired sector %v",
+							sector.SectorNumber)
+					}
```
anorth, line 873: Good find, I like it.

actors/builtin/multisig/multisig_actor.go
anorth: I noted this in #897

actors/builtin/miner/miner_actor.go
```diff
@@ -528,7 +545,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 
 type ProveCommitSectorParams struct {
 	SectorNumber abi.SectorNumber
-	Proof        []byte
+	Proof        []byte // PARCHECK: arbitrary size
```
Stebalien, line 550: It's usually fine to allow arbitrary size proofs. However, I'm a bit concerned in this case because prove-commit proofs are batched for async processing. That means we're not going to validate until later.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -109,14 +109,16 @@ func (a Actor) RemoveVerifier(rt vmr.Runtime, verifierAddr *addr.Address) *adt.E
 }
 
 type AddVerifiedClientParams struct {
-	Address   addr.Address
+	Address   addr.Address // PARCHECK: could be a bogus address
```
Stebalien: This will be solved along with https://github.com/filecoin-project/specs-actors/issues/556.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -47,7 +47,7 @@ func (a Actor) Constructor(rt vmr.Runtime, rootKey *addr.Address) *adt.EmptyValu
 }
 
 type AddVerifierParams struct {
-	Address   addr.Address
+	Address   addr.Address // PARCHECK: could be a bogus address
```
Stebalien: This will be solved along with https://github.com/filecoin-project/specs-actors/issues/556.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -109,14 +109,16 @@ func (a Actor) RemoveVerifier(rt vmr.Runtime, verifierAddr *addr.Address) *adt.E
...
 	Allowance DataCap
 }
 
 func (a Actor) AddVerifiedClient(rt vmr.Runtime, params *AddVerifiedClientParams) *adt.EmptyValue {
 	if params.Allowance.LessThan(MinVerifiedDealSize) {
 		rt.Abortf(exitcode.ErrIllegalArgument, "allowance %d below MinVerifiedDealSize for add verified client %v", params.Allowance, params.Address)
 	}
+	// The caller will be verified by checking the verifiers table below.
+	// PARCHECK: Should this be restricted to a specific type of ID?
```
Stebalien: This will be solved along with #556.

Stebalien:
@anorth I think I've resolved most of my questions (or punted them) but there are still two remaining (unresolved comments above).

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -262,6 +264,7 @@ func (a Actor) RestoreBytes(rt vmr.Runtime, params *RestoreBytesParams) *adt.Emp
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load verifiers")
 
 		// validate we are NOT attempting to do this for a verifier
+		// XXX: This seems expensive and pointless?
```
Stebalien, line 267: This is going to cost us some (expensive) state accesses, to check something that should be impossible. We can keep it, it just doesn't seem worth it.

actors/builtin/verifreg/verified_registry_actor.go
anorth: No. Please add a note pointing to #727

actors/builtin/verifreg/verified_registry_actor.go
anorth: Should be impossible because a verifier should not appear in the clients table? Yes, SGTM, just leave a comment noting that.

Stebalien: Not quite because we may have removed it from the clients table (if zero).

However, this method can only be called by cron when a verified deal times out.

Stebalien: Ok, I'm reverting my change here unless we resolve https://github.com/filecoin-project/specs-actors/pull/758/files#r470369407.

actors/builtin/miner/miner_actor.go
Stebalien: Done.

actors/builtin/miner/miner_actor.go
```diff
@@ -520,13 +539,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
-		// Require sector lifetime meets minimum by assuming activation happens at last epoch permitted for seal proof.
-		// This could make sector maximum lifetime validation more lenient if the maximum sector limit isn't hit first.
-		maxActivation := rt.CurrEpoch() + MaxSealDuration[params.SealProof]
-		validateExpiration(rt, maxActivation, params.Expiration, params.SealProof)
-
 		depositMinimum := big.Zero()
 		if params.ReplaceCapacity {
+			// NOTE: we don't validate that the replacement sector
+			// lives at the target deadline/partition. If it
+			// doesn't, sealing may still succeed but replacement will fail.
```
wadeAlexC, line 546: This is validated within `validateReplaceSector` -> `CheckSectorHealth`

wadeAlexC: And it probably should be validated if it isn't, because you might be able to queue up a replacement for a sector that hasn't been proven yet.

Stebalien: Hm. You're right. Although the target sector could expire before we even finish... but I guess that doesn't matter.

> because you might be able to queue up a replacement for a sector that hasn't been proven yet.

We _do_ check to make sure that the sector to-be-replaced has been proven.

PR #922, opened 2020-08-07
Use RequireNoErr when aborting on error

Stebalien:
This ensures that we use the correct error code. This patch also includes some additional information.


anorth:
Oops, we collided: #926

Could you aggregate into one PR?

Stebalien:
Done. I've pushed to your branch.

PR #926, opened 2020-08-07
Fix ignored exit codes in paych, and funky tests

anorth:
FYI @austinabell


actors/builtin/paych/paych_actor.go
```diff
@@ -46,13 +46,9 @@ func (pca *Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Em
 
 	// check that both parties are capable of signing vouchers
 	to, err := pca.resolveAccount(rt, params.To)
-	if err != nil {
-		rt.Abortf(exitcode.ErrIllegalArgument, err.Error())
-	}
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve to address")
```
anorth, line 49: In the current code, this will never fall back to `ErrIllegalState` because every possible error returned from `resolveAccount` carries its own exit code.

PR #930, opened 2020-08-07
More Miner tests

### LINKED ISSUE #479: More miner_test coverage
The following testing suggestions are taken from TODOs in builtin/miner/miner_test.go

`TestControlAddresses`
* changing worker (with delay)
* changing peer id

`TestCommitments`
* test insufficient funds when the precommit deposit is set above zero
* too early to prove sector
* seal rand epoch too old
* commitment expires before proof
* invalid seal proof (more thought needs to go into how the test should work)

ZenGround0:
Closes #479 

- Simple ChangePeerID happy path test
- A few new sector commitment edge case tests
- Aggregate some test ideas from todo and new test ideas into linked issue

Worker changes already handled elsewhere

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -1730,6 +1730,26 @@ func TestWithdrawBalance(t *testing.T) {
 	})
 }
 
+func TestChangePeerID(t *testing.T) {
+	periodOffset := abi.ChainEpoch(100)
+
+	setupFunc := func() (*mock.Runtime, *actorHarness) {
+		actor := newHarness(t, periodOffset)
+		builder := builderForHarness(actor).
+			WithBalance(bigBalance, big.Zero())
+		rt := builder.Build(t)
+
+		return rt, actor
+	}
+
+	t.Run("successfully change peer id", func(t *testing.T) {
+		rt, actor := setupFunc()
+		actor.constructAndVerify(rt)
+		newPID := tutil.MakePID("test-change-peer-id")
+		actor.changePeerID(rt, newPID)
+	})
+}
+
 func TestChangeWorkerAddress(t *testing.T) {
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1693,6 +1710,26 @@ func TestWithdrawBalance(t *testing.T) {
 	})
 }
 
+func TestChangePeerID(t *testing.T) {
+	periodOffset := abi.ChainEpoch(100)
+
+	setupFunc := func() (*mock.Runtime, *actorHarness) {
+		actor := newHarness(t, periodOffset)
```
anorth, line 1717: I don't like these setup funcs generally (because they couple different cases and are easy to abuse). But another reason is that this captures the wrong `t`. It must be a parameter to `setupFunc`.

actors/builtin/miner/miner_test.go
```diff
@@ -141,18 +141,19 @@ func TestControlAddresses(t *testing.T) {
...
 
 // Test for sector precommitment and proving.
 func TestCommitments(t *testing.T) {
 	periodOffset := abi.ChainEpoch(100)
 
 	// TODO more tests
+	// https://github.com/filecoin-project/specs-actors/issues/936
 	// - Concurrent attempts to upgrade the same CC sector (one should succeed)
-	// - Insufficient funds for pre-commit, for prove-commit
+	// - Insufficient funds for prove-commit (charged in power actor -- scenario testing candidate)
```
acruikshank, line 153: Since these are now covered in an issue (Thanks!), do they still need to be documented as comments? My concern is that this becomes a redundant and eventually inconsistent work queue.

ZenGround0: Yeah great point, I will delete

PR #939, opened 2020-08-10
VerifReg: Resolve Verifier and Client Addresses

### LINKED ISSUE #556: Verified Registry needs to resolve addresses

aarshkshah1992:
For #730 AND Closes #556.

@anorth To review after  #912  goes in.

### DIFF
--- a/actors/builtin/verifreg/verified_registry_actor.go
+++ b/actors/builtin/verifreg/verified_registry_actor.go
@@ -52,17 +52,20 @@ type AddVerifierParams struct {
 }
 
 func (a Actor) AddVerifier(rt vmr.Runtime, params *AddVerifierParams) *adt.EmptyValue {
-	if params.Allowance.LessThan(MinVerifiedDealSize) {
-		rt.Abortf(exitcode.ErrIllegalArgument, "Allowance %d below MinVerifiedDealSize for add verifier %v", params.Allowance, params.Address)
-	}
-
 	var st State
 	rt.State().Readonly(&st)
 	rt.ValidateImmediateCallerIs(st.RootKey)
 
+	verifierIdAddr, err := builtin.ResolveToIDAddr(rt, params.Address)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve verifier address %v to ID address", params.Address)
+
+	if params.Allowance.LessThan(MinVerifiedDealSize) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "Allowance %d below MinVerifiedDealSize for add verifier %v", params.Allowance, verifierIdAddr)
+	}
+
 	// TODO We need to resolve the verifier address to an ID address before making this comparison.
 	// https://github.com/filecoin-project/specs-actors/issues/556
-	if params.Address == st.RootKey {
+	if verifierIdAddr == st.RootKey {
 		rt.Abortf(exitcode.ErrIllegalArgument, "Rootkey cannot be added as verifier")
 	}
 	rt.State().Transaction(&st, func() {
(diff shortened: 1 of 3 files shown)


wadeAlexC:
LGTM

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -52,17 +52,20 @@ type AddVerifierParams struct {
...
 	var st State
 	rt.State().Readonly(&st)
 	rt.ValidateImmediateCallerIs(st.RootKey)
 
+	verifierIdAddr, err := builtin.ResolveToIDAddr(rt, params.Address)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve verifier address %v to ID address", params.Address)
+
+	if params.Allowance.LessThan(MinVerifiedDealSize) {
+		rt.Abortf(exitcode.ErrIllegalArgument, "Allowance %d below MinVerifiedDealSize for add verifier %v", params.Allowance, verifierIdAddr)
+	}
```
anorth, line 64: Leave the checks that don't require state above the state load call.

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -94,11 +97,14 @@ func (a Actor) RemoveVerifier(rt vmr.Runtime, verifierAddr *addr.Address) *adt.E
 	rt.State().Readonly(&st)
 	rt.ValidateImmediateCallerIs(st.RootKey)
 
+	verifierIdAddr, err := builtin.ResolveToIDAddr(rt, *verifierAddr)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve verifier address %v to ID address", *verifierAddr)
```
anorth, line 101: Check before loading state

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -114,16 +120,17 @@ type AddVerifiedClientParams struct {
 }
 
 func (a Actor) AddVerifiedClient(rt vmr.Runtime, params *AddVerifiedClientParams) *adt.EmptyValue {
+	vClientIdAddr, err := builtin.ResolveToIDAddr(rt, params.Address)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve verified client address %v", params.Address)
```
anorth, line 124: Move after the simple checks, but before the state load

actors/builtin/verifreg/verified_registry_actor.go
```diff
@@ -242,15 +252,16 @@ type RestoreBytesParams struct {
 func (a Actor) RestoreBytes(rt vmr.Runtime, params *RestoreBytesParams) *adt.EmptyValue {
 	rt.ValidateImmediateCallerIs(builtin.StorageMarketActorAddr)
 
+	vClientIdAddr, err := builtin.ResolveToIDAddr(rt, params.Address)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve verified client addr %v", params.Address)
+
 	if params.DealSize.LessThan(MinVerifiedDealSize) {
```
anorth, line 258: Check this first

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: Done.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: Done.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: Done.

actors/builtin/verifreg/verified_registry_actor.go
aarshkshah1992: Done.

aarshkshah1992:
@anorth Have made all the changes. Please take a look.

actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -184,6 +211,21 @@ func TestRemoveVerifier(t *testing.T) {
 
 		ac.removeVerifier(rt, va)
 	})
+
+	t.Run("add verifier with non ID address and then remove with it's ID address", func(t *testing.T) {
```
anorth, line 215: Nit: "its"

aarshkshah1992: This is done.

PR #941, opened 2020-08-10
Miner: Allow owner key to act as miner

### LINKED ISSUE #547: Allow owner key to act as miner [was: while key change is in progress]

aarshkshah1992:
Closes #547 .

@anorth Do we need to write a test for each of the functions we've touched by setting the caller as the owner and ensuring things work ? Looks like adding a lot of testing code to test a simple change. Let me know.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -198,7 +198,7 @@ func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyVa
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
 
-		rt.ValidateImmediateCallerIs(info.Worker)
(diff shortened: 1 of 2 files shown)


anorth:
@aarshkshah1992 no we don't need to duplicate all the tests, but please test _one_ of them, alongside a test that other addresses are rejected. Something simple like changing peer ids or compacting partitions.

aarshkshah1992:
@anorth Have addressed both your comments.

Issue #888, opened 2020-08-04
Record miner fee debt when penalties exceed unlocked balance


anorth:
Since #808, a miner pays fees first out of locked funds, then from the unlocked funds covering their initial pledge. When they reach the point where only funds covering pre-commit deposit (if any) remain, they are out of money to pay any subsequent penalty.

Issue #627 proposed terminating the miner at this point, but we don't need to be that harsh. We can allow a distressed miner to go into fee debt (not to be confused with IP debt), so long as they repay the debt before recovering full capabilities (see #886). As an example, if a miner with one sector has a fault, and runs out of funds after 7 days of consecutive fees, we allow them to stay faulty and accumulate fee debt, but recover the sector on the 13th day if they repay all fees. The miner would need to pay back the fee debt, and additionally restore the IP requirement.

- Add a new field to miner state recording the amount of fee debt
- When a miner needs to pay fees in excess of available balance, record that as fee debt, instead of ignoring it as we currently do
- When evaluating the condition of IP debt
    - If the miner has fee debt and any unlocked balance, burn the balance and reduce the fee debt appropriately
    - Consider a miner to be in IP debt if they have any fee debt, even if IP requirement is zero (thus restricted in capabilities as for #886)

I don't think we need a method for the miner operator to call explicitly to inspect the fee repayment, but there might turn out to be a case where we can't do it automatically.

An explicit flow to exercise in tests is a miner going into fee debt with a faulty sector, then calling DeclareFaultsRecovered in a message carrying enough balance to pay back the debt and recover IP requirements.

A miner in fee debt is not expected to receive any funds through `AddLockedFunds` because they can't win an election.

@zixuanzh @nicola @davidad please review and request any changes to this spec

zixuanzh:
LGTM except I dont understand the statement

> I don't think we need a method for the miner operator to call explicitly to inspect the fee repayment, but there might turn out to be a case where we can't do it automatically.

anorth:
Ok this is good to go.

PR #943, opened 2020-08-10
Feat/update ip debt restrictions

### LINKED ISSUE #886: Restrict miner capabilities while in IP debt
A miner is considered in "IP debt" when their unlocked balance does not cover the initial pledge requirements for all committed and non-terminated sectors. 

At present, when in IP debt, the following methods are disallowed:
- ProveCommitSector
- WithdrawFunds

It's also my understanding that the miner should become ineligible to win leader election though that it outside actors scope (though we could help by adding a static helper method similar to `MinerNominalPowerMeetsConsensusMinimum`)

IP debt should be a bad enough state that a miner is incentivised to top-up pledge collateral, without triggering a cascading failure that makes the situation worse. In particular, we definitely want to continue allowing Window PoSt so the miner does not necessarily incur more penalties.

Which other methods should be restricted in order to make repayment the miner's top priority? I propose:
- PreCommitSector ("no growth while in debt")
- DeclareFaultsRecovered ("pay penalties as condition of recovery")

Methods we should explicitly still allow:
- DeclareFaults (don't exacerbate the cost of miner operational issues)
- TerminateSectors (ditto)
- ReportConsensusFault (called by someone else, not the miner operator)

Other candidates that don't seem high value, so default to allowing normal operation:
- ExtendSectorExpiration
- CompactPartitions

@zixuanzh @nicola Please ack or add your opinions.

ZenGround0:
Closes #886 

In preparation for #888 I'm moving the requirements check within a transaction because we will soon be doing a state mutation to pay off debt every time we do the verification check.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -448,6 +448,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	var st State
 	newlyVested := big.Zero()
 	rt.State().Transaction(&st, func() {
+		verifyPledgeMeetsInitialRequirements(rt, &st)
+
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(info.Worker, info.Owner)
 		if params.SealProof != info.SealProofType {
@@ -547,16 +549,18 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 
 	store := adt.AsStore(rt)
 	var st State
-	rt.State().Readonly(&st)
-
-	verifyPledgeMeetsInitialRequirements(rt, &st)
-
+	var precommit *SectorPreCommitOnChainInfo
 	sectorNo := params.SectorNumber
-	precommit, found, err := st.GetPrecommittedSector(store, sectorNo)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-committed sector %v", sectorNo)
-	if !found {
-		rt.Abortf(exitcode.ErrNotFound, "no pre-committed sector %v", sectorNo)
-	}
+	rt.State().Transaction(&st, func() {
+		verifyPledgeMeetsInitialRequirements(rt, &st)
+		var found bool
+		var err error
+		precommit, found, err = st.GetPrecommittedSector(store, sectorNo)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-committed sector %v", sectorNo)
+		if !found {
+			rt.Abortf(exitcode.ErrNotFound, "no pre-committed sector %v", sectorNo)
+		}
+	})
 
 	msd, ok := MaxSealDuration[precommit.Info.SealProof]
 	if !ok {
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -544,21 +546,18 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
...
-
+	var precommit *SectorPreCommitOnChainInfo
 	sectorNo := params.SectorNumber
-	precommit, found, err := st.GetPrecommittedSector(store, sectorNo)
-	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-committed sector %v", sectorNo)
-	if !found {
-		rt.Abortf(exitcode.ErrNotFound, "no pre-committed sector %v", sectorNo)
-	}
+	rt.State().Transaction(&st, func() {
+		verifyPledgeMeetsInitialRequirements(rt, &st)
```
acruikshank, line 552: This is dropping the comment about verify funds. I believe that information is still correct and relevant. The comment should move to the `verifyPledgeMeetsIntialRequirements` declaration.

actors/builtin/miner/miner_test.go
```diff
@@ -1353,6 +1366,37 @@ func TestDeclareFaults(t *testing.T) {
 	})
 }
 
+func TestDeclareRecoveries(t *testing.T) {
```
acruikshank, line 1369: This is great. Since your at it, a quick happy path test of declare recoveries would be nice. Should be the the same test without the IP Debt part and a state assertion that the sector has been declared recovered.

actors/builtin/miner/miner_actor.go
ZenGround0: We no longer vest IPs so I believe the whole comment is now incorrect and misleading.

actors/builtin/miner/miner_test.go
```diff
@@ -1353,6 +1366,60 @@ func TestDeclareFaults(t *testing.T) {
...
+	periodOffset := abi.ChainEpoch(100)
+	actor := newHarness(t, periodOffset)
+	builder := builderForHarness(actor).
+		WithBalance(bigBalance, big.Zero())
+
+	t.Run("recovery happy path", func(t *testing.T) {
```
acruikshank, line 1375: ❤️ 🙏

actors/builtin/miner/miner_actor.go
```diff
@@ -445,6 +445,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	var st State
 	newlyVested := big.Zero()
 	rt.State().Transaction(&st, func() {
+		verifyPledgeMeetsInitialRequirements(rt, &st)
```
anorth, line 448: This method name is getting dusty. How about `verifyBalanceCoversPledge`?

actors/builtin/miner/miner_actor.go
anorth: I think there's still something to call out here.
- It's still true that this doesn't compute vesting (keep it that way)
- A miner could be in IP debt but have locked funds(*), and this won't unlock them
- So it's not in the miner's favour any more

(*) Maybe this is in fact impossible in the current code, but if so, it's a brittle condition depending on the implementation of AwardBlockReward, and could easily change.

actors/builtin/miner/miner_actor.go
ZenGround0: Agree it could use a new name.  It's going to need to change again in the upcoming fee debt pr so I'll put off the the change until that one.

PR #949, opened 2020-08-10
Add a kinded inline type

Stebalien:
This special type will auto-inline values under the inline limit. Values over the inline limit will be stored in an external object and linked by a CID.

Pros:

* Doesn't require any buy in from IPLD and other implementations. This is just another ADT.
* It doesn't require any inline CID support in blockstores.
* We can use this to auto-inline specific things like the sector ID allocation bitfield.

Cons:

* This requires explicitly auto-inlining every CID that might be relevant.
* It won't transparently apply to HAMTs, AMTs, etc and explicitly applying this special ADT inside those structures would _massively_ complicate them.

In short, I'm not really a fan of this solution, but it's _a_ solution if inline CIDs don't work out for some reason.

### DIFF
--- a/actors/util/adt/inline.go
+++ b/actors/util/adt/inline.go
@@ -1,0 +1,84 @@
+package adt
+
+import (
+	"bytes"
+	"io"
+
+	"github.com/ipfs/go-cid"
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+const InlineLimit = 64
+
+type AutoInline struct {
+	link   cid.Cid
+	inline cbg.Deferred
+}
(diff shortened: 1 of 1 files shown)


Stebalien:
cc @mikeal.

mikeal:
Love it!

anorth:
Thanks, demonstrations much appreciated.

> It won't transparently apply to HAMTs, AMTs, etc

🤞 that inline CIDs work out.

Stebalien:
We've decided to not go with this approach as it adds more complexity than it's worth.

PR #954, opened 2020-08-11
Miner: Multiple control addresses

### LINKED ISSUE #942: Multiple control addresses for miner actors
Having an additional set of control addresses in MinerInfo would allow mining software to very easily set different gas fee caps depending on message types (e.g. set way higher fee caps on WindowPost messages vs [Pre]Comits) - currently if a miner has a bunch of other messages queued in mempool, to quickly submit a windowPost they have to either pay a lot higher fee for those queued messages (depending on chain conditions 10-100x higher than necessary), or replace one of the queued messages with the windowPoSt message, which is very complicated, and not atomic at all.

All operations which are based on the worker key would still use it, control addresses could only be used for submitting those messages

https://github.com/filecoin-project/specs-actors/pull/941 looks somewhat related, but I don't think I want to 'exploit' this, and say, always use the owner key to submit WindowPoSts (though I guess it would be simpler)

---

Alternatively we could put an optional field in at least some message params for worker signature, and allow submitting things from any address as long as that signature is present, but that may create some message replay problems if not designed correctly (quite bad), and will end up putting lots of signatures which can't be aggregated on chain (deal breaker bad)

aarshkshah1992:
Closes #942.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -168,21 +168,22 @@ type ChangeWorkerAddressParams struct {
 	NewControlAddrs []addr.Address
 }
 
+// ChangeWorkerAddress will ALWAYS overwrite the existing control addresses with the control addresses passed in the params.
+// If a nil addresses slice is passed, the control addresses will be cleared.
+// A worker change will be scheduled if the worker passed in the params is different from the existing worker.
 func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams) *adt.EmptyValue {
 	var effectiveEpoch abi.ChainEpoch
 
-	worker := addr.Undef
-	if params.NewWorker != addr.Undef {
-		worker = resolveWorkerAddress(rt, params.NewWorker)
-	}
+	worker := resolveWorkerAddress(rt, params.NewWorker)
 
-	controlAddrs := make([]addr.Address, 0, len(params.NewControlAddrs))
+	var controlAddrs []addr.Address
 	for _, ca := range params.NewControlAddrs {
 		resolved := resolveControlAddress(rt, ca)
 		controlAddrs = append(controlAddrs, resolved)
 	}
 
 	var st State
+	isWorkerChange := false
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
 
@@ -191,15 +192,14 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 
 		{
 			// save the new control addresses
-			if len(controlAddrs) != 0 {
-				info.ControlAddresses = controlAddrs
-			}
+			info.ControlAddresses = controlAddrs
 		}
 
 		{
 			// save worker addr key change request
 			// This may replace another pending key change.
-			if worker != addr.Undef {
+			if worker != info.Worker {
+				isWorkerChange = true
 				effectiveEpoch = rt.CurrEpoch() + WorkerKeyChangeDelay
 
 				info.PendingWorkerKey = &WorkerKeyChange{
@@ -215,7 +215,7 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 
 	// we only need to enroll the cron event for worker key change as we change the control
 	// addresses immediately
-	if worker != addr.Undef {
+	if isWorkerChange {
 		cronPayload := CronEventPayload{
 			EventType: CronEventWorkerKeyChange,
 		}
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -208,7 +208,9 @@ func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyVa
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
 
-		rt.ValidateImmediateCallerIs(info.Worker, info.Owner)
+		allowed := append(info.ControlAddresses, info.Owner, info.Worker)
+		rt.ValidateImmediateCallerIs(allowed...)
+
 		info.PeerId = params.NewID
 		err := st.SaveInfo(adt.AsStore(rt), info)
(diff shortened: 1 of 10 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -201,7 +208,9 @@ func (a Actor) ChangePeerID(rt Runtime, params *ChangePeerIDParams) *adt.EmptyVa
 	rt.State().Transaction(&st, func() {
 		info := getMinerInfo(rt, &st)
 
-		rt.ValidateImmediateCallerIs(info.Worker, info.Owner)
+		allowed := append(info.ControlAddresses, info.Owner, info.Worker)
+		rt.ValidateImmediateCallerIs(allowed...)
```
anorth, line 212: Please collapse these all to one-liners.

actors/builtin/miner/miner_actor.go
```diff
@@ -1998,6 +2023,22 @@ func resolveOwnerAddress(rt Runtime, raw addr.Address) addr.Address {
 	return resolved
 }
 
+// Resolves a control address to an ID address and verifies that it is address of an account actor.
+func resolveControlAddress(rt Runtime, raw addr.Address) addr.Address {
```
anorth, line 2027: Rename `resolveOwnerAddress` to `resolveControlAddress` and use that. I don't see any reason to exclude multisigs here; that could even be useful.

actors/builtin/miner/miner_state.go
```diff
@@ -85,6 +85,10 @@ type MinerInfo struct {
 	// The associated pubkey-type address is used to sign blocks and messages on behalf of this miner.
 	Worker addr.Address // Must be an ID-address.
 
+	// ControlAddresses are miner control addresses that are used to submit messages.
```
anorth, line 88: "Additional addresses that are permitted to sign messages to this actor"

anorth:
I just realised this is missing the ability to change the control addresses.

actors/builtin/miner/miner_actor.go
```diff
@@ -151,42 +157,71 @@ func (a Actor) ControlAddresses(rt Runtime, _ *adt.EmptyValue) *GetControlAddres
...
-		// Only the Owner is allowed to change the worker address.
-		// Allowing the worker to do this does not make sense because the owner would usually do it
-		// only if the worker keys are lost.
+		// Only the Owner is allowed to change the worker and control addresses.
 		rt.ValidateImmediateCallerIs(info.Owner)
 
-		effectiveEpoch = rt.CurrEpoch() + WorkerKeyChangeDelay
+		{
+			// save the new control addresses
+			if len(controlAddrs) != 0 {
```
anorth, line 194: The miner must be able to remove all these addresses. Remove this check, just replace the addresses with whatever was provided.

actors/builtin/miner/miner_actor.go
```diff
@@ -151,42 +157,71 @@ func (a Actor) ControlAddresses(rt Runtime, _ *adt.EmptyValue) *GetControlAddres
...
+	NewWorker       addr.Address
+	NewControlAddrs []addr.Address
 }
 
 func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams) *adt.EmptyValue {
 	var effectiveEpoch abi.ChainEpoch
-	worker := resolveWorkerAddress(rt, params.NewWorker)
+
+	worker := addr.Undef
+	if params.NewWorker != addr.Undef {
```
anorth, line 175: This can never happen: `addr.Undef` does not serialize so the miner will have to specify their existing worker address.

Always resolve this, and check at 202 against the current worker address.

actors/builtin/miner/miner_actor.go
```diff
@@ -151,42 +157,71 @@ func (a Actor) ControlAddresses(rt Runtime, _ *adt.EmptyValue) *GetControlAddres
...
-	NewWorker addr.Address
+	NewWorker       addr.Address
+	NewControlAddrs []addr.Address
 }
 
 func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams) *adt.EmptyValue {
 	var effectiveEpoch abi.ChainEpoch
-	worker := resolveWorkerAddress(rt, params.NewWorker)
+
+	worker := addr.Undef
```
anorth, line 174: Rename to `newWorker`.

actors/builtin/miner/miner_test.go
```diff
@@ -1764,16 +1829,18 @@ func TestChangeWorkerAddress(t *testing.T) {
 		return rt, actor
 	}
 
-	t.Run("successfully change a worker address", func(t *testing.T) {
+	t.Run("successfully change ONLY the worker address", func(t *testing.T) {
```
anorth, line 1832: After the change requested above, replace this with a test that changing to an empty array removes all control addresses.

aarshkshah1992:
@anorth  Have made the changes. Please take a look.

actors/builtin/miner/miner_actor.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_actor.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_test.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_actor.go
aarshkshah1992: This is done.

actors/builtin/miner/miner_test.go
```diff
@@ -1791,6 +1858,90 @@ func TestChangeWorkerAddress(t *testing.T) {
...
+
+		currentEpoch := abi.ChainEpoch(5)
+		rt.SetEpoch(currentEpoch)
+		effectiveEpoch := currentEpoch + miner.WorkerKeyChangeDelay
+		actor.changeWorkerAddress(rt, newWorker, effectiveEpoch, []addr.Address{c1, c2})
+
+		// set current epoch to effective epoch and ask to change the address
+		actor.cronWorkerAddrChange(rt, effectiveEpoch, newWorker)
+
+		// assert control addresses are unchanged
```
anorth, line 1912: Stale comment

PR #956, opened 2020-08-11
Feat/fee debt

### LINKED ISSUE #888: Record miner fee debt when penalties exceed unlocked balance

ZenGround0:
Closes #888 

1. Record miner fee debt when not enough balance to pay
2. Check and repay fee debt along with IP requirements checking
3. Test fee repayment on recovery

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -506,14 +506,15 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	var st State
 	var err error
 	newlyVested := big.Zero()
-	feeDebtPenalty := abi.NewTokenAmount(0)
+	feeToBurn := abi.NewTokenAmount(0)
 	rt.State().Transaction(&st, func() {
 		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
-		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
-
-		feeDebtPenalty = verifyPledgeRequirementsAndRepayDebts(rt, &st)
-		unlockedBalance = big.Sub(unlockedBalance, feeDebtPenalty)
+		// available balance already accounts for fee debt so it is correct to call
+		// this before VerifyPledgeRequirementsAndRepayDebts. We would have to
+		// subtract fee debt explicitly if we called this after.
+		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
+		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
 
 		info := getMinerInfo(rt, &st)
 
@@ -562,7 +563,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 			PreCommitDepositForPower(rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, sectorWeight),
 			depositMinimum,
 		)
-		if unlockedBalance.LessThan(depositReq) {
+		if availableBalance.LessThan(depositReq) {
 			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds for pre-commit deposit: %v", depositReq)
 		}
 
@@ -592,10 +593,7 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add pre-commit expiry to queue")
 	})
 
-	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
-		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
-		builtin.RequireSuccess(rt, code, "failed to burn debt")
-	}
+	burnFunds(rt, feeToBurn)
 
 	notifyPledgeChanged(rt, newlyVested.Neg())
 
@@ -625,9 +623,9 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 	var st State
 	var precommit *SectorPreCommitOnChainInfo
 	sectorNo := params.SectorNumber
-	feeDebtPenalty := abi.NewTokenAmount(0)
+	feeToBurn := abi.NewTokenAmount(0)
 	rt.State().Transaction(&st, func() {
-		feeDebtPenalty = verifyPledgeRequirementsAndRepayDebts(rt, &st)
+		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
 		var found bool
 		var err error
 		precommit, found, err = st.GetPrecommittedSector(store, sectorNo)
@@ -656,10 +654,7 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 		RegisteredSealProof: precommit.Info.SealProof,
 	})
 
-	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
-		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
-		builtin.RequireSuccess(rt, code, "failed to burn debt")
-	}
+	burnFunds(rt, feeToBurn)
 
 	_, code := rt.Send(
 		builtin.StoragePowerActorAddr,
@@ -1234,11 +1229,11 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 
 	store := adt.AsStore(rt)
 	var st State
-	feeDebtPenalty := abi.NewTokenAmount(0)
+	feeToBurn := abi.NewTokenAmount(0)
 	rt.State().Transaction(&st, func() {
 		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
 		// and repay fee debt now.
-		feeDebtPenalty = verifyPledgeRequirementsAndRepayDebts(rt, &st)
+		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
 
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
@@ -1271,10 +1266,7 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save deadlines")
 	})
 
-	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
-		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
-		builtin.RequireSuccess(rt, code, "failed to burn debt")
-	}
+	burnFunds(rt, feeToBurn)
 
 	// Power is not restored yet, but when the recovered sectors are successfully PoSted.
 	return nil
@@ -1470,8 +1462,8 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 	}
 	var info *MinerInfo
 	newlyVested := big.Zero()
-	feeDebtPenalty := big.Zero()
-	unlockedBalance := big.Zero()
+	feeToBurn := big.Zero()
+	availableBalance := big.Zero()
 	rt.State().Transaction(&st, func() {
 		var err error
 		info = getMinerInfo(rt, &st)
@@ -1493,27 +1485,26 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 		if err != nil {
 			rt.Abortf(exitcode.ErrIllegalState, "failed to vest fund: %v", err)
 		}
-		unlockedBalance = st.GetUnlockedBalance(rt.CurrentBalance())
+		// available balance already accounts for fee debt so it is correct to call
+		// this before VerifyPledgeRequirementsAndRepayDebts. We would have to
+		// subtract fee debt explicitly if we called this after.
+		availableBalance = st.GetAvailableBalance(rt.CurrentBalance())
 
 		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
 		// and repay fee debt now.
-		feeDebtPenalty = verifyPledgeRequirementsAndRepayDebts(rt, &st)
-		unlockedBalance = big.Sub(unlockedBalance, feeDebtPenalty)
+		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
 	})
 
-	amountWithdrawn := big.Min(unlockedBalance, params.AmountRequested)
+	amountWithdrawn := big.Min(availableBalance, params.AmountRequested)
 	Assert(amountWithdrawn.GreaterThanEqual(big.Zero()))
-	Assert(amountWithdrawn.LessThanEqual(unlockedBalance))
+	Assert(amountWithdrawn.LessThanEqual(availableBalance))
 
 	if amountWithdrawn.GreaterThan(abi.NewTokenAmount(0)) {
 		_, code := rt.Send(info.Owner, builtin.MethodSend, nil, amountWithdrawn)
 		builtin.RequireSuccess(rt, code, "failed to withdraw balance")
 	}
 
-	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
-		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
-		builtin.RequireSuccess(rt, code, "failed to burn debt")
-	}
+	burnFunds(rt, feeToBurn)
 
 	pledgeDelta := newlyVested.Neg()
 	notifyPledgeChanged(rt, pledgeDelta)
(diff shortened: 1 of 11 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1414,10 +1414,59 @@ func TestDeclareRecoveries(t *testing.T) {
...
 		dlIdx, pIdx, err := st.FindSector(rt.AdtStore(), oneSector[0].SectorNumber)
 		require.NoError(t, err)
-		rt.ExpectAbortContainsMessage(exitcode.ErrInsufficientFunds, "does not cover pledge requirements", func () {
-			actor.declareRecoveries(rt, dlIdx, pIdx, bf(uint64(oneSector[0].SectorNumber)))
+		rt.ExpectAbortContainsMessage(exitcode.ErrInsufficientFunds, "does not cover pledge requirements", func() {
+			actor.declareRecoveries(rt, dlIdx, pIdx, bf(uint64(oneSector[0].SectorNumber)), big.Zero())
 		})
 	})
+
+	t.Run("recovery must pay back fee debt", func(t *testing.T) {
```
acruikshank, line 1422: This is great. It would be nice if the new functionality in `PreCommitSector`, `ProveCommitSector`, `ConfirmSectorProofsValid` and `WithdrawBalance` was also covered.

actors/builtin/miner/miner_actor.go
```diff
@@ -1126,8 +1140,11 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 
 	store := adt.AsStore(rt)
 	var st State
+	debtToBurn := abi.NewTokenAmount(0)
 	rt.State().Transaction(&st, func() {
-		verifyPledgeMeetsInitialRequirements(rt, &st)
+		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
+		// and repay fee debt now.
+		debtToBurn = verifyPledgeRequirementsAndRepayDebts(rt, &st)
```
acruikshank, line 1147: I'm a little surprised we don't vest before this. Is it because we know vesting happened in prove commit within the same epoch?

actors/builtin/miner/miner_actor.go
```diff
@@ -1375,18 +1399,27 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
...
 			rt.Abortf(exitcode.ErrIllegalState, "failed to vest fund: %v", err)
 		}
+		unlockedBalance = st.GetUnlockedBalance(rt.CurrentBalance())
 
-		// Verify InitialPledgeRequirement does not exceed unlocked funds
-		verifyPledgeMeetsInitialRequirements(rt, &st)
+		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
+		// and repay fee debt now.
+		debtToBurn = verifyPledgeRequirementsAndRepayDebts(rt, &st)
+		unlockedBalance = big.Sub(unlockedBalance, debtToBurn)
```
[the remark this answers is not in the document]
ZenGround0: > It should return it.

I landed with this while fostering simplicity with the fund accounting because it is getting stressfully complicated and adding another return value to the interface gives me another calculation I need to rely on happening correctly in a subcall that I would rather see here directly.

actors/builtin/miner/miner_actor.go
```diff
@@ -446,9 +446,16 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 
 	store := adt.AsStore(rt)
 	var st State
+	var err error
 	newlyVested := big.Zero()
+	debtToBurn := abi.NewTokenAmount(0)
```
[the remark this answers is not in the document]
ZenGround0: `feeDebtPenalty`?

actors/builtin/miner/miner_state_test.go
```diff
@@ -733,6 +733,26 @@ func TestSectorNumberAllocation(t *testing.T) {
 	})
 }
 
+func TestPenalizationDebt(t *testing.T) {
```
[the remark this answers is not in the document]
ZenGround0: now tracked in #966

actors/builtin/miner/miner_state_test.go
```diff
@@ -733,6 +733,26 @@ func TestSectorNumberAllocation(t *testing.T) {
...
+	harness := constructStateHarness(t, abi.ChainEpoch(0))
+
+	currentBalance := abi.NewTokenAmount(300)
+	fee := abi.NewTokenAmount(1000)
+	_, _, err := harness.s.PenalizeFundsInPriorityOrder(harness.store, abi.ChainEpoch(0), fee, currentBalance)
```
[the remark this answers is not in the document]
ZenGround0: Now tracked in #966

actors/builtin/miner/miner_test.go
ZenGround0: Yes, I will add a test for `PreCommitSector` and `WithdrawBalance`.  `ProveCommitSector` verification is going away (#961) so I'll drop that one.  I don't think this PR changes `ConfirmSectorProofsValid` but let me know if I'm missing a hidden change trickling from precommit.

actors/builtin/miner/miner_actor.go
ZenGround0: > Is it because we know vesting happened in prove commit within the same epoch?

Did you think this was `ConfirmSectorProofsValid`?  It's `DeclareFaultsRecovered` so there is no relationship with a prove commit.  

We could vest before but in the past we haven't vested during these verifications.  See the comment above `verifyPledgeRequirementsAndPayDebts` that I lifted and modified from @anorth's original comment in the ProveCommit invocation of verification.  

From that comment: "Computing vesting here would be almost always redundant since vesting is quantized to ~daily units".  It doesn't seem like we are consistent about that everywhere though and maybe vesting has become less expensive?  @anorth does the infrequent vesting still make sense?

acruikshank: Yeah, I was confused, and that's why I thought ConfirmSectorProofsValid needed to be tested. It does seem like we're vesting inconsistently.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
acruikshank: Well.. you're already relying on that calculation happening correctly in that subcall You just don't see it.

actors/builtin/miner/miner_actor.go
```diff
@@ -2021,12 +2054,27 @@ func requestCurrentTotalPower(rt Runtime) *power.CurrentTotalPowerReturn {
...
+			unlockedBalance, st.FeeDebt,
+		)
+	}
+	debtToRepay := st.FeeDebt
+	st.FeeDebt = big.Zero()
+	currBalance = big.Sub(currBalance, debtToRepay)
+	unlockedBalance = big.Sub(unlockedBalance, debtToRepay)
+	AssertMsg(currBalance.GreaterThanEqual(big.Zero()), "debt repayment can not take balance below zero")
+
 	if !st.MeetsInitialPledgeCondition(rt.CurrentBalance()) {
```
anorth, line 2072: Blocking: Using `rt.CurrentBalance()` after the calculations above looks like a bug.

This is what I'm getting at with the requests for tests. We need to test _this_ code directly, not as part of complicated whole actor flow tests.

This `verifyPledgeRequirementsAndRepayDebts` probably deserves to be elevated to something we can test directly, maybe broken out to a `fees.go` file (or into `monies.go`) along with the things it calls.

As a structural note, we're trying as hard as we can not to manipulate the state objects from actor code, except through methods, so they can enforce their own invariants. The setting of `st.FeeDebt` here goes against that a little. I would like to see structure that would let us replace this method with something like:
```
toBurn, err := st.RepayDebt(rt.CurrentBalance())
// if err ...
Assert(st.FeeDebt == big.Zero)
if (rt.CurrentBalance() - toBurn).LessThan(st.InitialPledgeRequirement) {
...
}

```

actors/builtin/miner/miner_actor.go
```diff
@@ -485,9 +485,16 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
+	var err error
 	newlyVested := big.Zero()
+	feeDebtPenalty := abi.NewTokenAmount(0)
 	rt.State().Transaction(&st, func() {
-		verifyPledgeMeetsInitialRequirements(rt, &st)
+		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+
+		feeDebtPenalty = verifyPledgeRequirementsAndRepayDebts(rt, &st)
```
anorth, line 496: It took me quite some reading to understand `feeDebtPenalty`. A name like `feeToBurn` would help.

actors/builtin/miner/miner_actor.go
```diff
@@ -485,9 +485,16 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
...
 	store := adt.AsStore(rt)
 	var st State
+	var err error
 	newlyVested := big.Zero()
+	feeDebtPenalty := abi.NewTokenAmount(0)
 	rt.State().Transaction(&st, func() {
-		verifyPledgeMeetsInitialRequirements(rt, &st)
+		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
```
anorth, line 494: Nit: Move this next to line 497, after the fee debt is set to zero. I think it would be clearer as

```
feeDebtPenalty := verifyPledgeRequirementsAndRepayDebts
unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance() - feeDebtPenalty)
```

Blocking: It seems wrong to me that this is using unlocked, rather than available balance. Balance already covering IP cannot be used here. If it's right, please explain in comments.

actors/builtin/miner/miner_actor.go
```diff
@@ -572,6 +575,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add pre-commit expiry to queue")
 	})
 
+	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
+		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
+		builtin.RequireSuccess(rt, code, "failed to burn debt")
+	}
```
anorth, line 581: Call the existing `burnFunds()`

actors/builtin/miner/miner_actor.go
```diff
@@ -622,6 +631,11 @@ func (a Actor) ProveCommitSector(rt Runtime, params *ProveCommitSectorParams) *a
 		RegisteredSealProof: precommit.Info.SealProof,
 	})
 
+	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
+		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
+		builtin.RequireSuccess(rt, code, "failed to burn debt")
+	}
```
anorth, line 637: `burnFunds()`

actors/builtin/miner/miner_actor.go
```diff
@@ -1202,6 +1219,11 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save deadlines")
 	})
 
+	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
+		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
+		builtin.RequireSuccess(rt, code, "failed to burn debt")
+	}
```
anorth, line 1225: `burnFunds()`

actors/builtin/miner/miner_actor.go
```diff
@@ -1417,18 +1441,27 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
...
 			rt.Abortf(exitcode.ErrIllegalState, "failed to vest fund: %v", err)
 		}
+		unlockedBalance = st.GetUnlockedBalance(rt.CurrentBalance())
 
-		// Verify InitialPledgeRequirement does not exceed unlocked funds
-		verifyPledgeMeetsInitialRequirements(rt, &st)
+		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
+		// and repay fee debt now.
+		feeDebtPenalty = verifyPledgeRequirementsAndRepayDebts(rt, &st)
+		unlockedBalance = big.Sub(unlockedBalance, feeDebtPenalty)
```
anorth, line 1449: Similar comment to above: repay fees, then calculate available balance.

And again, unlocked rather than available here seems wrong.

actors/builtin/miner/miner_actor.go
```diff
@@ -1417,18 +1441,27 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
...
-	builtin.RequireSuccess(rt, code, "failed to withdraw balance")
+	if amountWithdrawn.GreaterThan(abi.NewTokenAmount(0)) {
+		_, code := rt.Send(info.Owner, builtin.MethodSend, nil, amountWithdrawn)
+		builtin.RequireSuccess(rt, code, "failed to withdraw balance")
+	}
+
+	if feeDebtPenalty.GreaterThan(abi.NewTokenAmount(0)) {
+		_, code := rt.Send(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, feeDebtPenalty)
+		builtin.RequireSuccess(rt, code, "failed to burn debt")
+	}
```
anorth, line 1464: `burnFunds()`

actors/builtin/miner/miner_actor.go
anorth: Infrequent still makes sense. We now do it every deadline, so any additional calls are likely redundant. It's worth still doing on withdrawal, though.

actors/builtin/miner/miner_actor.go
```diff
@@ -2021,12 +2054,27 @@ func requestCurrentTotalPower(rt Runtime) *power.CurrentTotalPowerReturn {
 // may be slightly lower than the true amount.
 // Computing vesting here would be almost always redundant since vesting is quantized to ~daily units.
 // Vesting will be at most one proving period old if computed in the cron callback.
-func verifyPledgeMeetsInitialRequirements(rt Runtime, st *State) {
+func verifyPledgeRequirementsAndRepayDebts(rt Runtime, st *State) abi.TokenAmount {
```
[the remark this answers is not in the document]
ZenGround0: > Note that there's partial debt repayment

That is not my understanding of what's happeing

>  the function name and call sites imply that if the whole debt can't be repaid, the call aborts so there's no point recording partial repayment.

This is my understanding of what's happening

actors/builtin/miner/miner_state.go
```diff
@@ -917,8 +926,8 @@ func (st *State) AssertBalanceInvariants(balance abi.TokenAmount) {
 }
 
 func (st *State) MeetsInitialPledgeCondition(balance abi.TokenAmount) bool {
-	available := st.GetUnlockedBalance(balance)
-	return available.GreaterThanEqual(st.InitialPledgeRequirement)
+	unlockedBalance := st.GetUnlockedBalance(balance)
+	return unlockedBalance.GreaterThanEqual(st.InitialPledgeRequirement)
```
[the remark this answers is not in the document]
ZenGround0: > Note: This is policy, rather than pure state, and so perhaps belongs in actor code or a free function rather than a state method. 

I can get on board with that

> I don't think we want to say a miner is meeting their pledge condition if fee debt is nonzero.

I mildly disagree.  As `verifyPledgeRequirementsXXX` expands in scope to include constraints on recent consensus faults and potentially consensus fee related balance I think it helps clarity to unravel eligibility requirements into a set with no blurry overlap.  Having a condition for "unlocked balance fulfills pledge requirement" and another for "fee debt is zero" and then combining the two where needed feels better.

actors/builtin/miner/miner_state_test.go
```diff
@@ -733,6 +733,26 @@ func TestSectorNumberAllocation(t *testing.T) {
 	})
 }
 
+func TestPenalizationFundsInPriorityOrder(t *testing.T) {
```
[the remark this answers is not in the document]
ZenGround0: #966 tracks this (and I will be enthusiastic about doing this work once latest feature changes are up)

actors/builtin/miner/miner_state.go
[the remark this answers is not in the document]
ZenGround0: Resolution, IP reqs only met if fee debt is zero.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
anorth: Oops, I meant to write "no partial debt repayment"

actors/builtin/miner/miner_state.go
```diff
@@ -761,11 +763,31 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
 }
 
+// Repays the full miner actor fee debt.  Returns the amount that must be
+// repaid and an error if there are not sufficient funds to cover repayment.
+// Miner state repays from unlocked funds, potentially violating IP requirements
+// and bringing actor into IP debt.  FeeDebt should be  zero after calling.
+func (st *State) RepayDebt(currBalance abi.TokenAmount) (abi.TokenAmount, error) {
+	unlockedBalance := st.GetUnlockedBalance(currBalance)
+	if !st.CanRepayFeeDebt(unlockedBalance) {
+		return big.Zero(), xerrors.Errorf("unlocked balance can not repay fee debt (%v < %v)", unlockedBalance, st.FeeDebt)
```
anorth, line 784: I think this is an appropriate place for `exitcode.ErrInsufficientFunds.Wrapf("unlock....")`

actors/builtin/miner/miner_state.go
```diff
@@ -761,11 +763,31 @@ func (st *State) PenalizeFundsInPriorityOrder(store adt.Store, currEpoch abi.Cha
...
+	remaining = big.Sub(remaining, fromBalance)
+	if remaining.GreaterThan(big.Zero()) {
+		st.FeeDebt = big.Add(st.FeeDebt, remaining)
+	}
+
 	return fromVesting, fromBalance, nil
 }
 
+// Repays the full miner actor fee debt.  Returns the amount that must be
+// repaid and an error if there are not sufficient funds to cover repayment.
```
anorth, line 778:
```suggestion
// burnt and an error if there are not sufficient funds to cover repayment.
```

actors/builtin/miner/monies.go
```diff
@@ -111,3 +113,25 @@ func InitialPledgeForPower(qaPower abi.StoragePower, baselinePower abi.StoragePo
 
 	return big.Add(ipBase, additionalIP)
 }
+
+// Verifies that the total unlocked balance covers the amount needed to both cover
+// the pledge requirement and repay all fee debt.  If not aborts.
```
anorth, line 118: Maybe "Repays all fee debt and then verifies that ..... Returns an amount that must be burnt by the actor"

actors/builtin/miner/miner_test.go
```diff
@@ -1479,10 +1513,59 @@ func TestDeclareRecoveries(t *testing.T) {
...
+		ff := miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, ongoingPwr.QA)
+		advanceDeadline(rt, actor, &cronConfig{
+			ongoingFaultsPenalty: big.Zero(), // fee is instead added to debt
+		})
+
+		st = getState(rt)
+		assert.Equal(t, ff, st.FeeDebt)
+
+		// Recovery fails when it can't pay back fee debt
+		rt.ExpectAbortContainsMessage(exitcode.ErrInsufficientFunds, "unlocked balance can not repay fee debt", func() {
```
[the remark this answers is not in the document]
ZenGround0: The contains message differentiates it for us.

Issue #887, opened 2020-08-04
Add static state methods for miner election eligibility


anorth:
A miner's eligibility for elections is conditional on some properties of state. The power state has `MinerNominalPowerMeetsConsensusMinimum` that is never invoked internally, but from the node, encoding the property of a miner meeting the minimum power requirement.

The miner state has `MeetsInitialPledgeCondition`, but we added that recently and Lotus probably aren't using it yet.

Add a single `MinerEligibleForElection` function that takes parameters as necessary (power state, miner state, current balance) to encapsulate all the relevant state predicates, including if we change them in the future.

FYI @magik6k @arajasek, or let us know if this would not be helpful.

anorth:
Check also fee debt (#888) in the predicates.

anorth:
Check minimum IP requirement: #933

anorth:
Check not recently reported for consensus fault: #934

ZenGround0:
After @anorth and @nicola have synced this issue is now expanding in scope to include running consensus eligibility conditions (5x reward of funds available + no faults in finality epochs) inside of the existing pledge / fee debt predicate.  

The `verifyXXX` function should expand to take in all of the necessary parameters and return an error so that the endpoint we give to lotus can just gather inputs from states and pass them to this function and pass along the error.

anorth:
Please don't fold the enforcement of restrictions in miner methods with the checks for election eligibility. They are similar, but likely slightly different computations for different contexts. They will no doubt share some pieces, but we should be able to update one independently of the other.

anorth:
Re-opening because #988 did not include the check on minimum consensus power, `MinerNominalPowerMeetsConsensusMinimum`.

PR #958, opened 2020-08-11
Consensus Fault Update

### LINKED ISSUE #934: Consensus fault renders miner ineligible for election until finality
The penalty for consensus faults is to be reduced from complete miner termination down to a fee and ineligibility for elections until finality.

The consensus fault penalty (CFP) is to be `5.0` times the current per-win-count total block reward (irrespective of miner's power share), as specified in #933.

Add a `ConsensusFaultReported abi.ChainEpoch` to miner state, initially `-1`. Upon a valid consensus fault report, instead of terminating the miner:
- set `ConsensusFaultReported` to the current epoch
- charge the consensus fault penalty

The miner must have the CFP in unlocked balance because that is a condition of election eligibility.

ZenGround0:
Closes #934 

1. Consensus fault no longer terminates a miner
2. Introduce consensus fault fee
3. Track last consensus fault epoch in miner state


actors/builtin/miner/miner_actor.go
```diff
@@ -1363,24 +1363,33 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
...
-	slasherReward := RewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
-	_, code := rt.Send(reporter, builtin.MethodSend, nil, slasherReward)
-	builtin.RequireSuccess(rt, code, "failed to reward reporter")
-
 	var st State
 	rt.State().Readonly(&st)
+	unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+	slasherReward := RewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
+	if slasherReward.GreaterThan(unlockedBalance) {
+		rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient unlocked balance to pay reward")
```
acruikshank, line 1371: This is surprising. It seems to reward the miner (by reverting the slash) and penalize the slasher (by not paying them) for the miner being under-collateralized. Is this right? Can you add a comment explaining why?

actors/builtin/miner/miner_actor.go
```diff
@@ -1931,17 +1926,6 @@ func getVerifyInfo(rt Runtime, params *SealVerifyStuff) *abi.SealVerifyInfo {
 	}
 }
 
-// Closes down this miner by erasing its power, terminating all its deals and burning its funds
-func terminateMiner(rt Runtime) {
```
acruikshank, line 1935: 🎉

actors/builtin/power/power_actor.go
```diff
@@ -237,40 +234,6 @@ func (a Actor) UpdatePledgeTotal(rt Runtime, pledgeDelta *abi.TokenAmount) *adt.
 	return nil
 }
 
-func (a Actor) OnConsensusFault(rt Runtime, pledgeAmount *abi.TokenAmount) *adt.EmptyValue {
```
acruikshank, line 240: 🎉

actors/builtin/miner/miner_actor.go
ZenGround0: Glad you pointed it out.  Before the change this was deducting a fixed portion of the actor balance without any checks on whether that amount was less than the unlocked funds so it was possible to violate the balance invariants and fail in an assertion (which I stumbled upon while testing).  This state is just an explicit version of what we had before so now the uncanny behavior is visible and we will be motivated to design it away.

The simplest alternative is to take the reward from the miner actor in the normal penalization order.  I didn't want to do that in this PR because there is something potentially bad about allowing the reward to come from locked funds -- if miners slash themselves they could unvest  their rewards early.  This might warrant coming up with a penalization order that does not draw from locked funds.

I am planning to file an issue about this to gather input from red team and crypto econ.

anorth, line 1371: Blocking: I don't think we want to prevent the reporting of a consensus fault, even if no reward can be paid.

actors/builtin/miner/miner_actor.go
```diff
@@ -1363,24 +1363,33 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
...
-
 	var st State
 	rt.State().Readonly(&st)
+	unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+	slasherReward := RewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
+	if slasherReward.GreaterThan(unlockedBalance) {
+		rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient unlocked balance to pay reward")
+	}
+	_, code := rt.Send(reporter, builtin.MethodSend, nil, slasherReward)
+	builtin.RequireSuccess(rt, code, "failed to send reward")
```
anorth, line 1374: While this should never fail, if it does, we still want to record the fault. Log instead.

actors/builtin/miner/miner_state.go
```diff
@@ -106,6 +106,8 @@ type MinerInfo struct {
 	// The number of sectors in each Window PoSt partition (proof).
 	// This is computed from the proof type and represented here redundantly.
 	WindowPoStPartitionSectors uint64
+
+	ConsensusFaultReported abi.ChainEpoch
```
anorth, line 110: Please document this.

Please also add a `policy.go` parameter of `ConsensusFaultIneligibilityDuration = Finality` in preparation for #887.

Thinking about this yesterday, it might be simpler to set this epoch as the first epoch _after_ the penalty period, when they become eligible again (`ConsensusFaultElapsed`). That way, if we init to `-1`, there's no other tests or arithmetic necessary to determine whether a miner is currently ok. But your call.

actors/builtin/miner/miner_test.go
```diff
@@ -2108,44 +2108,56 @@ func TestReportConsensusFault(t *testing.T) {
...
-		allDeals := []abi.DealID{}
-		for _, ids := range dealIDs {
-			allDeals = append(allDeals, ids...)
-		}
-		actor.reportConsensusFault(rt, addr.TestAddress, params, allDeals)
+		actor.reportConsensusFault(rt, addr.TestAddress, params)
 	})
 
-	t.Run("miner batches termination requests when number of deals exceeds limit", func(t *testing.T) {
+	t.Run("Fault reporting fails when balance can't cover reward", func(t *testing.T) {
```
anorth, line 2114: I don't think we want this.

actors/builtin/miner/monies.go
```diff
@@ -33,6 +33,9 @@ var UndeclaredFaultProjectionPeriod = abi.ChainEpoch(5) * builtin.EpochsInDay
 // Maximum number of days of BR a terminated sector can be penalized
 const TerminationLifetimeCap = abi.ChainEpoch(70)
 
+// Number of whole block rewards covered by consensus fault penalty
```
anorth, line 36: whole per-winner rewards

actors/builtin/miner/miner_actor.go
```diff
@@ -1363,24 +1363,33 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
...
 
 	// Reward reporter with a share of the miner's current balance.
-	slasherReward := RewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
-	_, code := rt.Send(reporter, builtin.MethodSend, nil, slasherReward)
-	builtin.RequireSuccess(rt, code, "failed to reward reporter")
-
 	var st State
 	rt.State().Readonly(&st)
+	unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+	slasherReward := RewardForConsensusSlashReport(faultAge, rt.CurrentBalance())
```
anorth, line 1369: I realise this was totally unaddressed by the issue, but I was expecting the amount of which the reporter earns a share to be the `ConsensusFaultPenalty`, not a function of the miner's size. So we'd compute the penalty, then split it into reward share and burn share. `ConsensusFaultPenalty` is the maximum total lost by the miner.

@zixuanzh @nicola could you please confirm this interpretation?

If I'm wrong, and we do want to be taking the reward from the miner's balance over and above the consensus fault penalty, then we should only consider that part of the balance that is available for paying fees: the unlocked balance plus the vesting funds. I think it would follow the same logic as `PenalizeFundsInPriorityOrder`. The fact of a miner having a large pre-commit deposit lockup should not enter into it.

nicola: Correct, only the consensus fault penalty is lost by the miner.

actors/builtin/miner/miner_actor.go
ZenGround0: This will be resolved in this PR now that I understand the path forward

actors/builtin/miner/miner_actor.go
ZenGround0: One minor remaining problem is that now we are trying to pay an actual account for fees that could possibly only be covered by debt.  Fee debt is currently only repayable to the burnt funds actor.  Extending it to be repayable to any slashing account does not seem worth the state tracking cost.

Proposal: reward reporter with between `0` and `slasherReward` FIL (slasherReward when the penalty does not include any fee debt, 0 when the penalty is totally covered by fee debt).  The reporter never gets paid back out of fee debt.

actors/builtin/miner/miner_state.go
ZenGround0: Yeah I like the first unpenalized epoch idea.

actors/builtin/miner/miner_actor.go
```diff
@@ -1366,25 +1366,35 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
...
-	builtin.RequireSuccess(rt, code, "failed to notify power actor on consensus fault")
+	rewardStats := requestCurrentEpochBlockReward(rt)
+	faultPenalty := ConsensusFaultPenalty(rewardStats.ThisEpochReward)
+	slasherReward := RewardForConsensusSlashReport(faultAge, faultPenalty)
+	pledgeDelta := big.Zero()
+	rt.State().Transaction(&st, func() {
+		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
+		penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(adt.AsStore(rt), rt.CurrEpoch(), faultPenalty, unlockedBalance)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock unvested funds")
+		faultPenalty = big.Add(penaltyFromVesting, penaltyFromBalance)
```
anorth, line 1380: Either signpost this with a comment about how it's updating to be the amount actually payable, which might be less, or use a new variable name.

actors/builtin/miner/miner_actor.go
```diff
@@ -1366,25 +1366,35 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
...
+		pledgeDelta = big.Add(pledgeDelta, penaltyFromVesting.Neg())
 
-	// close deals and burn funds
-	terminateMiner(rt)
+		// clamp slasherReward at faultPenalty
+		slasherReward = big.Min(faultPenalty, slasherReward)
+		// reduce faultPenalty by slasherReward
+		faultPenalty = big.Sub(faultPenalty, slasherReward)
```
anorth, line 1386: Use different names, maybe `rewardAmount, burnAmount`.

actors/builtin/miner/miner_state.go
anorth: Thanks for the comment, please rename it too

actors/builtin/miner/miner_actor.go
ZenGround0: FYI @nicola ^

PR #959, opened 2020-08-11
use an AMT for lane tracking

whyrusleeping:
Still a bit WIP, but should be most of the way there.


hannahhoward:
I actually took a look at the referenced PR and what the "mutator pattern" is was not entirely clear to me.

PR #964, opened 2020-08-12
Remove duplicate miner constructor params and serialization

austinabell:
Feel free to close if this is necessary for some reason. I made the change to make sure that it was holding nothing up, but the extra serialization and deserialization and copy didn't make any sense.


austinabell:
nvm I guess it's needed now because of the control addresses added just now. Ready to close this issue, but is this for sure correct? Because it doesn't seem like control addresses are ever added in a functional call outside of tests (maybe I'm missing something)

anorth:
I would rather keep the CreateMinerParams and the miner constructor params as distinct objects anyway, even if they share all the same fields. Future changes may result in additions to CreateMinerParams.

PR #967, opened 2020-08-12
Remove duplicate slice in batch proof verif

austinabell:
Isn't this logically the same? Just want to avoid the seemingly unnecessary slice allocation


austinabell:
ah, looks like your CI doesn't like iterating over a go map (rationale would be very interesting to me still)

anorth:
You would have to match Go map iteration order in other programming languages (and future versions of Go) to match the behaviour (such as gas consumed).

austinabell:
> You would have to match Go map iteration order in other programming languages (and future versions of Go) to match the behaviour (such as gas consumed).

ah, of course, thanks !

PR #969, opened 2020-08-13
Missing tests for Exported methods

### LINKED ISSUE #596: Cover each exported actor method with at least one test

aarshkshah1992:
Closes #596 .

@anorth I think the `TestCompactPartitions` success case can be tested better but I do not understand how it works. Please can you give some ideas around how to test that one better ?


actors/builtin/miner/miner_test.go
```diff
@@ -2778,6 +2892,30 @@ func (h *actorHarness) changeWorkerAddress(rt *mock.Runtime, newWorker addr.Addr
...
+func (h *actorHarness) checkSectorProven(rt *mock.Runtime, sectorNum abi.SectorNumber) {
+	param := &miner.CheckSectorProvenParams{sectorNum}
+
+	rt.ExpectValidateCallerAny()
+
+	rt.Call(h.a.CheckSectorProven, param)
+	rt.Verify()
+}
+
+func (h *actorHarness) changeMultiAddrs(rt *mock.Runtime, newAddrs []abi.Multiaddrs) {
```
Stebalien, line 2904: I think I conflicted with you here a bit. I've added some tests for multiaddrs/peer ids as part of #920.

actors/builtin/miner/miner_test.go
```diff
@@ -2052,15 +2052,129 @@ func TestWithdrawBalance(t *testing.T) {
...
+		rt.ExpectAbort(exitcode.ErrForbidden, func() {
+			actor.compactPartitions(rt, 1, bitfield.New())
+		})
+	})
+
+	t.Run("fails if partition count is above limit", func(t *testing.T) {
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		// partition limit is 4 for the default construction
```
Stebalien, line 2117: Where is this set? It should be 200.

Stebalien: Hm. I see. Because we might need to load a bunch of sector infos. That's kind of annoying. We could actually just enforce this as we go along.

Stebalien: Actually, we can't. I thought we could count live sectors only, but we need to address dead sectors as well. The current logic is correct, if a bit unfortunate.

actors/builtin/miner/miner_test.go
```diff
@@ -2052,15 +2052,129 @@ func TestWithdrawBalance(t *testing.T) {
...
 	t.Run("successfully change peer id", func(t *testing.T) {
+		rt := builder.Build(t)
+		actor.constructAndVerify(rt)
+
+		newPID := tutil.MakePID("test-change-peer-id")
+		actor.changePeerID(rt, newPID)
+	})
+}
+
+func TestCompactPartitions(t *testing.T) {
```
Stebalien, line 2068: CompactPartitions garbage collects terminated sectors. Specifically, it:

1. Deletes the specified partitions.
2. Sorts the sectors into a live set and a dead set (terminated).
3. _Deletes_ the sector infos for the dead set.
4. Re-adds the live set back to the deadline they came from.

Tests here should test:

* We cannot compact partitions with faults.
* We cannot compact partitions with pending early terminations (e.g., a fault expired and the fee hasn't yet been payed). This may be a bit difficult to test.
* If we compact a partition with 5 live sectors and 5 dead, the 5 dead should be deleted (from the state's Sectors array) and the 5 live should be still be present (check with `FindSector`) in the deadline at which they started.

actors/builtin/miner/miner_test.go
```diff
@@ -2052,15 +2052,129 @@ func TestWithdrawBalance(t *testing.T) {
...
+	})
+
+	t.Run("fails is sector is not found", func(t *testing.T) {
+		actor := newHarness(t, periodOffset)
+		rt := builderForHarness(actor).
+			WithBalance(bigBalance, big.Zero()).
+			Build(t)
+		actor.constructAndVerify(rt)
+
+		rt.ExpectAbort(exitcode.ErrNotFound, func() {
```
Stebalien, line 2148: Forwarding advice from @acruikshank, `ExpectAbortContainsMessage` is a great way to make sure you're getting the _right_ error.

actors/builtin/miner/miner_test.go
```diff
@@ -2052,15 +2052,178 @@ func TestWithdrawBalance(t *testing.T) {
...
+		partitions := bitfield.NewFromSet([]uint64{partId})
+		actor.compactPartitions(rt, deadlineId, partitions)
+
+		st := getState(rt)
+		assertSectorExists(rt.AdtStore(), st, sector2, partId, deadlineId)
+		assertSectorExists(rt.AdtStore(), st, sector3, partId, deadlineId)
+		assertSectorExists(rt.AdtStore(), st, sector4, partId, deadlineId)
+
+		// TODO Why is this working ?
+		assertSectorExists(rt.AdtStore(), st, sector1, partId, deadlineId)
```
aarshkshah1992, line 2120: @Stebalien I am not sure what's going wrong here.

This sector was terminated above and thus should not exist after compacting the partition. Please can you review the test ?

actors/builtin/miner/miner_test.go
aarshkshah1992: Will use this going ahead.

actors/builtin/miner/miner_test.go
aarshkshah1992: @Stebalien Never mind. I'll rebase after your PR goes in.

actors/builtin/miner/miner_test.go
aarshkshah1992: @Stebalien I've added tests for **1** AND **3** but am running into this problem:

https://github.com/filecoin-project/specs-actors/pull/969#discussion_r471251481.

aarshkshah1992:
@Stebalien Have addressed your changes. Please take a look.

actors/builtin/miner/miner_test.go
Stebalien: Because the code wasn't saving the state... I've pushed a fix.

Issue #973, opened 2020-08-13
Check deferred method "params" when proposed


Stebalien:
When a miner proposes a method invocation, we can, and probably should, validate that:

1. The target actor & method exist.
2. The passed parameters are valid for the target actor & method.

This isn't necessary, but it would help guard against potential issues with invalid parameters.

anorth:
What do you mean by "propose" here? Are you talking about actors, or the VM?

anorth:
I think you're talking about when block producers are putting messages in blocks. That's outside the scope of the actors project, but, in brief:
- the target actor doesn't need to exist if it will be an account actor
- the VM checks that the method exists
- the VM decodes into the target method's structure (*)

This last point is a fact of convenience, not design. Logically, an actor method should receive raw bytes and inspect or decode them as it wishes. We happen to (a) always use CBOR and (b) lean on the VM to decode this for us, but it's not a design goal. "Are valid for the target method" is 100% subjective for the method in question.

anorth:
Basically the VM what is necessary, in fact does more than is necessary, and I don't think we should change anything.

anorth:
@Stebalien is there still an actors-side change that you'd propose here?

Stebalien:
No, thanks, this has been resolved.

PR #974, opened 2020-08-14
Validate CBOR

### LINKED ISSUE #972: CBORBytes does not validate
Unlike `cbg.Deferred`, `runtime.CBORBytes` does not validate the input CBOR. It looks like we may be fine at the moment, but anything like the following would be really dangerous:

```go
type Params struct {
    ...
    Params runtime.CBORBytes
    ...
}
...
p = Params{..., Params: runtime.CBORBytes(userDefinedBytes), ...}
```

I'd feel more comfortable if we used deferred (always checks on deserialize), but this would be a pretty large refactor.

Stebalien:
~This change validates CBOR whenever accepted from the user as an arbitrary byte array.

Note:

1. For the multisig actor, we could _technically_ skip validating the CBOR on approval because it has already been validated on propose. But this shouldn't be that expensive.
2. We don't actually need to validate the CBOR in the cron actor. I'm happy to remove it, I just wanted to start with something paranoid and walk back from there.~

This change validates CBOR when "sending" it to other actor methods. Technically, actor message params are arbitrary bytes. Ideally, we'd just pass the raw bytes in.

However, for convenience, the runtime accepts any object implementing `CBORMarshaler`. Unfortunately, passing raw bytes as a `CBORMarshaler` without _checking_ them could lead to bugs down the road if we decide to treat this `CBORMarshaler` object as a valid CBOR object, so we validate them.

fixes #972, #973

### DIFF
--- a/actors/builtin/init/init_actor.go
+++ b/actors/builtin/init/init_actor.go
@@ -57,6 +57,13 @@ func (a Actor) Exec(rt runtime.Runtime, params *ExecParams) *ExecReturn {
 		rt.Abortf(exitcode.ErrForbidden, "caller type %v cannot exec actor type %v", callerCodeCID, params.CodeCID)
 	}
 
+	// Technically, actors take _bytes_ as their method params. However,
+	// given that we pass them around as a cbor-able object, we need to
+	// ensure it correctly encodes to CBOR.
+	// See: https://github.com/filecoin-project/specs-actors/issues/972
+	ctorParams, err := runtime.CheckCBOR(params.ConstructorParams)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "params must be valid CBOR")
+
 	// Compute a re-org-stable address.
 	// This address exists for use by messages coming from outside the system, in order to
 	// stably address the newly created actor even if a chain re-org causes it to end up with
@@ -77,7 +84,7 @@ func (a Actor) Exec(rt runtime.Runtime, params *ExecParams) *ExecReturn {
 	rt.CreateActor(params.CodeCID, idAddr)
 
 	// Invoke constructor.
-	_, code := rt.Send(idAddr, builtin.MethodConstructor, runtime.CBORBytes(params.ConstructorParams), rt.Message().ValueReceived())
+	_, code := rt.Send(idAddr, builtin.MethodConstructor, ctorParams, rt.Message().ValueReceived())
 	builtin.RequireSuccess(rt, code, "constructor failed")
 
 	return &ExecReturn{idAddr, uniqueAddress}
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -401,10 +401,19 @@ func (a Actor) processDeferredCronEvents(rt Runtime) {
 	})
 	failedMinerCrons := make([]addr.Address, 0)
 	for _, event := range cronEvents {
+		// Technically, actors take _bytes_ as their method params. However,
+		// given that we pass them around as a cbor-able object, we need to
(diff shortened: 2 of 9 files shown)


anorth:
I disagree with the premise here, though I may be missing some discussion that led to this work. I'll take discussion to #973.

Deserializing binary method params into actor structures is conceptually something the actor does. We happen to do it in the VM for convenience now, but logically an actor should directly receive the bytes of the message parameters.

Stebalien:
The problem is with the interplay between the `CBORBytes` type and our CBOR libraries:

1. Our CBOR libraries don't validate CBOR on write.
2. CBORBytes doesn't validate it's inputs.

Example:

1. Client invokes method with some pre-serialized parameter (should be CBOR, but we don't validate).
2. Actor wraps these "should be CBOR bytes" in a `CBORBytes` object.
3. Actor passes this `CBORBytes` object back to the runtime.
4. The runtime embeds this object within some other CBOR object, then marshals the other object.

At this point, `CBORBytes` will write out whatever the user happened to pass in, even if it's not valid CBOR. This _happens_ to be fine in all cases I can find because the runtime never seems to embed this "untrusted" CBOR in any trusted CBOR objects, but the following _would not_ be fine.

```go
type Invocation struct {
    Method Method
    Params CBORMarshaler
    Amount TokenAmount
}
func Call(method Method, params CBORMarshaler) error {
    var buf bytes.Buffer
    inv := Invocation{Method, params, Amount: 0}
    if err := inv.MarshalCBOR(buf); err != nil {
        return err
    }
    return remoteCall(buf.Bytes())
}
```

The "params" would be able to overwrite the `Amount`.

Stebalien:
@anorth I've changed this to check CBOR on send only.

anorth (approved):

Stebalien:
@magik6k is this change going to cause problems for lotus tests?

actors/builtin/init/init_actor.go
```diff
@@ -57,6 +57,13 @@ func (a Actor) Exec(rt runtime.Runtime, params *ExecParams) *ExecReturn {
 		rt.Abortf(exitcode.ErrForbidden, "caller type %v cannot exec actor type %v", callerCodeCID, params.CodeCID)
 	}
 
+	// Technically, actors take _bytes_ as their method params. However,
+	// given that we pass them around as a cbor-able object, we need to
+	// ensure it correctly encodes to CBOR.
+	// See: https://github.com/filecoin-project/specs-actors/issues/972
+	ctorParams, err := runtime.CheckCBOR(params.ConstructorParams)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "params must be valid CBOR")
```
anorth, line 65: Actually, this should probably be `ErrSerialization` (in the other places too).

actors/builtin/power/power_actor.go
```diff
@@ -401,10 +401,19 @@ func (a Actor) processDeferredCronEvents(rt Runtime) {
 	})
 	failedMinerCrons := make([]addr.Address, 0)
 	for _, event := range cronEvents {
+		// Technically, actors take _bytes_ as their method params. However,
+		// given that we pass them around as a cbor-able object, we need to
+		// ensure it correctly encodes to CBOR.
+		// See: https://github.com/filecoin-project/specs-actors/issues/972
+		params, err := vmr.CheckCBOR(event.CallbackPayload)
+		if err != nil {
+			rt.Log(vmr.WARN, "OnDeferredCronEvent failed for due to invalid callback payload: %s", err)
```
ZenGround0, line 410: Since this is a send failure we need to add to failed miner crons

Stebalien:
At this point, there's probably no need for this.

PR #976, opened 2020-08-14
Change the WPoSt commit randomness lookback to match the proof challenge epoch

Stebalien:
This change gets rid of the variable ChainCommitRand parameter and fixes it to be `DeadlineOpen - WPoStChallengeLookback`.

This makes it easy for lotus to extract randomness from the chain at the beginning of the challenge window such that:

1. The randomness is stable and won't change due to reverts. The WPoStChallengeLookback should ensure this.
2. No WPoSt message will become _invalid_ before the challenge window ends.

Before, if lotus picked a commit epoch that was too far in the past, it ran the risk of not being able to submit the proof in time. However, if it picked a recent epoch, it ran the risk of a revert changing the chain.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -290,9 +290,8 @@ type SubmitWindowedPoStParams struct {
 	// Array of proofs, one per distinct registered proof type present in the sectors being proven.
 	// In the usual case of a single proof type, this array will always have a single element (independent of number of partitions).
 	Proofs []abi.PoStProof
-	// The epoch at which these proofs is being committed to a particular chain.
-	ChainCommitEpoch abi.ChainEpoch
-	// The ticket randomness on the chain at the ChainCommitEpoch on the chain this post is committed to
+	// The ticket randomness on the chain at the challenge epoch (WPoStChallengeLookback before the
+	// challenge window opens).
 	ChainCommitRand abi.Randomness
 }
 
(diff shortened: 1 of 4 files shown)


Stebalien:
I have a spec LGTM from @irenegia.

actors/builtin/miner/miner_actor.go
```diff
@@ -287,8 +287,6 @@ type SubmitWindowedPoStParams struct {
 	// Array of proofs, one per distinct registered proof type present in the sectors being proven.
 	// In the usual case of a single proof type, this array will always have a single element (independent of number of partitions).
 	Proofs []abi.PoStProof
-	// The epoch at which these proofs is being committed to a particular chain.
-	ChainCommitEpoch abi.ChainEpoch
 	// The ticket randomness on the chain at the ChainCommitEpoch on the chain this post is committed to
```
anorth, line 290: Fix the comment about which epoch this is from

actors/builtin/miner/policy.go
```diff
@@ -22,7 +22,7 @@ var WPoStChallengeWindow = abi.ChainEpoch(30 * 60 / builtin.EpochDurationSeconds
 const WPoStPeriodDeadlines = uint64(48)
 
 // WPoStMaxChainCommitAge is the maximum distance back that a valid Window PoSt must commit to the current chain.
-var WPoStMaxChainCommitAge = WPoStChallengeWindow
+var WPoStMaxChainCommitAge = WPoStChallengeWindow + WPoStChallengeLookback
```
anorth, line 25: Delete this, now unused.

anorth (approved):

Issue #874, opened 2020-08-03
Constants should not hardcoded but derived and constraints


nicola:
During the audits we found the following problems:
- unclear what are the constraints on the parameters (e.g. was this param set due to scaling constraints or to reach a security goal?)
- unclear how parameters are derived (e.g. does this param depend on the block time or on the block height?)

The current audit checked that values were correct, however we should do the following:
- Add assertion in the code that guarantee that parameters meet our constraints
- Write an explicit formula (when possible) or link to the analysis of how a parameter was decided
- Ideally parameters are not hard coded but derived by such formulas

If we decide not to do this in code, we could try to do this in ubercalc.

ZenGround0:
Relevant to broad rethinking of parameterization: #353

anorth:
I certainly agree with deriving values with in-code arithmetic where possible.

Did you document the constraints that the values should meet in your audit?

nicola:
We have notes for most.

I will give an extra pass when things solidify and write it explicitly in the code

ZenGround0:
@nicola do you still have these notes?

nicola:
I do not, but we can likely get them in parts from the audit issues. #695

PR #984, opened 2020-08-17
Miner: Limit partitions per deadline

### LINKED ISSUE #945: Miner: Limit max number of partitions per deadline
From Slack: 

> [There is] not a hard limit, but the expectation is that no miner will have more than 1EiB of storage. Although we should probably set a hard limit just in case.

aarshkshah1992:
Closes #945 .


actors/abi/sector.go
```diff
@@ -35,6 +35,8 @@ const MaxSectorNumber = math.MaxInt64
 // }
 type SectorSize uint64
 
+const OneEib = SectorSize(1152921504606846976)
```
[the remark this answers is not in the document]
Stebalien: Although, I'd honestly, pick a fixed "partitions per deadline" maximum _based_ on this calculation, then hard-code that (or something close). That way, we always have a consistent number.

I believe the limit should be 300 partitions per deadline.

```
300*(32<<30)*2349*48 -> 1.008 EiB
```

actors/builtin/miner/deadline_assignment.go
```diff
@@ -185,12 +198,22 @@ func assignDeadlines(
...
+			return changes, xerrors.Errorf("maxPartitions limit %d reached for all deadlines", maxPartitions)
+		}
+
 		changes[info.index] = append(changes[info.index], sector)
 		info.liveSectors++
 		info.totalSectors++
 
+		nPartitions := info.totalSectors / partitionSize
+		if nPartitions == maxPartitions {
+			info.maxPartitionsReached = true
```
Stebalien, line 211: nit: I wouldn't actually have this bool. I'd just check `if info.totalSectors >= partitionSize*maxPartitions { return }`.

actors/abi/sector.go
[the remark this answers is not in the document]
Stebalien: Basically, we're trying to find a reasonable limit such that a miner can scale to 1EiB. We might as well pick a nice, memorable number.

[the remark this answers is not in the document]
aarshkshah1992: This is done.

actors/builtin/miner/deadline_assignment.go
aarshkshah1992: I'd like to. keep this around  as it makes the code easier to read in all the places this is used.

aarshkshah1992:
@Stebalien Have made the changes. Please take a look.

wadeAlexC:
Just a heads up, I don't currently understand deadline assignment well enough to provide a good review of this PR.

actors/builtin/miner/deadline_assignment.go
Stebalien: Then we'd need to set it on initialization in case the deadline was already full. But I _really_ think we should get rid of it so we don't forget to set it somewhere.

What about turning it into a helper function on the `deadlineAssignmentInfo` type?

actors/builtin/miner/miner_state.go
```diff
@@ -109,6 +109,9 @@ type MinerInfo struct {
 	// This is computed from the proof type and represented here redundantly.
 	WindowPoStPartitionSectors uint64
 
+	// The maximum number of partitions allowed per Post deadline.
+	MaxPartitionsPerDeadline uint64
```
Stebalien, line 113: This is a global. Is there any reason to include it in the miner state?

Stebalien:
@anorth unless I'm mistaken, we'll need to allow 300 partitions per deadline to reach 1EiB, but that's larger than our current 200 partitions "max addressed". Can we raise the max addressed limit?

actors/builtin/miner/deadline_assignment.go
aarshkshah1992: This makes sense. Done.

actors/builtin/miner/miner_state.go
aarshkshah1992: Have moved this to the abi package.

aarshkshah1992:
@Stebalien @anorth Have made the changes. Please take a look.

actors/builtin/miner/miner_state.go
Stebalien: But why are we storing this in the miner state as well? This is a global limit, not a per-miner limit.

actors/abi/sector.go
```diff
@@ -22,6 +22,10 @@ func (s SectorNumber) String() string {
 // Raising this would require modifying our AMT implementation.
 const MaxSectorNumber = math.MaxInt64
 
+// MaxPartitionsPerDeadline is the maximum number of partitions that will be assigned to a deadline.
+// The current value allows upto 1 Eib storage for a miner.
+const MaxPartitionsPerDeadline = 300
```
anorth, line 27: This is in the wrong place, it should be in `miner/policy.go`.

Also this value is way too small. 1EiB is a _minimum_ to support.

Note: since this number is derived from other numbers, that might change, we should do the arithmetic here either in comment, or, better, the value itself (#874). But if we set it to a value way higher, that's less important.

I propose:
- Document that `48 * 32GiB * 2349 * 300 = 1.00808144 EiB` ([link](https://www.google.com/search?q=48+*+32GiB+*+2349+*+300+in+EiB)) 
- Set this to 3000 (i.e. 10EiB)

If you use units of GiB, you can probably do the arithmetic with int64s to derive this in place.

anorth:
FYI @nicola another parameter.

actors/abi/sector.go
Stebalien: If we set this to 3000, we'll need to process up to 3000 partitions at the end of a deadline. I wonder if we should bump the max addressed partitions limit to match.

Stebalien: > Note: since this number is derived from other numbers, that might change, we should do the arithmetic here either in comment, or, better, the value itself (#874).

Really, there are two constraints here:

1. Maximum supported miner size. This is set as a product requirement.
2. Maximum number of partitions per deadline. This is bounded above by how much work we can do per epoch, and below by how many partitions we need to support 1.

anorth: > bump the max addressed partitions limit to match

Yes, that seems fair.

aarshkshah1992: @Stebalien What is this "ax addressed partitions" that you speak of ?

aarshkshah1992:
@anorth @Stebalien Have made the changes. Please take a look.

actors/abi/sector.go
aarshkshah1992: This is done.

Stebalien:
https://github.com/filecoin-project/specs-actors/pull/984/files#r472276620

actors/builtin/miner/policy.go
```diff
@@ -39,9 +45,8 @@ const SectorsMax = 32 << 20 // PARAM_FINISH
 
 // The maximum number of partitions that may be required to be loaded in a single invocation.
 // This limits the number of simultaneous fault, recovery, or sector-extension declarations.
-// With 48 deadlines (half-hour), 200 partitions per declaration permits loading a full EiB of 32GiB
-// sectors with 1 message per epoch within a single half-hour deadline. A miner can of course submit more messages.
-const AddressedPartitionsMax = 200
+// We set this to same as MaxPartitionsPerDeadline so we can process that many partitions every deadline.
+const AddressedPartitionsMax = 3000
```
Stebalien, line 49: I wonder if we should just set this to `= MaxPartitionsPerDeadline`.

go.mod
```diff
@@ -10,17 +10,27 @@ require (
 	github.com/ipfs/go-block-format v0.0.2
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-hamt-ipld v0.1.1
-	github.com/ipfs/go-ipld-cbor v0.0.4
+	github.com/ipfs/go-ipfs-util v0.0.2 // indirect
+	github.com/ipfs/go-ipld-cbor v0.0.5-0.20200428170625-a0bd04d3cbdf
+	github.com/ipfs/go-ipld-format v0.2.0 // indirect
+	github.com/ipfs/go-log/v2 v2.1.2-0.20200626104915-0016c0b4b3e4 // indirect
+	github.com/kr/pretty v0.2.0 // indirect
```
Stebalien, line 17: Where are these coming from?

actors/builtin/miner/policy.go
aarshkshah1992: Done.

aarshkshah1992:
@Stebalien I think I'd messed up with the go mod changes. It's fixed now. Have also remove the `MaxPartitionsForDeadline` from the miner state.

aarshkshah1992:
@anorth @Stebalien Please can I get an approval on this if it looks good so we can land it ?

anorth (approved):

PR #988, opened 2020-08-17
Feat/static method election eligibility

### LINKED ISSUE #887: Add static state methods for miner election eligibility

### LINKED ISSUE #933: Require minimum total initial pledge for leader election eligibility
As part of simplifying and radically reducing the penalty for consensus faults, we wish to introduce a minimum initial pledge requirement as an eligibility condition for winning the leader election. (That the miner must also have enough funds to cover this pledge is captured in #888).

Implement a free function of miner and reward state that tests whether the miner has a total initial pledge requirement of at least `5.0 * epoch-whole-block-reward`.

The epoch whole block reward value refers to the reward paid to a winning miner per win count, irrespective of the miner's share of power. I.e. `reward.State.ThisEpochReward` [edit: `reward.State.ThisEpochReward / ExpectedLeadersPerEpoch`].

@zixuanzh should this use the `ThisEpochRewardSmoothed`?

ZenGround0:
Closes #887 
Closes #933 

Static function of miner state, reward state, miner balance and current epoch that returns election eligibility status of the provided miner at the provided epoch.


ZenGround0:
The intended caller is the lotus node during block validation.  FYI @magik6k

actors/builtin/miner/election.go
```diff
@@ -0,0 +1,31 @@
+package miner
+
+import (
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/filecoin-project/specs-actors/actors/builtin/reward"
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
+)
+
+func MinerEligibleForElection(store adt.Store, mSt *State, rSt *reward.State, minerActorBalance abi.TokenAmount, currEpoch abi.ChainEpoch) (bool, error) {
```
acruikshank, line 9: My impulse would have been to make this a method on miner state. Is there a reason why you didn't? I don't feel strongly about it, although I'll be a little unhappy about adding another file to the miner directory if it never gets more than one free function.

acruikshank: It would be better to pass in rSt.ThisEpochReward as a parameter than to expose this method to all of reward state.

actors/builtin/miner/election.go
```diff
@@ -0,0 +1,31 @@
...
+	// No active consensus faults
+	mInfo, err := mSt.GetInfo(store)
+	if err != nil {
+		return false, err
+	}
+	if ConsensusFaultActive(mInfo, currEpoch) {
+		return false, nil
+	}
+
+	// IP requirement exceeds minimum for election
```
acruikshank, line 24:
```suggestion
	// IP requirement is sufficient to cover fee for a consensus fault
```

actors/builtin/miner/election_test.go
```diff
@@ -0,0 +1,104 @@
...
+		mSt.InitialPledgeRequirement = miner.ConsensusFaultPenalty(rSt.ThisEpochReward)
+		minerBalance := big.Sub(mSt.InitialPledgeRequirement, abi.NewTokenAmount(1))
+		currEpoch := abi.ChainEpoch(100000)
+
+		eligible, err := miner.MinerEligibleForElection(store, mSt, rSt, minerBalance, currEpoch)
+		require.NoError(t, err)
+		assert.False(t, eligible)
+	})
+
+	t.Run("ip requirement below threshold", func(t *testing.T) {
```
acruikshank, line 90:
```suggestion
	t.Run("ip requirement below consensus fault penalty", func(t *testing.T) {
```

actors/builtin/miner/election.go
ZenGround0: I'm going off of @anorth's cue from the issue this closes:
> Add a single MinerEligibleForElection function that takes parameters as necessary (power state, miner state, current balance)

I think the idea is to make it easier for lotus so they don't have to do any state processing.

Regardless of how we structure we can move to another file.  Any suggestions where to put it?

acruikshank: Lotus will need to do an equivalent amount of state processing whether it's a method on miner state or miner state is passed in as a parameter. I was thinking it should simply return to `miner_state.go` unless there's more anticipated election related logic.

actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +957,30 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
...
 }
 
+func MinerEligibleForElection(store adt.Store, mSt *State, thisEpochReward abi.TokenAmount, minerActorBalance abi.TokenAmount, currEpoch abi.ChainEpoch) (bool, error) {
+	// IP requirements are met.  This includes zero fee debt
+	if !mSt.MeetsInitialPledgeCondition(minerActorBalance) {
+		return false, nil
+	}
+
+	// No active consensus faults
+	mInfo, err := mSt.GetInfo(store)
```
anorth, line 967: Consider taking the info as a parameter instead of the store. The caller likely has it already.

actors/builtin/miner/policy.go
```diff
@@ -239,3 +239,9 @@ func RewardForConsensusSlashReport(elapsedEpoch abi.ChainEpoch, collateral abi.T
 	denom := big.Mul(slasherShareDenominator, consensusFaultReporterInitialShare.denominator)
 	return big.Min(big.Div(num, denom), big.Div(big.Mul(collateral, maxReporterShareNum), maxReporterShareDen))
 }
+
+func ConsensusFaultActive(info *MinerInfo, currEpoch abi.ChainEpoch) bool {
+	// For penalization period to last for exactly finality epochs
+	// consensus faults are active until currEpoch exceeds ConsensusFaultElapsed
+	return currEpoch <= info.ConsensusFaultElapsed
```
anorth, line 246: `ConsensusFaultElapsed` is poorly named if this is the semantic. I think the name is good, and we should change what value we put there such that this becomes `<` and the miner is no longer in fault once the "fault elapsed" epoch begins.

PR #989, opened 2020-08-17
miner create->PoSt scenario test

acruikshank:
### Motivation

We want to have confidence that the flow from creating a miner through committing a sector, all the way through submitting a PoSt work. This test does that at a high level. It confirms that All the method invocations involved in that flow and all their sub-invocations succeed. It does this without getting into the details of what the logic does or specifying inputs and outputs which should be handled by various unit tests.

### Proposed Changes

1. Add `TestCommitPoStFlow` with happy path and a few error cases.
2. Fix bug in VM that returns the wrong current balance.
3. Fix bug in construction of market actor in testing setup.
4. Add functions to advance time by miner deadlines and get current epoch.

Error cases are forthcoming.

### DIFF
--- a/actors/test/miner_scenario_test.go
+++ b/actors/test/miner_scenario_test.go
@@ -14,6 +14,7 @@ import (
 	vm "github.com/filecoin-project/specs-actors/support/vm"
 
 	"github.com/filecoin-project/go-bitfield"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -23,8 +24,6 @@ func TestCommitPoStFlow(t *testing.T) {
 	addrs := vm.CreateAccounts(ctx, t, v, 1, big.Mul(big.NewInt(10_000), big.NewInt(1e18)), 93837778)
 
 	minerBalance := big.Mul(big.NewInt(10_000), vm.FIL)
-	sectorNumber := abi.SectorNumber(100)
-	sealedCid := tutil.MakeCID("100", &miner.SealedCIDPrefix)
 	sealProof := abi.RegisteredSealProof_StackedDrg32GiBV1
 
 	// create miner
@@ -47,6 +46,11 @@ func TestCommitPoStFlow(t *testing.T) {
 	//
 	// precommit sector
 	//
+
+	sectorNumber := abi.SectorNumber(100)
+	sealedCid := tutil.MakeCID("100", &miner.SealedCIDPrefix)
+	sectorSize, err := sealProof.SectorSize()
+	require.NoError(t, err)
 
 	preCommitParams := miner.SectorPreCommitInfo{
 		SealProof:     sealProof,
@@ -71,6 +75,9 @@ func TestCommitPoStFlow(t *testing.T) {
 			{To: builtin.StorageMarketActorAddr, Method: builtin.MethodsMarket.VerifyDealsForActivation}},
 	}.Matches(t, v.Invocations()[0])
 
+	balances := vm.GetMinerBalances(t, v, minerAddrs.IDAddress)
+	assert.True(t, balances.PreCommitDeposit.GreaterThan(big.Zero()))
+
 	// find information about precommited sector
 	var minerState miner.State
 	err = v.GetState(minerAddrs.IDAddress, &minerState)
@@ -82,7 +89,7 @@ func TestCommitPoStFlow(t *testing.T) {
 
 	// advance time to max seal duration
 	proveTime := v.GetEpoch() + miner.MaxSealDuration[sealProof]
-	v, dlInfo := vm.AdvanceTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
+	v, dlInfo := vm.AdvanceByDeadlineTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
 
 	//
 	// overdue precommit
@@ -90,7 +97,7 @@ func TestCommitPoStFlow(t *testing.T) {
 
 	t.Run("missed prove commit results in precommit expiry", func(t *testing.T) {
 		// advanced one more deadline so precommit is late
-		tv, _ := v.WithEpoch(dlInfo.Close)
+		tv, err := v.WithEpoch(dlInfo.Close)
 		require.NoError(t, err)
 
 		// run cron which should expire precommit
@@ -117,6 +124,18 @@ func TestCommitPoStFlow(t *testing.T) {
 				{To: builtin.StorageMarketActorAddr, Method: builtin.MethodsMarket.CronTick},
 			},
 		}.Matches(t, tv.Invocations()[0])
+
+		// precommit deposit has been reset
+		balances := vm.GetMinerBalances(t, tv, minerAddrs.IDAddress)
+		assert.Equal(t, big.Zero(), balances.InitialPledge)
+		assert.Equal(t, big.Zero(), balances.PreCommitDeposit)
+
+		// no power is added
+		networkStats := vm.GetNetworkStats(t, tv)
+		assert.Equal(t, big.Zero(), networkStats.TotalBytesCommitted)
+		assert.Equal(t, big.Zero(), networkStats.TotalPledgeCollateral)
+		assert.Equal(t, big.Zero(), networkStats.TotalRawBytePower)
+		assert.Equal(t, big.Zero(), networkStats.TotalQualityAdjPower)
 	})
 
 	//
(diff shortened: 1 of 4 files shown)


acruikshank:
cc @wadeAlexC

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+	err = v.GetState(minerAddrs.IDAddress, &minerState)
+	require.NoError(t, err)
+
+	precommit, found, err := minerState.GetPrecommittedSector(v.Store(), sectorNumber)
+	require.NoError(t, err)
+	require.True(t, found)
+
+	// advance time to max seal duration
+	proveTime := v.GetEpoch() + miner.MaxSealDuration[sealProof]
+	v, dlInfo := vm.AdvanceTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
```
wadeAlexC, line 85: It seems like this doesn't actually create a vm at epoch `proveTime` - I think it ends up 1 deadline after `proveTime`. Is that correct/intended?

wadeAlexC: ^ Maybe not... but I think the predicate in `AdvanceTillEpoch` may need to be `dlInfo.Close < e`, not `dlInfo.Close <= e`

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+	//
+	// prove and verify
+	//
+
+	// Prove commit sector after max seal duration
+	v, err = v.WithEpoch(proveTime)
+	require.NoError(t, err)
+	proveCommitParams := miner.ProveCommitSectorParams{
+		SectorNumber: sectorNumber,
+	}
```
wadeAlexC, line 131: Why does `Proof == nil` work here?

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+			}},
+			{To: builtin.StorageMarketActorAddr, Method: builtin.MethodsMarket.CronTick},
+		},
+	}.Matches(t, v.Invocations()[1])
+
+	//
+	// Submit PoSt
+	//
+
+	// find information about committed sector
```
wadeAlexC, line 175: I think the best way to do this may be the miner method `CheckSectorProven`

actors/test/miner_scenario_test.go
acruikshank: The miner_actor simply passes the `Proof` into it's verify syscall and the scenario test VM syscall simply ignores the input and does nothing indicating the proof is valid. Scenario testing would get extremely complicated if we needed real proofs, and the unit tests cover that the miner handles the proof correctly.

wadeAlexC: Makes sense. Do you think it'd be useful to at least check that syscalls are made? Something like `ExpectInvocation`, but with important syscalls.

wadeAlexC:
These are awesome! I think you struck a really good balance testing invocations with checking actor state. Left a few comments. I'm excited to see more of these!

actors/test/miner_scenario_test.go
wadeAlexC: I may just be wrong about this. Feel free to ignore :)

acruikshank: I fiddled with this quite a bit yesterday trying to get something that kind of works. The important thing for the advance is that the vm lands in a deadline period that matches the `dlInfo` returned. Because of the way we use the `CurrentDeadline` field in miner state, getting deadline info for an epoch doesn't mean the epoch is actually in that deadline.

Note that we get a VM at `proveTime` before calling `ProveCommitSector`.

actors/test/miner_scenario_test.go
acruikshank: It would be easy to do. Those syscalls are well covered in the unit tests, so I haven't bothered here. I'm going to state that things that we can test well in unit tests we should avoid in scenario tests just to keep complexity down and reduce the burden of change. I could easily be convinced otherwise if you or someone else were to make the case.

actors/test/miner_scenario_test.go
acruikshank: I'm not quite sure what you mean. I need the deadline and partition indices for the sector to send a `SubmitPoSt` at the right time with the right parameters. And I need the `OnChainSectorInfo` to compute the sector's power. I could add a call to `CheckSectorProven` before and after the test cases.

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+					{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.CurrentTotalPower},
+					{To: builtin.StorageMarketActorAddr, Method: builtin.MethodsMarket.ActivateDeals},
+					{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.UpdateClaimedPower},
+					{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.UpdatePledgeTotal},
+				}},
+				{To: builtin.RewardActorAddr, Method: builtin.MethodsReward.UpdateNetworkKPI},
+			}},
+			{To: builtin.StorageMarketActorAddr, Method: builtin.MethodsMarket.CronTick},
+		},
+	}.Matches(t, v.Invocations()[1])
```
ZenGround0, line 169: Having to track the index adds some friction.  Maybe adding a `v.LastInvocation` would help

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+		sectorSize, err := sector.SealProof.SectorSize()
+		require.NoError(t, err)
+		sectorPower := miner.PowerForSector(sectorSize, sector)
+		updatePowerParams := &power.UpdateClaimedPowerParams{
+			RawByteDelta:         sectorPower.Raw.Neg(),
+			QualityAdjustedDelta: sectorPower.QA.Neg(),
+		}
+
+		vm.ExpectInvocation{
+			// Original send to storage power actor
```
ZenGround0, line 289: nit: original send is to cron

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+	v, dlInfo := vm.AdvanceTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
+
+	//
+	// overdue precommit
+	//
+
+	t.Run("missed prove commit results in precommit expiry", func(t *testing.T) {
+		// advanced one more deadline so precommit is late
+		tv, _ := v.WithEpoch(dlInfo.Close)
+		require.NoError(t, err)
```
anorth, line 94: `err` is not set as expected here.

Higher level, I think we should strive to remove error returns from the test helpers. fail the test or panic instead so the test code can be tighter.

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+						// The call to burnt funds indicates the overdue precommit has been penalized
+						{To: builtin.BurntFundsActorAddr, Method: builtin.MethodSend, Value: vm.ExpectAttoFil(precommit.PreCommitDeposit)},
+						{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.EnrollCronEvent},
+					}},
+					//{To: minerAddrs.IDAddress, Method: builtin.MethodsMiner.ConfirmSectorProofsValid},
+					{To: builtin.RewardActorAddr, Method: builtin.MethodsReward.UpdateNetworkKPI},
+				}},
+				{To: builtin.StorageMarketActorAddr, Method: builtin.MethodsMarket.CronTick},
+			},
+		}.Matches(t, tv.Invocations()[0])
```
anorth, line 119: I would like to see some high-level state checks, e.g. that the miner's IP is still zero and PCD is now also 0, and power actor total pledge also zero. (We need _some_ state inspection in lieu of read-only methods). Possibly extract them behind a method (like Go-filecoin's state view) so that future changes are easy to ingest here.

actors/test/miner_scenario_test.go
```diff
@@ -0,0 +1,315 @@
...
+				PoStProof: abi.RegisteredPoStProof_StackedDrgWindow32GiBV1,
+			}},
+			ChainCommitEpoch: tv.GetEpoch() - 1,
+			ChainCommitRand:  []byte("not really random"),
+		}
+		_, code = tv.ApplyMessage(addrs[0], minerAddrs.RobustAddress, big.Zero(), builtin.MethodsMiner.SubmitWindowedPoSt, &submitParams)
+		require.Equal(t, exitcode.Ok, code)
+
+		vm.ExpectInvocation{
+			// Original send to storage power actor
```
anorth, line 211: Stale comment

support/vm/testing.go
```diff
@@ -226,6 +227,45 @@ func ParamsForInvocation(t *testing.T, vm *VM, idxs ...int) interface{} {
 	return invocation.Msg.params
 }
 
+//
+// Advancing Time while updating state
+//
+
+type advanceDeadlinePredicate func(dlInfo *miner.DeadlineInfo) bool
+
+func AdvanceByMinerDeadline(t *testing.T, v *VM, minerIDAddr address.Address, predicate advanceDeadlinePredicate) (*VM, *miner.DeadlineInfo) {
```
anorth, line 236: Please document what this does

anorth (approved):
Good stuff. 

I think there'll be room to make these more concise as we develop examples.

One thing I would like us to try is more checking of balances and transfers. Calculating penalties etc can make this tricky, but it's one of the most important high-level outcomes to be checking.

PR #996, opened 2020-08-18
Multisig: Allow configurable vesting Start

### LINKED ISSUE #511: Configurable vesting start for multisig
From @ianjdarrow

Currently, vesting always starts from the block in which the actor is created. It would be nice from a product perspective to be able to specify the start block (including blocks less than the current height and negative heights) to allow for things like time-locks, partial pre-vesting, and the like.

aarshkshah1992:
Closes #511.

@anorth Turns out we were missing tests related to the vesting/locking of funds. Have added them in this PR. Please take a look.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -68,6 +68,7 @@ type ConstructorParams struct {
 	Signers               []addr.Address
 	NumApprovalsThreshold uint64
 	UnlockDuration        abi.ChainEpoch
+	StartEpoch            abi.ChainEpoch
 }
 
 func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.EmptyValue {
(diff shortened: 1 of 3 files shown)


actors/builtin/multisig/multisig_actor.go
```diff
@@ -117,7 +118,7 @@ func (a Actor) Constructor(rt vmr.Runtime, params *ConstructorParams) *adt.Empty
 	if params.UnlockDuration != 0 {
 		st.InitialBalance = rt.Message().ValueReceived()
 		st.UnlockDuration = params.UnlockDuration
-		st.StartEpoch = rt.CurrEpoch()
+		st.StartEpoch = params.StartEpoch
```
anorth, line 121: We're changing behaviour here where there's probably nontrivial external code already written. If, at epoch 100000, someone sets a vesting duration of 10 days but forgets to set StartEpoch, they probably don't intend everything to immediately vest. 

We should probably either
- default to current epoch if start epoch is zero, or
- require it to be >= the current epoch

We can't simply require start epoch to be nonzero since that's a valid value during genesis state construction.

@whyrusleeping do you have a preference?

If not, go with defaulting to current epoch.

whyrusleeping: hrm... zero should be a valid value. I think if they forget to set the value and it defaults to zero, then they can just make a new multisig? No strong opinions here though

anorth: Ok, no change required then.

anorth (approved):

PR #997, opened 2020-08-18
Cap 32GiB sector initial pledge to 1FIL for Space Race (hotfix)

anorth:
Computes a per-byte initial pledge cap so that a 32GiB sector is fractionally under 1 FIL.

- Pre-commit deposit remains BR(20)
- Termination fee is unchanged; still includes BR(20) plus the approximation of rewards earned.

The cap is a `var`, and evaluated only when a pledge calculation is made, so Lotus can override this, e.g. effectively disabling it by setting a higher value.

This PR is against the `release/v0.9` branch, not `master`. After merge, a new release can be made from this branch head.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -240,7 +240,7 @@ func TestCommitments(t *testing.T) {
 		assert.Equal(t, big.NewInt(int64(sectorSize/2)), onChainPrecommit.VerifiedDealWeight)
 
 		qaPower := miner.QAPowerForWeight(sectorSize, precommit.Expiration-precommitEpoch, onChainPrecommit.DealWeight, onChainPrecommit.VerifiedDealWeight)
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -240,7 +240,7 @@ func TestCommitments(t *testing.T) {
 		assert.Equal(t, big.NewInt(int64(sectorSize/2)), onChainPrecommit.VerifiedDealWeight)
 
 		qaPower := miner.QAPowerForWeight(sectorSize, precommit.Expiration-precommitEpoch, onChainPrecommit.DealWeight, onChainPrecommit.VerifiedDealWeight)
-		expectedDeposit := miner.InitialPledgeForPower(qaPower, actor.baselinePower, actor.networkPledge, actor.epochRewardSmooth, actor.epochQAPowerSmooth, rt.TotalFilCircSupply())
+		expectedDeposit := miner.PreCommitDepositForPower(actor.epochRewardSmooth, actor.epochQAPowerSmooth, qaPower)
```
anorth, line 243: This was previously setting the wrong expectation, although on master they align when the circulating supply is zero.

PR #998, opened 2020-08-18
Miner: Limit control addresses

### LINKED ISSUE #977: Miner: Limit size of ControlAddresses slice
When the Miner's info struct is loaded, it must load the entire `info.ControlAddresses` slice, which has no upper bound on size.

In particular, this may increase strain on the reward actor during its invocation of `Miner.AddLockedFund`:

https://github.com/filecoin-project/specs-actors/blob/0d2ebd356c45acb4c3d340ab9d81b20eb65fea03/actors/builtin/miner/miner_actor.go#L1365-L1375

aarshkshah1992:
Closes #977.


wadeAlexC:
LGTM

Issue #981, opened 2020-08-14
Miner/Market: Clean up deal weight calculation during precommit


wadeAlexC:
`Miner.PreCommitSector` invokes `Market.ValidateDealsForActivation` in order to calculate the total deal weight for all referenced deal:

https://github.com/filecoin-project/specs-actors/blob/0d2ebd356c45acb4c3d340ab9d81b20eb65fea03/actors/builtin/miner/miner_actor.go#L503

`ValidateDealsForActivation` is provided an incorrect `SectorStart` epoch by the miner (`CurrEpoch`), as the sector's start epoch is not known at this point. This `SectorStart` is used to perform validation that the deal is in an acceptable state, but is irrelevant to the calculation of `DealWeight`:

https://github.com/filecoin-project/specs-actors/blob/0d2ebd356c45acb4c3d340ab9d81b20eb65fea03/actors/builtin/market/market_actor.go#L630-L642

#### Recommendation

Given that this validation doesn't accomplish much, I propose removing validation from `Market.ValidateDealsForActivation` in favor of a dead-simple deal weight calculation.

anorth:
We've determined that this isn't high importance for network launch. It's a nice clean-up that we'll implement in a discretionary upgrade later.

cc @wadeAlexC

PR #999, opened 2020-08-18
Miner: Batch call to market.ActivateDeals

### LINKED ISSUE #474: ConfirmSectorProofsValid: batch VerifyDealsOnSectorProveCommit
From builtin/miner/miner_actor.go:ConfirmSectorProofsValid:

```

		// Check (and activate) storage deals associated to sector. Abort if checks failed.
		// return DealWeight for the deal set in the sector
		// TODO: we should batch these calls...
```

I believe this is referring to batching multiple sectors in a single call from miner to market instead of one call per sector.

aarshkshah1992:
Closes #474.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -278,9 +278,11 @@ func (A Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
 
 	var st State
 	rt.State().Readonly(&st)
-	store := adt.AsStore(rt)
 
-	dealWeight, verifiedWeight, err := ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, params.SectorStart)
+	msm, err := st.mutator(adt.AsStore(rt)).withDealProposals(ReadOnlyPermission).build()
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deal state")
+
+	dealWeight, verifiedWeight, err := ValidateDealsForActivation(msm.dealProposals, params.DealIDs, minerAddr, params.SectorExpiry, params.SectorStart)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to validate dealProposals for activation")
 
 	return &VerifyDealsForActivationReturn{
@@ -311,7 +313,6 @@ func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParam) *ActivateDe
 	currEpoch := rt.CurrEpoch()
 
 	var st State
-	store := adt.AsStore(rt)
 
 	activatedSectors := make([]abi.SectorNumber, 0, len(params.Requests))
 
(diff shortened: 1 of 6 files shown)


aarshkshah1992:
@anorth @ZenGround0 Please can I get some quick feedback on the market refactor before I push the miner and test changes ? Does this refactor make sense ?

actors/builtin/market/market_actor.go
```diff
@@ -289,65 +289,118 @@ func (A Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
...
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal state %d", dealID)
+			if success {
+				for _, dealID := range req.DealIDs {
+					err := msm.dealStates.Set(dealID, &DealState{
+						SectorStartEpoch: currEpoch,
+						LastUpdatedEpoch: epochUndefined,
+						SlashEpoch:       epochUndefined,
+					})
+
+					builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal state for activation, dealId: %d, sector number: %v, err: %w",
```
aarshkshah1992, line 389: @anorth @ZenGround0 

Unfortunately, I think we need to abort here for atomicity.

If we don't do this, we could end up in a situation where some deals are set and some others fail which would leave the sector "fragmented".

ZenGround0: I'm in support of propagating this illegal state error to ConfirmSectorProofsValid and failing there instead of just logging.  If our prove commits expose illegal market state there is a big problem that should be brought to light rather than patched over in the interest of saving the miner some funds.

actors/builtin/market/market_actor.go
```diff
@@ -289,65 +289,118 @@ func (A Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
...
+			_, _, err := ValidateDealsForActivation(&st, store, req.DealIDs, minerAddr, req.SectorExpiry, currEpoch)
+			if err != nil {
+				rt.Log(vmr.ERROR, "failed to validate deals for activation, sector number: %v, err: %w", req.SectorNumber, err)
+				continue
 			}
 
-			proposal, err := getDealProposal(msm.dealProposals, dealID)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get dealId %d", dealID)
+			success := true
+			for _, dealID := range req.DealIDs {
```
Stebalien, line 332: Can we break this into a function?

actors/builtin/market/market_actor.go
```diff
@@ -289,65 +289,118 @@ func (A Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
...
 
-		for _, dealID := range params.DealIDs {
-			// This construction could be replaced with a single "update deal state" state method, possibly batched
-			// over all deal ids at once.
-			_, found, err := msm.dealStates.Get(dealID)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get state for dealId %d", dealID)
-			if found {
-				rt.Abortf(exitcode.ErrIllegalArgument, "deal %d already included in another sector", dealID)
+		for _, req := range params.Requests {
+			_, _, err := ValidateDealsForActivation(&st, store, req.DealIDs, minerAddr, req.SectorExpiry, currEpoch)
```
Stebalien, line 325: We're loading state here after we start mutating. This isn't going to be correct. I think we need to inline the logic from this function.

actors/builtin/market/market_actor.go
```diff
@@ -289,65 +289,118 @@ func (A Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
...
 			}
 
-			err = msm.dealStates.Set(dealID, &DealState{
-				SectorStartEpoch: currEpoch,
-				LastUpdatedEpoch: epochUndefined,
-				SlashEpoch:       epochUndefined,
-			})
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to set deal state %d", dealID)
+			if success {
+				for _, dealID := range req.DealIDs {
```
Stebalien, line 382: This requires https://github.com/filecoin-project/specs-actors/pull/975 for correctness as we need to check for duplicate deals. As noted above, I think it would be better to inline the relevant logic, including the duplicate detection logic

aarshkshah1992:
@ZenGround0 @Stebalien I've incorporated your comments and this PR is now complete and ready for review. Please take a look.

actors/builtin/market/market_actor.go
```diff
@@ -289,65 +291,67 @@ func (A Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
...
 
-		for _, dealID := range params.DealIDs {
-			// This construction could be replaced with a single "update deal state" state method, possibly batched
-			// over all deal ids at once.
-			_, found, err := msm.dealStates.Get(dealID)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get state for dealId %d", dealID)
-			if found {
-				rt.Abortf(exitcode.ErrIllegalArgument, "deal %d already included in another sector", dealID)
+		for _, req := range params.Requests {
+			_, _, err := ValidateDealsForActivation(msm.dealProposals, req.DealIDs, minerAddr, req.SectorExpiry, currEpoch)
```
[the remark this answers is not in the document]
anorth: The logic in validateDealCanActivate must be in the path somewhere.

anorth (commented):
This has introduced an arbitrary-sizes message in the `ActivateDealsParam`. The batching needs to be more careful in order to bound the size of each request.

This is looking like a much bigger change than I was hoping might be possible. I'm no longer confident in attempting it this late.

I propose abandoning this for now, but instead doing the very simple thing of avoiding the call to `ActivateDeals` when there are no deals. The rest of the optimization is not worth the risk right now.

This might be easier after we do #981.

aarshkshah1992:
Closing as suggested by @anorth .

PR #1000, opened 2020-08-18
Disallow miner actions after consensus fault

### LINKED ISSUE #965: Restrict miner capabilities around consensus fault
Following on from #886, we wish also to restrict the same capabilities that are restricted in IP debt when a miner is reported for a consensus fault, or would be unable to be penalised for one.

Thus, in order to be eligible to pre-commit, recover faults, etc, a miner must:
- have balance fully covering IP requirement
- have _balance_ (but not IP requirement) above the minimum required by #933
- have a consensus fault mark (#934) that is empty or has expired before the current epoch

Note a subtlety with regard to the minimum IP requirement. The leader election requirement is expressed in terms of minimum IP requirement, but we can't prevent a miner from committing new sectors with a low IP requirement since committing new sectors is how they increase that value. So we instead require the miner to have an unlocked balance exceeding that IP requirement. This is imperfect, since a miner could pre-commit sectors, withdraw balance down to PCD+IPR (below minimum), and commit a consensus fault all in the same epoch.

@nicola I must express some doubts that this is really a good idea. 
- It amounts to a minimum pledge before a miner can participate at all (by committing their first sectors), even though they can't win a block anyway. Is that the actual design goal?
- Preventing election win by a miner which couldn't pay the fee makes sense, but how does preventing pre-commit or recovery help? A miner with IPR below the minimum can't possibly commit a consensus fault anyway.
- It's not well enforceable. If it's important, we'll need to do something more complicated for that minimum participation pledge

cc @ZenGround0

ZenGround0:
Closes #965 

A miner cannot pre-commit, withdraw funds or recover sectors after a consensus fault.  @nicola @anorth I think we are mostly all aligned here after discussion on issue.  The one remaining thing to hash out is whether these are the correct methods to be disallowing.  I kept the same set that we do IP requirement / fee debt checking on.  However in the issue @nicola mentioned ProveCommit which isn't included here and didn't mention WithdrawBalance which is included here.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -631,6 +631,22 @@ func TestCommitments(t *testing.T) {
 		st.FeeDebt = big.Zero()
 		rt.ReplaceState(st)
 		rt.Reset()
+
+		// Try to precommit with an active consensus fault
+		st = getState(rt)
+		params := &miner.ReportConsensusFaultParams{
+			BlockHeader1:     nil,
+			BlockHeader2:     nil,
+			BlockHeaderExtra: nil,
+		}
+
+		actor.reportConsensusFault(rt, addr.TestAddress, params)
+		rt.ExpectAbortContainsMessage(exitcode.ErrForbidden, "precommit not allowed during active consensus fault", func() {
+			actor.preCommitSector(rt, actor.makePreCommit(102, challengeEpoch, expiration, nil))
+		})
+		// reset state back to normal
+		rt.ReplaceState(st)
+		rt.Reset()
 	})
 
 	t.Run("valid committed capacity upgrade", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -608,6 +608,22 @@ func TestCommitments(t *testing.T) {
...
+
+		// Try to precommit with an active consensus fault
+		st = getState(rt)
+		params := &miner.ReportConsensusFaultParams{
+			BlockHeader1:     nil,
+			BlockHeader2:     nil,
+			BlockHeaderExtra: nil,
+		}
+
+		actor.reportConsensusFault(rt, addr.TestAddress, params)
```
anorth, line 620: Nit: push the params construction down inside reportConsensusFault so this isn't so noisy.

anorth (approved):
This looks good, but I don't see a reason to prevent withdrawal.

Issue #966, opened 2020-08-12
More tests of miner state fee accounting


ZenGround0:
Hit all the edge cases with money moving functions of miner state.  Motivated by [this suggestion](https://github.com/filecoin-project/specs-actors/pull/956/files#r468961333) to extend test of PenalizeFundsInPriorityOrder but there are definitely other edge cases we are not testing.

ZenGround0:
Also see [this suggestion](https://github.com/filecoin-project/specs-actors/pull/956#discussion_r469646161) for motivation on further tests.

PR #1004, opened 2020-08-19
IP accounting refactor

ZenGround0:
After a proposal from @wadeAlexC and brief discussion with @acruikshank and @anorth I wanted to see how this refactor looks.  This is to help make the discussion concrete -- we're not necessarily going to merge this.

Pros: the simplification is real, both in the actor code, the mental model and in the reduction of separate cases to test.  By doing all fee debt repayment lazily very little new code needs to be added.

Cons: one upside that's come to light with the current strategy of applying initial pledge to fee payment up front is that dead miners that have incurred fees always pay all of their initial pledge requirement to burnt funds without intervention.  After this refactor it is possible for a miner that is no longer worth maintaining to be abandoned with lots of fee debt and lots of funds free in its balance as initial pledge is unlocked on sector expiry.  

We could fix this easily with an unpermissioned `RepayFee` method that burns free balance up to FeeDebt, or some more complicated automatic payment.  Not sure if that is worth the trade off.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -425,8 +425,11 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 
 		// Note: We could delay this charge until end of deadline, but that would require more accounting state.
 		totalPenaltyTarget := big.Add(undeclaredPenaltyTarget, declaredPenaltyTarget)
-		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
-		vestingPenaltyTotal, balancePenaltyTotal, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, totalPenaltyTarget, unlockedBalance)
+		err = st.ApplyPenalty(totalPenaltyTarget)
+		builtin.RequireNoErr(rt, err, exitcode.Unwrap(err, exitcode.ErrIllegalState), "failed to apply penalty")
+
+		// Pay penalty
+		vestingPenaltyTotal, balancePenaltyTotal, err := st.RepayPartialDebtInPriorityOrder(store, currEpoch, rt.CurrentBalance())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty for %v", undeclaredPenaltyPower)
 		penaltyTotal = big.Add(vestingPenaltyTotal, balancePenaltyTotal)
 		pledgeDelta = big.Sub(pledgeDelta, vestingPenaltyTotal)
@@ -507,10 +510,10 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 		// available balance already accounts for fee debt so it is correct to call
-		// this before VerifyPledgeRequirementsAndRepayDebts. We would have to
+		// this before RepayDebts. We would have to
 		// subtract fee debt explicitly if we called this after.
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
-		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
+		feeToBurn = RepayDebtsOrAbort(rt, &st)
 
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
@@ -590,6 +593,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	})
 
 	burnFunds(rt, feeToBurn)
+	rt.State().Readonly(&st)
+	st.AssertBalanceInvariants(rt.CurrentBalance())
 
 	notifyPledgeChanged(rt, newlyVested.Neg())
 
@@ -818,7 +823,7 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds for aggregate initial pledge requirement %s, available: %s", totalPledge, availableBalance)
 		}
 
-		st.AddInitialPledgeRequirement(totalPledge)
+		st.AddInitialPledge(totalPledge)
 		st.AssertBalanceInvariants(rt.CurrentBalance())
 	})
 
@@ -1227,7 +1232,7 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 	rt.State().Transaction(&st, func() {
 		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
 		// and repay fee debt now.
-		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
+		feeToBurn = RepayDebtsOrAbort(rt, &st)
 
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
@@ -1264,6 +1269,8 @@ func (a Actor) DeclareFaultsRecovered(rt Runtime, params *DeclareFaultsRecovered
 	})
 
 	burnFunds(rt, feeToBurn)
+	rt.State().Readonly(&st)
+	st.AssertBalanceInvariants(rt.CurrentBalance())
 
 	// Power is not restored yet, but when the recovered sectors are successfully PoSted.
 	return nil
@@ -1425,7 +1432,8 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 	}
 
 	// Elapsed since the fault (i.e. since the higher of the two blocks)
-	faultAge := rt.CurrEpoch() - fault.Epoch
+	currEpoch := rt.CurrEpoch()
+	faultAge := currEpoch - fault.Epoch
 	if faultAge <= 0 {
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid fault epoch %v ahead of current %v", fault.Epoch, rt.CurrEpoch())
 	}
@@ -1444,10 +1452,13 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 	burnAmount := big.Zero()
 	rewardAmount := big.Zero()
 	rt.State().Transaction(&st, func() {
-		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
-		penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(adt.AsStore(rt), rt.CurrEpoch(), faultPenalty, unlockedBalance)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock unvested funds")
-		// Burn the amount actually payable. Any difference in this and faultPenalty recorded as FeeDebt
+		err := st.ApplyPenalty(faultPenalty)
+		builtin.RequireNoErr(rt, err, exitcode.Unwrap(err, exitcode.ErrIllegalState), "failed to apply penalty")
+
+		// Pay penalty
+		penaltyFromVesting, penaltyFromBalance, err := st.RepayPartialDebtInPriorityOrder(adt.AsStore(rt), currEpoch, rt.CurrentBalance())
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to pay fees")
+		// Burn the amount actually payable. Any difference in this and faultPenalty already recorded as FeeDebt
 		burnAmount = big.Add(penaltyFromVesting, penaltyFromBalance)
 		pledgeDelta = big.Add(pledgeDelta, penaltyFromVesting.Neg())
 
@@ -1506,13 +1517,13 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
 			rt.Abortf(exitcode.ErrIllegalState, "failed to vest fund: %v", err)
 		}
 		// available balance already accounts for fee debt so it is correct to call
-		// this before VerifyPledgeRequirementsAndRepayDebts. We would have to
+		// this before RepayDebts. We would have to
 		// subtract fee debt explicitly if we called this after.
 		availableBalance = st.GetAvailableBalance(rt.CurrentBalance())
 
 		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
 		// and repay fee debt now.
-		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
+		feeToBurn = RepayDebtsOrAbort(rt, &st)
 	})
 
 	amountWithdrawn := big.Min(availableBalance, params.AmountRequested)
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/monies.go
```diff
@@ -124,17 +124,11 @@ func InitialPledgeForPower(qaPower abi.StoragePower, baselinePower abi.StoragePo
 // may be slightly lower than the true amount. Computing vesting here would be
 // almost always redundant since vesting is quantized to ~daily units.  Vesting
 // will be at most one proving period old if computed in the cron callback.
-func VerifyPledgeRequirementsAndRepayDebts(rt Runtime, st *State) abi.TokenAmount {
+func RepayDebts(rt Runtime, st *State) abi.TokenAmount {
```
ZenGround0, line 127: Not sure about this name, just trying out something much simpler

actors/builtin/miner/miner_state.go
```diff
@@ -858,34 +858,31 @@ func (st *State) CheckVestedFunds(store adt.Store, currEpoch abi.ChainEpoch) (ab
...
+	unlockedBalance := big.Subtract(actorBalance, st.LockedFunds, st.PreCommitDeposits, st.InitialPledge)
 	Assert(unlockedBalance.GreaterThanEqual(big.Zero()))
 	return unlockedBalance
 }
 
-// Unclaimed funds.  Actor balance - (locked funds, precommit deposit, ip requirement, fee debt)
+// Unclaimed funds.  Actor balance - (locked funds, precommit deposit, initial pledge, fee debt)
 // Can go negative if the miner is in IP debt
 func (st *State) GetAvailableBalance(actorBalance abi.TokenAmount) abi.TokenAmount {
 	unlockedBalance := st.GetUnlockedBalance(actorBalance)
```
anorth, line 872: It's unfortunate that we still end up with both these methods. Can one or the other be removed?

anorth (commented):
Thanks for the draft.

I'm still very uncertain about this. I think I'd have to spend much more time with the PR, manually checking all call sites of the balance methods, etc.

anorth:
@ZenGround0 I think we'd want pledge funds that free up to automatically pay off fee debt. I think we want to avoid a bunch of funds being effectively burnt by being locked in an abandoned miner without accounting for them in money supply calculations.

ZenGround0:
> I think we'd want pledge funds that free up to automatically pay off fee debt.

Yeah @wadeAlexC pointed out to me earlier that this would actually be simpler.  We just need a new function to repay partial fee debt and call it after sectors expire and reduce InitialPledge.

zixuanzh:
> We just need a new function to repay partial fee debt and call it after sectors expire and reduce InitialPledge.

Partial fee debt should probably be called first before any InitialPledge is reduced, the other case will be termination.

> I think we want to avoid a bunch of funds being effectively burnt by being locked in an abandoned miner without accounting for them in money supply calculations.

Not sure how we plan to do this, abandoned miners with fee debt could also happen before this refactor. My understanding is that FIL_Burned will not include fee debt to avoid double counting. We can clean up a miner if there is no proving sectors and the miner is still in fee debt after exhausting all initial pledge. 

Lastly, can we confirm that FIL_Locked = TotalInitialPledgeRequirement + TotalLockedFunds (maybe +TotalPCDs too) after the refactor? Let me know if this behavior is intended to change.

zixuanzh:
> We just need a new function to repay partial fee debt and call it after sectors expire and reduce InitialPledge.

Partial fee debt should probably be called first before any InitialPledge is reduced, the other case will be termination.

> I think we want to avoid a bunch of funds being effectively burnt by being locked in an abandoned miner without accounting for them in money supply calculations.

Not sure how we plan to do this, abandoned miners with fee debt could also happen before this refactor. My understanding is that FIL_Burned will not include fee debt to avoid double counting. We can clean up a miner if there is no more proving sectors and the miner is still in fee debt after exhausting all initial pledge. cc @davidad 

Lastly, can we confirm that FIL_Locked = TotalInitialPledgeRequirement + TotalLockedFunds (maybe +TotalPCDs too) after the refactor? Let me know if this behavior is intended to change.

zixuanzh:
> We just need a new function to repay partial fee debt and call it after sectors expire and reduce InitialPledge.

Partial fee debt should probably be called first before any InitialPledge is reduced, the other case will be termination.

> I think we want to avoid a bunch of funds being effectively burnt by being locked in an abandoned miner without accounting for them in money supply calculations.

Not sure how we plan to do this, abandoned miners with fee debt could also happen before this refactor. My understanding is that FIL_Burned will not include fee debt to avoid double counting. We can clean up a miner if there is no more proving sectors and the miner is still in fee debt after exhausting all initial pledge. cc @davidad @jzimmerman 

Lastly, can we confirm that FIL_Locked = TotalInitialPledgeRequirement + TotalLockedFunds (maybe +TotalPCDs too) after the refactor? Let me know if this behavior is intended to change.

ZenGround0:
> Not sure how we plan to do this

We will do this by burning partial fee debt after a sector expires (covers termination case).  So as soon as IP is unlocked it will go to paying off fee debt.

actors/builtin/miner/miner_actor.go
```diff
@@ -1621,11 +1621,17 @@ func processEarlyTerminations(rt Runtime) (more bool) {
 		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
 		penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, rt.CurrEpoch(), penalty, unlockedBalance)
```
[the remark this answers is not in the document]
anorth: Please do that clean-up.

actors/builtin/miner/miner_state.go
```diff
@@ -793,6 +793,12 @@ func (st *State) RepayDebt(currBalance abi.TokenAmount) (abi.TokenAmount, error)
 	return debtToRepay, nil
 }
 
+func (st *State) RepayPartialDebt(unlockedBalance abi.TokenAmount) abi.TokenAmount {
+	debtToRepay := big.Min(unlockedBalance, st.FeeDebt)
+	st.FeeDebt = big.Sub(st.FeeDebt, debtToRepay)
+	return debtToRepay
+}
```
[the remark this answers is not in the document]
anorth: I would really like it if we could remove the `unlockedBalance` parameter here, in favour of `currentBalance`. It seems really fragile to be passing that into a state method.

actors/builtin/miner/miner_state.go
anorth: Ok, please document it as such. Perhaps we move it to a free function rather than a state method, matching some other only-for-nodes functions?

actors/builtin/miner/monies.go
anorth: I don't think this needs to be exported, which means it probably moves to `miner_actor.go` as `repayDebts`.

anorth (commented):
I appreciate the effort here to keep the change small, but in this case the review effort is dominated by making sure it's correct. I'm not evaluating this as a "what's changed?", but "is this whole thing correct as written?". A change that was larger but more obviously correct would be easier. I think that means doing some of #1020 and #1026 right away.

actors/builtin/miner/miner_actor.go
```diff
@@ -1751,7 +1757,13 @@ func handleProvingDeadline(rt Runtime) {
...
 			// when the early termination is processed.
 			pledgeDelta = big.Sub(pledgeDelta, expired.OnTimePledge)
-			st.AddInitialPledgeRequirement(expired.OnTimePledge.Neg())
+			unlockedBalance = big.Add(unlockedBalance, expired.OnTimePledge)
+			st.AddInitialPledge(expired.OnTimePledge.Neg())
+
+			// Use unlocked pledge to pay down outstanding fee debt
+			feeToBurn := st.RepayPartialDebt(unlockedBalance)
+			unlockedBalance = big.Sub(unlockedBalance, feeToBurn) //nolint:ineffassign
+			penaltyTotal = big.Add(penaltyTotal, feeToBurn)
```
anorth, line 1766: Note there's a collision here with #1009, and I think that should probably land first.

actors/builtin/miner/miner_actor.go
```diff
@@ -1730,15 +1737,14 @@ func handleProvingDeadline(rt Runtime) {
...
-
 		}
 		{
 			// Record faulty power for penalisation of ongoing faults, before popping expirations.
 			// This includes any power that was just faulted from missing a PoSt.
 			penaltyTarget := PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, deadline.FaultyPower.QA)
 			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, unlockedBalance)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
-			unlockedBalance = big.Sub(unlockedBalance, penaltyFromBalance) //nolint:ineffassign
+			unlockedBalance = big.Sub(unlockedBalance, penaltyFromBalance)
```
anorth, line 1747: While reviewing #1009 I determined that the `unlockedBalance` calculated above is incorrect, as it doesn't take into account balance earmarked for burning from expired precommits.

actors/builtin/miner/miner_actor.go
```diff
@@ -1621,11 +1621,18 @@ func processEarlyTerminations(rt Runtime) (more bool) {
 		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
 		penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, rt.CurrEpoch(), penalty, unlockedBalance)
```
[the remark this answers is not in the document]
ZenGround0: We've had to do this in the past because we needed a way to track all of the funds we have already burned from our balance.  I'm going to do #1026 which should clean this up a lot but tracking all our penalties in FeeDebt without restriction.

actors/builtin/miner/miner_actor.go
```diff
@@ -1621,11 +1621,18 @@ func processEarlyTerminations(rt Runtime) (more bool) {
...
 		// Remove pledge requirement.
-		st.AddInitialPledgeRequirement(totalInitialPledge.Neg())
+		unlockedBalance = big.Add(unlockedBalance, totalInitialPledge)
+		st.AddInitialPledge(totalInitialPledge.Neg())
 		pledgeDelta = big.Add(totalInitialPledge, penaltyFromVesting).Neg()
+
+		// Use unlocked pledge to pay down outstanding fee debt
+		feeToBurn := st.RepayPartialDebt(unlockedBalance)
+		unlockedBalance = big.Sub(unlockedBalance, feeToBurn) //nolint:ineffassign
+		penalty = big.Add(penalty, feeToBurn)
```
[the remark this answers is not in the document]
ZenGround0: > We should really do the same thing any time funds unlock

This is true but it's out of scope of this issue.  The problem of vested and unlocked PCD not going towards fee debt was introduced with fee debt, not with the accounting cleanup.

> this is worth factoring out

I don't see how to factor this out beyond `RepayParitalFeeDebt` in a way that applies to all four cases.  Open for suggestions.

actors/builtin/miner/miner_actor.go
ZenGround0: will fix with #1026

actors/builtin/miner/miner_state.go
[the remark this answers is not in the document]
ZenGround0: Done with the move to RepayPartialDebtInPriorityOrder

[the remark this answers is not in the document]
ZenGround0: (and the tweak to RepayDebt -- now RepayDebtsOrAbort

actors/builtin/miner/miner_actor.go
```diff
@@ -507,10 +507,10 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 		// available balance already accounts for fee debt so it is correct to call
-		// this before VerifyPledgeRequirementsAndRepayDebts. We would have to
+		// this before RepayDebts. We would have to
 		// subtract fee debt explicitly if we called this after.
 		availableBalance := st.GetAvailableBalance(rt.CurrentBalance())
-		feeToBurn = VerifyPledgeRequirementsAndRepayDebts(rt, &st)
+		feeToBurn = RepayDebts(rt, &st)
```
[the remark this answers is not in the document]
ZenGround0: Did 1 and 3 but planning to address 2 by closing out 1020 separate from this PR unless you block approval on that

actors/builtin/miner/miner_state.go
ZenGround0: Will do this in 1020

actors/builtin/miner/miner_state_test.go
```diff
@@ -761,20 +761,30 @@ func TestSectorNumberAllocation(t *testing.T) {
...
 	harness := constructStateHarness(t, abi.ChainEpoch(0))
 
 	currentBalance := abi.NewTokenAmount(300)
 	fee := abi.NewTokenAmount(1000)
-	_, _, err := harness.s.PenalizeFundsInPriorityOrder(harness.store, abi.ChainEpoch(0), fee, currentBalance)
+	err := harness.s.ApplyPenalty(fee)
 	require.NoError(t, err)
 
+	assert.Equal(t, harness.s.FeeDebt, fee)
+	penaltyFromVesting, penaltyFromBalance, err := harness.s.RepayPartialDebtInPriorityOrder(harness.store, abi.ChainEpoch(0), currentBalance)
```
anorth, line 773: Nit: need more tests here later for non-zero locked funds.

actors/builtin/miner/monies.go
```diff
@@ -126,17 +126,11 @@ func InitialPledgeForPower(qaPower, baselinePower abi.StoragePower, rewardEstima
 // may be slightly lower than the true amount. Computing vesting here would be
 // almost always redundant since vesting is quantized to ~daily units.  Vesting
 // will be at most one proving period old if computed in the cron callback.
-func VerifyPledgeRequirementsAndRepayDebts(rt Runtime, st *State) abi.TokenAmount {
+func RepayDebtsOrAbort(rt Runtime, st *State) abi.TokenAmount {
 	currBalance := rt.CurrentBalance()
-	toBurn, err := st.RepayDebt(currBalance)
+	toBurn, err := st.repayDebts(currBalance)
 	builtin.RequireNoErr(rt, err, exitcode.Unwrap(err, exitcode.ErrIllegalState), "unlocked balance can not repay fee debt")
```
anorth, line 132: This was already here, but the manual `Unwrap` is a bad smell, shouldn't be necessary. `RequireNoErr` already does it.

anorth (approved):
Thanks for this.

A good property that I think I'm observing is that methods now invoke `rt.CurrentBalance` at most once, which means they are less likely to accidentally be failing to take to-be-burnt funds into account. To that end, the call hidden inside `RepayDebtsOrAbort` may be better pulled out to an argument.

actors/builtin/miner/miner_actor.go
```diff
@@ -1673,7 +1686,9 @@ func handleProvingDeadline(rt Runtime) {
 		{
 			depositToBurn, err := st.ExpirePreCommits(store, currEpoch)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expire pre-committed sectors")
-			penaltyTotal = big.Add(penaltyTotal, depositToBurn)
+
+			err = st.ApplyPenalty(depositToBurn)
+			builtin.RequireNoErr(rt, err, exitcode.Unwrap(err, exitcode.ErrIllegalState), "failed to apply penalty")
```
wadeAlexC, line 1691: This is never explicitly burned, just added to `FeeDebt`. The var name suggests it should be explicitly burned, but I think just applying it works?

wadeAlexC: Yeah, PCD is decreased by the amount added to FeeDebt, so it stands to reason that this'll be burned. Nice.

actors/builtin/miner/miner_actor.go
```diff
@@ -590,6 +593,8 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 	})
 
 	burnFunds(rt, feeToBurn)
+	rt.State().Readonly(&st)
+	st.AssertBalanceInvariants(rt.CurrentBalance())
```
wadeAlexC, line 597: Nice, but there's now an extra `AssertBalanceInvariants` about 30 lines up.

actors/builtin/miner/miner_actor.go
```diff
@@ -1701,13 +1716,14 @@ func handleProvingDeadline(rt Runtime) {
...
-				penaltyTotal = big.Sum(penaltyTotal, penaltyFromVesting, penaltyFromBalance)
-				pledgeDeltaTotal = big.Sub(pledgeDeltaTotal, penaltyFromVesting)
-			}
+
+			err = st.ApplyPenalty(penaltyTarget)
+			builtin.RequireNoErr(rt, err, exitcode.Unwrap(err, exitcode.ErrIllegalState), "failed to apply penalty")
+
+			penaltyFromVesting, penaltyFromBalance, err := st.RepayPartialDebtInPriorityOrder(store, currEpoch, rt.CurrentBalance())
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
+			penaltyTotal = big.Add(penaltyFromVesting, penaltyFromBalance)
```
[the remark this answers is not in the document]
wadeAlexC: The reason this is `big.Add` is because (IMO) `penaltyTotal` is misnamed - it's actually the amount to burn, rather than the total amount penalized. See a similar misnomer above, where `depositToBurn` is never explicitly burned, but rather added to `FeeDebt`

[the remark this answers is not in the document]
ZenGround0: As @wadeAlexC is getting at I explicitly changed from Sum to Add for clarity.  This should be the only place funds to burn are incremented.  The design from #1026 means that calls to `ApplyPenalty` will capture any additional penalizations in the state's fee debt.

I can change the name from penaltyTotal to penaltyToBurn

actors/builtin/miner/miner_state_test.go
ZenGround0: Tracked in #966

ZenGround0:
Follow up work tracked in #1032

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
KRS246269: هلا

KRS246269:
اهلين

PR #1009, opened 2020-08-19
Refactor deadline handling

### LINKED ISSUE #983: Miner: reduce operations performed in handleProvingDeadline where possible
Each Miner represents at least 48 `CronEventProvingDeadline` callbacks per-day. This callback executes `Miner.handleProvingDeadline`, which performs many operations that may not always be necessary. Especially for small / newly-created miners, this should be avoided to circumvent potential DoS vectors.

#### Recommendation

`handleProvingDeadline` should implement short-circuit logic where possible. Here are some possibilities:

* `len(st.AllocatedSectors) == 0`: the entire method may be skipped
* `st.InitialPledgeRequirement == 0`: most of the method may be skipped. precommit expiries should still be handled.
* `st.PreCommitDeposits == 0`: precommit expiries can be skipped
* If the current deadline has no proven sectors, most of the method can be skipped

Stebalien:
* Moves most logic into helper functions.
* Avoids loading deadlines, vesting table, etc. unnecessarily.
* Loads the vesting table at most twice. Once to unlock vested funds, once to unlock unvested funds for penalties.

fixes #983

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1655,141 +1655,59 @@ func handleProvingDeadline(rt Runtime) {
 
 	hadEarlyTerminations := false
 
-	powerDelta := PowerPair{big.Zero(), big.Zero()}
+	powerDeltaTotal := NewPowerPairZero()
 	penaltyTotal := abi.NewTokenAmount(0)
-	pledgeDelta := abi.NewTokenAmount(0)
+	pledgeDeltaTotal := abi.NewTokenAmount(0)
 
 	var st State
 	rt.State().Transaction(&st, func() {
-		var err error
+		penaltyTarget := abi.NewTokenAmount(0)
+
 		{
 			// Vest locked funds.
 			// This happens first so that any subsequent penalties are taken
 			// from locked vesting funds before funds free this epoch.
 			newlyVested, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
-			pledgeDelta = big.Add(pledgeDelta, newlyVested.Neg())
+			pledgeDeltaTotal = big.Add(pledgeDeltaTotal, newlyVested.Neg())
 		}
 
 		{
-			// expire pre-committed sectors
-			expiryQ, err := LoadBitfieldQueue(store, st.PreCommittedSectorsExpiry, st.QuantSpecEveryDeadline())
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sector expiry queue")
-
-			bf, modified, err := expiryQ.PopUntil(currEpoch)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to pop expired sectors")
-
-			if modified {
-				st.PreCommittedSectorsExpiry, err = expiryQ.Root()
-				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save expiry queue")
-			}
-
-			depositToBurn, err := st.checkPrecommitExpiry(store, bf)
+			depositToBurn, err := st.ExpirePreCommits(store, currEpoch)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to expire pre-committed sectors")
-			penaltyTotal = big.Add(penaltyTotal, depositToBurn)
+			penaltyTarget = big.Add(penaltyTarget, depositToBurn)
 		}
 
 		// Record whether or not we _had_ early terminations in the queue before this method.
 		// That way, don't re-schedule a cron callback if one is already scheduled.
 		hadEarlyTerminations = havePendingEarlyTerminations(rt, &st)
 
-		// Note: because the cron actor is not invoked on epochs with empty tipsets, the current epoch is not necessarily
-		// exactly the final epoch of the deadline; it may be slightly later (i.e. in the subsequent deadline/period).
-		// Further, this method is invoked once *before* the first proving period starts, after the actor is first
-		// constructed; this is detected by !dlInfo.PeriodStarted().
-		// Use dlInfo.PeriodEnd() rather than rt.CurrEpoch unless certain of the desired semantics.
-		dlInfo := st.DeadlineInfo(currEpoch)
-		if !dlInfo.PeriodStarted() {
-			return // Skip checking faults on the first, incomplete period.
-		}
-		deadlines, err := st.LoadDeadlines(store)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
-		deadline, err := deadlines.LoadDeadline(store, dlInfo.Index)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %d", dlInfo.Index)
-		quant := dlInfo.QuantSpec()
-		unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
-
 		{
-			// Detect and penalize missing proofs.
-			faultExpiration := dlInfo.Last() + FaultMaxAge
+			penalty, pledgeDelta, powerDelta, err := st.AdvanceDeadline(
+				store, currEpoch,
+				epochReward.ThisEpochRewardSmoothed,
+				pwrTotal.QualityAdjPowerSmoothed,
+			)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to advance deadline")
 
-			partPowerDelta, penalizedPower, err := deadline.ProcessDeadlineEnd(store, quant, faultExpiration)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to process end of deadline %d", dlInfo.Index)
+			penaltyTarget = big.Add(penaltyTarget, penalty)
+			powerDeltaTotal = powerDeltaTotal.Add(powerDelta)
+			pledgeDeltaTotal = big.Add(pledgeDeltaTotal, pledgeDelta)
+		}
 
-			powerDelta = powerDelta.Add(partPowerDelta)
-
-			// Unlock sector penalty for all undeclared faults.
-			penaltyTarget := PledgePenaltyForUndeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, penalizedPower.QA)
-			// Subtract the "ongoing" fault fee from the amount charged now, since it will be added on just below.
-			penaltyTarget = big.Sub(penaltyTarget, PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, penalizedPower.QA))
+		if !penaltyTarget.IsZero() {
+			unlockedBalance := st.GetUnlockedBalance(rt.CurrentBalance())
 			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, unlockedBalance)
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
-			unlockedBalance = big.Sub(unlockedBalance, penaltyFromBalance)
-			penaltyTotal = big.Sum(penaltyTotal, penaltyFromVesting, penaltyFromBalance)
-			pledgeDelta = big.Sub(pledgeDelta, penaltyFromVesting)
-
-		}
-		{
-			// Record faulty power for penalisation of ongoing faults, before popping expirations.
-			// This includes any power that was just faulted from missing a PoSt.
-			penaltyTarget := PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, deadline.FaultyPower.QA)
-			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, unlockedBalance)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
-			unlockedBalance = big.Sub(unlockedBalance, penaltyFromBalance) //nolint:ineffassign
-			penaltyTotal = big.Sum(penaltyTotal, penaltyFromVesting, penaltyFromBalance)
-			pledgeDelta = big.Sub(pledgeDelta, penaltyFromVesting)
-		}
-		{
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +991,140 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
...
+		networkQAPowerEstimate,
+		totalFaultyPower.QA,
+	)
+	// Compute the overlap so we can avoid double-charging undeclared faulty
+	// power as declared faulty power.
+	overlapPenalty := PledgePenaltyForDeclaredFault(
+		rewardEstimate,
+		networkQAPowerEstimate,
+		undeclaredFaultyPower.QA,
+	)
```
Stebalien, line 1123: I'd prefer to have compute `undeclaredFaultyPower` and `declaredFaultyPower` (`totalFaultyPower-undeclaredFaultyPower`) up-front, then compute two penalties. However, that changes rounding. Was there a reason for doing this one way versus the other?

actors/builtin/miner/miner_actor.go
```diff
@@ -1653,141 +1653,59 @@ func handleProvingDeadline(rt Runtime) {
...
 	var st State
 	rt.State().Transaction(&st, func() {
-		var err error
+		penaltyTarget := abi.NewTokenAmount(0)
+
 		{
 			// Vest locked funds.
 			// This happens first so that any subsequent penalties are taken
 			// from locked vesting funds before funds free this epoch.
 			newlyVested, err := st.UnlockVestedFunds(store, rt.CurrEpoch())
```
Stebalien, line 1668: This is a no-op now if there are no vesting funds.

actors/builtin/miner/miner_actor.go
```diff
@@ -1653,141 +1653,59 @@ func handleProvingDeadline(rt Runtime) {
...
-			bf, modified, err := expiryQ.PopUntil(currEpoch)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to pop expired sectors")
-
-			if modified {
-				st.PreCommittedSectorsExpiry, err = expiryQ.Root()
-				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save expiry queue")
-			}
-
-			depositToBurn, err := st.checkPrecommitExpiry(store, bf)
+			depositToBurn, err := st.ExpirePreCommits(store, currEpoch)
```
Stebalien, line 1674: This is a no-op now if there are no precommits (well, if there is no precommit deposit). Can we assume that pre-commits always have non-zero deposits?

actors/builtin/miner/miner_actor.go
```diff
@@ -1653,141 +1653,59 @@ func handleProvingDeadline(rt Runtime) {
...
-			// Unlock sector penalty for all undeclared faults.
-			penaltyTarget := PledgePenaltyForUndeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, penalizedPower.QA)
-			// Subtract the "ongoing" fault fee from the amount charged now, since it will be added on just below.
-			penaltyTarget = big.Sub(penaltyTarget, PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, penalizedPower.QA))
-			penaltyFromVesting, penaltyFromBalance, err := st.PenalizeFundsInPriorityOrder(store, currEpoch, penaltyTarget, unlockedBalance)
-			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty")
-			unlockedBalance = big.Sub(unlockedBalance, penaltyFromBalance)
-			penaltyTotal = big.Sum(penaltyTotal, penaltyFromVesting, penaltyFromBalance)
-			pledgeDelta = big.Sub(pledgeDelta, penaltyFromVesting)
+			penalty, pledgeDelta, powerDelta, err := st.AdvanceDeadline(
```
Stebalien, line 1684: This will short-circuit if:

* The period hasn't started.
* There is no "initial pledge" (no sectors). Can we assume that every sector will have a non-zero initial pledge?
* We load the deadline and find it has no sectors.

actors/builtin/miner/miner_actor.go
anorth: Hmm, it's not my favourite thing. PCD are zero if the block reward expectation is zero, which will happen one day. I think we should load the queue.

actors/builtin/miner/miner_actor.go
anorth: As for PCD, I am not comfortable coupling a test for the existence of any sectors to the presence of pledge.

anorth: IMO this has gone too far in pushing things down into the state. I do like the manipulations being pushed there, but do not think that "policy" logic like how much penalty to pay should be there.

actors/builtin/miner/miner_state.go
```diff
@@ -918,9 +929,32 @@ func (st *State) AddPreCommitExpiry(store adt.Store, expireEpoch abi.ChainEpoch,
 	return nil
 }
 
-func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (depositToBurn abi.TokenAmount, err error) {
+func (st *State) ExpirePreCommits(store adt.Store, currEpoch abi.ChainEpoch) (depositToBurn abi.TokenAmount, err error) {
 	depositToBurn = abi.NewTokenAmount(0)
 
+	// If we don't have any deposit, we don't have any pre-commits.
+	if !st.PreCommitDeposits.IsZero() {
```
anorth, line 936: 👎

actors/builtin/miner/miner_state.go
anorth: No, computing power first is preferable.

anorth (changes requested):
I would prefer to revert most of the `AdvanceDeadline` refactor. There may be small pure manipulations that could be pushed down, but it shouldn't take parameters like the reward estimate or return values like penalty amounts.

actors/builtin/miner/miner_actor.go
Stebalien: SGTM.

Stebalien: Yeah, you're right. I've moved fee calculation back up.

Stebalien:
> I would prefer to revert most of the AdvanceDeadline refactor. There may be small pure manipulations that could be pushed down, but it shouldn't take parameters like the reward estimate or return values like penalty amounts.

Most of the `AdvanceDeadline` refactor doesn't actually have anything to do with penalty, just power changes. Let me know what you think about it now.

actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +985,133 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
...
+// - Returns the changes to power & pledge, and faulty power (both declared and undeclared).
+func (st *State) AdvanceDeadline(store adt.Store, currEpoch abi.ChainEpoch) (*AdvanceDeadlineResult, error) {
+	pledgeDelta := abi.NewTokenAmount(0)
+	powerDelta := NewPowerPairZero()
+
+	undeclaredFaultyPower := NewPowerPairZero()
+	declaredFaultyPower := NewPowerPairZero()
+
+	// Note: because the cron actor is not invoked on epochs with empty tipsets, the current epoch is not necessarily
+	// exactly the final epoch of the deadline; it may be slightly later (i.e. in the subsequent deadline/period).
```
anorth, line 1006: This comment is now incorrect.

actors/builtin/miner/miner_actor.go
```diff
@@ -1652,141 +1652,68 @@ func handleProvingDeadline(rt Runtime) {
...
-			powerDelta = powerDelta.Add(partPowerDelta)
+			result, err := st.AdvanceDeadline(
+				store, currEpoch,
+			)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to advance deadline")
 
-			// Unlock sector penalty for all undeclared faults.
-			penaltyTarget := PledgePenaltyForUndeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, penalizedPower.QA)
-			// Subtract the "ongoing" fault fee from the amount charged now, since it will be added on just below.
-			penaltyTarget = big.Sub(penaltyTarget, PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, penalizedPower.QA))
```
anorth, line 1721: This "subtract FF" comment and line are really important, and I can't see where the corresponding logic is in the new version.

actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +985,133 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
 	return depositToBurn, nil
 }
 
+type AdvanceDeadlineResult struct {
+	PledgeDelta                                abi.TokenAmount
+	PowerDelta                                 PowerPair
+	UndeclaredFaultyPower, DeclaredFaultyPower PowerPair
```
anorth, line 991: It took me quite a while to figure out how (I think) these are being used. I think:
- UndeclaredFaultyPower is the faults triggered by missing posts
- DeclaredFaultyPower was not just declared here, but is the ongoing faulty power to be penalised

I.e. one is an "edge" and one is a "gauge", just reading a value from state. "Declared" is misleading here, since nothing is being declared and also it includes power from faults that were both previously, and in this call, undeclared.

Maybe names could help, but it definitely needs explanation. Maybe
- `DetectedFaultyPower` sum of power of sectors from partitions which were not proven this deadline
- `OngoingFaultyPower` sum of power of sectors that are (or were) faulty

Does the latter include the former? Maybe `PreviouslyFaultyPower` or `AlreadyFaultyPower` then.

actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +985,133 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
...
+	quant := dlInfo.QuantSpec()
+	{
+		// Detect and penalize missing proofs.
+		faultExpiration := dlInfo.Last() + FaultMaxAge
+
+		powerDelta, undeclaredFaultyPower, err = deadline.ProcessDeadlineEnd(store, quant, faultExpiration)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to process end of deadline %d: %w", dlInfo.Index, err)
+		}
+		declaredFaultyPower = deadline.FaultyPower.Sub(undeclaredFaultyPower)
```
anorth, line 1056: I don't understand this, or at least it's appearance here. No "declared faults" can happen while processing missed PoSts. In combination with my confusion above, this line would make a bit more sense at the bottom of this method. But if it's just trying to count the faulty power before the new faults were added .... just read `deadline.FaultyPower` before processing the new faults?

actors/builtin/miner/miner_test.go
```diff
@@ -1207,7 +1207,7 @@ func TestWindowPost(t *testing.T) {
 		actor.submitWindowPoSt(rt, dlinfo, partitions, infos, cfg)
 
 		// expect declared fee to be charged during cron
-		dlinfo = advanceDeadline(rt, actor, &cronConfig{ongoingFaultsPenalty: declaredFee})
+		dlinfo = advanceDeadline(rt, actor, &cronConfig{declaredFaultsPenalty: declaredFee})
```
anorth, line 1210: I see some of my earlier confusion now. Here, the original is better IMO. Nothing is being declared, and some of the power may have been an undeclared fault.

actors/builtin/miner/miner_test.go
```diff
@@ -1526,16 +1526,14 @@ func TestDeadlineCron(t *testing.T) {
...
 		retractedPwr := miner.PowerForSectors(actor.sectorSize, allSectors[1:])
 		retractedPenalty := miner.PledgePenaltyForUndeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, retractedPwr.QA)
-		// subtract ongoing penalty, because it's charged below (this prevents round-off mismatches)
-		retractedPenalty = big.Sub(retractedPenalty, miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, retractedPwr.QA))
 
 		// Un-recovered faults are charged as ongoing faults
-		ongoingPwr := miner.PowerForSectors(actor.sectorSize, allSectors)
-		ongoingPenalty := miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, ongoingPwr.QA)
+		declaredPwr := miner.PowerForSectors(actor.sectorSize, allSectors[:1])
+		declaredPenalty := miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, declaredPwr.QA)
```
anorth, line 1532: Ditto, this rename confuses me.

actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +985,134 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
...
+	quant := dlInfo.QuantSpec()
+	{
+		// Detect and penalize missing proofs.
+		faultExpiration := dlInfo.Last() + FaultMaxAge
+
+		powerDelta, detectedFaultyPower, err = deadline.ProcessDeadlineEnd(store, quant, faultExpiration)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to process end of deadline %d: %w", dlInfo.Index, err)
+		}
+		totalFaultyPower = deadline.FaultyPower
```
anorth, line 1057: I would prefer to do away with this `totalFaultyPower` var completely, and just use `deadline.FaultyPower` in the return value.

actors/builtin/miner/miner_state.go
```diff
@@ -957,6 +985,134 @@ func (st *State) checkPrecommitExpiry(store adt.Store, sectors abi.BitField) (de
...
+		if err != nil {
+			return nil, xerrors.Errorf("failed to count early terminations: %w", err)
+		}
+		if !noEarlyTerminations {
+			st.EarlyTerminations.Set(dlInfo.Index)
+		}
+
+		// The termination fee is paid later, in early-termination queue processing.
+		// We could charge at least the undeclared fault fee here, which is a lower bound on the penalty.
+		// https://github.com/filecoin-project/specs-actors/issues/674
```
anorth, line 1089: Nit: remove this comment, I don't think it's worthwhile any more.

anorth (approved):

actors/builtin/miner/miner_state.go
anorth: I made a really bad suggestion here. `deadline.FaultyPower` is subsequently mutated in `PopExpiredSectors`, which causes it to forget about faulty-and-expiring power.

PR #1013, opened 2020-08-20
Dep/update hamt imports

### LINKED ISSUE #995: Move HAMT into filecoin-project and take over ownership
The HAMT implementation in https://github.com/ipfs/go-hamt-ipld/pull/52 is specialized for, and used only by, Filecoin. And within Filecoin, it's used primarily by this repo: the only other use I'm aware of is as the root of the state tree (which perhaps should be in this repo too #117)

@rvagg and @Stebalien have proposed some solid improvements, which we should shepherd in.

Moving the repo will cause a Go package path change which downstream repos will need to integrate.

ZenGround0:
Closes #995 changing hamt import to new filecoin project repo.

### DIFF
--- a/actors/builtin/market/set_multimap.go
+++ b/actors/builtin/market/set_multimap.go
@@ -4,7 +4,7 @@ import (
 	"reflect"
 
 	cid "github.com/ipfs/go-cid"
-	"github.com/ipfs/go-hamt-ipld"
+	"github.com/filecoin-project/go-hamt-ipld"
 	errors "github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


anorth (approved):
Please sort imports so the new `filecoin-project` one is next to the others.

ZenGround0:
This is now only blocking on [this HAMT fix](https://github.com/filecoin-project/go-hamt-ipld/pull/65).  If I point this PR at that go-hamt-ipld branch tests all pass.

PR #1014, opened 2020-08-20
Miner: Do not call activate deals in ConfirmSectorProofsValid if sector has no deals

aarshkshah1992:
For #474 .


aarshkshah1992:
@anorth @Stebalien There is a small problem here because of which the tests are failing. 

We can no longer commit a sector with no deals as `ConfirmSectorProofsValid` will fail if it is called with sectors with no deals. 

However, at the same time,  in `PrecommitSector`, we will fail in the call to `validateReplaceSector` if the sector we want to replace has deals.

So, how do we tests this flow now ?

PR #1017, opened 2020-08-20
Remove ThisEpochReward from ThisEpochRewardReturn, and use smoothed value for conesnsus fault penalty

anorth:
I think using the instantaneous version was an oversight, since all other penalties use the smoothed version.

Removed the now-unused instantaneous version from the return struct to prevent accidental use.


ZenGround0:
@zixuanzh checking this is ok with you

zixuanzh:
thanks for checking, since we can't use the most accurate number which is the exact epoch reward when consensus fault happens, using the smoothed state is fine too from my pov. cc @davidad @nicola

PR #1019, opened 2020-08-20
Test replacing a committed capacity sector twice.

Stebalien:
When a committed capacity sector is replaced by two new sectors, both new sectors should succeed, but the old sector should only be replaced once.

Most of this test is copy/paste, but reduplicating the logic is probably more trouble than it's worth.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -734,7 +734,7 @@ func TestCommitments(t *testing.T) {
 		// Fail to submit PoSt. This means that both sectors will be detected faulty.
 		// Expect the old sector to be marked as terminated.
 		bothSectors := []*miner.SectorOnChainInfo{oldSector, newSector}
-		lostPower := actor.powerPairForSectors(bothSectors[1:]).Neg() // new sector not active yet.
+		lostPower := actor.powerPairForSectors(bothSectors[:1]).Neg() // new sector not active yet.
 		faultPenalty := actor.undeclaredFaultPenalty(bothSectors)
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -865,6 +865,173 @@ func TestCommitments(t *testing.T) {
...
+		upgradeParams1.ReplaceSectorDeadline = dlIdx
+		upgradeParams1.ReplaceSectorPartition = partIdx
+		upgradeParams1.ReplaceSectorNumber = oldSector.SectorNumber
+		upgrade1 := actor.preCommitSector(rt, upgradeParams1)
+
+		// Check new pre-commit in state
+		assert.True(t, upgrade1.Info.ReplaceCapacity)
+		assert.Equal(t, upgradeParams1.ReplaceSectorNumber, upgrade1.Info.ReplaceSectorNumber)
+		// Require new sector's pledge to be at least that of the old sector.
+		assert.Equal(t, oldSector.InitialPledge, upgrade1.PreCommitDeposit)
```
ZenGround0, line 908: Not your bug but the comment is incorrect and its hiding that something is wrong.  The assertion captures the correct invariant, new sectors PCD is at least old initial pledge.  Looking at the code I don't think we enforce this requirement on the upgrade's IP.  I think this is different than the expected behavior.  I will file an issue.

Stebalien: I _think_ it's implicit because the new sector will live longer, but I'm not sure as pledge requirements change over time...

ZenGround0: Discussion in #1021

PR #1024, opened 2020-08-21
Miner: Validate sector params duing confirm

### LINKED ISSUE #913: Miner: Drop precommits from prove commit set if power/pledge/reward values aren't sane
https://github.com/filecoin-project/specs-actors/blob/8735710c0586c16054724aea77808860531096e5/actors/builtin/miner/miner_actor.go#L671-L677

Each of these values (`duration`, `power`, `dayReward`, `storagePledge`, `initialPledge`) should be sanity-checked before a sector is proven. **At a minimum, none of these values should be zero**, but each of them probably also have some additional constraints. For example,  `duration` should be sanity checked against minimum sector lifetime (see #795).

I'm not sure what bounds checks may apply to `dayReward`, `storagePledge`, and `initialPledge` - but I suspect these have at least some trivial relationships - like `storagePledge > dayReward`.

aarshkshah1992:
Closes #913 .

@Stebalien I do not understand these values enough to determine invalid equations among them. Please can you give some ideas ?

Also, do we really need to move this code (the call to `validatePreCommitSectorParams`) to the above loop where we call `ActivateDeals`? I think it's a bigger refactor and this is fine as it is.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -775,6 +775,10 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 			initialPledge := InitialPledgeForPower(power, rewardStats.ThisEpochBaselinePower, pwrTotal.PledgeCollateral,
 				rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, circulatingSupply)
 
+			if !validatePreCommitSectorParams(rt, precommit.Info.SectorNumber, power, dayReward, storagePledge, initialPledge) {
+				continue
+			}
+
 			totalPrecommitDeposit = big.Add(totalPrecommitDeposit, precommit.PreCommitDeposit)
 			totalPledge = big.Add(totalPledge, initialPledge)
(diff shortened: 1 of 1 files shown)


anorth:
@wadealexC I have objections to most of these, I think it needs a little more thought (sorry @aarshkshah1992).

If we deterministically drop a proof here, the miner will lose their pre-commit deposit, which is a big deal. This should only happen when it's the miner's fault, not our fault. These checks are all defensive, seemingly against programming errors. That's fine, but any check that could be instead made at pre-commit time should be made there. Once we accept the pre-commit, we must accept the proof when it comes. If some error means that, say, the sector comes online with no power, it's still much better for us to let that happen than to deny the sector (also no power) _and_ fine the miner for it.

We can't make perfect checks at pre-commit time because we don't know the activation epoch, but if something significant changes between pre-commit and proof that would cause an unexpected result, I think we have to accept it. 

Retaining the log warnings would still make sense though.

@aarshkshah1992 how many of these are reasonable to move to pre-commit time, without loading any more state?

wadeAlexC:
Totally valid concerns. I wish activation epoch could be determined at precommit, but I think that change may be too much at this point.

ZenGround0:
Time to drop this one

PR #1030, opened 2020-08-25
miner capacity upgrade scenario test

acruikshank:
### Motivation

We want to integration test more of the actor code, and a capacity upgrade for a sector containing deals is a good way to hit a lot of different actor functionality including the verified registry and market actor

### Proposed Changes

1. Add `TestReplaceCommittedCapacitySectorWithDealLadenSector`
2. Move scenario tests into their own files named after flow under test instead of primary actor.
3. Add public `GetClaim` method to `power.State` to permit retrieval power.
4. Various small changes to test vm to ease testing.
5. Additions to vm testing to advance till a sectors proving period and methods to inspect power state.

cc @wadeAlexC

### DIFF
--- a/actors/test/committed_capacity_scenario_test.go
+++ b/actors/test/committed_capacity_scenario_test.go
@@ -1,4 +1,4 @@
-package test
+package test_test
 
 import (
 	"context"
--- a/actors/test/committed_capacity_scenario_test.go
+++ b/actors/test/committed_capacity_scenario_test.go
@@ -342,5 +342,11 @@ func publishDeal(t *testing.T, v *vm.VM, provider, dealClient, minerID addr.Addr
 		})
 	}
 
+	vm.ExpectInvocation{
+		To:             builtin.StorageMarketActorAddr,
+		Method:         builtin.MethodsMarket.PublishStorageDeals,
+		SubInvocations: expectedPublishSubinvocations,
+	}.Matches(t, v.LastInvocation())
+
 	return ret.(*market.PublishStorageDealsReturn)
 }
--- a/actors/test/committed_capacity_scenario_test.go
+++ b/actors/test/committed_capacity_scenario_test.go
@@ -46,10 +46,6 @@ func TestReplaceCommittedCapacitySectorWithDealLadenSector(t *testing.T) {
 	minerAddrs, ok := ret.(*power.CreateMinerReturn)
 	require.True(t, ok)
 
-	// advance vm so we can have seal randomness epoch in the past
-	v, err := v.WithEpoch(200)
-	require.NoError(t, err)
-
 	//
 	// Precommit, prove and PoSt empty sector (more fully tested in TestCommitPoStFlow)
 	//
@@ -71,7 +67,7 @@ func TestReplaceCommittedCapacitySectorWithDealLadenSector(t *testing.T) {
 	v, _ = vm.AdvanceByDeadlineTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
 
 	// Prove commit sector after max seal duration
-	v, err = v.WithEpoch(proveTime)
+	v, err := v.WithEpoch(proveTime)
 	require.NoError(t, err)
 	proveCommitParams := miner.ProveCommitSectorParams{
 		SectorNumber: sectorNumber,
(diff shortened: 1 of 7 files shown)


actors/test/committed_capacity_scenario_test.go
```diff
@@ -0,0 +1,352 @@
+package test
```
anorth, line 1: test_test?

actors/test/committed_capacity_scenario_test.go
```diff
@@ -0,0 +1,352 @@
...
+		Peer:          abi.PeerID("not really a peer id"),
+	}
+	ret, code := v.ApplyMessage(addrs[0], builtin.StoragePowerActorAddr, minerBalance, builtin.MethodsPower.CreateMiner, &params)
+	require.Equal(t, exitcode.Ok, code)
+
+	minerAddrs, ok := ret.(*power.CreateMinerReturn)
+	require.True(t, ok)
+
+	// advance vm so we can have seal randomness epoch in the past
+	v, err := v.WithEpoch(200)
```
anorth, line 50: This shouldn't be necessary: negative epoch randomness is still well defined.

actors/test/committed_capacity_scenario_test.go
```diff
@@ -0,0 +1,352 @@
...
+	//
+
+	// precommit sector
+	preCommitParams := miner.SectorPreCommitInfo{
+		SealProof:     sealProof,
+		SectorNumber:  sectorNumber,
+		SealedCID:     sealedCid,
+		SealRandEpoch: v.GetEpoch() - 1,
+		DealIDs:       nil,
+		Expiration:    v.GetEpoch() + miner.MinSectorExpiration + miner.MaxProveCommitDuration[sealProof] + 100,
```
anorth, line 64: If this expiration isn't that important, just use a big round number.

actors/test/committed_capacity_scenario_test.go
```diff
@@ -0,0 +1,352 @@
...
+		ReplaceSectorPartition: pIdx,
+		ReplaceSectorNumber:    sectorNumber,
+	}
+
+	// prove commit, verify and PoSt
+	_, code = v.ApplyMessage(addrs[0], minerAddrs.RobustAddress, big.Zero(), builtin.MethodsMiner.PreCommitSector, &preCommitParams)
+	require.Equal(t, exitcode.Ok, code)
+
+	// assert successful precommit invocation
+	none := []vm.ExpectInvocation{}
```
anorth, line 180: Can you use `nil`? I'm guessing `nil` is interpreted as "don't care", but I think that would be fine here.

support/vm/testing.go
```diff
@@ -33,6 +33,15 @@ import (
 )
 
 var FIL = big.NewInt(1e18)
+var VerifregRoot address.Address
+
+func init() {
+	var err error
+	VerifregRoot, err = address.NewIDAddress(80)
```
anorth, line 40: Could you please follow up to put this in `singletons.go` too?

support/vm/testing.go
```diff
@@ -275,6 +309,24 @@ func AdvanceByDeadlineTillIndex(t *testing.T, v *VM, minerIDAddr address.Address
 	})
 }
 
+// Advance to the epoch when the sector is due to be proven.
+// Returns the deadline info for proving deadline for sector, partition index of sector, and a VM at the opening of
+// the deadline (ready for SubmitWindowedPoSt).
+func AdvanceTillProvingPeriod(t *testing.T, v *VM, minerIDAddress address.Address, sectorNumber abi.SectorNumber) (*miner.DeadlineInfo, uint64, *VM) {
```
anorth, line 315: `AdvanceTillProvingDeadline`

anorth (approved):
This is good.

I think we should pause after this to write some helpers, especially around common-case params construction, to reduce noise. If we make a stateful component (i.e. that knows it's own address, next sector number etc), with no deps, it could be a component of the miner agent later.

actors/test/committed_capacity_scenario_test.go
acruikshank: I started out leaving subinvocations at the leaves untested, but I wasn't sure whether the leaves were actually leaves. This guarantees there are no unexpected calls between actors.

wadeAlexC:
Nice! Looks like you hit mostly the happy paths. It'd be good to also test the unhappy paths I mentioned - especially the one where deal init timeout results in a `RestoreBytes` invocation. That one's primarily Market actor functionality (`CronTick`), so it'd be good to cover it as well.

support/vm/testing.go
acruikshank: #1031

PR #1033, opened 2020-08-25
Explicitly limit the number of declarations in miner methods

Stebalien:
The total number of operations is implicitly specified in the partition limits, but we might as well limit the number of declarations as well. Otherwise, a miner could force the chain to perform additional work handling inefficiently encoded messages.

part of #416


actors/builtin/miner/miner_actor.go
```diff
@@ -879,8 +879,8 @@ type ExpirationExtension struct {
 // The sector must not be terminated or faulty.
 // The sector's power is recomputed for the new expiration.
 func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpirationParams) *adt.EmptyValue {
-	if uint64(len(params.Extensions)) > AddressedPartitionsMax {
-		rt.Abortf(exitcode.ErrIllegalArgument, "too many declarations %d, max %d", len(params.Extensions), AddressedPartitionsMax)
+	if uint64(len(params.Extensions)) > DeclarationsMax {
```
Stebalien, line 882: (renamed for accuracy)

anorth (approved):

PR #1034, opened 2020-08-25
add some unhappy scenarios for cc upgrade

acruikshank:
follow up to #1030

### Motivation

CC upgrades have a happy-path scenario test. We want to check some of the unhappy scenarios as well.

### Proposed Changes

1. Add test for when a miner fails to prove a sector containing verified deals that the bytes are returned to registry when deals are deactivated.
2. Add test for when a miner misses first PoSt for cc upgrade sector that power for both sectors is removed.
3. Add test for when a miner skips a CC upgrade sector in first PoSt that sector is faulted and power is unchanged.

### DIFF
--- a/actors/test/committed_capacity_scenario_test.go
+++ b/actors/test/committed_capacity_scenario_test.go
@@ -167,8 +167,6 @@ func TestReplaceCommittedCapacitySectorWithDealLadenSector(t *testing.T) {
 		ReplaceSectorPartition: pIdx,
 		ReplaceSectorNumber:    sectorNumber,
 	}
-
-	// prove commit, verify and PoSt
 	_, code = v.ApplyMessage(addrs[0], minerAddrs.RobustAddress, big.Zero(), builtin.MethodsMiner.PreCommitSector, &preCommitParams)
 	require.Equal(t, exitcode.Ok, code)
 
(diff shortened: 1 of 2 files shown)


anorth:
> when a miner misses first PoSt for cc upgrade sector that power for both sectors is removed

The power for the new sector is not added until the first PoSt, so it should not be removed.

I might be confused whether "cc upgrade sector" refers to the upgraded (replaced) or new sector, tho

wadeAlexC:
LGTM. As alex said, power for the new sector isn't added till the first PoSt.

I wonder if it'd be helpful to try and test some of the params for `UpdateClaimedPower` and values for `BurntFunds.MethodSend`. Especially where multiple sectors are concerned, it'd be good to know that these invocations correspond to some specific action.

acruikshank:
I think both new and old sectors are faulting in the same deadline resulting in the miner losing power for the old sector. I'll try to tease them apart to show that the old sector isn't faulted until it misses its own PoSt.

actors/test/committed_capacity_scenario_test.go
```diff
@@ -239,9 +270,96 @@ func TestReplaceCommittedCapacitySectorWithDealLadenSector(t *testing.T) {
...
+	t.Run("miner skips replacing sector in first PoSt", func(t *testing.T) {
+		tv, err := v.WithEpoch(v.GetEpoch()) // create vm copy
+		require.NoError(t, err)
+
+		submitParams = miner.SubmitWindowedPoStParams{
+			Deadline: dlInfo.Index,
+			Partitions: []miner.PoStPartition{{
+				Index: pIdx,
+				// skip cc upgrade
+				Skipped: bitfield.NewFromSet([]uint64{uint64(ccSectorNumber)}),
```
anorth, line 332: The name`ccSectorNumber` is quite confusing here. A committed capacity sector is a sector that has no deals. This, however, is the a sector that has deals that is replacing a committed capacity sector.

Please rename `sectorNumber` and `ccSectorNumber`. Perhaps:
- ccSectorNumber, upgrade/replacementSectorNumber?
- oldSectorNumber, newSectorNumber?
Use those terms consistently in the comments and other code (you already use old/new a lot).

Name `sectorPower` to match.

anorth (approved):

Issue #1035, opened 2020-08-25
multiple cc upgrades to the same sector


acruikshank:
### Problem

As pointed out in #1021, multiple sectors can claim to be a CC upgrade for the same sector. Since the only benefit of a CC upgrade is to allow a sector to expire early, more than one CC upgrade for the same sector does little for the miner. It could however result in an initial pledge and termination fee increase that is unnecessarily high. The questions for this issue:

1. Do we care at all? Miners should be able to avoid replacing the same sector.
2. Do we care enough to fix it? Detecting this condition for sectors proven in the same epoch is relatively easy, but detecting it in multiple sectors will probably require additional miner state.

anorth:
Thanks for opening the discussion. @nicola @zixuanzh do you have opinion here?

My opinion is that miners should be able to avoid it, the expected cost is thus low, and so not worth adding complexity for.

zixuanzh:
I see, we are less concerned about this given that sector upgrades require a reseal. It will be a bigger problem if sector upgrade becomes much cheaper.

PR #1036, opened 2020-08-25
remove requirement for pcd>=replaced.pcd and add one for initialPledge

### LINKED ISSUE #1021: CC Upgrade PCD and IP issues
There are two potential bugs:

1. The precommit deposit of the new sector is lower bounded by the initial pledge of the old sector.  
- Should there be a precommit lower bound based on the old sector?
- If so should it be the old initial pledge, which is greater than or equal to precommit deposit?

2. The initial pledge of the new sector is not impacted by the old sector
- I think it should be lower bounded by the old sector's pledge 
- We'll need to add a field to `SectorPreCommitOnChainInfo` to do this

There is also an underspecification we should think through:

1. If you precommit many sectors as upgrading from one old sector all sectors will be added but only one will really "take the place of the old sector" since only the single old sector will expire.  Should all new sectors inherit constraints from the old sector.  Currently all such sectors will have termination fees impacted by the old sector's age and BR.  After this issue they will probably all have initial pledges lower bounded by the old sector's initial pledge.  Should only one sector inherit this baggage or are we ok giving it to all sectors.

acruikshank:
closes #1021 
work unaddressed in #1035

### Motivation

We currently require a precommit deposit for cc upgrades greater or equal to the initial pledge of the replaced sector. We do not put any requirements on the cc upgrade's initial pledge. This is backwards. PCD should have no bound and initial pledge should have a lower bound of the replaced sector's IP.

### Proposed Changes

1. Remove lower bound for PCD for cc upgrades.
2. Add lower bound for initial pledge for cc upgrades.
3. Fix tests to reflect this.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -552,19 +552,13 @@ func (a Actor) PreCommitSector(rt Runtime, params *SectorPreCommitInfo) *adt.Emp
 			rt.Abortf(exitcode.ErrIllegalState, "sector %v already committed", params.SectorNumber)
 		}
 
-		depositMinimum := big.Zero()
 		if params.ReplaceCapacity {
-			replaceSector := validateReplaceSector(rt, &st, store, params)
-			// Note the replaced sector's initial pledge as a lower bound for the new sector's deposit
-			depositMinimum = replaceSector.InitialPledge
+			validateReplaceSector(rt, &st, store, params)
 		}
 
 		duration := params.Expiration - rt.CurrEpoch()
 		sectorWeight := QAPowerForWeight(info.SectorSize, duration, dealWeight.DealWeight, dealWeight.VerifiedDealWeight)
-		depositReq := big.Max(
-			PreCommitDepositForPower(rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, sectorWeight),
-			depositMinimum,
-		)
+		depositReq := PreCommitDepositForPower(rewardStats.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, sectorWeight)
 		if availableBalance.LessThan(depositReq) {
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -780,8 +774,8 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 				pwrTotal.QualityAdjPowerSmoothed, circulatingSupply)
 
 			totalPrecommitDeposit = big.Add(totalPrecommitDeposit, precommit.PreCommitDeposit)
+			initialPledge, replacedAge, replacedDayReward := replacedSectorParameters(rt, precommit, initialPledge, replacedBySectorNumber)
```
anorth, line 777: This overwrite of `initialPledge` is subtle, and a bit confusing with the name `replacedSectorParameters`. It would be clearer if this just returned the replaced sectors pledge as `replacedPledge`, and the `max()` performed here.

actors/builtin/miner/miner_test.go
```diff
@@ -889,8 +890,6 @@ func TestCommitments(t *testing.T) {
 		// Check new pre-commit in state
 		assert.True(t, upgrade1.Info.ReplaceCapacity)
 		assert.Equal(t, upgradeParams1.ReplaceSectorNumber, upgrade1.Info.ReplaceSectorNumber)
-		// Require new sector's pledge to be at least that of the old sector.
-		assert.Equal(t, oldSector.InitialPledge, upgrade1.PreCommitDeposit)
```
anorth, line 893: Blocking: we're losing important coverage here. The old check isn't right any more, but can you `assert.Equal(t, oldSector.InitialPledge, upgrade1.InitialPledge)` after the prove-commit? Same for upgrade2.

Code earlier on is explicitly manipulating the reward to ensure this property.

Edit: I see code in the `confirmSectorProofsValid` wrapper setting the expectation, but still think we should assert this concrete expectation here. We know exactly what pledge to expect.

anorth (approved):

anorth:
I'll land this so I can build on it.

Issue #1038, opened 2020-08-26
Deal weight should be recalculated when sector is proven


anorth:
Deal weights are calculated in PreCommitSector in order to estimate the sector's power and hence share of block reward in order to set the pre-commit deposit. A weight is `bytes * epochs` and so depends on the expected sector lifetime. We don't know the activation epoch exactly at pre-commit, but the estimate of the current epoch is good enough for setting PCD. The weights are stored in the SectorPreCommitInfo.

These same weights are then used as inputs to the sector power computation in ConfirmSectorProofsValid, but alongside a different duration. Weights calculated for one duration are invalid for a different (shorter) duration: the sum of the weights will come out greater than the sector's total `space x time`.

We already call the market actor to activate the deals with the now-known activation epoch. We must take the deal weight return values and use them for the power calculation.

anorth:
When a sector is in fact full with deals, this causes an assertion failure in `QualityForWeight`:
```
Assert(sectorSpaceTime.GreaterThanEqual(totalDealSpaceTime))
```

anorth:
My diagnosis here was wrong: the deal weight is computed from the deal start epoch, not the current epoch. So the calculation should not change with time. In order for the deal weight to exceed the sector spacetime, a deal's duration and/or size would need to exceed the sector's, which should result in rejection.

I did notice that nothing validates that the deals' total size doesn't exceed the sector's.

- The market actor should receive the sector size and validate that the sectors fit
- The market actor should check that the total deal spacetime fits inside the sector's
- The miner actor should also check that the received spacetime fits inside the sector

PR #1039, opened 2020-08-26
Skip requesting deal weight for no deals; fix confusing deal weight in tests

anorth:
The miner actor harness previously reported an amount of verified deal weight even when a sector has no deals, which confused me a lot when debugging some pre-commit deposit calculations. It now records non-zero deal weight only if the sector has a deal.

The weight reported was an odd amount: the sector size. I fixed this to be a product of sector size and duration, and discovered #1038.

I also removed an unnecessary call to the market actor to compute deal weight for no deals.

- [x] Depends on #1036 for passing tests.

### DIFF
--- a/actors/abi/big/int.go
+++ b/actors/abi/big/int.go
@@ -340,3 +340,7 @@ func (bi *Int) IsZero() bool {
 func (bi *Int) Nil() bool {
 	return bi.Int == nil
 }
+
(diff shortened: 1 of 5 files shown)


anorth:
I hear you. We can definitely lean on scenario tests more - I just need to jump the barrier of writing my own first one :-)

PR #1043, opened 2020-08-26
confirm deal space does not exceed sector size in precommit.

### LINKED ISSUE #1038: Deal weight should be recalculated when sector is proven

acruikshank:
closes #1038 

### Motivation

That the total spacetime of a sector's deals does not exceed the spacetime of the sector is an invariant of the system. We assert this in `QAPowerForWeight` to avoid a nonsensical result. Our market actor deal verification ensures that a miner cannot publish a deal for a sector that exceeds the sector's time bounds (either estimated at precommit, or actual at prove commit). The syscall to compute the unsealed CID must fail if the space of all deals exceeds the sector size (else it would be forced to compute with negative padding), so sector commitment will fail at `ProveCommitSector` if deal space is too large.

It is still possible for a miner to violate the deal spacetime invariant in precommit, however by publishing deals that are too large prior to calling `PreCommitSector`. In this case, our assertion in `QAPowerForWeight` will fail.

The solution in this PR is to return the total deal size along with deal weight and verified deal weight, and have precommit check that it does not exceed sector size. 

### Proposed Changes

1. Return computed deal space from market.VerifyDealsForActivation (necessitating a new field in return type).
2. Add check in `PreCommitSector` to ensure deal space does not exceed sector size.
3. Test.
4. Remove `gen/command-line-arguments` from repo. This build product of `gen/gen.go` doesn't belong in the repo. It's out of date and confusing.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -294,9 +294,14 @@ type ActivateDealsParams struct {
 	SectorExpiry abi.ChainEpoch
 }
 
+type ActivateDealsReturn struct {
+	DealWeight         abi.DealWeight
+	VerifiedDealWeight abi.DealWeight
+}
+
 // Verify that a given set of storage deals is valid for a sector currently being ProveCommitted,
 // update the market's internal state accordingly.
-func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *adt.EmptyValue {
+func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *ActivateDealsReturn {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 	minerAddr := rt.Message().Caller()
 	currEpoch := rt.CurrEpoch()
@@ -304,9 +309,12 @@ func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *adt.Empty
 	var st State
 	store := adt.AsStore(rt)
 
+	ret := &ActivateDealsReturn{}
+
 	// Update deal dealStates.
 	rt.State().Transaction(&st, func() {
-		_, _, err := ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, currEpoch)
+		var err error
+		ret.DealWeight, ret.VerifiedDealWeight, err = ValidateDealsForActivation(&st, store, params.DealIDs, minerAddr, params.SectorExpiry, currEpoch)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to validate dealProposals for activation")
 
 		msm, err := st.mutator(adt.AsStore(rt)).withDealStates(WritePermission).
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -3090,6 +3090,11 @@ type actorHarness struct {
 
 	epochRewardSmooth  *smoothing.FilterEstimate
 	epochQAPowerSmooth *smoothing.FilterEstimate
+
+	precommitDealWeight         abi.DealWeight
+	precommitVerifiedDealWeight abi.DealWeight
+	dealWeight                  abi.DealWeight
+	verifiedDealWeight          abi.DealWeight
```
anorth, line 3097: I was running into this on #1039 too, but determined we should really make these more easily configurable from the  caller, with a `preCommitConfig` or similar. I shied away from the churn initially, but would prefer we go that path than add these here. You I can do it in my PR.

anorth (commented):
Thanks for this. As you identified and we discussed, it looks like my diagnosis of the problem was off. The market actor won't return a different value for the deal weight at activation, and either the weight sum should be less than the sector weight or some deal must start before the sector does, or end after it does, both of which should cause a failure. So ultimately the change to recalculate the weights isn't needed after all.

The changes to market_test are still good, I'd like to preserve them.

I'd also like us to add a defensive check in the miner_actor that will abort if a broken market actor does return deal weights exceeding the sector spacetime. It would signify a bug in our code, but we should abort rather than allow the assertion failure. The miner actor on its own can't ensure this property, so I think it's appropriate that it defend itself against the market actor.

actors/builtin/market/market_actor.go
```diff
@@ -609,45 +611,48 @@ func deleteDealProposalAndState(dealId abi.DealID, states *DealMetaArray, propos
 
 // Validates a collection of deal dealProposals for activation, and returns their combined weight,
 // split into regular deal weight and verified deal weight.
-func ValidateDealsForActivation(st *State, store adt.Store, dealIDs []abi.DealID, minerAddr addr.Address,
-	sectorExpiry, currEpoch abi.ChainEpoch) (big.Int, big.Int, error) {
+func ValidateDealsForActivation(
+	st *State, store adt.Store, dealIDs []abi.DealID, minerAddr addr.Address, sectorExpiry, currEpoch abi.ChainEpoch,
+) (big.Int, big.Int, abi.SectorSize, error) {
```
anorth, line 616: I think `uint64`  is a more appropriate type than `SectorSize`

anorth (approved):

anorth:
Thanks @acruikshank , could you please edit the PR description to reflect the simpler change while landing this?

PR #1049, opened 2020-08-27
Reorganize seal proof metadata

### LINKED ISSUE #1044: Refactor registered proof tables into a data structure
We have a bunch of switch statements inside methods on the registered proof enum for properties like the associated other proof types, sector sizes etc. This disperses information about one proof type widely across the file, and is hard to visually parse.

Re-arrange the per-proof-type information into a struct, and add a mapping of seal proof types to the associated metadata.

FYI @nicola

anorth:
Re-arranged to group the information for each proof type together, but retained the existing accessor methods to reduce churn. 

Closes #1044. FYI @nicola

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -66,10 +66,8 @@ func NewStoragePower(n int64) StoragePower {
 	return big.NewInt(n)
 }
 
-type RegisteredProof = int64
-
-// This ordering, defines mappings to UInt in a way which MUST never change.
-type RegisteredSealProof RegisteredProof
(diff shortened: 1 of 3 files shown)


actors/abi/sector.go
```diff
@@ -66,10 +66,8 @@ func NewStoragePower(n int64) StoragePower {
 	return big.NewInt(n)
 }
 
-type RegisteredProof = int64
```
anorth, line 69: Lotus doesn't use this.

nicola (approved):
Much better and thank you!

PR #1050, opened 2020-08-27
Remove change worker cron and add confirm update worker key method

### LINKED ISSUE #806: Limit Miner.ChangeWorkerAddress to avoid cron event queue load
Currently, this method may be repeatedly invoked by a Miner to append cron events to the Power actor's CronEventQueue.

Consider restricting this method to Miners with some minimum power requirement.

Related: #802

### LINKED ISSUE #982: Miner: Remove ChangeWorkerAddress reliance on cron
`Miner.ChangeWorkerAddress` relies on a cron callback after `WorkerKeyChangeDelay` epochs to commit the pending key change. This should be avoided, as cron is a critically-important system; callbacks should only be added if they are absolutely necessary.

#### Recommendation

There are a few possible ways to implement this. Here's an option that requires relatively few changes:
* If `params.NewWorker` is `nil`, treat the invocation as a "commit" action, and commit the pending change (as long as the appropriate delay has passed).

Resolving this will make #806 redundant.

acruikshank:
closes #982
closes #806

### Motivation

Cron calls are costly. We don't need to be adding them for worker key updates. Instead we can check for worker updates in existing cron calls at deadline end and and provide an additional actor call if more precision is required.

### Proposed Changes

1. Add `miner.ConfirmUpdateWorkerKey` method to immediately enact a worker key change if one is ready to be enacted.
2. Modify `ChangeWorkerAddress` to skip enrolling a cron job to enact the worker change.
3. Modify `ChangeWorkerAddress` to cancel an existing change if a change exists and the worker address is the same as the original.
4. Process PendingWorkerKeys in `provingPeriodUpdate` by check if the effective time has past and updating worker address if so.
5. Remove code related to `commitWorkerKeyChange` cron call.
6. test

cc @wadeAlexC, @nicola , @sternhenri

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -31,7 +31,7 @@ type Runtime = vmr.Runtime
 type CronEventType int64
 
 const (
-	CronEventProvingDeadline CronEventType = iota
+	CronEventProvingDeadline CronEventType = iota + 1
 	CronEventProcessEarlyTerminations
 )
 
@@ -202,9 +202,6 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 				NewWorker:   newWorker,
 				EffectiveAt: rt.CurrEpoch() + WorkerKeyChangeDelay,
 			}
-		} else {
-			// Allow owner to cancel existing update by supplying existing worker address
-			info.PendingWorkerKey = nil
 		}
 
 		err := st.SaveInfo(adt.AsStore(rt), info)
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -196,8 +196,7 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
 		info.ControlAddresses = controlAddrs
 
 		// save newWorker addr key change request
-		// This may replace another pending key change.
-		if newWorker != info.Worker {
+		if newWorker != info.Worker && info.PendingWorkerKey == nil {
 			info.PendingWorkerKey = &WorkerKeyChange{
 				NewWorker:   newWorker,
 				EffectiveAt: rt.CurrEpoch() + WorkerKeyChangeDelay,
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -31,8 +31,7 @@ type Runtime = vmr.Runtime
 type CronEventType int64
 
 const (
-	CronEventWorkerKeyChange CronEventType = iota
-	CronEventProvingDeadline
+	CronEventProvingDeadline CronEventType = iota
 	CronEventProcessEarlyTerminations
 )
 
@@ -70,6 +69,7 @@ func (a Actor) Exports() []interface{} {
 		18:                        a.ChangeMultiaddrs,
 		19:                        a.CompactPartitions,
 		20:                        a.CompactSectorNumbers,
+		21:                        a.ConfirmUpdateWorkerKey,
 	}
 }
 
@@ -177,8 +177,6 @@ type ChangeWorkerAddressParams struct {
 func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams) *adt.EmptyValue {
 	checkControlAddresses(rt, params.NewControlAddrs)
 
-	var effectiveEpoch abi.ChainEpoch
-
 	newWorker := resolveWorkerAddress(rt, params.NewWorker)
 
 	var controlAddrs []addr.Address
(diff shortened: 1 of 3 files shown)


nicola:
It's awesome to see how fast this moved. Thank you for making this happen!

actors/builtin/miner/miner_actor.go
```diff
@@ -188,44 +186,45 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
+		// save newWorker addr key change request
+		// This may replace another pending key change.
+		if newWorker != info.Worker {
+			info.PendingWorkerKey = &WorkerKeyChange{
+				NewWorker:   newWorker,
+				EffectiveAt: rt.CurrEpoch() + WorkerKeyChangeDelay,
 			}
+		} else {
+			// Allow owner to cancel existing update by supplying existing worker address
+			info.PendingWorkerKey = nil
```
sternhenri, line 207: good idea

actors/builtin/miner/miner_actor.go
```diff
@@ -1685,6 +1682,12 @@ func handleProvingDeadline(rt Runtime) {
 			pledgeDeltaTotal = big.Add(pledgeDeltaTotal, newlyVested.Neg())
 		}
 
+		{
+			// Process pending worker change if any
+			info := getMinerInfo(rt, &st)
+			processPendingWorker(info, rt, &st)
+		}
+
```
sternhenri, line 1690: At a high level, when will `processPendingWorker` be called by `handleProvingDeadline` rather than `ConfirmUpdateWorkerKey`?

acruikshank: It's called by `handleProvingDeadline` which is executed at the end of every deadline (30 mins or 48 times a day). This is the same average frequency we would expect for SubmitPoSt if the miner has sectors in every deadline and sends proofs for all partitions in one message.

actors/builtin/miner/miner_actor.go
```diff
@@ -188,44 +186,45 @@ func (a Actor) ChangeWorkerAddress(rt Runtime, params *ChangeWorkerAddressParams
...
-				info.PendingWorkerKey = &WorkerKeyChange{
-					NewWorker:   newWorker,
-					EffectiveAt: effectiveEpoch,
-				}
+		// save newWorker addr key change request
+		// This may replace another pending key change.
+		if newWorker != info.Worker {
+			info.PendingWorkerKey = &WorkerKeyChange{
+				NewWorker:   newWorker,
+				EffectiveAt: rt.CurrEpoch() + WorkerKeyChangeDelay,
```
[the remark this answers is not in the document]
sternhenri: ack, we should simply ensure that miners can easily determine the `EffectiveAt` Epoch when staging the change, given this potentially has security/operational implications for them.

actors/builtin/miner/miner_actor.go
sternhenri: So when would the `ConfirmUpdateWorkerKey` happen?

acruikshank: It's called by `handleProvingDeadline` which is executed at the end of every deadline (30 mins or 48 times a day). This is the same average frequency we would expect for SubmitPoSt if the miner has sectors in every deadline and sends proofs for all partitions in one message.

acruikshank: `ConfirmUpdateWorkerKey` can be called by the miner's owner any time, but it will only have an effect when a pending worker key has been set and the current epoch is at or past it's effective epoch. It allows the miner's owner to spend a little gas to update the key as soon as possible.

sternhenri: I see! If @wadeAlexC's suggestion of quantizing the `EffectiveAt` time below is applied, then the call should be removed. 

Though again, if there is a security issue (ie compromised keys), it's definitely nice for the miner to have the option to do this.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
acruikshank: Just to clarify the existing situation:
The actual delay experienced depends on when the change happens in the deadline. It will be bounded by `WorkerKeyChangeDelay` and `WorkerKeyChangeDelay+WPoStChallengeWindow-1` or `900 - 951` epochs with current parameters. The `ConfirmUpdateWorkerKey` method gives the miner's owner the option of lowering reducing the upper bound to `WorkerKeyChangeDelay` (900) if they don't want to wait for deadline end.

If we quantize `EffectiveAt` there's no use for `ConfirmUpdateWorkerKey`. The timing of the change will be determined solely by the time the change is made within the deadline cycle within the 900-951 range. Is this what we want?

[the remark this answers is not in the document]
sternhenri: I expect current behavior is better...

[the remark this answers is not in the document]
wadeAlexC: That makes sense. I would be fine with either approach.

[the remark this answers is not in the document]
anorth: I think the current approach is better. Quantizing wouldn't make the change happen any sooner, it would just mean that the epoch recorded is the epoch that the change will be automatically applied. This can not be sooner than the "ideal" earliest epoch.

actors/builtin/miner/miner_actor.go
```diff
@@ -31,8 +31,7 @@ type Runtime = vmr.Runtime
 type CronEventType int64
 
 const (
-	CronEventWorkerKeyChange CronEventType = iota
-	CronEventProvingDeadline
+	CronEventProvingDeadline CronEventType = iota
 	CronEventProcessEarlyTerminations
```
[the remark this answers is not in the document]
anorth: Great, the +1 here also lets us avoid changing the values for existing constants, which may have been duplicated by code external to us by now.

actors/builtin/miner/miner_actor.go
anorth: This has broken the prior behaviour of supporting "no change" to the worker key, while submitting an update to the control addresses. The old way didn't support cancelling a worker key change, but I'm not sure it should: cancelling a change (without delay) amounts to getting an instantaneous choice between two worker keys at the epoch before the change takes effect.

Please revert this for now so the rest of this PR isn't hung up on discussion about it. And to save the next person, add a comment explaining that specifying the worker key equal to the current key allows setting the control addreses without interfering with the worker key or any pending change to it.

anorth (approved):
LGTM without the new change-cancellation behaviour. Please do extract that to a separate PR/issue so we can discuss what we want there.

actors/builtin/miner/miner_actor.go
sternhenri: Ooh. This is right @anorth. Good catch.

acruikshank: @sternhenri, @anorth the old behavior already has this instantaneous choice. If a miner had initiated a change to `addr2` 899 epochs ago, the miner could either allow the change to happen next epoch or make a change to `addr3` which would effectively cancel the change for another 900 epochs. 

If we want to avoid canceling, the condition should if a change exists rather than if the change is to the existing address. This means that an incorrect change will take a minimum of 1799 epochs to correct no matter if it's a revert or a change to a third address.

sternhenri: mmh fair enough. Sorry hadn't paged in this exact thing. I expected we prevented any update to workerKey while a pending change was in progress...

Worth picking up in a new issue after merger I reckon

acruikshank: I've pushed a change to prevent altering the pending worker key until the change has been effected. I can revert it. I've added an issue for future work at #1066.

anorth (approved):
Great work @acruikshank, thanks.

PR #1054, opened 2020-08-28
Randomize deadline assignment

### LINKED ISSUE #432: Randomise partition assignment to equally-full deadlines
Randomise the assignment of a partition to a new deadline among equally-full candidates. This randomisation should be seeded from chain randomness. We're seeking a really simple pseudo-random shuffle here, it's not first order security-critical. A Fiestel network may work, but is probably way overkill. A simple seeded PRNG that we can implement directly is probably sufficient.

Stebalien:
This commit randomizes deadline assignment by starting deadline assignment at a random offset.

fixes #432


Stebalien:
We've decided to drop (or, at least punt) this feature.

PR #1058, opened 2020-08-28
Restrict election eligibility on min power

### LINKED ISSUE #887: Add static state methods for miner election eligibility

ZenGround0:
Closes #887

- Election eligibility depends on the power state check for minimum power
- Fix obo in that power state check (zero power should not be eligible)
- Test

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -11,6 +11,7 @@ import (
 	errors "github.com/pkg/errors"
 	xerrors "golang.org/x/xerrors"
 
+	power "github.com/filecoin-project/specs-actors/actors/builtin/power"
 	. "github.com/filecoin-project/specs-actors/actors/util"
 
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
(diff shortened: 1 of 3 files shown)


ZenGround0:
FYI @magik6k `MinerEligibleForElection` should now wrap all state machine election eligibility checks.  This PR introduces power state as a parameter.  I haven't been following what power table lookback you are using for checking min miner power.  In case there is any confusion this function assumes the power state you pass in is coming from that lookback.

Related heads up: another PR just moved min power from ~1 TiB to ~100 TiB

actors/builtin/miner/miner_state.go
```diff
@@ -1110,7 +1111,16 @@ func (st *State) AdvanceDeadline(store adt.Store, currEpoch abi.ChainEpoch) (*Ad
 	}, nil
 }
 
-func MinerEligibleForElection(store adt.Store, mSt *State, thisEpochReward abi.TokenAmount, currEpoch abi.ChainEpoch) (bool, error) {
```
anorth, line 1113: Nit: this doesn't really belong with the miner actor any more, though I won't object much if you land it here.

I think it belongs directly under the `builtin` package, either in one of the existing files like `shared.go`, `network.go`, or a new one for node state utility functions.

anorth (approved):

actors/builtin/miner/miner_state.go
ZenGround0: Filed in #1061 since this is something we can do post mainnet launch no problem.  I would have just cut and pasted but I remember from trying to write tests without the miner actor unit test harness that there is a bit of extra state juggling that isn't worth it right now.

PR #1073, opened 2020-09-01
swap order of batch proof verifies and deferred cron events

### LINKED ISSUE #909: Power: Process batch proof verifications before deferred cron events
From OnEpochTickEnd: https://github.com/filecoin-project/specs-actors/blob/0f9f59a4a6c302014a4dddc12fdff4effe409885/actors/builtin/power/power_actor.go#L200-L201

This would also involve some changes to the Miner actor. For example, precommit expiry bound could remove the additional "+1": https://github.com/filecoin-project/specs-actors/blob/0f9f59a4a6c302014a4dddc12fdff4effe409885/actors/builtin/miner/miner_actor.go#L516-L520

acruikshank:
closes #909

### Motivation

Sectors replaced by cc upgrades should terminate at the end of their proving deadline. This works unless the sector is replaced at the end of its proving period deadline (i.e. it is replaced and expected to be terminated in the same epoch). This due to the ordering of cron events triggered in power actor. The sector is flagged for termination in batch proof verifies, but that is run after `handleProvingPeriod` that actually processes all the pending terminations. Swapping the order of batch proof verifies and deferred cron events are triggered corrects the problem.

### Proposed Changes

1. Swap order of `processBatchProofVerifies()` and `processDeferredCronEvents()` calls in power actor `OnEpochTickEnd`.
2. Add scenario test that demonstrates the problem and its correction.

cc @wadeAlexC

### DIFF
--- a/actors/test/cron_catches_expiries_scenario_test.go
+++ b/actors/test/cron_catches_expiries_scenario_test.go
@@ -18,6 +18,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var fakeChainRandomness = []byte("not really random")
+
 func TestCronCatchedCCExpirationsAtDeadlineBoundary(t *testing.T) {
 	ctx := context.Background()
 	v := vm.NewVMWithSingletons(ctx, t)
@@ -36,7 +38,7 @@ func TestCronCatchedCCExpirationsAtDeadlineBoundary(t *testing.T) {
 		SealProofType: sealProof,
 		Peer:          abi.PeerID("not really a peer id"),
 	}
-	ret, code := v.ApplyMessage(addrs[0], builtin.StoragePowerActorAddr, minerBalance, builtin.MethodsPower.CreateMiner, &params)
+	ret, code := v.ApplyMessage(worker, builtin.StoragePowerActorAddr, minerBalance, builtin.MethodsPower.CreateMiner, &params)
 	require.Equal(t, exitcode.Ok, code)
 
 	minerAddrs, ok := ret.(*power.CreateMinerReturn)
@@ -55,7 +57,7 @@ func TestCronCatchedCCExpirationsAtDeadlineBoundary(t *testing.T) {
 	require.Equal(t, exitcode.Ok, code)
 
 	// advance time to max seal duration
-	proveTime := v.GetEpoch() + miner.MaxProveCommitDuration[sealProof]
+	proveTime := v.GetEpoch() + miner.PreCommitChallengeDelay + 1
 	v, _ = vm.AdvanceByDeadlineTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
 
 	// Prove commit sector after max seal duration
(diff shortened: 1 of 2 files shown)


actors/test/cron_catches_expiries_scenario_test.go
```diff
@@ -0,0 +1,160 @@
...
+	sealProof := abi.RegisteredSealProof_StackedDrg32GiBV1
+
+	// create miner
+	params := power.CreateMinerParams{
+		Owner:         worker,
+		Worker:        worker,
+		SealProofType: sealProof,
+		Peer:          abi.PeerID("not really a peer id"),
+	}
+	ret, code := v.ApplyMessage(addrs[0], builtin.StoragePowerActorAddr, minerBalance, builtin.MethodsPower.CreateMiner, &params)
```
ZenGround0, line 39: nit: already assigned to `worker` so use for clarity

actors/test/cron_catches_expiries_scenario_test.go
```diff
@@ -0,0 +1,160 @@
...
+	submitParams := miner.SubmitWindowedPoStParams{
+		Deadline: dlInfo.Index,
+		Partitions: []miner.PoStPartition{{
+			Index:   pIdx,
+			Skipped: bitfield.New(),
+		}},
+		Proofs: []abi.PoStProof{{
+			PoStProof: abi.RegisteredPoStProof_StackedDrgWindow32GiBV1,
+		}},
+		ChainCommitRand: []byte("not really random"),
```
ZenGround0, line 86: This should be factored out into a constant "FakeVMRandomness" to prevent confusion and breakage if the value ever changes.

actors/test/cron_catches_expiries_scenario_test.go
```diff
@@ -0,0 +1,160 @@
...
+		ReplaceCapacity:        true,
+		ReplaceSectorDeadline:  dlInfo.Index,
+		ReplaceSectorPartition: pIdx,
+		ReplaceSectorNumber:    sectorNumber,
+	}
+	_, code = v.ApplyMessage(addrs[0], minerAddrs.RobustAddress, big.Zero(), builtin.MethodsMiner.PreCommitSector, &preCommitParams)
+	require.Equal(t, exitcode.Ok, code)
+
+	// advance time to end of original sector's first proving deadline after minimum prove time has past
+	proveTime = v.GetEpoch() + miner.PreCommitChallengeDelay + 1
```
ZenGround0, line 125: Above you use `proveTime := v.GetEpoch() + miner.MaxProveCommitDuration[sealProof]` which seems right.  Why is this delay between precommit and prove commit different?

anorth (approved):

actors/test/cron_catches_expiries_scenario_test.go
acruikshank: The first advances to the end of the prove commit time range. That's a little arbitrary. This one advances to the beginning of the prove commit range, then advances till the next proving deadline of the original sector, which should still be in the proving period range. Prove commit of the replacing sector at the last epoch in the original sector's proving deadline is what we need to test.

I'm going to switch the first prove commit to be the minimum and augment the comment for this advancement.

acruikshank:
@ZenGround0 The replaced sector will still be in the deadline's expiration epochs, so it should get popped off and processed the next proving period. I haven't verified this, though.

PR #1077, opened 2020-09-01
Feat/penalize all the way

### LINKED ISSUE #1071: Propagate penalty > reward from reward to miner actor
From @Kubuxu: 
> @anorth @zenground0 We need to figure out how to not cap the Penalty to block reward (in other words 0) that is passed to Reward actor.
Currently if miner gets a lot of penalties they just get block reward of 0. This is really bad in some cases.
> If message FeeCap is too low then the miner pays the difference: BaseFee - message.FeeCap
> This means that if block reward is around 7FIL, and Base Fee is higher than 7e-10 it is worth for the miner to wipe out their block reward and include all of their messages.
In essence, it caps network fee at blockReward/10e9 for miners because then the whole reward is wiped, and they don't pay anything more.

The offending line is the penalty cap in the reward actor [here](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/reward/reward_actor.go#L100)

In order to penalize the full amount I am proposing we send the penalty and reward over to the miner actor and do a full ordered penalization dropping funds from the miner's account up to imposing fee debt.

1. `AddLockedFunds` --> `ApplyRewards`
2. Add a penalty field to the ApplyRewards params
3. ApplyPenalty(penalty) after adding locked funds
4. RepayPartialDebtInPriorityOrder at the end of method's state transaction

ZenGround0:
Closes #1071

Reward actor does not deduct penalty and cap at reward but sends over to miner actor which applys the penalty to the miner state fee debt

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1441,8 +1441,12 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *adt.
 		pledgeDeltaTotal = big.Sub(pledgeDeltaTotal, newlyVested)
 		pledgeDeltaTotal = big.Add(pledgeDeltaTotal, params.Reward)
 
+		// If the miner incurred block mining penalties charge these to miner's fee debt
 		err = st.ApplyPenalty(params.Penalty)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to apply penalty")
+		// Attempt to repay all fee debt in this call. In most cases the miner will have enough
+		// funds in the *reward alone* to cover the penalty. In the rare case a miner incurs more
+		// penalty than it can pay for with reward and existing funds, it will go into fee debt.
 		penaltyFromVesting, penaltyFromBalance, err := st.RepayPartialDebtInPriorityOrder(store, rt.CurrEpoch(), rt.CurrentBalance())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to repay penalty")
 		pledgeDeltaTotal = big.Sub(pledgeDeltaTotal, penaltyFromVesting)
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1423,8 +1423,10 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *adt.
 
 	var st State
 	pledgeDeltaTotal := big.Zero()
+	toBurn := big.Zero()
 	rt.State().Transaction(&st, func() {
 		var err error
+		store := adt.AsStore(rt)
 		rt.ValidateImmediateCallerIs(builtin.RewardActorAddr)
 
 		// This ensures the miner has sufficient funds to lock up amountToLock.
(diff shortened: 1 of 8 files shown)


ZenGround0:
FYI @Kubuxu

actors/builtin/shared.go
```diff
@@ -87,3 +87,8 @@ func ResolveToIDAddr(rt runtime.Runtime, address addr.Address) (addr.Address, er
 
 	return idAddr, nil
 }
+
+type ApplyRewardParams struct {
+	Reward  *abi.TokenAmount
+	Penalty *abi.TokenAmount
```
anorth, line 93: Neither of these should be pointers here. 

(When the single TokenAmount was the parameter, it needed to be a pointer to satisfy the CBORMarshaller interface)

actors/builtin/miner/miner_test.go
```diff
@@ -3049,7 +3049,7 @@ func TestAddLockedFund(t *testing.T) {
 
 		// Lock some funds with AddLockedFund
 		amt := abi.NewTokenAmount(600_000)
-		actor.addLockedFunds(rt, amt)
```
anorth, line 3052: Please tests the case of non-zero penalty.

actors/builtin/reward/reward_actor.go
```diff
@@ -77,8 +77,9 @@ func (a Actor) AwardBlockReward(rt vmr.Runtime, params *AwardBlockRewardParams)
 	if !ok {
 		rt.Abortf(exitcode.ErrNotFound, "failed to resolve given owner address")
 	}
+	_ = minerAddr
```
anorth, line 80: Unnecessary?

anorth (commented):
This looks good, but see my comment in the issue. I don't think we can leave the miner with newly-incurred debt if the reward would have covered it.

Of the two resolutions, restoring the reward actor's behaviour of reducing the reward introduces the least code and change into this flow.

ZenGround0:
@anorth ready for another round of review.  I went with the option of immediately repaying fees from within the miner actor.

acruikshank (approved):

actors/builtin/miner/miner_actor.go
```diff
@@ -1413,30 +1413,44 @@ func (a Actor) CompactSectorNumbers(rt Runtime, params *CompactSectorNumbersPara
...
+			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", unlockedBalance, params.Reward)
 		}
 
-		newlyVested, err = st.AddLockedFunds(adt.AsStore(rt), rt.CurrEpoch(), *amountToLock, &RewardVestingSpec)
+		newlyVested, err := st.AddLockedFunds(store, rt.CurrEpoch(), params.Reward, &RewardVestingSpec)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to lock funds in vesting table")
+		pledgeDeltaTotal = big.Sub(pledgeDeltaTotal, newlyVested)
+		pledgeDeltaTotal = big.Add(pledgeDeltaTotal, params.Reward)
+
+		err = st.ApplyPenalty(params.Penalty)
```
anorth, line 1444: Nit: signpost what's happening high level here with some comments.

anorth (approved):

PR #1079, opened 2020-09-02
allow/require PoSt in miner's first proving period

### LINKED ISSUE #946: Prove Commit before ProvingPeriodStart
Currently, if a miner tries to prove commit a sector before their proving period starts, they won't be able to actually _prove_ it till after their proving period starts.

One fix is to set the proving period start in the past instead of in the future. This _shouldn't_ have any negative repercussions as far as I know.

acruikshank:
closes #946

### Motivation

Previously, a new miner's first (partial) proving period was special in that we did not run the proving period cron until the end of the first proving period after the miner's creation. This creates an undesirable situation where a miner can pre-commit and prove-commit sectors before they are in a proving period. They cannot however PoSt for these sectors or gain power from them, but they would not be faulted for failing to PoSt when a PoSt is due.

This PR eliminates the special period between miner's creation and the first proving period. This is done by initializing miner state with a `ProvingPeriodStart` that precedes the current epoch and setting the `CurrentDeadline` to the index appropriate to the current epoch. We then set the next proving period cron for the last epoch in the current deadline.

### Proposed Changes

1. Initialize miner state with `ProvingPeriodStart` and `CurrentDeadline` consistent with those expected for the current epoch.
2. Set miner's first proving period cron to last epcoh of current deadline.
3. Remove logic in `miner.State.AdvanceDeadline` that skips most of deadline processing for the first call (it will no longer be called).
4. Fix test that broke with specific expectations for miner state initialization.
5. Add test that sector can be proven before the end of a miner's first proving period.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -2155,6 +2155,7 @@ func currentProvingPeriodStart(currEpoch abi.ChainEpoch, offset abi.ChainEpoch) 
 // Computes the deadline index for the current epoch for a given period start.
 // currEpoch must be within the proving period that starts at provingPeriodStart to produce a valid index.
 func currentDeadlineIndex(currEpoch abi.ChainEpoch, periodStart abi.ChainEpoch) uint64 {
+	Assert(currEpoch >= periodStart)
 	return uint64((currEpoch - periodStart) / WPoStChallengeWindow)
 }
 
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -126,19 +126,21 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
 	currEpoch := rt.CurrEpoch()
 	offset, err := assignProvingPeriodOffset(rt.Message().Receiver(), currEpoch, rt.Syscalls().HashBlake2b)
 	builtin.RequireNoErr(rt, err, exitcode.ErrSerialization, "failed to assign proving period offset")
-	periodStart := nextProvingPeriodStart(currEpoch, offset)
-	Assert(periodStart > currEpoch)
+	periodStart := currentProvingPeriodStart(currEpoch, offset)
+	deadlineIndex := uint64((currEpoch - periodStart) / WPoStChallengeWindow)
+	Assert(deadlineIndex < WPoStPeriodDeadlines)
 
 	info, err := ConstructMinerInfo(owner, worker, controlAddrs, params.PeerId, params.Multiaddrs, params.SealProofType)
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -126,19 +126,21 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *adt.EmptyValu
...
 	infoCid := rt.Store().Put(info)
 
-	state, err := ConstructState(infoCid, periodStart, emptyBitfieldCid, emptyArray, emptyMap, emptyDeadlinesCid, emptyVestingFundsCid)
+	state, err := ConstructState(infoCid, periodStart, deadlineIndex, emptyBitfieldCid, emptyArray, emptyMap, emptyDeadlinesCid, emptyVestingFundsCid)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to construct state")
 	rt.State().Create(state)
 
-	// Register first cron callback for epoch before the first proving period starts.
-	enrollCronEvent(rt, periodStart-1, &CronEventPayload{
+	// Register first cron callback for epoch before the next proving period starts.
```
anorth, line 141:
```suggestion
	// Register first cron callback for epoch before the next proving deadline starts.
```

actors/builtin/miner/miner_actor.go
```diff
@@ -2118,11 +2120,17 @@ func nextProvingPeriodStart(currEpoch abi.ChainEpoch, offset abi.ChainEpoch) abi
...
-	Assert(periodStart > currEpoch)
+	periodStart := currEpoch - periodProgress
+	Assert(periodStart <= currEpoch)
 	return periodStart
 }
 
+// Computes the deadline index for the current epoch for a given period start.
+// currEpoch must be within the proving period that starts at provingPeriodStart to produce a valid index.
+func currentDeadlineIndex(currEpoch abi.ChainEpoch, periodStart abi.ChainEpoch) uint64 {
+	return uint64((currEpoch - periodStart) / WPoStChallengeWindow)
```
anorth, line 2131: Assert that currEpoch >= periodStart.

actors/builtin/miner/miner_test.go
```diff
@@ -659,6 +663,29 @@ func TestCommitments(t *testing.T) {
 		rt.Reset()
 	})
 
+	t.Run("prove commit just after period start permits PoSt", func(t *testing.T) {
+		actor := newHarness(t, periodOffset)
+		rt := builderForHarness(actor).
+			WithBalance(bigBalance, big.Zero()).
+			Build(t)
+
+		// at epoch 101, the miner will get an initial proving period start thousands of epochs in the future
```
anorth, line 672: I don't understand this comment.

anorth (approved):

PR #1080, opened 2020-09-02
Canonicalise import order and remove redundant import aliases

anorth:
This is a prefactor to reduce noise in forthcoming upgrade mechanisms (where we need to rename a lot of imports)

- The import ordering was mostly achieved with GoLand's import optimization, with a few manual helps
- Redudant aliases detected manually, removed with a find/replace script

There are a few odd import aliases lingering.

If approved, please merge this for me to reduce the branch overhead.


actors/builtin/miner/miner_state.go
```diff
@@ -11,13 +11,13 @@ import (
 	errors "github.com/pkg/errors"
 	xerrors "golang.org/x/xerrors"
 
-	power "github.com/filecoin-project/specs-actors/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
 	. "github.com/filecoin-project/specs-actors/actors/util"
```
ZenGround0, line 16: We should delete this new line

PR #1082, opened 2020-09-02
Add big.NewFromGo constructor

anorth:
Avoid having to push through the abstraction every time. This was prompted by version/migration concerns.

This follows the docs recommendations about copying. Please lmk if you think this should take a `*big.Int` instead.

### DIFF
--- a/actors/util/smoothing/alpha_beta_filter.go
+++ b/actors/util/smoothing/alpha_beta_filter.go
@@ -61,10 +61,10 @@ func init() {
 	}
 	constBigs := math.Parse(constStrs)
 	_ = math.Parse(constStrs)
-	DefaultAlpha = big.NewFromGo(*constBigs[0])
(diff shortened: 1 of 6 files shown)


actors/util/smoothing/alpha_beta_filter.go
```diff
@@ -208,8 +208,8 @@ func lnBetweenOneAndTwo(x big.Int) big.Int {
 	num := math.Polyval(lnNumCoef, x.Int)     // Q.128
 	denom := math.Polyval(lnDenomCoef, x.Int) // Q.128
 
-	num = num.Lsh(num, math.Precision)       // Q.128 => Q.256
-	return big.Int{Int: num.Div(num, denom)} // Q.256 / Q.128 => Q.128
+	num = num.Lsh(num, math.Precision)        // Q.128 => Q.256
```
ZenGround0, line 211: nit: align comments

PR #1083, opened 2020-09-02
remove ValidateImmediateCallerType from market WithdrawBalance

### LINKED ISSUE #1075: Market withdrawal always fails due to repeated caller validation
#718 has introduced a bug which would make all withdrawals abort. It invokes caller validation twice. The behaviour of this is not well documented, but it will abort in Lotus.

The check for `rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)` should be removed.

acruikshank:
closes #1075

### Motivation

Although nothing specifically prevents it in actor code, validating caller types more than once in a method is prohibited and causes an abort in Lotus. This double validation occurs in `market.Actor.WithdrawBalance`. The call to `ValidateImmediateCallerType` is unnecessary, so we should remove it.

### Proposed Changes

1. Simulate multiple validation calls abort in scenario test vm.
2. Add scenario test for adding and withdrawing collateral to demonstrate the problem.
3. Delete `ValidateImmediateCallerType` in `market.Actor.WithrdrawBalance`.

### DIFF
--- a/actors/test/market_withdrawl_test.go
+++ b/actors/test/market_withdrawl_test.go
@@ -1,0 +1,46 @@
+package test_test
+
+import (
+	"context"
+	"github.com/filecoin-project/specs-actors/actors/builtin/market"
+	"testing"
+
+	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	vm "github.com/filecoin-project/specs-actors/support/vm"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMarketWithdrawl(t *testing.T) {
+	ctx := context.Background()
+	v := vm.NewVMWithSingletons(ctx, t)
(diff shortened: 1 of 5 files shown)


actors/test/market_withdrawl_test.go
```diff
@@ -0,0 +1,46 @@
+package test_test
+
+import (
+	"context"
+	"github.com/filecoin-project/specs-actors/actors/builtin/market"
```
ZenGround0, line 5: nit: move this guy to the bottom group (after @anorth's latest PR we're trying to be strict about import conventions)

actors/test/market_withdrawl_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/filecoin-project/specs-actors/actors/abi/big"
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+	vm "github.com/filecoin-project/specs-actors/support/vm"
+
+	"github.com/stretchr/testify/assert"
```
ZenGround0, line 13: nit: these should be in the middle, specs actors at the end

PR #1084, opened 2020-09-02
add method to repay debt

### LINKED ISSUE #1072: Method for manual fee debt repayment
A miner with non-zero fee debt is ineligible for elections, sector pre-commit, fault recovery, and withdrawal. For the latter three, debt is repaid automatically by the miner code when that action is attempted (via `RepayDebtsOrAbort`). But there's no way for a miner to deposit funds and repay debt to regain election eligibility immediately. We should provide a direct means for that, else miners will be penalised by ineligibility for too long.

- Note that withdrawal of zero funds would trigger it, but this is available only to the owner address
- Declaring zero recoveries will also do this, but by chance rather than design; it would be easy to lose this in the future

The smallest change would be to allow WithdrawFunds to be called by the worker iff the requested amount is zero. Otherwise I think we need a new exported method. I'm ok with either.

acruikshank:
closes #1072

### Motivation

If a miner goes into fee debt, the worker can pay it off by adding balance and then pre-committing a sector, recovering a fault. If the worker doesn't have a sector to precommit or a recovered fault they must wait for the deadline end for the debt to be paid. This PR adds a `RepayDebt` method to trigger that mechanism immediately.

### Proposed Changes

1. Add `miner.RepayDebt` that calls `miner.State.RepayPartialDebtInPriorityOrder` and burns the appropriate funds and notifies of any pledge change due to fee payment from vesting funds.
2. Add unit tests for new method.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1601,8 +1601,7 @@ func (a Actor) RepayDebt(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
 
-		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
-		// and repay fee debt now.
+		// Repay as much fee debt as possible.
 		fromVesting, fromBalance, err = st.RepayPartialDebtInPriorityOrder(adt.AsStore(rt), rt.CurrEpoch(), rt.CurrentBalance())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock fee debt")
 	})
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -70,6 +70,7 @@ func (a Actor) Exports() []interface{} {
 		19:                        a.CompactPartitions,
 		20:                        a.CompactSectorNumbers,
 		21:                        a.ConfirmUpdateWorkerKey,
+		22:                        a.RepayDebt,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1572,6 +1573,26 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *adt.E
...
 
+func (a Actor) RepayDebt(rt Runtime, params *adt.EmptyValue) *adt.EmptyValue {
+	var st State
+	var fromVesting, fromBalance abi.TokenAmount
+	rt.State().Transaction(&st, func() {
+		var err error
+		info := getMinerInfo(rt, &st)
+		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
+
+		// Verify unlocked funds cover both InitialPledgeRequirement and FeeDebt
```
anorth, line 1584: This first part of the comment doesn't make sense after @ZenGround0 's recent changes.  I don't think this will or should fail of only part of the debt can be paid off.

anorth (approved):

actors/builtin/miner/miner_test.go
```diff
@@ -4159,6 +4238,26 @@ func (h *actorHarness) withdrawFunds(rt *mock.Runtime, amountRequested, amountWi
 	rt.Verify()
 }
 
+func (h *actorHarness) repayDebt(rt *mock.Runtime, value, expectedRepayedFromVest, expectedRepaidFromBalance abi.TokenAmount) {
```
ZenGround0, line 4241: Love the explicit split.  Side note -- one of the next ways to improve miner test harness is to come up with patterns that make this easy to do everywhere.  For example it is very difficult to specify this with the current cron config without heroic contortions in the calling test code.

ZenGround0 (approved):

PR #1085, opened 2020-09-02
adjust min miner size for 64 g

### LINKED ISSUE #1027: Adjust minimum miner size for 64GiB miners
The minimum miner size calculations was made for 32GiB sectors. It depends on sector size for PoSt security. The minimum size for a 64GiB miner is double that of a 32GiB miner. 

The power actor must be updated to record a miner's sector size alongside its power claim record. The size-dependent threshold should be used everywhere that the current one is, in particular calculations of the total network power and eligibility for elections.

acruikshank:
closes #1027

### Motivation

The security of a miner's eligibility for election is affected by the total power the miner has on the network, but also on the miner's sector size. This PR modifies all consensus power threshold calculations to be a function of miner sector size (proof type) and sets the value for 64GiB sectors to twice the previous minimum which is still appropriate for 32GiB sectors.

### Proposed Changes

1. Add `ConsensusMinerMinPower` to seal proof info.
2. Set values for the above: 32GiB is unchanged, 64GiB is 2x, all the rest (unused in production) are zero.
3. Add `SealProofType` to power actor's `Claim` type.
4. Initialize above to miner `SealProofType` to miner's proof type when miner is created. Copy it for each claim update.
5. Use above to derive minimum power in `power.State.addToClaim` and `power.State.MinerNominalPowerMeetsConsensusMinimum`.
6. Fix tests that did not use 32GiB proof type and computed minimum consensus power.
7. Augment tests to check the new behavior in `power.State.addToClaim` and `power.State.MinerNominalPowerMeetsConsensusMinimum`.

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -99,6 +99,7 @@ type SealProofInfo struct {
 	WinningPoStProof           RegisteredPoStProof
 	WindowPoStProof            RegisteredPoStProof
 	SectorMaxLifetime          ChainEpoch
+	ConsensusMinerMinPower     StoragePower
 }
 
 // For all Stacked DRG sectors, the max is 5 years
@@ -114,6 +115,7 @@ var SealProofInfos = map[RegisteredSealProof]*SealProofInfo{
 		WinningPoStProof:           RegisteredPoStProof_StackedDrgWinning2KiBV1,
 		WindowPoStProof:            RegisteredPoStProof_StackedDrgWindow2KiBV1,
 		SectorMaxLifetime:          fiveYears,
+		ConsensusMinerMinPower:     NewStoragePower(0),
 	},
 	RegisteredSealProof_StackedDrg8MiBV1: {
 		SectorSize:                 8 << 20,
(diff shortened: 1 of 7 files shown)


actors/abi/sector.go
```diff
@@ -114,34 +115,39 @@ var SealProofInfos = map[RegisteredSealProof]*SealProofInfo{
...
 		SectorMaxLifetime:          fiveYears,
+		ConsensusMinerMinPower:     NewStoragePower(0),
 	},
 	RegisteredSealProof_StackedDrg512MiBV1: {
 		SectorSize:                 512 << 20,
 		WindowPoStPartitionSectors: 2,
 		WinningPoStProof:           RegisteredPoStProof_StackedDrgWinning512MiBV1,
 		WindowPoStProof:            RegisteredPoStProof_StackedDrgWindow512MiBV1,
 		SectorMaxLifetime:          fiveYears,
+		ConsensusMinerMinPower:     NewStoragePower(0),
```
anorth, line 134: For all except the 2KiB sectors, let's set this >0 . Say at 2 * sector size?

actors/builtin/miner/miner_state_test.go
```diff
@@ -881,6 +881,23 @@ func TestMinerEligibleForElection(t *testing.T) {
...
 		assert.False(t, eligible)
+
+		// get minimum
+		pow32GiBMin, err := abi.RegisteredSealProof_StackedDrg32GiBV1.ConsensusMinerMinPower()
+		require.NoError(t, err)
+
+		// with enough miners above minimum, power must be above min
+		pow := big.Sub(pow32GiBMin, big.NewInt(1))
+		pSt = constructPowerStateWithPowerAtAddr(t, pow, actor.receiver, abi.RegisteredSealProof_StackedDrg32GiBV1, rt)
+		pSt.MinerAboveMinPowerCount = power.ConsensusMinerMinMiners
```
anorth, line 892: Looks like you're missing some assertions after here, because this is then overwritten.

anorth (approved):
FYI @ZenGround0 @magik6k

PR #1087, opened 2020-09-03
Separate seal proof type metadata from policy

anorth:
Separates the conceptually fixed associations between proof types from the policy about how they are used (that could conceivably change in a protocol upgrade).

This is in preparation for integrating https://github.com/filecoin-project/go-state-types/pull/1, which moves the values and metadata to a new repository.

### DIFF
--- a/actors/abi/sector.go
+++ b/actors/abi/sector.go
@@ -159,7 +159,7 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 type SealProofPolicy struct {
 	WindowPoStPartitionSectors uint64
 	SectorMaxLifetime          ChainEpoch
-  ConsensusMinerMinPower     StoragePower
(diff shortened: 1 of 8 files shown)


actors/abi/sector.go
```diff
@@ -95,53 +95,35 @@ const (
 // Metadata about a seal proof type.
 type SealProofInfo struct {
 	SectorSize                 SectorSize
-	WindowPoStPartitionSectors uint64
```
Stebalien, line 98: Is this not a limitation of the proof type?

Stebalien (approved):

ZenGround0 (approved):

actors/abi/sector.go
anorth: I'm pretty sure it's policy, in that it was chosen based on security analysis, and new analysis could change it. It would have to be matched in the proofs library, but it's a feasible upgrade.

PR #1088, opened 2020-09-03
Repackage as specs-actors/v1

anorth:
This PR changes the package path for the repo to `github.com/filecoin-project/specs-actors/v1`. This is in preparation for eventually cutting the mainnet-candidate `release-1.0` branch and release series. We intend to version functional upgrades to actors with release branches and corresponding Go package major versions, such that a downstream project can import multiple different versions at once.

This repackaging does not imply that no more functional changes should be landing; we still have some fixes to land before mainnet and consumers should not assume state or functionality is fixed until at least the release branch is cut.

This is a purely mechanical renaming of the package.


anorth:
Note: in order to ease testing an upgrade between the `0.9` branch and this, a subsequent change will import `specs-actors v0.9.3` into this repo and then alias a bunch of types that haven't changed onto their `0.9` versions.

Stebalien:
The `/vN` prefix only applies at `N>=2`. Go actually rejects v1.

PR #1089, opened 2020-09-03
abort consensus fault reports when miner is already under fault

### LINKED ISSUE #1008: Miner: Verify that duplicate submissions of consensus faults are not processed
From the code: 

https://github.com/filecoin-project/specs-actors/blob/ca1b7724d3016313f76c583c4f1110fe1e58ef7d/actors/builtin/miner/miner_actor.go#L1414-L1417

Given that miners are no longer terminated on consensus faults, this method should ensure that submitted faults are unique.

acruikshank:
closes #1008

### Motivation

A miner should only be penalized (and reported rewarded) once for a consensus fault. The most straightforward way to do this is to prevent any consensus fault reports for a miner until the miner's existing consensus fault penalization period has expired.

### Proposed Changes

1. Abort `ReportConsensusFault` if current epoch < `miner.Info.ConsensusFaultElapsed`.
2. Test that this prevents duplicate consensus fault reports.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1511,8 +1511,8 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 		info := getMinerInfo(rt, &st)
 
 		// verify miner hasn't already been faulted
-		if rt.CurrEpoch() < info.ConsensusFaultElapsed {
-			rt.Abortf(exitcode.ErrForbidden, "consensus fault has already been reported")
+		if fault.Epoch < info.ConsensusFaultElapsed {
+			rt.Abortf(exitcode.ErrForbidden, "miner has existing fault that has not expired (this may be a duplicate)")
 		}
 
 		err := st.ApplyPenalty(faultPenalty)
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1508,6 +1508,13 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 	burnAmount := big.Zero()
 	rewardAmount := big.Zero()
 	rt.State().Transaction(&st, func() {
+		info := getMinerInfo(rt, &st)
+
+		// verify miner hasn't already been faulted
+		if rt.CurrEpoch() < info.ConsensusFaultElapsed {
+			rt.Abortf(exitcode.ErrForbidden, "consensus fault has already been reported")
+		}
+
 		err := st.ApplyPenalty(faultPenalty)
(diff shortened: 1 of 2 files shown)


acruikshank:
cc @wadeAlexC

actors/builtin/miner/miner_actor.go
```diff
@@ -1498,6 +1498,13 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 	burnAmount := big.Zero()
 	rewardAmount := big.Zero()
 	rt.State().Transaction(&st, func() {
+		info := getMinerInfo(rt, &st)
+
+		// verify miner hasn't already been faulted
+		if rt.CurrEpoch() < info.ConsensusFaultElapsed {
+			rt.Abortf(exitcode.ErrForbidden, "consensus fault has already been reported")
```
anorth, line 1505: Blocking: it's the `fault.Epoch` that must be after the exclusion period, in order to prevent replay.

Also the message isn't quite right - it might be a different report. Say something about expired or too old instead.

anorth (changes requested):

actors/builtin/miner/miner_test.go
```diff
@@ -3131,6 +3131,43 @@ func TestReportConsensusFault(t *testing.T) {
...
+		rt.Reset()
+
+		// new consensus faults are forbidden until original has elapsed
+		rt.SetEpoch(endInfo.ConsensusFaultElapsed) // actor.reportConsensusFault reports a fault for previous epoch
+		rt.ExpectAbortContainsMessage(exitcode.ErrForbidden, "miner has existing fault that has not expired", func() {
+			actor.reportConsensusFault(rt, addr.TestAddress)
+		})
+		rt.Reset()
+
+		// a new consensus fault can be reported once fault has elapsed
```
anorth, line 3165: Test that an old consensus fault cannot be reported (you'll need to mock the returned `Fault.Epoch`).

actors/builtin/miner/miner_actor.go
```diff
@@ -1498,6 +1498,13 @@ func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultPara
 	burnAmount := big.Zero()
 	rewardAmount := big.Zero()
 	rt.State().Transaction(&st, func() {
+		info := getMinerInfo(rt, &st)
+
+		// verify miner hasn't already been faulted
+		if fault.Epoch < info.ConsensusFaultElapsed {
+			rt.Abortf(exitcode.ErrForbidden, "miner has existing fault that has not expired (this may be a duplicate)")
```
anorth, line 1505: Nit: the message is still misleading. `ConsensusFaultElapsed` may have expired, but the fault is older.

`fault epoch %d is too old, last exclusion period ended at %d`

anorth (approved):

ZenGround0 (approved):

PR #1092, opened 2020-09-03
delete miner claim after cron failure

### LINKED ISSUE #1056: Power: Explicitly delete miner claim on failing cron callback
#### Background

If a miner's cron callback aborts, their power is removed from the Power actor table:

https://github.com/filecoin-project/specs-actors/blob/d08932735838703b671d6bb8535539d14fe1fdd7/actors/builtin/power/power_actor.go#L435-L440

This feature was added following discussion in #452, and exists to defend against a broken cron callback.

From the above issue, a broken cron callback means there is something wrong in the miner actor code. Whether the miner exploited it maliciously or not: if a broken cron callback occurs once, it can occur again. Therefore, a hard fork is required to fix the issue.

Because we know a hard fork is required, we don't need to take drastic steps (like deleting the miner, terminating their deals, etc) - decisions like these can be deferred to the hard fork itself and will likely be determined by a social consensus that the miner was (or was not) acting maliciously. Instead, our priority should be to isolate the problem and mitigate potential further damage to the system.

#### Effects of a broken cron callback in the current implementation

The effects of a broken cron callback depend on which callback failed (`CronEventProvingDeadline` or `CronEventProcessEarlyTerminations`), but generally we can assume:

* **The miner may no longer have a functioning proving period cycle.** If the `CronEventProvingDeadline` callback fails, the miner's deadline and proving period will not be updated, and a future `CronEventProvingDeadline` will not be scheduled. This callback is responsible for keeping miners accountable (via penalties, sector/precommit expiries, and debt payment). Without this callback, the miner never has to worry about submitting proofs on time, paying debts, and more.
* **For the most part, the miner is still able to function.** If a miner's cron callback aborts, their power will be set to zero, making them ineligible for leader election. However, the miner's methods still function to a large extent. If a miner is able to gain power somehow (perhaps through submission of PoSts), they may again be eligible for leader election.

Although it may seem that a failing `CronEventProvingDeadline` is the "worst-case scenario" for failing cron callbacks, this may not be the case. Let's assume `CronEventProcessEarlyTerminations` fails, leaving the miner with zeroed-out power, but a functional proving period cycle. Then, on the next `CronEventProvingDeadline` callback, a missed PoSt results in a negative power delta:

https://github.com/filecoin-project/specs-actors/blob/d08932735838703b671d6bb8535539d14fe1fdd7/actors/builtin/miner/partition_state.go#L658-L660

This is communicated to the power actor via `UpdateClaimedPower`, which adds the delta directly to the miner's claim in `power_state.addToClaim`:

https://github.com/filecoin-project/specs-actors/blob/d08932735838703b671d6bb8535539d14fe1fdd7/actors/builtin/power/power_actor.go#L153-L154

Because the miner's power was zeroed out, the new claim will have negative power, causing a panic here:

https://github.com/filecoin-project/specs-actors/blob/d08932735838703b671d6bb8535539d14fe1fdd7/actors/builtin/power/power_state.go#L191-L194

Because the failing assertion was reached during cron execution, this would result in an unprocessable tipset.

#### Conclusion

The existing solution (zeroing out the miner's power) is insufficient in mitigating the effects of a broken cron callback. Regardless of which callback breaks, the miner is still able to interact with the system to a large degree. It seems likely that if this does occur, the miner could cause significant damage.

#### Recommendation

In an effort to keep proposed changes small, I've organized these by priority:

* **Explicitly delete a miner's claim after a failed callback**: This is a tiny change that will address most concerns stated above, as existing code handles "claim not found" well. Specifically, `UpdatedClaimedPower` will abort with `ErrNotFound`, rather than panicking after attempting to reduce a miner's power to a negative value. 

Deleting a miner's claim will make it far more difficult for a broken miner to cause more damage, but there is still some room for a miner to function post-deletion:
  * The miner is still free to use methods that may update power (as long as their resulting `powerDelta` is zero), as this will not invoke `UpdateClaimedPower`, which would abort and revert state changes.
  * `EnrollCronEvent`, `UpdatePledgeTotal`, and `SubmitPoRepForBulkVerify` do not query a miner's claim, and so will continue to function.
  * Additionally, `OnEpochTickEnd` will continue to process events for the miner. `processBatchProofVerifies` may call into `Miner.ConfirmSectorProofsValid`, and `processDeferredCronEvents` may call into `Miner.OnDeferredCronEvent`.

The following recommendation seeks to address these issues:

* **Check whether a miner has a claim in other Power methods**: We don't want a miner with a broken cron callback to continue interacting with the power actor. Short of making changes within the miner actor, we can ensure this doesn't happen by implementing the above claim deletion, and adding a check for a miner's claim to other Power methods: 
  * `EnrollCronEvent`, `UpdatePledgeTotal`, and `SubmitPoRepForBulkVerify` should each abort if the miner doesn't have a claim.
  * `processBatchProofVerifies` should check if a given miner has a claim. If they do not, the method should skip verification and emit a log message.
  * `processDeferredCronEvents` should check if a given miner has a claim. If they do not, the method should skip execution of the cron event and emit a log message.

The above two changes will ensure a broken miner cannot interact with the power actor. It may be that this is sufficient for the majority of potential cases, but there are still a few things to consider:

* The miner can continue to interact with the market actor; `Market.ActivateDeals` and `Market.OnMinerSectorsTerminate` can still be invoked. Given that `Market.ActivateDeals` is only invoked from `Miner.ConfirmSectorProofsValid`, the above recommendations should prevent this. For `Market.OnMinerSectorsTerminate`, the above recommendations should cause aborts when the miner updates either pledge or power as a result of terminations. Personally, I am comfortable with the market remaining blind to a miner's status in the power actor (as long as the above changes are made).
* The miner is still able to use a few methods, but nothing that will result in a change to either power or pledge. This limits the miner's capability significantly, though there are a few points to note:
  * `Miner.PreCommitSector` is still usable as long as the miner doesn't vest any funds, which is easy to accomplish as a broken miner won't be earning block rewards.
  * `Miner.DeclareFaultsRecovered` still seems usable, but I'm not sure that's a bad thing
  * `Miner.ReportConsensusFault` is usable, as long as the miner doesn't have vesting funds
  * `Miner.WithdrawBalance` is usable, as long as the miner doesn't have vesting funds

If any of these final points are concerning, consider implementing a flag similar to `info.ConsensusFaultElapsed`, and using it to safeguard the remaining methods.

acruikshank:
closes #1056

### Motivation

When a miner's cron tasks fail, the miner is in a bad state and could potentially put power in a progressively worse state as time progresses. Zeroing out miner power helps, but it is not enough. This PR deletes the miner claim altogether when cron fails and makes the existence of a claim a requirement for most interactions between the miner and the power actor.

### Proposed Changes

1. On failure in `processDeferredCronEvents` delete miner's claim altogether.
2. In `UpdatePledgeTotal` and `SubmitPoRepForBulkVerify`, abort if miner has no claim.
3. In `processBatchProofVerifies` and `processDeferredCronEvents` skip and log actions for miner's that have no claim.
4. Fix existing power tests that mostly broke because miner was never initialized.
5. Add tests for new functionality.
6. Skip transaction and claims initialization and save if no cron failures were detected. This is a small optimization, for the case were there are no failures (hopefully always).

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -2,7 +2,6 @@ package power
 
 import (
 	"bytes"
-
 	"github.com/filecoin-project/go-address"
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
@@ -229,6 +228,7 @@ func (a Actor) UpdatePledgeTotal(rt Runtime, pledgeDelta *abi.TokenAmount) *adt.
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 	var st State
 	rt.State().Transaction(&st, func() {
+		validateMinerHasClaim(rt, st, rt.Message().Caller())
 		st.addPledgeTotal(*pledgeDelta)
 	})
 	return nil
@@ -245,6 +245,8 @@ func (a Actor) SubmitPoRepForBulkVerify(rt Runtime, sealInfo *proof.SealVerifyIn
 
 	var st State
 	rt.State().Transaction(&st, func() {
+		validateMinerHasClaim(rt, st, minerAddr)
+
 		store := adt.AsStore(rt)
 		var mmap *adt.Multimap
 		if st.ProofValidationBatch == nil {
@@ -302,6 +304,17 @@ func (a Actor) CurrentTotalPower(rt Runtime, _ *adt.EmptyValue) *CurrentTotalPow
 // Method utility functions
 ////////////////////////////////////////////////////////////////////////////////
 
+func validateMinerHasClaim(rt Runtime, st State, minerAddr address.Address) {
+	claims, err := adt.AsMap(adt.AsStore(rt), st.Claims)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load claims")
+
+	found, err := claims.Has(AddrKey(minerAddr))
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to look up claim")
+	if !found {
+		rt.Abortf(exitcode.ErrForbidden, "unknown miner %s forbidden to interact with power actor", minerAddr)
+	}
+}
+
 func (a Actor) processBatchProofVerifies(rt Runtime) {
 	var st State
 
@@ -316,9 +329,20 @@ func (a Actor) processBatchProofVerifies(rt Runtime) {
 		mmap, err := adt.AsMultimap(store, *st.ProofValidationBatch)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load proofs validation batch")
 
+		claims, err := adt.AsMap(adt.AsStore(rt), st.Claims)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load claims")
+
 		err = mmap.ForAll(func(k string, arr *adt.Array) error {
 			a, err := address.NewFromBytes([]byte(k))
 			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to parse address key")
+
+			// refuse to process proofs for miner with no claim
+			found, err := claims.Has(AddrKey(a))
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to look up claim")
+			if !found {
+				rt.Log(vmr.WARN, "skipping batch verifies for unknown miner %s", a)
+				return nil
+			}
 
 			miners = append(miners, a)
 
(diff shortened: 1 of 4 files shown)


anorth (approved):
Thanks for the tests.

anorth:
@wadealexc

wadeAlexC:
Looks good for the most part. One thing that may be missing - `addToClaim` updates several state variables (`TotalQABytesCommitted`, `TotalBytesCommitted`, ...) - `claims.Delete` does not.

acruikshank:
good call, @wadeAlexC. I'm now using `addToClaim` to subtract out miner power before deleting the claim. This should keep the accounting more sane if this code ever gets called.

PR #1096, opened 2020-09-04
Empty the abi package

anorth:
Moves everything remaining in the `abi` package elsewhere. This avoids package name conflicts with the `abi` package now in `go-state-types`.


actors/builtin/sector.go
```diff
@@ -1,12 +1,10 @@
-package abi
+package builtin
```
anorth, line 1: It's hard to find a good spot for this. It's really miner policy, but used by the power actor. Since miner depends on power already, anything used by both needs to be in power, or builtin etc.

actors/builtin/miner/expiration_queue.go
```diff
@@ -60,12 +60,12 @@ func (es *ExpirationSet) Add(onTimeSectors, earlySectors bitfield.BitField, onTi
 // Removes sectors and power from the expiration set in place.
 func (es *ExpirationSet) Remove(onTimeSectors, earlySectors bitfield.BitField, onTimePledge abi.TokenAmount, activePower, faultyPower PowerPair) error {
 	// Check for sector intersection. This could be cheaper with a combined intersection/difference method used below.
-	if found, err := aabi.BitFieldContainsAll(es.OnTimeSectors, onTimeSectors); err != nil {
+	if found, err := util.BitFieldContainsAll(es.OnTimeSectors, onTimeSectors); err != nil {
```
Stebalien, line 63: We should rally just move these to the bitfield package, eventually.

actors/builtin/sector.go
Stebalien: We could introduce a new "sector" package, but this seems fine.

Stebalien (approved):

ZenGround0 (approved):

PR #1101, opened 2020-09-05
Implement cap of 1 FIL / 32GiB cap on initial pledge

anorth:
This was implemented on the 0.9 branch for Space Race in #997, and is to be preserved for mainnet.

This is a cherry-pick, followed by a fix to a test with a now-broken assumption.


anorth:
FYI @zixuanzh @nicola

ZenGround0 (approved):

nicola:
Where is the motivation for this change? @zixuanzh?

zixuanzh:
@nicola we dont expect initial pledge to go above 1 FIL/32GiB qa power with a reasonable initial network size. The other case could be that there is no power added to the network for a long time, in which case a max would be useful.

PR #1104, opened 2020-09-05
Reduce vesting cliff, penalties, min deal collateral on network v1

anorth:
This is an inline upgrade for Space Race, adjusting some crypto-economic parameters to reduce penalties and deal collateral. From version `1` the actor code will behave differently, but no existing state will be migrated.

This PR is against the release/v0.9 branch. The changes are to be subsequently cherry-picked to master, where we can drop the handling of network v0 (which will be handled by major release versioning).

- Adds `NetworkVersion()` call to Runtime
- v1: Reduce reward vesting cliff to zero
- v1: Reduce undeclared fault penalty to 3BR
- v1: Halve the sector age penalised in termination fee
- v1: Reduce min provider deal collateral to 1% of share, based on _raw_ power

- [ ] Wait for @nicola to ack the reduced undeclared fault penalty

FYI @whyrusleeping @magik6k @Stebalien

### DIFF
--- a/actors/runtime/runtime.go
+++ b/actors/runtime/runtime.go
@@ -32,9 +32,21 @@ const (
 	ERROR
 )
 
+// Enumeration of network upgrades where actor behaviour can change (without necessarily
+// vendoring and versioning the whole actor codebase).
+type NetworkVersion uint
+
+const (
+	NetworkVersion0 = NetworkVersion(iota) // specs-actors v0.9.3
+	NetworkVersion1                        // specs-actors v0.9.?
+)
+
 // Runtime is the VM's internal runtime object.
 // this is everything that is accessible to actors, beyond parameters.
 type Runtime interface {
+	// The network protocol version number at the current epoch.
+	NetworkVersion() NetworkVersion
+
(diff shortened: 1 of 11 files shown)


ZenGround0 (approved):
I missed the memo that we wanted to do this as part of the small upgrade but from context that must be what we are trying to do.  (If that's not the case this approach does not make sense)

zixuanzh (approved):

actors/runtime/runtime.go
```diff
@@ -32,9 +32,21 @@ const (
 	ERROR
 )
 
+// Enumeration of network upgrades where actor behaviour can change (without necessarily
+// vendoring and versioning the whole actor codebase).
+type NetworkVersion uint
+
+const (
+	NetworkVersion0 = NetworkVersion(iota) // specs-actors v0.9.3
+	NetworkVersion1                        // specs-actors v0.9.?
```
arajasek, line 41: Wanna call it v0.9.7, and then cut the 0.9.7 release as soon as this is merged?

arajasek (approved):

PR #1114, opened 2020-09-07
Remove PartitionKey type

anorth:
This was useful when there were many miner actor methods which accumulated things by this key, but now that's all pushed down into deadline and partition state objects.


acruikshank (approved):
Looks good.

PR #1115, opened 2020-09-07
WIP state migration functions

anorth:
TODO:
- [x] Rewrite all HAMTs
- [x] Calculate miner fee debt
- [x] Ensure miner balance >= locked
- [x] Migrate reward baseline function state directly to new initial values
- [x] Normalize addresses in verified registry state (possibly creating accounts to do so).
- [x] Verify addresses in payment channel state are all normalized
- [x] Add miner proof type to power claims


actors/states/tree_top.go
```diff
@@ -92,3 +92,15 @@ func (t *TreeTop) LookupID(addr address.Address) (address.Address, error) {
 
 	return a, nil
 }
+
+func (t *TreeTop) ForEach(ctx context.Context, fn func(addr address.Address, actor *Actor) error) error {
+	var val Actor
+	return t.m.ForEach(&val, func(key string) error {
+		addr, err := address.NewFromString(key)
```
anorth, line 99: @ZenGround0 this isn't the right converstion. We should put something with the new go-state-types keys that does it right (`NewFromBytes`)

actors/migration/top.go
```diff
@@ -17,65 +23,159 @@ type StateMigration interface {
...
 	},
 }
+
+// A phoenix tracks the unburning of funds
+type phoenix struct {
+	burntBalance abi.TokenAmount
+}
+
+func (p phoenix) load(ctx context.Context, actorsIn *states.TreeTop) error {
+	burntFundsActor, err := actorsIn.GetActor(ctx, builtin0.BurntFundsActorAddr)
```
anorth, line 83: I don't know that this is where the funds will come from. All funds were unburnt in a previous migration, back to mining rewards wallets or something. This needs to be easily configurable.

actors/migration/top.go
```diff
@@ -17,65 +23,159 @@ type StateMigration interface {
...
+	burntFundsActor.Balance = p.burntBalance
+	return actorsOut.SetActor(ctx, builtin.BurntFundsActorAddr, burntFundsActor)
+}
+
+// Migrates the filecoin state tree starting from the global state tree and upgrading all actor state.
+func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid.Cid) (cid.Cid, error) {
+	// first migrate the global state tree hamt to something v2 can work with
+	// if this is very slow (likely) we can cherry-pick states.TreeTop to v0.9
+	// to avoid this step.
+	stateRootInTweaked, err := migrateHAMTRaw(ctx, store, stateRootIn)
```
anorth, line 117: I don't understand why you're doing this first. We're going to rewrite the whole thing again anyway. Why re-write it to something that needs to be rewritten yet again. This will be slow and pollute the store.

actors/states/tree_top.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"context"
+
+	address "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	init_ "github.com/filecoin-project/specs-actors/v2/actors/builtin/init"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
```
anorth, line 11: Within the migration package, I think every actors import should be explicitly versioned, e.g. `adt2`

actors/states/tree_top.go
```diff
@@ -0,0 +1,94 @@
...
+	cid "github.com/ipfs/go-cid"
+	xerrors "golang.org/x/xerrors"
+)
+
+var ErrActorNotFound = xerrors.Errorf("Actor not found")
+
+type Actor struct {
+	Code    cid.Cid
+	Head    cid.Cid
+	Nonce   uint64
```
anorth, line 21: CallSeqNum

actors/states/tree_top.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	init_ "github.com/filecoin-project/specs-actors/v2/actors/builtin/init"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+	cid "github.com/ipfs/go-cid"
+	xerrors "golang.org/x/xerrors"
+)
+
+var ErrActorNotFound = xerrors.Errorf("Actor not found")
```
anorth, line 16: Please return a boolean instead

actors/migration/top.go
ZenGround0: The reason is that the new hamt dependency will possibly be unable to read the old HAMT because with high probability there are keys out of order.  I think you've realized this with the all to pull out the state tree changes on v0.9.  As the comment states, if we do that we can avoid this migration.

actors/states/tree_top.go
anorth: Resolved in #1147

actors/migration/top.go
anorth: Ok, #1147

actors/states/tree_top.go
anorth: Resolved in #1147

actors/states/tree_top.go
anorth: Resolved in #1147

actors/migration/top.go
```diff
@@ -118,26 +114,40 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
+	pm := migrations[builtin0.StoragePowerActorCodeID].StateMigration.(*powerMigrator)
+	pm.actorsIn = actorsIn
 
 	// Iterate all actors in old state root
 	// Set new state root actors as we go
 	err = actorsIn.ForEach(ctx, func(addr address.Address, actorIn *states.Actor) error {
 		migration := migrations[actorIn.Code]
+
+		// This will be migrated at the end
+		if actorIn.Code == builtin0.VerifiedRegistryActorCodeID {
```
anorth, line 148: You can just check for the singleton address (`t080`), and if the address is not that, assert it's not a verified registry.

actors/migration/top.go
ZenGround0: Glad you pointed this out.  I will investigate.

ZenGround0: From @whyrusleeping its ok as policy to do this from burnt funds.  The concern is that we really want to make sure burnt funds has enough to do all of these so if it is a possibility that burnt funds won't cover ip debt we should take from the miner reserve.

anorth:
@ZenGround0 apart from lacking tests, I think this is ready to go. I've marked it ready for review. When it passes your manual checks, let's land it and keep building on top.

ZenGround0 (approved):

PR #1119, opened 2020-09-08
Reduce vesting cliff, penalties, min deal collateral.

anorth:
This is #1104, interpreted for v2.

- Add network version to runtime
- Implement reduced sector penalties and reward vesting cliff
- Reduce deal min provider collateral
- Fix tests missing expectations about rewards vesting after only 1 proving period

This version does not implement the `if rt.NetworkVersion() >= 1` tests because actors v2 is never intended to run on any before 2. To evaluate network versions 0 and 1, use `release/v0.9`. This demonstrates that those inline upgrades do not accumulate over time.

Based on c6ce28641f59c44af1594058f92bac67edb17664 from release/v0.9.

FYI @Stebalien @arajasek @magik6k

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -174,7 +174,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 
 	resolvedAddrs := make(map[addr.Address]addr.Address, len(params.Deals))
 	baselinePower := requestCurrentBaselinePower(rt)
-	networkQAPower := requestCurrentNetworkQAPower(rt)
+	networkRawPower, networkQAPower := requestCurrentNetworkPower(rt)
 
 	var newDealIds []abi.DealID
 	var st State
@@ -186,7 +186,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 
 		// All storage dealProposals will be added in an atomic transaction; this operation will be unrolled if any of them fails.
 		for di, deal := range params.Deals {
-			validateDeal(rt, deal, baselinePower, networkQAPower)
+			validateDeal(rt, deal, baselinePower, networkRawPower, networkQAPower)
 
 			if deal.Proposal.Provider != provider && deal.Proposal.Provider != providerRaw {
 				rt.Abortf(exitcode.ErrIllegalArgument, "cannot publish deals from different providers at the same time")
@@ -680,7 +680,7 @@ func validateDealCanActivate(proposal *DealProposal, minerAddr addr.Address, sec
 	return nil
 }
 
-func validateDeal(rt Runtime, deal ClientDealProposal, baselinePower, networkQAPower abi.StoragePower) {
+func validateDeal(rt Runtime, deal ClientDealProposal, baselinePower, networkRawPower, networkQAPower abi.StoragePower) {
 	if err := dealProposalIsInternallyValid(rt, deal); err != nil {
 		rt.Abortf(exitcode.ErrIllegalArgument, "Invalid deal proposal: %s", err)
 	}
@@ -722,7 +722,7 @@ func validateDeal(rt Runtime, deal ClientDealProposal, baselinePower, networkQAP
 	}
 
 	minProviderCollateral, maxProviderCollateral := DealProviderCollateralBounds(proposal.PieceSize, proposal.VerifiedDeal,
-		networkQAPower, baselinePower, rt.TotalFilCircSupply())
+		networkRawPower, networkQAPower, baselinePower, rt.TotalFilCircSupply(), rt.NetworkVersion())
 	if proposal.ProviderCollateral.LessThan(minProviderCollateral) || proposal.ProviderCollateral.GreaterThan(maxProviderCollateral) {
 		rt.Abortf(exitcode.ErrIllegalArgument, "Provider collateral out of bounds.")
 	}
@@ -783,11 +783,11 @@ func requestCurrentBaselinePower(rt Runtime) abi.StoragePower {
 }
 
 // Requests the current network total power and pledge from the power actor.
-func requestCurrentNetworkQAPower(rt Runtime) abi.StoragePower {
+func requestCurrentNetworkPower(rt Runtime) (rawPower, qaPower abi.StoragePower) {
 	pwret, code := rt.Send(builtin.StoragePowerActorAddr, builtin.MethodsPower.CurrentTotalPower, nil, big.Zero())
 	builtin.RequireSuccess(rt, code, "failed to check current power")
 	var pwr power.CurrentTotalPowerReturn
 	err := pwret.Into(&pwr)
 	builtin.RequireNoErr(rt, err, exitcode.ErrSerialization, "failed to unmarshal power total value")
-	return pwr.QualityAdjPower
+	return pwr.RawBytePower, pwr.QualityAdjPower
 }
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -186,7 +186,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 
 		// All storage dealProposals will be added in an atomic transaction; this operation will be unrolled if any of them fails.
 		for di, deal := range params.Deals {
-			validateDeal(rt, deal, baselinePower, networkRawPower, networkQAPower)
+			validateDeal(rt, deal, networkRawPower, networkQAPower, baselinePower)
 
 			if deal.Proposal.Provider != provider && deal.Proposal.Provider != providerRaw {
 				rt.Abortf(exitcode.ErrIllegalArgument, "cannot publish deals from different providers at the same time")
@@ -680,7 +680,7 @@ func validateDealCanActivate(proposal *DealProposal, minerAddr addr.Address, sec
 	return nil
 }
 
-func validateDeal(rt Runtime, deal ClientDealProposal, baselinePower, networkRawPower, networkQAPower abi.StoragePower) {
+func validateDeal(rt Runtime, deal ClientDealProposal, networkRawPower, networkQAPower, baselinePower abi.StoragePower) {
 	if err := dealProposalIsInternallyValid(rt, deal); err != nil {
 		rt.Abortf(exitcode.ErrIllegalArgument, "Invalid deal proposal: %s", err)
 	}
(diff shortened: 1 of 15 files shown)


actors/builtin/market/policy.go
```diff
@@ -5,15 +5,16 @@ import (
...
 )
 
 // The number of epochs between payment and other state processing for deals.
 const DealUpdatesInterval = builtin.EpochsInDay // PARAM_SPEC
 
 // The percentage of normalized cirulating
 // supply that must be covered by provider collateral in a deal
-var ProviderCollateralSupplyTarget = builtin.BigFrac{
-	Numerator:   big.NewInt(5), // PARAM_SPEC
+var ProviderCollateralSupplyTargetV1 = builtin.BigFrac{
```
anorth, line 16: I chose to keep naming these after the network version at which they came into effect, but we don't need to do this.

actors/builtin/market/policy.go
```diff
@@ -35,17 +36,16 @@ func dealPricePerEpochBounds(_ abi.PaddedPieceSize, _ abi.ChainEpoch) (min abi.T
 	return abi.NewTokenAmount(0), builtin.TotalFilecoin // PARAM_FINISH
 }
 
-func DealProviderCollateralBounds(pieceSize abi.PaddedPieceSize, verified bool, networkQAPower, baselinePower abi.StoragePower, networkCirculatingSupply abi.TokenAmount) (min abi.TokenAmount, max abi.TokenAmount) {
+func DealProviderCollateralBounds(pieceSize abi.PaddedPieceSize, verified bool, networkRawPower, networkQAPower, baselinePower abi.StoragePower,
+	networkCirculatingSupply abi.TokenAmount, _ runtime.NetworkVersion) (min, max abi.TokenAmount) {
```
anorth, line 40: I preserved the network version plumbing, even though it's unused, to aid using it in a possible future inline upgrade.

actors/builtin/miner/miner_test.go
```diff
@@ -3795,6 +3796,12 @@ func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.PreCommit
 		rt.ExpectSend(builtin.StorageMarketActorAddr, builtin.MethodsMarket.VerifyDealsForActivation, &vdParams, big.Zero(), &vdReturn, exitcode.Ok)
 	}
 	st := getState(rt)
+
+	pledgeDelta :=  immediatelyVestingFunds(rt, st).Neg()
```
anorth, line 3800: Our habit of stacking the vesting funds in order to pay penalties means we now have to deal with these funds actually vesting after 1 day.

actors/builtin/miner/monies.go
```diff
@@ -87,25 +96,27 @@ func PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate smoo
...
 	return big.Max(
-		PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate, qaSectorPower),
+		PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate, qaSectorPower, networkVersion),
 		big.Add(
 			twentyDayRewardAtActivation,
 			big.Div(
-				expectedReward,
-				big.NewInt(builtin.EpochsInDay)))) // (epochs*AttoFIL/day -> AttoFIL)
+				penalizedReward,
+				big.Mul(big.NewInt(builtin.EpochsInDay), TerminationRewardFactorV2.Denominator)))) // (epochs*AttoFIL/day -> AttoFIL)
```
anorth, line 119: Please check carefully

actors/builtin/miner/monies_test.go
```diff
@@ -121,19 +122,25 @@ func TestPledgePenaltyForTermination(t *testing.T) {
...
-		fee := miner.PledgePenaltyForTermination(dayReward, replacementAge, twentyDayReward, powerEstimate, power, rewardEstimate, dayReward, sectorAge)
+		oldPenalty := big.Div(
+			big.Product(oldDayReward, big.NewInt(oldSectorAgeInDays), miner.TerminationRewardFactorV2.Numerator),
+			miner.TerminationRewardFactorV2.Denominator,
+		)
+		newPenalty := big.Div(
+			big.Product(dayReward, big.NewInt(replacementAgeInDays), miner.TerminationRewardFactorV2.Numerator),
+			miner.TerminationRewardFactorV2.Denominator,
+		)
+		expectedFee := big.Sum(initialPledge, oldPenalty, newPenalty)
```
anorth, line 141: This was tricky. As far as I can tell, the prior `expectedFee` calculation was wrong. Note the second-last parameter was `dayReward`, I changed it to `oldDayReward`. And how was the +initialPledge excluded? 

Does this look right to you @acruikshank

actors/runtime/types.go
```diff
@@ -6,6 +6,18 @@ import (
...
+// vendoring and versioning the whole actor codebase).
+type NetworkVersion uint // FIXME move to types
+
+const (
+	NetworkVersion0 = NetworkVersion(iota) // specs-actors v0.9.3
+	NetworkVersion1                        // specs-actors v0.9.7
+	NetworkVersion2                        // specs-actors v2.0.?
+
+	NetworkVersionLatest = NetworkVersion2
+)
```
anorth, line 19: We might want this out in `go-state-types`, right now it's duplicated between branches.

actors/builtin/miner/monies_test.go
acruikshank: Yes, the new code looks right. I would replace `initialPledge` with `twentyDayReward` for clarity in the final sum (they happen to be equal).

That the old version even passed appears to be an unlucky coincidence that sectorAgeInDays happened to equal the pledge factor and oldBR was chosen to be double day reward (which produced a difference between dayReward and oldDayReward equal to IP).

acruikshank (approved):
There's a lot here, but it looks right AFAICT.

I'd like you to consider carefully adding version suffixes to constants, which add more noise than signal to my eyes. If we can't eliminate them when not required by an inline update, we should at least document exactly what they mean, what constants need to have them and what are the rules for updating them.

actors/builtin/market/policy.go
ZenGround0: I don't see the value in keeping that plumbing here.  It's not obvious that past upgrade sites are good predictors for future upgrade sites.  This plumbing is very easy to do lazily.  If we keep it around without actually referencing it we add friction to tracing through the code to find existing upgrade sites.

actors/builtin/market/market_actor.go
```diff
@@ -680,7 +680,7 @@ func validateDealCanActivate(proposal *DealProposal, minerAddr addr.Address, sec
 	return nil
 }
 
-func validateDeal(rt Runtime, deal ClientDealProposal, baselinePower, networkQAPower abi.StoragePower) {
+func validateDeal(rt Runtime, deal ClientDealProposal, baselinePower, networkRawPower, networkQAPower abi.StoragePower) {
```
ZenGround0, line 683: nit: would be slightly better to rearrange inputs so that the "raw, qa, baseline" order is the same everywhere to prevent swapping mistakes.  But if there are a lot of changes necessary forget about it.

actors/builtin/miner/miner_test.go
ZenGround0: Creating a better fee stacking and penalty payment expectation assertion interface feels like the ripest of the low hanging fruits growing on the miner test harness tree. 🥝 🍌 🍒 🍏 🍓

actors/builtin/market/policy.go
ZenGround0: Similar thoughts about the version suffix on constants.  When reading the code looking at specs-actors/v2 gives the context for which protocol epochs these constants apply to.  If we need to go back and refer to previous constants we can always add on suffixes as needed here since we'd be doing an upgrade anyway.  We can also make use of importing constants from previous branches.

actors/builtin/miner/monies.go
```diff
@@ -76,8 +83,10 @@ func PledgePenaltyForDeclaredFault(rewardEstimate, networkQAPowerEstimate smooth
 // The penalty for a newly faulty sector that was been declared in advance.
 // It is a projection of the expected reward earned by the sector.
 // Also known as "SP(t)"
-func PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate smoothing.FilterEstimate, qaSectorPower abi.StoragePower) abi.TokenAmount {
-	return ExpectedRewardForPower(rewardEstimate, networkQAPowerEstimate, qaSectorPower, UndeclaredFaultProjectionPeriod)
+func PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate smoothing.FilterEstimate,
+	qaSectorPower abi.StoragePower, _ runtime.NetworkVersion) abi.TokenAmount {
+	projectionPeriod := UndeclaredFaultProjectionPeriodV1
```
ZenGround0, line 88: nit: unnecessary assignment

actors/builtin/miner/monies.go
```diff
@@ -87,25 +96,27 @@ func PledgePenaltyForUndeclaredFault(rewardEstimate, networkQAPowerEstimate smoo
 func PledgePenaltyForTermination(dayReward abi.TokenAmount, sectorAge abi.ChainEpoch,
 	twentyDayRewardAtActivation abi.TokenAmount, networkQAPowerEstimate smoothing.FilterEstimate,
 	qaSectorPower abi.StoragePower, rewardEstimate smoothing.FilterEstimate, replacedDayReward abi.TokenAmount,
-	replacedSectorAge abi.ChainEpoch,
+	replacedSectorAge abi.ChainEpoch, networkVersion runtime.NetworkVersion,
 ) abi.TokenAmount {
-	// max(SP(t), BR(StartEpoch, 20d) + BR(StartEpoch, 1d)*min(SectorAgeInDays, 70))
+	// max(SP(t), BR(StartEpoch, 20d) + BR(StartEpoch, 1d) * factor * min(SectorAgeInDays, 140))
```
ZenGround0, line 101: nit: maybe terminationRewardFactor or something along those lines?

ZenGround0 (approved):
I moderately disagree with keeping version clutter around after the upgrade.  But this looks sound.

actors/builtin/market/policy.go
ZenGround0: > We could introduce version suffixes only for inline updates and insist that versioned constants become unversioned when we remove inline code at the next release update.

👍

ZenGround0:
Relevant question -- how should we version specs actors branches in the face of several inline protocol version bumps?  This time we got lucky because go mod hates v1 so we naturally skipped `specs-actors/v1` and now our branch name matches the protocol version.

We could try to keep matching versions with branch names.  To do so we would need to skip a major version every time there is an inline upgrade.  

The alternative is to have specs-actors branch numbers that march forward independently of protocol version.

actors/builtin/market/policy.go
anorth: Ok

actors/builtin/market/policy.go
anorth: Ok, I'll remove the plumbing

actors/builtin/market/market_actor.go
anorth: Thanks, I had intended that

PR #1122, opened 2020-09-08
Rework election eligibilty predicates

### LINKED ISSUE #1102: Miner eligibility check does not check for min miner size
As I explain in this comment https://github.com/filecoin-project/lotus/issues/3378#issuecomment-687526380, there are three checks to be done by lotus:

- Check that miner is valid at current epoch
- Check that miner is eligible at current epoch
- Check that miner had min miner size at WinningPoStSectorSet lookback

Check 1 and 3 are currently already integrated in lotus, check 2 must be added (https://github.com/filecoin-project/lotus/issues/3378 is tracking this addition).

Check 2 should not check min miner size (hence these lines should be removed)

Why? Eligibility and validity check are not related to power accounting, but to debts and faults. Anything related to power accounting must happen with the lookback.

anorth:
The miner election eligibility predicates need to be split into one that operates on the parent state, and one at the Winning PoSt lookback.

- Move the predicates into a new `states` package intended for use from outside the actors themselves
- In `MinerEligibleForElection` change the consensus min power check to a non-zero claim check
- Add `MinerEligibleForElectionAtPoStLookback` which checks for consensus min power
- Rework tests to not use the mock runtime or other higher level stuff

Closes #1102
FYI @nicola @whyrusleeping @magik6k

### DIFF
--- a/actors/states/election.go
+++ b/actors/states/election.go
@@ -1,0 +1,55 @@
+package states
+
+import (
+	addr "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/reward"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+)
+
+// Checks for miner election eligibility.
+// A miner must satisfy conditions on both the immediate parent state, as well as state at the
+// Winning PoSt election lookback.
+
+// Tests whether a miner is eligible to win an election given the immediately prior state of the
+// miner and system actors.
+func MinerEligibleForElection(store adt.Store, mstate *miner.State, pstate *power.State, rstate *reward.State, maddr addr.Address, currEpoch abi.ChainEpoch) (bool, error) {
+	// Non-empty power claim.
+	if claim, found, err := pstate.GetClaim(store, maddr); err != nil {
+		return false, err
+	} else if !found {
+		return false, err
+	} else if claim.QualityAdjPower.LessThanEqual(big.Zero()) {
+		return false, err
+	}
+
+	// No fee debt.
+	if !mstate.IsDebtFree() {
+		return false, nil
+	}
+
+	// No active consensus faults.
+	if mInfo, err := mstate.GetInfo(store); err != nil {
+		return false, err
+	} else if miner.ConsensusFaultActive(mInfo, currEpoch) {
+		return false, nil
+	}
+
+	// IP requirement is sufficient to cover fee for a consensus fault
+	electionRequirement := miner.ConsensusFaultPenalty(rstate.ThisEpochReward)
+	if mstate.InitialPledge.LessThan(electionRequirement) {
+		return false, nil
+	}
+
+	return true, nil
+}
+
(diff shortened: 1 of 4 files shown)


actors/states/election_test.go
```diff
@@ -0,0 +1,253 @@
...
+	ctx := context.Background()
+	tenFIL := big.Mul(big.NewInt(1e18), big.NewInt(10))
+	store := ipld.NewADTStore(ctx)
+	proofType := abi.RegisteredSealProof_StackedDrg32GiBV1
+	pwr := abi.NewStoragePower(1)
+
+	owner := tutil.NewIDAddr(t, 100)
+	maddr := tutil.NewIDAddr(t, 101)
+
+	// Construct valid power state
```
ZenGround0, line 35: nit: out of date comment

actors/states/election_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	t.Run("miner eligible", func(t *testing.T) {
+		mstate := constructMinerState(ctx, t, store, owner)
+		pstate := constructPowerStateWithMiner(t, store, maddr, pwr, proofType)
+		rstate := constructRewardState(t, tenFIL)
```
ZenGround0, line 40: love it

actors/states/election_test.go
```diff
@@ -0,0 +1,253 @@
...
+		mstate.InitialPledge = miner.ConsensusFaultPenalty(rstate.ThisEpochReward)
+		mstate.FeeDebt = abi.NewTokenAmount(1000)
+
+		currEpoch := abi.ChainEpoch(100000)
+		eligible, err := states.MinerEligibleForElection(store, mstate, pstate, rstate, maddr, currEpoch)
+		require.NoError(t, err)
+		assert.False(t, eligible)
+	})
+
+	t.Run("ip requirement below consensus fault penalty", func(t *testing.T) {
```
ZenGround0, line 92: nit: ip requirement is now just ip or initial pledge since accounting refactor

ZenGround0 (approved):
LGTM

actors/states/election.go
```diff
@@ -0,0 +1,55 @@
...
+	electionRequirement := miner.ConsensusFaultPenalty(rstate.ThisEpochReward)
+	if mstate.InitialPledge.LessThan(electionRequirement) {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// Tests whether a miner is eligible for election given a Winning PoSt lookback state.
+func MinerEligibleForElectionAtPoStLookback(store adt.Store, pstate *power.State, mAddr addr.Address) (bool, error) {
```
acruikshank, line 52: Nit. This function returning true doesn't imply the miner was eligible at PoSt lookback. Also this method doesn't deal with lookback directly. `MinerMeetsConsensusMinimumAtPoStLookback`would be more accurate. `MinerMeetsPoStLookbackEligibilityRequirement`? I'm not convinced this function adds much over calling `pstate.MinerNominalPowerMeetsConsensusMinimum(store, mAddr)` directly. 

The comment kind of gets at it, but it would be better to to directly call out "pstate must be the prior state of the power actor at the PoSt lookback epoch". This is an unusual thing in the actor code.

acruikshank (approved):

actors/states/election.go
anorth: What this adds is that when the eligibility requirements change, we can encapsulate that change here (or at least force downstream ack of it through a type signature change). Such a change is likely to be accompanied by other actor changes so we can cover the whole thing and reduce coordination with Lotus.

Thanks for the name suggestion.

PR #1129, opened 2020-09-09
prevent faults from double subtracting cc upgrade power

### LINKED ISSUE #1100: Miner: Declaring a replaced CC sector faulty can result in a sector existing in an expiration queue twice
Replaced CC sectors get their expiration rescheduled in `ConfirmSectorProofsValid`:

https://github.com/filecoin-project/specs-actors/blob/b1153df2ab47c2db86b46b04fa94592d4adf0599/actors/builtin/miner/miner_actor.go#L770-L772

This means that, within the replaced sectors' partitions, they are removed from `partition.ExpirationsEpochs` at their "on-time expiry," and re-added at a `newExpiration` (the end of their next deadline window):

https://github.com/filecoin-project/specs-actors/blob/b1153df2ab47c2db86b46b04fa94592d4adf0599/actors/builtin/miner/expiration_queue.go#L177-L190

Note that replaced CC sectors cannot be faulty at the time they are replaced. However, they can be declared faulty after the fact. When a replaced CC sector is declared faulty, `expiration_queue.RescheduleAsFaults` is invoked, which iterates over fault declarations by on-time expiration. For each expiration, the queue is queried for a corresponding `ExpirationSet`:

https://github.com/filecoin-project/specs-actors/blob/b1153df2ab47c2db86b46b04fa94592d4adf0599/actors/builtin/miner/expiration_queue.go#L201-L207

The sectors in question may, or may not exist in this `ExpirationSet`; no existence check is made.

#### Results

If the now-faulty sector's on-time expiration is less than or equal to the fault expiration, the expiration set's active and faulty power totals are adjusted erroneously:

https://github.com/filecoin-project/specs-actors/blob/b1153df2ab47c2db86b46b04fa94592d4adf0599/actors/builtin/miner/expiration_queue.go#L208-L214

If the now-faulty sector's on-time expiration would come _after_ the fault expiration, this method reschedules the sector for expiration at the fault expiration epoch. This rescheduling is done by first removing the sector from `ExpirationSet.OnTimeSectors` (note that this doesn't do anything - the sector isn't there):

https://github.com/filecoin-project/specs-actors/blob/b1153df2ab47c2db86b46b04fa94592d4adf0599/actors/builtin/miner/expiration_queue.go#L216-L222

Then, the sector is re-added to the queue at the fault expiry epoch:

https://github.com/filecoin-project/specs-actors/blob/b1153df2ab47c2db86b46b04fa94592d4adf0599/actors/builtin/miner/expiration_queue.go#L234-L241

In this case, the sector ends up in the expiration queue twice: at the end of its next deadline window, and at the fault expiry epoch. This should never happen and may have several unintended consequences - improper power accounting, cron callback aborts, and more.

#### Conclusion

This method of inserting a replaced CC sector into an expiration queue twice is possible via `Miner.DeclareFaults` as well as `Miner.SubmitWindowedPoSt` (where the sector is marked "skipped"). 

The primary cause of the issue is a failure to check for the existence of referenced sectors within `expiration_queue.RescheduleAsFaults`.

I did not find any other miner methods that handle replaced CC sectors incorrectly, but **further investigation is warranted.** In particular, the expiration queue implementation often performs operations without making checks to ensure that certain invariants are maintained. Consider reviewing the file as a whole and adding checks like those mentioned in #1064, as well as existence checks throughout.

acruikshank:
closes #1100

### Motivation

All sectors live within an expiration set within an expiration queue within a partition within a deadline. This expiration set is the set at the sectors `Expiration` field. When we reschedule expirations (say for a CC upgrade or a fault), we move it into a different set to trigger an earlier expiration. We do not change the sector's `Expiration` field, so the new set can not necessarily be determined from the sector's on chain info.

Whenever we perform an operation that affects sector expirations, we group the sectors by expiration and operate on the set _en mass_. If sectors have been rescheduled, we end up altering expiration set stats for sectors that are no longer there, and failing to update the stats for the sets to which they have moved. In the case of declaring a fault for a cc upgraded sector, the end result is that we subtract the sectors power twice from the miner's claim (among other accounting mistakes).

The solution implemented in this PR is to check whether sectors expected to be in an expiration set are actually there. Sectors that are not are grouped by iterating through the expiration queue's sets instead. Not adding sectors to groups for expiration sets don't contain them should be sufficient to avoid breaking accounting invariants, whereas finding the sectors in the correct sets should deliver the intended behavior.

### Proposed Changes

1. Implement `findSectorsByExpiration` that groups sectors in the correct expiration set even if they have been rescheduled.
2. Use `findSectorsByExpiration` in `ExpirationQueue.RescheduleAsFaults` and `ExpirationQueue.removeActiveSectors`.
3. Fix regression discovered in this PR that prevented faulty expiring sectors from being penalized in their last proving period cron.
4. Add test for cc upgrading a sector before it is proven.
5. Add test for declaring a fault for a cc replaced sector before it expires.
6. Add test for skipping a cc replaced sector in its last PoSt.
7. Add test for missed PoSt in proving deadline where cc replaced sector expires.
8. Add test for terminating cc replaced sector before it expires.
9. Add test for extending expiration of cc replaced sector before it expires.
10. Add test for faulting and recovering a cc replaced sector before it expires.
11. Rearranged miner tests so tests related to cc upgrades have their own section.

Without the changes to grouping, only 4 and 7 would have behaved as expected.

__Note 11.__: Tests that do not appear in the list have been moved without change and don't require the scrutiny of the rest.

### DIFF
--- a/actors/builtin/miner/expiration_queue.go
+++ b/actors/builtin/miner/expiration_queue.go
@@ -224,7 +224,12 @@ func (q ExpirationQueue) RescheduleAsFaults(newExpiration abi.ChainEpoch, sector
 	rescheduledPower := NewPowerPairZero()
 
 	// Group sectors by their target expiration, then remove from existing queue entries according to those groups.
-	for _, group := range groupSectorsByExpiration(ssize, sectors, q.quant) {
+	groups, err := q.groupSectorsByExpirationWithRescheduled(ssize, sectors, q.quant)
+	if err != nil {
+		return NewPowerPairZero(), err
+	}
+
+	for _, group := range groups {
 		var err error
 		var es ExpirationSet
 		if err = q.mustGet(group.epoch, &es); err != nil {
@@ -739,4 +744,113 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
 		return sectorEpochSets[i].epoch < sectorEpochSets[j].epoch
 	})
 	return sectorEpochSets
+}
+
+// Works like groupSectorsByExpiration but finds all sectors rescheduled out their expected expiration set and groups them as well
+func (q *ExpirationQueue) groupSectorsByExpirationWithRescheduled(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo, quant QuantSpec) ([]sectorEpochSet, error) {
+	sectorsByExpiration := make(map[abi.ChainEpoch][]*SectorOnChainInfo)
+
+	for _, sector := range sectors {
+		qExpiration := quant.QuantizeUp(sector.Expiration)
+		sectorsByExpiration[qExpiration] = append(sectorsByExpiration[qExpiration], sector)
+	}
+
+	sectorEpochSets := make([]sectorEpochSet, 0, len(sectorsByExpiration))
+	allRescheduled := bitfield.New()
+
+	// This map iteration is non-deterministic but safe because we sort by epoch below.
+	for expiration, epochSectors := range sectorsByExpiration { //nolint:nomaprange // result is subsequently sorted
+		var es ExpirationSet
+		if err := q.mustGet(expiration, &es); err != nil {
+			return nil, err
+		}
+
+		sectorNumbers := []uint64{}
+		totalPower := NewPowerPairZero()
+		totalPledge := big.Zero()
+		for _, sector := range epochSectors {
+			if found, err := es.OnTimeSectors.IsSet(uint64(sector.SectorNumber)); err != nil {
+				return nil, err
+			} else if found {
+				sectorNumbers = append(sectorNumbers, uint64(sector.SectorNumber))
+				totalPower = totalPower.Add(PowerForSector(sectorSize, sector))
+				totalPledge = big.Add(totalPledge, sector.InitialPledge)
+			} else {
+				allRescheduled.Set(uint64(sector.SectorNumber))
+			}
+		}
+		if len(sectorNumbers) > 0 {
+			sectorEpochSets = append(sectorEpochSets, sectorEpochSet{
+				epoch:   expiration,
+				sectors: sectorNumbers,
+				power:   totalPower,
+				pledge:  totalPledge,
+			})
+		}
+	}
+
+	if empty, err := allRescheduled.IsEmpty(); err != nil {
+		return nil, err
+	} else if !empty {
+		err := q.traverseMutate(func(epoch abi.ChainEpoch, es *ExpirationSet) (changed, keepGoing bool, err error) {
+			intersect, err := bitfield.IntersectBitField(es.OnTimeSectors, allRescheduled)
+			if err != nil {
+				return false, false, err
+			}
+
+			group, err := groupSectorSet(sectorSize, sectors, &intersect, epoch)
+			if err != nil {
+				return false, false, err
+			}
+			sectorEpochSets = append(sectorEpochSets, group)
+
+			allRescheduled, err = bitfield.SubtractBitField(allRescheduled, intersect)
+			if err != nil {
+				return false, false, err
+			}
+
+			empty, err := allRescheduled.IsEmpty()
+			if err != nil {
+				return false, false, err
+			}
+			return false, !empty, nil
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	sort.Slice(sectorEpochSets, func(i, j int) bool {
+		return sectorEpochSets[i].epoch < sectorEpochSets[j].epoch
+	})
+	return sectorEpochSets, nil
+}
+
+// Takes a slice of sector infos a bitfield of sector numbers and returns a single group for all bitfield sectors
+func groupSectorSet(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo, set *bitfield.BitField, expiration abi.ChainEpoch) (sectorEpochSet, error) {
+	var filteredSectors []*SectorOnChainInfo
+	if filter, err := set.AllMap(uint64(len(sectors))); err != nil {
+		return sectorEpochSet{}, err
+	} else {
+		for _, sector := range sectors {
+			if _, ok := filter[uint64(sector.SectorNumber)]; ok {
+				filteredSectors = append(filteredSectors, sector)
+			}
+		}
+	}
+
+	sectorNumbers := make([]uint64, len(filteredSectors))
+	totalPower := NewPowerPairZero()
+	totalPledge := big.Zero()
+	for i, sector := range filteredSectors {
+		sectorNumbers[i] = uint64(sector.SectorNumber)
+		totalPower = totalPower.Add(PowerForSector(sectorSize, sector))
+		totalPledge = big.Add(totalPledge, sector.InitialPledge)
+	}
+	return sectorEpochSet{
+		epoch:   expiration,
+		sectors: sectorNumbers,
+		power:   totalPower,
+		pledge:  totalPledge,
+	}, nil
 }
--- a/actors/builtin/miner/expiration_queue.go
+++ b/actors/builtin/miner/expiration_queue.go
@@ -794,7 +794,7 @@ func (q *ExpirationQueue) findSectorsByExpiration(sectorSize abi.SectorSize, sec
 	if len(allRemaining) > 0 {
 		err := q.traverse(func(epoch abi.ChainEpoch, es *ExpirationSet) (bool, error) {
 			// If this set's epoch is one of our declared epochs, we've already processed it in the loop above,
-			// so skip processing here.
+			// so skip processing here. Sectors rescheduled to this epoch would have been included in the earlier processing.
 			if _, found := declaredExpirations[epoch]; found {
 				return true, nil
 			}
--- a/actors/builtin/miner/expiration_queue.go
+++ b/actors/builtin/miner/expiration_queue.go
@@ -224,7 +224,7 @@ func (q ExpirationQueue) RescheduleAsFaults(newExpiration abi.ChainEpoch, sector
 	rescheduledPower := NewPowerPairZero()
 
 	// Group sectors by their target expiration, then remove from existing queue entries according to those groups.
-	groups, err := q.groupSectorsByExpirationWithRescheduled(ssize, sectors, q.quant)
+	groups, err := q.groupSectorsByExpirationWithRescheduled(ssize, sectors)
 	if err != nil {
 		return NewPowerPairZero(), err
 	}
@@ -615,7 +615,12 @@ func (q ExpirationQueue) removeActiveSectors(sectors []*SectorOnChainInfo, ssize
 	noFaultyPower := NewPowerPairZero()
 
 	// Group sectors by their expiration, then remove from existing queue entries according to those groups.
-	for _, group := range groupSectorsByExpiration(ssize, sectors, q.quant) {
+	groups, err := q.groupSectorsByExpirationWithRescheduled(ssize, sectors)
+	if err != nil {
+		return bitfield.BitField{}, NewPowerPairZero(), big.Zero(), err
+	}
+
+	for _, group := range groups {
 		sectorsBf := bitfield.NewFromSet(group.sectors)
 		if err := q.remove(group.epoch, sectorsBf, noEarlySectors, group.power, noFaultyPower, group.pledge); err != nil {
 			return bitfield.BitField{}, NewPowerPairZero(), big.Zero(), err
@@ -747,11 +752,11 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
 }
 
 // Works like groupSectorsByExpiration but finds all sectors rescheduled out their expected expiration set and groups them as well
-func (q *ExpirationQueue) groupSectorsByExpirationWithRescheduled(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo, quant QuantSpec) ([]sectorEpochSet, error) {
+func (q *ExpirationQueue) groupSectorsByExpirationWithRescheduled(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo) ([]sectorEpochSet, error) {
 	sectorsByExpiration := make(map[abi.ChainEpoch][]*SectorOnChainInfo)
 
 	for _, sector := range sectors {
-		qExpiration := quant.QuantizeUp(sector.Expiration)
+		qExpiration := q.quant.QuantizeUp(sector.Expiration)
 		sectorsByExpiration[qExpiration] = append(sectorsByExpiration[qExpiration], sector)
 	}
 
--- a/actors/builtin/miner/expiration_queue.go
+++ b/actors/builtin/miner/expiration_queue.go
@@ -318,15 +318,15 @@ func (q ExpirationQueue) RescheduleAllAsFaults(faultExpiration abi.ChainEpoch) e
 		return nil
 	}
 
-	// Add rescheduled sectors to new expiration as early-terminating and faulty.
-	allRescheduled, err := bitfield.MultiMerge(rescheduledSectors...)
+	// Add missing sectors to new expiration as early-terminating and faulty.
+	allNotFound, err := bitfield.MultiMerge(rescheduledSectors...)
 	if err != nil {
 		return xerrors.Errorf("failed to merge rescheduled sectors: %w", err)
 	}
 	noOnTimeSectors := bitfield.New()
 	noActivePower := NewPowerPairZero()
 	noOnTimePledge := abi.NewTokenAmount(0)
-	if err = q.add(faultExpiration, noOnTimeSectors, allRescheduled, noActivePower, rescheduledPower, noOnTimePledge); err != nil {
+	if err = q.add(faultExpiration, noOnTimeSectors, allNotFound, noActivePower, rescheduledPower, noOnTimePledge); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


anorth (commented):
I haven't fully reviewed this yet, but I think this problem is more widespread. Every call to groupSectorsByExpiration is suspect if it expects to find the sectors there. Checking the call tree:
- EQ.removeActiveSectors, must be updated to support replaced sectors for:
    - EQ.RemoveSectors (from TerminateSectors)
    - EQ.ReplaceSectors (from ExtendSectorExpiration)
    - EQ.RescheduleExpirations (from CC upgrade - should work but do nothing)
- EQ.AddActiveSectors: ok, these sectors aren't in the queue yet
- EQ.RescheduleAsFaults: handled in this PR already

I think we should rename for clarity:
- `groupSectorsByExpiration` -> `groupNewSectorsByDeclaredExpiration` (only for AddActiveSectors)
- `groupSectorsByExpirationWithRescheduled ` -> `eq.findSectorsByExpiration`

Tests we need:
- Terminate a replaced sector before it expires
- Extend a replaced sector's expiration
- Replace a sector that's already been replaced before it expires
- Declare fault on a replaced sector before it expires
- Fail to prove a replaced sector before it expires (detected fault)

cc @wadealexc

actors/builtin/miner/miner_test.go
```diff
@@ -948,6 +948,127 @@ func TestCommitments(t *testing.T) {
...
+		})
+
+		// At proving period cron expect to pay declared fee for new sector
+		// and to have its pledge requirement deducted indicating it has expired.
+		//oldPower := miner.QAPowerForSector(actor.sectorSize, oldSector)
+		//expectedPowerDelta := miner.NewPowerPair(big.NewInt(int64(actor.sectorSize)), oldPower).Neg()
+		//expectedFee := miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, oldPower)
+		//actor.applyRewards(rt, expectedFee, big.Zero())
+		actor.onDeadlineCron(rt, &cronConfig{
+			//detectedFaultsPenalty:     expectedFee,
```
anorth, line 1066: I think the fix is to revert my ill-advised comment [here](https://github.com/filecoin-project/specs-actors/pull/1009#discussion_r475382622).

wadeAlexC:
I don't have the bandwidth to review this fully yet (but I will make time before the end of the week).

Regarding @anorth's comments, yes - I agree that the problem is more widespread than what was mentioned in #1100:

- `TerminateSectors` invokes `EQ.RemoveSectors`, which doesn't support replaced CC (I believe it aborts)
- `ExtendSectorExpiration` invokes `EQ.ReplaceSectors`, which doesn't support replaced CC (I believe it aborts)
- `SubmitWindowedPoSt` invokes:
  - `EQ.RescheduleAsFaults` - handled in this PR
  - `EQ.RescheduleRecovered` - doesn't support replacements. This means that declaring a replacement faulty, then recovering it would result in its expiry being set to its original, on-time expiry. (via `EQ.AddActiveSectors`)
- `ConfirmSectorProofsValid` invokes `EQ.RescheduleExpirations`, which won't find replacements (I think it aborts). I think this one is working as intended (but a test would be nice).
- `CompactPartitions` invokes `EQ.AddActiveSectors` - I didn't get a chance to review this method fully, so tests would be good to make sure this works as intended.
- `handleProvingDeadline` invokes:
  - `EQ.RescheduleAllAsFaults`: this iterates over the EQ rather than using `groupSectorsByExpiration`. I _think_ it works, but a test is needed.
  - `EQ.PopUntil`: (when popping expired sectors). This iterates over the EQ rather than grouping by expiry. I _think_ it works, but a test is needed.

Edge cases to consider:
- The "new expiry" calculated for a replaced CC sector may be either before, equal to, or even after its expected expiry
- Replaced CC sectors may be `Unproven`

Tests needed (@anorth's tests plus a few more):
- In the same epoch a CC sector is proven (`ConfirmSPValid`), precommit a replacement for it
- Extend a replaced sector's expiry via `ExtendSectorExpiration`
- Terminate a replaced sector before it expires
- Replace a sector that's already been replaced before it expires
- Replace an Unproven sector
- Declare a replaced sector faulty via `DeclareFaults`
  - Then declare it recovered
- "Declare" a replaced sector faulty via "skipping" (`SubmitWindowedPoSt`)
  - Then declare it recovered
- "Declare" a replaced sector faulty via missed post (picked up in `handleProvingDeadline`)
  - Then declare it recovered

acruikshank:
@wadeAlexC I believe I've added all the tests you've requested. "Replace a sector that's already been replaced before it expires" was already in place.

actors/builtin/miner/expiration_queue.go
```diff
@@ -698,3 +712,122 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
...
+			intersect, err := bitfield.IntersectBitField(es.OnTimeSectors, allRescheduled)
+			if err != nil {
+				return false, err
+			}
+
+			group, err := groupSectorSet(sectorSize, sectors, &intersect, es, epoch)
+			if err != nil {
+				return false, err
+			}
+			sectorEpochSets = append(sectorEpochSets, group)
```
anorth, line 778: What if there's already a sectorEpochSet for this epoch, because some other sector had a declared expiration there? I think this will produce another ses for the epoch.

It might be that callers are ok with this, but it's a bit of a trap. Also the `sort.Slice` below is unstable so non-deterministic with equal keys.

I think best to accumulate these in a map by epoch instead, and handle the possibility of the ses already existing at this point.

actors/builtin/miner/expiration_queue.go
```diff
@@ -698,3 +712,122 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
...
+
+	// if sectors are missed because they have been rescheduled, traverse expiration sets to find and group them.
+	if empty, err := allRescheduled.IsEmpty(); err != nil {
+		return nil, err
+	} else if !empty {
+		err := q.traverse(func(epoch abi.ChainEpoch, es *ExpirationSet) (bool, error) {
+			intersect, err := bitfield.IntersectBitField(es.OnTimeSectors, allRescheduled)
+			if err != nil {
+				return false, err
+			}
```
anorth, line 772: Blocking: this needs to check `es.EarlySectors` as well. Especially for the case where the sector has moved because it was rescheduled. I'm a bit surprised this escaped the great testing.

I think to hit this case we need to fault a sector (thus rescheduling it) and then terminate it (the other call path to this point, since it can't be faulted again).

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+			rt.Reset()
+		}
+
+		{ // Target must not be terminated
+			params := *upgradeParams
+			st := getState(rt)
+			prevState := *st
+			quant := st.QuantSpecForDeadline(dlIdx)
+			deadlines, err := st.LoadDeadlines(rt.AdtStore())
+			require.NoError(t, err)
```
anorth, line 1074: The harness has helpers like `getDeadlines`, `getDeadline` to avoid all this noise.

Edit: I see you just moved this. Minimising the moves would help make the real delta here clear.

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+		// The new one expires when the old one used to.
+		// The partition is registered with an expiry at both epochs.
+		pIdx := uint64(0)
+		dQueue := actor.collectDeadlineExpirations(rt, deadline)
+		dlInfo := miner.NewDeadlineInfo(st.ProvingPeriodStart, dlIdx, rt.Epoch())
+		quantizedExpiration := miner.QuantSpecForDeadline(dlInfo).QuantizeUp(oldSector.Expiration)
+		assert.Equal(t, map[abi.ChainEpoch][]uint64{
+			dlInfo.NextNotElapsed().Last(): {pIdx},
+			quantizedExpiration:            {pIdx},
+		}, dQueue)
```
anorth, line 1246: We could thin down lots of these state inspections that are adequately covered in other tests, and focus just on the specific flow

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+		// advance to sector proving deadline
+		dlInfo = actor.deadline(rt)
+		for dlIdx != dlInfo.Index {
+			advanceDeadline(rt, actor, &cronConfig{})
+			dlInfo = actor.deadline(rt)
+		}
+
+		// submit post for new sector
+		// oldSector is faulty, so expect newSector twice in validation; once for its proof and once to replace oldSector
+		// power is added for new sector and no penalties are paid yet
```
anorth, line 1286: I don't understand this comment about "twice in validation".
Edit: oh I get it, referring to the new sector info masking the old one. But the power delta is just for new sector

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+		newPower := miner.QAPowerForSector(actor.sectorSize, newSector)
+		partitions := []miner.PoStPartition{
+			{Index: pIdx, Skipped: bitfield.New()},
+		}
+		actor.submitWindowPoSt(rt, dlInfo, partitions, []*miner.SectorOnChainInfo{newSector, newSector}, &poStConfig{
+			expectedPowerDelta: miner.NewPowerPair(big.NewInt(int64(actor.sectorSize)), newPower),
+			expectedPenalty:    big.Zero(),
+		})
+
+		// At proving period cron expect to pay declared fee for new sector
```
anorth, line 1297: for old sector

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+		// and to have its pledge requirement deducted indicating it has expired.
+		// Importantly, power is NOT removed, because it was taken when fault was declared.
+		oldPower := miner.QAPowerForSector(actor.sectorSize, oldSector)
+		expectedFee := miner.PledgePenaltyForDeclaredFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, oldPower)
+		actor.applyRewards(rt, expectedFee, big.Zero())
+		actor.onDeadlineCron(rt, &cronConfig{
+			detectedFaultsPenalty:     expectedFee,
+			expiredSectorsPledgeDelta: oldSector.InitialPledge.Neg(),
+			expectedEnrollment:        rt.Epoch() + miner.WPoStChallengeWindow,
```
anorth, line 1306: Add the powerDelta = zero explicitly?

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+			{Index: pIdx, Skipped: bf(uint64(oldSector.SectorNumber))},
+		}
+		actor.submitWindowPoSt(rt, dlInfo, partitions, []*miner.SectorOnChainInfo{newSector, newSector}, &poStConfig{
+			expectedPowerDelta: expectedPowerDelta,
+			expectedPenalty:    expectedFee,
+		})
+
+		// At proving period cron expect to pay declared fee for old (now faulty) sector
+		// and to have its pledge requirement deducted indicating it has expired.
+		// Importantly, power is NOT removed, because it was taken when fault was declared.
```
anorth, line 1353: was taken when the sector was skipped in Window PoSt

anorth (commented):
Great work, nearly there, ❤️  the tests.

actors/builtin/miner/miner_test.go
```diff
@@ -899,53 +1057,509 @@ func TestCommitments(t *testing.T) {
...
+		quantizedExpiration = miner.QuantSpecForDeadline(dlInfo).QuantizeUp(oldSector.Expiration)
+
+		// deadling marks expirations for partition at expiration epoch
+		assert.Equal(t, map[abi.ChainEpoch][]uint64{
+			dlInfo.NextNotElapsed().Last(): {pIdx},
+			expectedReplacedExpiration:     {pIdx},
+			quantizedExpiration:            {pIdx},
+		}, dQueue)
+
+		// but partitions expiration set at that epoch is empty
```
wadeAlexC, line 1268: Why is it empty?

wadeAlexC: Shouldn't `DeclareFaults` place the old sector in `EarlySectors` at the fault expiration epoch?

wadeAlexC (commented):
Looks good for the most part. Left some comments.

actors/builtin/miner/expiration_queue.go
acruikshank: Good catch. I rework the algorithm altogether. It now just intersects `es.OnTimeSectors` with all the sectors it hasn't found yet to form the group and then computes the remainder. It iterates through the expected iterations first and then goes sequentially for what is left. I think it's a little more straightforward, and fixes this issue.

actors/builtin/miner/expiration_queue.go
acruikshank: I'm with @wadeAlexC that I don't think we could exercise this code. If we stop assuming that grouped sectors can be a mix of faulty and non-faulty sectors, our downstream accounting gets a lot more complicated. We could add an assertion that no sectors in the set are early. I tried it an the tests still pass.

actors/builtin/miner/expiration_queue.go
```diff
@@ -698,3 +716,126 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
...
+
+	expirationGroups := make([]sectorExpirationSet, 0, len(sectorsByExpiration))
+	allRescheduled := bitfield.New()
+
+	// This map iteration is non-deterministic but safe because we sort by epoch below.
+	for expiration, epochSectors := range sectorsByExpiration { //nolint:nomaprange // result is subsequently sorted
+		var es ExpirationSet
+		if err := q.mustGet(expiration, &es); err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
acruikshank: There is no consequence for using `mayGet` here. I went ahead and changed it. If a set is erroneously missing, the sector will be left in `allNotFound` and trigger an error at the end of the grouping.

actors/builtin/miner/miner_test.go
acruikshank: Since the sector had an on time expiration earlier than the fault expiration, it is left in its original expiration set, but its power is moved from active to faulty. This is checking that the expiration set at the fault expiration is empty, indicating it was not moved. Prior to this PR, it would have been moved to this expiration set and left in its original expiration set causing it to expire twice.

actors/builtin/miner/miner_test.go
acruikshank: I removed it. It was just in there because I found the expectation confusing. But this comment may be doing more harm than good on that front.

actors/builtin/miner/expiration_queue.go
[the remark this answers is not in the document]
anorth: Yes, I'm not sure mustGet is safe to use, now that we've internalized that we don't know where the queue entries are. I'm in favour of removing it (inlining into the one remaining call site).

actors/builtin/miner/expiration_queue.go
```diff
@@ -283,15 +284,15 @@ func (q ExpirationQueue) RescheduleAllAsFaults(faultExpiration abi.ChainEpoch) e
 		return nil
 	}
 
-	// Add rescheduled sectors to new expiration as early-terminating and faulty.
-	allRescheduled, err := bitfield.MultiMerge(rescheduledSectors...)
+	// Add missing sectors to new expiration as early-terminating and faulty.
+	allNotFound, err := bitfield.MultiMerge(rescheduledSectors...)
```
anorth, line 288: Sorry, I think the github diff collapsing caused me to request a rename there that's not very good. It was fine before.

actors/builtin/miner/expiration_queue.go
```diff
@@ -698,3 +716,126 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
...
+// Groups sectors into sets based on their Expiration field.
+// If sectors are not found in the expiration set corresponding to their expiration field
+// (i.e. they have been rescheduled) traverse expiration sets to for groups where these
+// sectors actually expire.
+// Groups will be returned in expiration order, earliest first.
+func (q *ExpirationQueue) findSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChainInfo) ([]sectorExpirationSet, error) {
+	declaredExpirations := make(map[abi.ChainEpoch]bool, len(sectors))
+	sectorsByNumber := make(map[uint64]*SectorOnChainInfo, len(sectors))
+	emptyBitfield := bitfield.New()
+	allNotFound := &emptyBitfield
```
anorth, line 729: We changed to pass Bitfields by value type a month or so back. Change`groupExpirationSet` to return a non-pointer to bitfield and remove the pointers here.

But also consider just using a `map[uint64]struct{}`, which will save the underlying unnecessary RLE

actors/builtin/miner/expiration_queue.go
```diff
@@ -698,3 +716,126 @@ func groupSectorsByExpiration(sectorSize abi.SectorSize, sectors []*SectorOnChai
...
+		}
+	}
+
+	// If sectors remain, traverse next in epoch order. Remaining sectors should be rescheduled to expire soon, so
+	// this traversal should exit early.
+	if empty, err := allNotFound.IsEmpty(); err != nil {
+		return nil, err
+	} else if !empty {
+		err := q.traverse(func(epoch abi.ChainEpoch, es *ExpirationSet) (bool, error) {
+			// skip if we've processed this epoch above
```
anorth, line 764: Please expand the comment, noting that if a sector was in this epoch due to rescheduling it would have already been included in the first loop.

actors/builtin/miner/expiration_queue.go
anorth: Ignoring `es.EarlySectors` (now down in groupExpirationSet) still makes me quite uncomfortable. Adding an error check that no such sectors are being sought would help.

anorth (approved):

actors/builtin/miner/expiration_queue.go
acruikshank: It's a little sad we can't efficiently do proper set math on bitfields. I spent some time trying to figure out how this could work. If we could copy `RunIterators`, we could use runs to perform multiple set operations and then convert back to bitfields efficiently (if necessary).

Using a map here ended up being less code anyway, so I went ahead and made that optimization.

dkkapur:
@wadeAlexC - this still shows up as a pending issue in the audit report (https://diligence.consensys.net/audits/private/2n87mi4ruvkxem/). Per the conversation above and the latest changes from @acruikshank, do you think the issue is resolved or is additional follow up needed?

acruikshank:
@dkkapur Yes, this PR fixes all the issues brought up in #1100. In our migration to V2 actors we now have a mechanism for correcting miner actors affected by this problem. We've added quite a bit of testing around the issue as well.

PR #1130, opened 2020-09-09
Add LockBalance method to set vesting schedule on non-vesting multisig (v0.9)

anorth:
In actors v2, a multisig's vesting start epoch can be set at construction, but in testnet it is always set to the epoch of construction. This PR adds a `LockBalance` method which allows the same vesting initialization for a multisig that was not initially configured with a locked amount. The vesting can only be configured via proposal from the multisig itself

This allows integration partners using the testnet to test out multisig vesting workflows etc prior to a full upgrade to v2.

~We can deploy this into the testnet as a backwards compatible change **if the  Lotus VM rejects method `9` for a multisig actor before the upgrade epoch**.~
Edit: Lotus does not yet prevent actors aborting with system exit codes, so I've added a test of the network version, and abort with `SysErrInvalidMethod` until network version 2.

I'll replicate this on `master` too, though we can consider it's use deprecated.

### DIFF
--- a/actors/builtin/multisig/multisig_state.go
+++ b/actors/builtin/multisig/multisig_state.go
@@ -38,10 +38,11 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 	if elapsedEpoch >= st.UnlockDuration {
 		return abi.NewTokenAmount(0)
 	}
-	if elapsedEpoch <= 0 {
+	if elapsedEpoch < 0 {
 		return st.InitialBalance
 	}
 
+	// TODO: fix division truncation https://github.com/filecoin-project/specs-actors/issues/1131
 	unitLocked := big.Div(st.InitialBalance, big.NewInt(int64(st.UnlockDuration)))
 	return big.Mul(unitLocked, big.Sub(big.NewInt(int64(st.UnlockDuration)), big.NewInt(int64(elapsedEpoch))))
 }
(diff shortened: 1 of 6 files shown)


actors/builtin/multisig/multisig_state.go
```diff
@@ -28,10 +28,19 @@ type State struct {
...
+	st.InitialBalance = lockedAmount
+}
+
 func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 	if elapsedEpoch >= st.UnlockDuration {
 		return abi.NewTokenAmount(0)
 	}
+	if elapsedEpoch <= 0 {
+		return st.InitialBalance
+	}
```
anorth, line 43: This is a bug in master: if the vesting hasn't started yet, `elapsedEpoch` will be negative and the expression below will calculate a locked balance larger than the total locked amount. The change is backwards compatible because it's impossible to configure a multisig with StartEpoch < current epoch until this change.

ZenGround0 (approved):
While reviewing I noticed the name InitialBalance could be improved by including information that this amount pertains only to a vesting amount as in `InitialBalanceVesting`.

anorth:
Yeah, the naming is not great, but I've refrained from changing it since other parties may be working with this as it is now and I don't want to mess up their efforts.

actors/builtin/multisig/multisig_state.go
anorth: @whyrusleeping prompted me to critically check the case of a multisig operation in the same epoch as creation, and I realised this is _not_ backwards compatible, because the division truncation (#1131) could cause the expression below to return a value that differs from `InitialBalance`.

I'll change the test here to `if elapsedEpoch < 0`, falling through to the slightly wrong behaviour below during the epoch of creation.

whyrusleeping (approved):

PR #1135, opened 2020-09-09
Change baseline function to 2.5EiB @ 100%

anorth:
New parameters from @zx.

Note for migration in testnet: we intend to just set the reward actor's baseline value to this new initial value directly, rather than work out what it would have been had the baseline value been set this way when testnet started.

The value will be higher, leading to a reduction in instantaneous rewards and penalties.


ZenGround0 (approved):
LGTM

zixuanzh (approved):

PR #1139, opened 2020-09-10
Feat/update runtime v0.9

ZenGround0:
Update v0.9 runtime to the agreed refactor
- [x] Inline sub interfaces
- [x] Modify send to have an output parameter
- [x] Tracing sub interface becomes single function
- [x] Extract log and cbor types importing latest go-state-types

Still WIP.  The inlining is done and all of the send modification for source.  Tests for send need fixing along with the last two items.


ZenGround0:
@anorth I know we won't merge this yet, but its ready for review.  Planning to cherrypick this over to master once review passes.

actors/runtime/runtime.go
```diff
@@ -141,8 +125,8 @@ type Runtime interface {
 	// should not be used by actor code directly.
 	Context() context.Context
 
-	// Starts a new tracing span. The span must be End()ed explicitly, typically with a deferred invocation.
-	StartSpan(name string) TraceSpan
+	// Starts a new tracing span. The span must be End()ed explicitly by invoking or deferring the returned end function
+	StartSpan(name string) func()
```
anorth, line 129: Can you use named return value syntax to name it `EndSpan`?

actors/runtime/runtime.go
```diff
@@ -152,15 +136,15 @@ type Runtime interface {
...
 	// Note events that may make debugging easier
-	Log(level LogLevel, msg string, args ...interface{})
+	Log(level rt.LogLevel, msg string, args ...interface{})
 }
 
 // Store defines the storage module exposed to actors.
 type Store interface {
 	// Retrieves and deserializes an object from the store into `o`. Returns whether successful.
-	Get(c cid.Cid, o CBORUnmarshaler) bool
+	Get(c cid.Cid, o cbor.Unmarshaler) bool
```
anorth, line 145: As embeds, I think these need longer names, `StoreGet` and `StorePut`.

actors/runtime/runtime.go
```diff
@@ -213,25 +197,19 @@ type Syscalls interface {
 // the return, in particular whether it has been serialized to bytes or just passed through.
 // Production code is expected to de/serialize, but test and other code may pass the value straight through.
 type SendReturn interface {
```
anorth, line 199: Delete SendReturn?

actors/runtime/runtime.go
```diff
@@ -213,25 +197,19 @@ type Syscalls interface {
...
-	End()
+	Into(cbor.Unmarshaler) error
 }
 
 // StateHandle provides mutable, exclusive access to actor state.
 type StateHandle interface {
 	// Create initializes the state object.
 	// This is only valid in a constructor function and when the state has not yet been initialized.
-	Create(obj CBORMarshaler)
+	Create(obj cbor.Marshaler)
```
anorth, line 207: Similar for naming: StateCreate, StateRead, StateTransact

The Message and Syscalls embeds are fine without longer names, though.

actors/runtime/runtime.go
```diff
@@ -305,3 +268,16 @@ func (b *CBORBytes) UnmarshalCBOR(r io.Reader) error {
 	*b = c.Bytes()
 	return err
 }
+
+// Discard is a helper
+type Discard struct{}
```
anorth, line 273: This is now an internal detail for actors, not about their interface to the VM. Note there are no more concrete types in this file.

Move to `builtin/shared.go` or similar.

anorth (approved):
Great work.

I'm not sure that removing the puppet actor is safe in this branch. Could you check with test-vectors team? I dunno if chain-validation is still in use in lotus master.

ZenGround0:
> I'm not sure that removing the puppet actor is safe in this branch. Could you check with test-vectors team? I dunno if chain-validation is still in use in lotus master.

Yup its off lotus master a few days back [PR here](https://github.com/filecoin-project/lotus/pull/3606/commits)

PR #1145, opened 2020-09-11
Feat/update runtime

ZenGround0:
Same as #1139 but on master.  Because v0.9 was released before our internal vm was born there is a bit more here than there


go.mod
```diff
@@ -19,5 +19,6 @@ require (
 	github.com/stretchr/testify v1.6.1
 	github.com/whyrusleeping/cbor-gen v0.0.0-20200812213548-958ddffe352c
 	github.com/xorcare/golden v0.6.0
+	golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8
```
anorth, line 22: This is probably leftover from an IDE auto-complete or something, but I don't think should be here.

anorth (approved):

PR #1152, opened 2020-09-14
terminate sectors scenario test

### LINKED ISSUE #810: Storage market integration/scenario tests
After #621 gives us a VM with which we can exercise actor interactions, develop integration tests demonstrating coordination between multiple actors.

acruikshank:
closes #810

### Motivation

We want to scenario test market actor methods. It turns out that a scenario involving adding deals then terminating them all is one stop shopping for all the market methods. Specifically this test hits:

* `market.AddBalance`
* `market.WithdrawBalance`
* `market.PublishStorageDeals`
* `market.VerifyDealsForActivation`
* `market.ActivateDeals`
* `market.OnMinerSectorsTerminate`
* `market.ComputeDataCommitment`
* `market.CronTick`

### Proposed Changes

1. Add `terminate_sectors_secnario_test.go`.
2. Various test helper functionality to support this test.

### DIFF
--- a/actors/test/terminate_sectors_scenario_test.go
+++ b/actors/test/terminate_sectors_scenario_test.go
@@ -25,7 +25,8 @@ func TestTerminateSectors(t *testing.T) {
 	ctx := context.Background()
 	v := vm.NewVMWithSingletons(ctx, t)
 	addrs := vm.CreateAccounts(ctx, t, v, 4, big.Mul(big.NewInt(10_000), vm.FIL), 93837778)
-	worker, verifier, unverifiedClient, verifiedClient := addrs[0], addrs[1], addrs[2], addrs[3]
+	owner, verifier, unverifiedClient, verifiedClient := addrs[0], addrs[1], addrs[2], addrs[3]
+	worker := owner
 
 	minerBalance := big.Mul(big.NewInt(1_000), vm.FIL)
 	sectorNumber := abi.SectorNumber(100)
@@ -34,7 +35,7 @@ func TestTerminateSectors(t *testing.T) {
 
 	// create miner
 	ret := vm.ApplyOk(t, v, addrs[0], builtin.StoragePowerActorAddr, minerBalance, builtin.MethodsPower.CreateMiner, &power.CreateMinerParams{
-		Owner:         worker,
+		Owner:         owner,
 		Worker:        worker,
 		SealProofType: sealProof,
 		Peer:          abi.PeerID("not really a peer id"),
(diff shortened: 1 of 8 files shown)


actors/test/terminate_sectors_scenario_test.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	_, code = v.ApplyMessage(vm.VerifregRoot, builtin.VerifiedRegistryActorAddr, big.Zero(), builtin.MethodsVerifiedRegistry.AddVerifier, &addVerifierParams)
+	require.Equal(t, exitcode.Ok, code)
+
+	addClientParams := verifreg.AddVerifiedClientParams{
+		Address:   verifiedClient,
+		Allowance: abi.NewStoragePower(32 << 40),
+	}
+	_, code = v.ApplyMessage(verifier, builtin.VerifiedRegistryActorAddr, big.Zero(), builtin.MethodsVerifiedRegistry.AddVerifiedClient, &addClientParams)
+	require.Equal(t, exitcode.Ok, code)
```
anorth, line 66: Nit: add an `ApplyOk` method that checks the return code. There are a lot of these "require ok" lines in this and other tests.

actors/test/terminate_sectors_scenario_test.go
```diff
@@ -0,0 +1,246 @@
...
+	sealProof := abi.RegisteredSealProof_StackedDrg32GiBV1
+
+	// create miner
+	params := power.CreateMinerParams{
+		Owner:         worker,
+		Worker:        worker,
+		SealProofType: sealProof,
+		Peer:          abi.PeerID("not really a peer id"),
+	}
+	ret, code := v.ApplyMessage(addrs[0], builtin.StoragePowerActorAddr, minerBalance, builtin.MethodsPower.CreateMiner, &params)
```
anorth, line 43: Nit: I think we could inline these params objects into the method calls, especially since it's the last parameter.

actors/test/terminate_sectors_scenario_test.go
```diff
@@ -0,0 +1,246 @@
...
+	// advance and run cron to complete processing of termination
+	v, err = v.WithEpoch(v.GetEpoch() + 1000)
+	require.NoError(t, err)
+	_, code = v.ApplyMessage(builtin.SystemActorAddr, builtin.CronActorAddr, big.Zero(), builtin.MethodsCron.EpochTick, nil)
+	require.Equal(t, exitcode.Ok, code)
+
+	// Verified client should be able to withdraw all all deal collateral.
+	// Client added 3 FIL balance and had 2 deals with 1 FIL collateral apiece.
+	// Should only be able to withdraw the full 2 FIL if deals have been slashed and balance was unlocked.
+	withdrawal := big.Mul(big.NewInt(2), vm.FIL)
```
anorth, line 229: What about the 3 FIL of own balance. It should be able to withdraw 5, I think.

anorth: Verify that the miner has lost escrow balance?

anorth (approved):

actors/test/terminate_sectors_scenario_test.go
acruikshank: > What about the 3 FIL of own balance. It should be able to withdraw 5, I think.

The verified client only added 3 FIL collateral to the market actor, so the market actor can't send more than that to verified client's account (which is what we're testing). Requesting 5 FIL actually sends 2999999994151043072 AttoFIL which is all the collateral added minus payment for the time the deals were stored before they were slashed.

PR #1154, opened 2020-09-15
Separate input and output stores

ZenGround0:
Separate input and output stores.

The actual migration will most likely pass the same store to both arguments.

The motivation is testing.  I would like to migrate 10s of gigabytes of old state without persisting 10s of gigs of new state in the same datastore.  I want to put new state in a temporary datastore so that it can be deleted without deleting chain data along with it.

### DIFF
--- a/actors/migration/account.go
+++ b/actors/migration/account.go
@@ -13,12 +13,12 @@ import (
 type accountMigrator struct {
 }
 
-func (m *accountMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid) (cid.Cid, error) {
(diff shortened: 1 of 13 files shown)


anorth:
This invalidates a shortcut that many migrations make of leaving an object intact if it does not need to change. For example, almost all AMTs are unchanged, and thus migrated by simply returning the existing AMT root CID. This may be part of an explanation of why migrations are faster than we naively expected. As you note, the real migration in a Lotus node will use the same store, because it needs to hold on to the old state too.

You could achieve your goal by wrapping the store with a write cache that directs writes to a different store, and reads from both (read order doesn't matter for content-addressed blocks). RAM might even be sufficient for the migrated state (given that the aforementioned non-migrated AMTs won't touch it).

ZenGround0:
Thanks for the review and solid workaround suggestion

PR #1157, opened 2020-09-15
lower consensus minimum power and use raw vs. qa power for requirement

### LINKED ISSUE #1153: Adjust Minimum Miner Size to 10TiB RawBytePower
Currently, Space Race is using 10TiB QA Power for block mining eligibility. Based on observations from the competition and security/cryptoecon analysis so far, we should adjust the mainnet parameter for Minimum Miner Size (aka eligibility to earn block rewards) from 100 TiB QA power --> **10 TiB RawBytePower**. 🎉

This helps increase network diversity and ensure many miners big and small can successfully participate in Filecoin mining.

acruikshank:
closes #1153

### Motivation

We're lowering the consensus minimum power requirement for miners from 100TiB quality adjusted power to 10TiB raw power.

### Proposed Changes

1. Change 32GiB consensus min to 10TiB and 64GiB consensus min to 20TiB.
2. Use raw power to alter consensus stats in `power.UpdateClaim` and `power.State.MinerNominalPowerMeetsConsensusMinimum`.
3. Adjust and skip tests to reflect the fact that we're using raw power

### Questions

`states.MinerEligibleForElection` will only return true if the miner's `QualityAdjPower` is greater than zero. This is unused in actor code, but is presumably used in block validation in Lotus. Are we merely lowering the consensus min or attempting to drop the requirement that miners have QA power altogether? If the latter, this should probably be changed.

cc @momack2

### DIFF
--- a/actors/builtin/power/power_test.go
+++ b/actors/builtin/power/power_test.go
@@ -259,9 +259,6 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
 	mul := func(a big.Int, b int64) big.Int {
 		return big.Mul(a, big.NewInt(b))
 	}
-	div := func(a big.Int, b int64) big.Int {
-		return big.Div(a, big.NewInt(b))
-	}
 	smallPowerUnit := big.NewInt(1_000_000)
 	require.True(t, smallPowerUnit.LessThan(powerUnit), "power.ConsensusMinerMinPower has changed requiring update to this test")
 	// Subtests implicitly rely on ConsensusMinerMinMiners = 3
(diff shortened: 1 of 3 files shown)


momack2:
@nicola can you comment? AFAIK a miner with 10 TiB RawBytePower without any verified client deals should have 1 QA Power, right, so we aren't removing the requirement that this is non-zero.

momack2:
(Note this should be in actors 1.0.0 so we can test it in Space Race)

actors/builtin/power/power_test.go
```diff
@@ -428,6 +425,7 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
 	})
 
 	t.Run("threshold only depends on qa power, not raw byte", func(t *testing.T) {
+		t.Skip("switching to consensus min based on raw byte power for space race")
```
anorth, line 428: Please adjust this and the tests below to instead verify that the raw byte power is consulted instead of the QA power.

actors/builtin/power/power_test.go
```diff
@@ -329,29 +326,29 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
 		actor.createMinerBasic(rt, owner, owner, miner4)
 		actor.createMinerBasic(rt, owner, owner, miner5)
 
-		actor.updateClaimedPower(rt, miner1, div(smallPowerUnit, 2), smallPowerUnit)
-		actor.updateClaimedPower(rt, miner2, div(smallPowerUnit, 2), smallPowerUnit)
+		actor.updateClaimedPower(rt, miner1, smallPowerUnit, smallPowerUnit)
```
anorth, line 329: Perhaps multiply the QA power by 10 so it's more clear that it's not being used? Ensure the QA power does exceed the threshold, but is ignored.

anorth:
@acruikshank interesting question about `states.MinerEligibleForElection`. QA power is still the "real" power measure for everything except this long-lookback consensus minimum, so I think it's ok to leave as-is.

anorth:
@momack2 a miner with 10 TiB RawBytePower without any verified client deals should have *10 TiB* QA Power, too. If the power was 100% from verified deals, they'd have 100 TiB QA Power.

anorth (approved):
LGTM after restoring and updating skipped tests

PR #1163, opened 2020-09-16
Export market policy functions

### LINKED ISSUE #1162: Make dealDurationBounds public
As in the title, this would be useful for miners to validate deals before accepting them (and finding out their bad at the publish deal stage).

anorth:
Fixes #1162


arajasek (approved):
Thanks!

PR #1167, opened 2020-09-17
InBalance func

ZenGround0:
We need a function for computing all the funds in an input state tree to do invariant checks.


acruikshank (approved):
Looks great. Maybe a clearer name like `V1Balance` or `InputTotalBalance` would be preferable.

Issue #1165, opened 2020-09-17
Implement actor state invariant checks (slow version) for other actors


anorth:
The `states/check.go` file is an entry point for exhaustive actor state verifications, for use in tests and migrations. A couple of actors have checks defined, but we need the rest too.

- [x] account
- [x] cron
- [x] init
- [x] market
- [x] miner
- [x] multisig
- [x] paych
- [x] power
- [x] reward
- [x] verifreg

anorth:
@acruikshank @ZenGround0 do you think we're all done here after #1240?

acruikshank:
@anorth Yes. We'll have a good base. From here we should probably add when we discover problems that could be recast and broken invariants.

PR #1170, opened 2020-09-17
state invariant tests for account, cron and market

acruikshank:
work towards #1165

### Motivation

For performance reasons, our actor state is represented over multiple data structures and values. We need to ensure these many views on state always stay consistent. This PR adds checks for the Account, Cron, and Market actors that traverse these data structures ensuring consistency and uses them to check the state after every successful unit test.

### Proposed Changes

1. Create `CheckStateInvariants` function for `Account`, `Cron` and `Market` actors.
2. Add a call to this at the end of every actor test function that succeeds that fails the test if there is an inconsistency.

### DIFF
--- a/actors/builtin/cron/testing.go
+++ b/actors/builtin/cron/testing.go
@@ -1,4 +1,4 @@
-package account
+package cron
 
 import (
 	"github.com/filecoin-project/go-address"
@@ -7,17 +7,17 @@ import (
 )
 
 type StateSummary struct {
-	PubKeyAddr address.Address
+	EntryCount int
 }
 
-// Checks internal invariants of account state.
+// Checks internal invariants of cron state.
 func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.MessageAccumulator, error) {
 	acc := &builtin.MessageAccumulator{}
-	acc.Require(
-		st.Address.Protocol() == address.BLS || st.Address.Protocol() == address.SECP256K1,
-		"actor address %v must be BLS or SECP256K1 protocol", st.Address)
+	for i, e := range st.Entries {
+		acc.Require(e.Receiver.Protocol() == address.ID, "entry %d receiver address %v must be ID protocol", i, e.Receiver)
+	}
 
 	return &StateSummary{
-		PubKeyAddr: st.Address,
+		EntryCount: len(st.Entries),
 	}, acc, nil
 }
--- a/actors/builtin/cron/testing.go
+++ b/actors/builtin/cron/testing.go
@@ -15,6 +15,7 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 	acc := &builtin.MessageAccumulator{}
 	for i, e := range st.Entries {
 		acc.Require(e.Receiver.Protocol() == address.ID, "entry %d receiver address %v must be ID protocol", i, e.Receiver)
+		acc.Require(e.MethodNum > 0, "entry %d has invalid method number %d", i, e.MethodNum)
 	}
 
 	return &StateSummary{
--- a/actors/builtin/market/testing.go
+++ b/actors/builtin/market/testing.go
@@ -1,14 +1,18 @@
 package market
 
 import (
+	"bytes"
+	"encoding/binary"
+
+	"github.com/ipfs/go-cid"
+	"github.com/pkg/errors"
+	cbg "github.com/whyrusleeping/cbor-gen"
+
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
 	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
-	"github.com/ipfs/go-cid"
-	cbg "github.com/whyrusleeping/cbor-gen"
-	"strconv"
 )
 
 type StateSummary struct {
@@ -57,9 +61,6 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 		if err != nil {
 			return err
 		}
-
-		_, found := proposalCids[pcid]
-		acc.Require(!found, "duplicate DealProposal %d found in proposals", dealID)
 
 		// keep some state
 		proposalCids[pcid] = struct{}{}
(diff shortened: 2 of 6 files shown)


actors/builtin/cron/testing.go
```diff
@@ -0,0 +1,23 @@
...
+
+type StateSummary struct {
+	EntryCount int
+}
+
+// Checks internal invariants of cron state.
+func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.MessageAccumulator, error) {
+	acc := &builtin.MessageAccumulator{}
+	for i, e := range st.Entries {
+		acc.Require(e.Receiver.Protocol() == address.ID, "entry %d receiver address %v must be ID protocol", i, e.Receiver)
```
anorth, line 17: Could require Method > 0 too

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+	maxDealID := int64(-1)
+	proposalCount := uint64(0)
+
+	proposals, err := adt.AsArray(store, st.Proposals)
+	if err != nil {
+		return nil, acc, err
+	}
+	var proposal DealProposal
+	err = proposals.ForEach(&proposal, func(dealID int64) error {
+		allIDs[abi.DealID(dealID)] = struct{}{}
```
anorth, line 58: Validate that both Provider and Client addresses are `ID` addresses.

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "deal ops has key that is not an int: %s", key)
+		}
+
+		dealOpEpochCount++
+		return dealOps.ForEach(abi.ChainEpoch(epoch), func(id abi.DealID) error {
+			_, found := allIDs[id]
+			acc.Require(found, "deal op found for deal id %d with missing proposal at epoch %d", id, epoch)
+			dealOpCount++
+			return nil
```
anorth, line 214: Either there should exist a corresponding deal state, or `CurrEpoch() >= deal.StartEpoch`. We can accept a current epoch parameter here too.

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+		pcid, err := pendingProposal.Cid()
+		if err != nil {
+			return err
+		}
+		acc.Require(pcid.Equals(proposalCID), "pending proposal's key does not match its CID %v != %v", pcid, proposalCID)
+
+		_, found := proposalCids[pcid]
+		acc.Require(found, "pending proposal with cid %v not fond within proposals %v", pcid, pendingProposals)
+
+		pendingProposalCount++
```
anorth, line 139: Can we check it's being garbage collected? I think there should be no pending deal with start date < `current - DealUpdatesInterval`

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+)
+
+type StateSummary struct {
+	ProposalCount        uint64
+	PendingProposalCount uint64
```
anorth, line 20: I think we'll want more info here for cross-checking with the miner. E.g. all the proposal IDs so we can confirm that the sector deal IDs are a subset. And the pending ones for pre-committed sectors.

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+	DealStateCount       uint64
+	LockTableCount       uint64
+	DealOpEpochCount     uint64
+	DealOpCount          uint64
+}
+
+// Checks internal invariants of market state.
+func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.MessageAccumulator, error) {
```
anorth, line 28: Please add a parameter for the actor's balance, and check that it's >= the sum of the lock-ups.

it's ok that this type signature won't match other actors (the return value doesn't anyway).

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+		// every entry in locked table should have a corresponding entry in escrow table that is at least as high
+		escrowAmount, err := escrowTable.Get(addr)
+		if err != nil {
+			return err
+		}
+		acc.Require(escrowAmount.GreaterThanEqual(lockedAmount),
+			"locked funds for %s, %s, greater than escrow amount, %s", addr, lockedAmount, escrowAmount)
+
+		lockTableCount++
+		return nil
```
anorth, line 178: Can we accumulate the per-client/provider collateral amounts from the deals and require the locked amounts here to be at least as large (should exactly match for provider, but client also locks up storage fee, which will be partially paid out and too complicated to recompute the expected value)?

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+			dealState.SectorStartEpoch >= 0,
+			"deal state start epoch undefined %d: %v", dealID, dealState)
+
+		acc.Require(
+			dealState.LastUpdatedEpoch == epochUndefined || dealState.LastUpdatedEpoch >= dealState.SectorStartEpoch,
+			"deal state last updated before sector start %d: %v", dealID, dealState)
+
+		acc.Require(
+			dealState.SlashEpoch == epochUndefined || dealState.SlashEpoch >= dealState.SectorStartEpoch,
+			"deal state slashed before sector start %d: %v", dealID, dealState)
```
anorth, line 102: There are a couple more assertions in `updatePendingDealState` that we could add here (some in terms of the current epoch, which you can add as a parameter)

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	return &StateSummary{
+		ProposalCount:        proposalCount,
+		PendingProposalCount: pendingProposalCount,
+		DealStateCount:       dealStateCount,
+		LockTableCount:       lockTableCount,
+		DealOpEpochCount:     dealOpEpochCount,
+		DealOpCount:          dealOpCount,
+	}, acc, nil
```
anorth, line 228: This function is big enough that it might be worth breaking up into some smaller ones, if the data flow allows it.

actors/builtin/market/testing.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		// keep some state
+		proposalCids[pcid] = struct{}{}
+		if dealID > maxDealID {
+			maxDealID = dealID
+		}
+		proposalCount++
+
+		return nil
```
anorth, line 72: I think there's something about proposal cleanup, where every proposal should either have an associated state, or be within `DealUpdatesInterval` of it's start date (about to be expired). Proposals more than `DealUpdatesInterval` after their EndDate should not exist, and neither should associated state. From `CronTick`.

The Proposal/State/OpEpoch inter-consistency might require loading all of those into Go maps for easier lookups.

anorth (approved):
Thanks, this is great. I think there's room for more paranoid checks, especially of the relationships between items in different containers and their lifecycles.

actors/builtin/market/testing.go
acruikshank: This one is kind of tough. Garbage collection happens in cron. So there technically can be a period between expiration and the next cron run where it doesn't hold, unless you assume crons are run for null blocks (which I think we are).

More to the point, we don't require our unit tests to run cron for every epoch, so adding this requirement causes many tests to fail, and it would be a fair amount of work and potentially a huge performance hit to start imposing that requirement.

actors/builtin/market/testing.go
acruikshank: I've added a map to track proposals expected to have an op and ensure that everything in that map has one.

PR #1171, opened 2020-09-18
Feat/parallel work

ZenGround0:
This aims to parallelize state migrations to make use of time spent waiting for datastore reads and writes as well and to leverage the multiple CPUs of miners' machines.  The idea is to run each actor migration in a different worker goroutine.  goroutines are bounded to a fixed size set (arbitrarily 16 here, need to empirically set it and it should be configurable for miners migrating).

@anorth this is still rough.  Apart from one clear bug it should work but hasn't been tested out yet so there might be other things wrong with it.
- We use a single shared miner migrator and set its balance and transfer state.  This is not threadsafe and is just broken.  I think we should differentiate the miner migration to have a different function signature to get rid of this problem simply.
- Concurrency pattern seems sound but there are likely much simpler approaches.  Maybe we could use a standard wait group instead of a fancy semaphore?  There should be some way we could do the semaphore release as a ~require~ defer instead of explicitly before every return statement.  The separate goroutine for accumulating transfers can probably go away with another approach. Etc.  Ideas welcome here.
- Global var semaphore is maybe a problem for multiple invocations
- Haven't separated out the synchronization code from the actors traversal so its getting crowded

### DIFF
--- a/actors/migration/top.go
+++ b/actors/migration/top.go
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	defaultMaxWorkers = 32
+	defaultMaxWorkers = 16
 	sem               *semaphore.Weighted
 	actOutMu          = &sync.Mutex{}
 )
@@ -93,50 +93,6 @@ var migrations = map[cid.Cid]ActorMigration{ // nolint:varcheck,deadcode,unused
 		OutCodeCID:     builtin.StorageMinerActorCodeID,
 		StateMigration: &minerMigrator{},
 	},
-}
-
-func migrateOneActor(ctx context.Context, store cbor.IpldStore, addr address.Address, actorIn states.Actor, actorsOut *states.Tree, transferCh chan big.Int, errCh chan error) {
-	var headOut, codeOut cid.Cid
-	var err error
-	transfer := big.Zero()
-	// This will be migrated at the end
-	if actorIn.Code == builtin0.VerifiedRegistryActorCodeID {
-		sem.Release(1)
-		return
-	} else {
-		migration := migrations[actorIn.Code]
-		codeOut = migration.OutCodeCID
-		headOut, transfer, err = migration.StateMigration.MigrateState(ctx, store, actorIn.Head, actorIn.Balance)
-	}
-
-	if err != nil {
-		err = xerrors.Errorf("state migration error on %s actor at addr %s: %w", builtin.ActorNameByCode(codeOut), addr, err)
-		sem.Release(1)
-		errCh <- err
-		return
-	}
-
-	// set up new state root with the migrated state
-	actorOut := states.Actor{
-		Code:       codeOut,
-		Head:       headOut,
-		CallSeqNum: actorIn.CallSeqNum,
-		Balance:    big.Add(actorIn.Balance, transfer),
-	}
-	actOutMu.Lock()
-	err = actorsOut.SetActor(addr, &actorOut)
-	actOutMu.Unlock()
-
-	// Release this worker goroutine before any channel sends to prevent deadlock
-	if err != nil {
-		sem.Release(1)
-		errCh <- err
-	} else if transfer.GreaterThan(big.Zero()) {
-		sem.Release(1)
-		transferCh <- transfer
-	} else {
-		sem.Release(1)
-	}
 }
 
 // Migrates the filecoin state tree starting from the global state tree and upgrading all actor state.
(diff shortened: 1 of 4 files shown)


actors/migration/top.go
```diff
@@ -75,41 +81,65 @@ var migrations = map[cid.Cid]ActorMigration{ // nolint:varcheck,deadcode,unused
...
-}
-
-func (p *phoenix) transfer(amt abi.TokenAmount) error {
-	p.burntBalance = big.Sub(p.burntBalance, amt)
-	if p.burntBalance.LessThan(big.Zero()) {
-		return xerrors.Errorf("migration programmer error burnt funds balance falls to %v, below zero", p.burntBalance)
+	headOut, err := migration.StateMigration.MigrateState(ctx, store, actorIn.Head)
+	if err != nil {
+		err = xerrors.Errorf("state migration error on %s actor at addr %s: %w", builtin.ActorNameByCode(migration.OutCodeCID), addr, err)
+		sem.Release(1)
```
ZenGround0, line 107: preventing deadlock in the rare case all worker goroutines are sending an error on channel and the main thread producing work is blocking on the semaphore.  This makes it so that there could be an extra worker thread but that seems like nbd, not being able to cleanly defer seems like more of a problem.

actors/migration/top.go
```diff
@@ -207,7 +239,21 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
 	}
 
 	// Track deductions to burntFunds actor's balance
-	if err := p.flush(ctx, actorsIn, actorsOut); err != nil {
+	transferCancel()
+	transferFromBurnt := <-transferOutCh
```
ZenGround0, line 243: must be the final value since we acquired all of the semaphore's weight earlier

actors/migration/top.go
```diff
@@ -207,7 +239,21 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
+		return cid.Undef, xerrors.Errorf("burnt funds actor not in tree")
+	}
+	burntFundsActor.Balance = big.Sub(burntFundsActor.Balance, transferFromBurnt)
+	if burntFundsActor.Balance.LessThan(big.Zero()) {
+		return cid.Undef, xerrors.Errorf("miner transfers send burnt funds actor balance below zero")
+	}
+	err = actorsOut.SetActor(builtin.BurntFundsActorAddr, burntFundsActor)
+	if err != nil {
 		return cid.Undef, err
 	}
```
ZenGround0, line 259: Might need to reset semaphore since its a global var.  Ideally we'd move from global to some local scope but that might make code less clean with lots of closures.  Passing sync objects in functions has burned me before, maybe it s ok this time?

Not using a semaphore would also solve this problem :)

actors/migration/top.go
```diff
@@ -129,60 +159,62 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
 
-		// set up new state root with the migrated state
-		actorOut := states.Actor{
-			Code:       migration.OutCodeCID,
-			Head:       headOut,
-			CallSeqNum: actorIn.CallSeqNum,
-			Balance:    actorIn.Balance,
+		// Hand off migration of one actor, blocking if we are out of worker goroutines
+		if err := sem.Acquire(ctx, 1); err != nil {
+			return err
```
anorth, line 198: Using a semaphore explicitly is a smell that we're not doing it right in Go.
- Consider that goroutines are _really_ cheap, try just not limiting them at all
- A bounded-capacity channel is probably a better pattern, with 1 producer that blocks when the channel is full

actors/migration/top.go
```diff
@@ -129,60 +159,62 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
-	}
+	// Extra actor setup
 	// power
 	pm := migrations[builtin0.StoragePowerActorCodeID].StateMigration.(*powerMigrator)
 	pm.actorsIn = actorsIn
 
+	// Setup synchronization
+	errCh := make(chan error)
+	transferInCh := make(chan big.Int)
+	transferOutCh := make(chan big.Int)
```
anorth, line 170: Many channels is really complicated. I can't quite follow what they're all doing, but prefer a single channel with a union-style struct value type over many channels with different value types.

actors/migration/top.go
```diff
@@ -75,41 +81,68 @@ var migrations = map[cid.Cid]ActorMigration{ // nolint:varcheck,deadcode,unused
...
+		errCh <- err
+		return
+	}
+	if transfer.GreaterThan(big.Zero()) {
+		sem.Release(1)
+		transferCh <- transfer
+		return
+	}
+	sem.Release(1)
+	return // nolint:gosimple
```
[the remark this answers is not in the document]
ZenGround0: I found a way that should work introducing only one new branch to if.  Not clear to me that it is simpler but the linter likes it better.

ZenGround0:
FYI @acruikshank

actors/migration/top.go
```diff
@@ -12,12 +13,30 @@ import (
...
 	cbor "github.com/ipfs/go-ipld-cbor"
+	"golang.org/x/sync/semaphore"
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
 	"github.com/filecoin-project/specs-actors/v2/actors/states"
 )
 
+var (
+	defaultMaxWorkers = 32
```
anorth, line 24: This is a bit higher than I expected.

actors/migration/top.go
anorth: Add that motivation to the code, not just the PR.

actors/migration/top.go
```diff
@@ -74,47 +89,58 @@ var migrations = map[cid.Cid]ActorMigration{ // nolint:varcheck,deadcode,unused
...
-	}
-	if !found {
-		return xerrors.Errorf("burnt funds actor not in tree")
+		sem.Release(1)
+		errCh <- err
+	} else if transfer.GreaterThan(big.Zero()) {
+		sem.Release(1)
+		transferCh <- transfer
+	} else {
+		sem.Release(1)
```
anorth, line 138: Why not move `sem.Release(1)` above the `if` block?

actors/migration/top.go
```diff
@@ -74,47 +89,58 @@ var migrations = map[cid.Cid]ActorMigration{ // nolint:varcheck,deadcode,unused
...
+		OutCodeCID:     builtin.StorageMinerActorCodeID,
+		StateMigration: &minerMigrator{},
+	},
 }
 
-// A phoenix tracks the unburning of funds
-type phoenix struct {
-	burntBalance abi.TokenAmount
-}
+func migrateOneActor(ctx context.Context, store cbor.IpldStore, addr address.Address, actorIn states.Actor, actorsOut *states.Tree, transferCh chan big.Int, errCh chan error) {
```
anorth, line 98: Move this utility method down below the main entry point.

anorth (approved):

actors/migration/top.go
ZenGround0: I haven't swept it rigorously with the latest optimizations but the overhead from going too high seems very marginal.  I'll bring it down to 16 and maybe tweak after the migration is complete.

PR #1173, opened 2020-09-21
Add balance and balance delta to migrator struct

anorth:
A prefactor for #1171, to aid the miner's balance transfer needs.


ZenGround0 (approved):
LGTM

PR #1175, opened 2020-09-22
Always allow transactions that send zero

anorth:
Multisig actors can now have a locked amount that exceeds their balance. But even in that state, we wish to allow messages which send no funds (e.g. proposals to call methods on the multisig itself to manipulate the signers).

### DIFF
--- a/actors/builtin/multisig/multisig_state.go
+++ b/actors/builtin/multisig/multisig_state.go
@@ -66,6 +66,10 @@ func (st *State) assertAvailable(currBalance abi.TokenAmount, amountToSpend abi.
 	if currBalance.LessThan(amountToSpend) {
 		return xerrors.Errorf("current balance %s less than amount to spend %s", currBalance.String(), amountToSpend.String())
 	}
+	if amountToSpend.IsZero() {
+		// Always permit a transaction that sends no value, even if the lockup exceeds the current balance.
+		return nil
+	}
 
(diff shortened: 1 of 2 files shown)


actors/builtin/multisig/multisig_state.go
```diff
@@ -66,6 +66,10 @@ func (st *State) assertAvailable(currBalance abi.TokenAmount, amountToSpend abi.
 	if currBalance.LessThan(amountToSpend) {
 		return xerrors.Errorf("current balance %s less than amount to spend %s", currBalance.String(), amountToSpend.String())
 	}
+	if amountToSpend.IsZero() {
+		// Always permit a transaction that sends no value, even if the lockup exceeds the current balance.
```
ZenGround0, line 70: Is this defensively allowing progress given a corrupt state we don't expect to reach or is the multisig actor expected to have a lockup exceeding balance.  If the latter what circumstances lead to this state?

acruikshank (approved):

actors/builtin/multisig/multisig_state.go
anorth: The actor can reach this state when the vesting schedule is installed after construction, with `LockFunds`. This permits locking more than the balance the actor has at the time, and allows flexibility in the sequence of initializing a multisig vs transferring balance to it.

Issue #1177, opened 2020-09-23
Repair state of miners broken by upgrade/fault collision #1100


anorth:
On testnet, a few miner actors have been broken when they encountered the condition described in #1100. They account for about 6PiB of testnet power.

While delicate, we believe it's possible to repair these miners' states so they can resume operation, assuming the operators have retained their sectors. This repair involves a state migration, implemented alongside an epoch-triggered network upgrade. The repair does not require changing actor code, so could be performed in-place before the network upgrades all the way to v2 (though we might want to run it again at that time to catch any new cases).

Decision about whether to resolve this with an in-place upgrade or with v2 is pending, but doesn't affect the work of creating and testing the algorithm.

anorth:
@acruikshank done?

acruikshank:
Yep it's done and tested. Seems to work.

PR #1178, opened 2020-09-23
Entry point and stubs for state migration in v0.9.

anorth:
We are considering performing another small upgrade on the release/v0.9 branch, ahead of the larger upgrade/migration to v2. We're still working out exactly what we can/should include, but probably:
- repair miners broken by #1100 
- reduce some Window PoSt penalties (spec coming in another issue)
- update reward baseline parameters (previously implemented in /v2)

This PR adds the entry point and stubs for the state migration part.

FYI @acruikshank we'll probably put the code for #1177 here.

### DIFF
--- a/actors/migration/nv3/migrators.go
+++ b/actors/migration/nv3/migrators.go
@@ -1,0 +1,31 @@
+package nv3
+
+import (
+	"context"
+
+	"github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+)
+
+type StateMigration interface {
+	// Loads an actor's state from an input store and writes new state to an output store.
+	// Returns the new state head CID.
+	MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid) (newHead cid.Cid, err error)
+}
+
+var migrators = map[cid.Cid]StateMigration{ // nolint:varcheck,deadcode,unused
+	builtin.RewardActorCodeID: &rewardMigrator{},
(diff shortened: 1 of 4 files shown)


actors/migration/nv3/migrators.go
```diff
@@ -0,0 +1,31 @@
+package nv3
```
ZenGround0, line 1: I get 3 for third in place, what does nv mean?

actors/migration/nv3/miner.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+
+	cid "github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+
+	miner "github.com/filecoin-project/specs-actors/actors/builtin/miner"
+)
+
+type minerMigrator struct {
+}
```
ZenGround0, line 13: TODO: we'll want to inject a way to set the power actor in here.

actors/migration/nv3/top.go
```diff
@@ -0,0 +1,78 @@
...
+			CallSeqNum: actorIn.CallSeqNum,
+			Balance:    actorIn.Balance,
+		}
+
+		return tree.SetActor(addr, &actorOut)
+	}); err != nil {
+		return cid.Undef, err
+	}
+
+	return tree.Flush()
```
ZenGround0, line 60: This looks good

ZenGround0 (approved):

actors/migration/nv3/migrators.go
anorth: "network version" (to distinguish from actors version)

PR #1179, opened 2020-09-23
Change reward actor baseline parameters to 2.5EiB/100% as in-place upgrade.

anorth:
This was previously implemented in #1135 for actors v2 (please cross-check compatibility). We're now planning a smaller migration to come ahead of that, involving a small state migration but no state schema changes, so we can bring this forward.

- [x] Land #1178 and rebase onto `release/v0.9`

After merge:
- [ ] Revert re-setting the baseline initial value in the v2 upgrade migration.

FYI @zixuanzh

### DIFF
--- a/actors/builtin/reward/reward_actor.go
+++ b/actors/builtin/reward/reward_actor.go
@@ -154,6 +154,7 @@ func (a Actor) UpdateNetworkKPI(rt runtime.Runtime, currRealizedPower *abi.Stora
 	if currRealizedPower == nil {
 		rt.Abortf(exitcode.ErrIllegalArgument, "arugment should not be nil")
 	}
+	networkVersion := rt.NetworkVersion()
 
 	var st State
 	rt.StateTransaction(&st, func() {
@@ -162,10 +163,10 @@ func (a Actor) UpdateNetworkKPI(rt runtime.Runtime, currRealizedPower *abi.Stora
 		// st.Epoch == rt.CurrEpoch()
 		for st.Epoch < rt.CurrEpoch() {
 			// Update to next epoch to process null rounds
-			st.updateToNextEpoch(*currRealizedPower)
+			st.updateToNextEpoch(*currRealizedPower, networkVersion)
 		}
 
-		st.updateToNextEpochWithReward(*currRealizedPower)
+		st.updateToNextEpochWithReward(*currRealizedPower, networkVersion)
 		// only update smoothed estimates after updating reward and epoch
 		st.updateSmoothedEstimates(st.Epoch - prev)
 	})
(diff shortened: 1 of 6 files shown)


actors/builtin/reward/reward_actor.go
```diff
@@ -162,10 +163,10 @@ func (a Actor) UpdateNetworkKPI(rt runtime.Runtime, currRealizedPower *abi.Stora
 		// st.Epoch == rt.CurrEpoch()
 		for st.Epoch < rt.CurrEpoch() {
 			// Update to next epoch to process null rounds
-			st.updateToNextEpoch(*currRealizedPower)
```
ZenGround0, line 165: Is this branch possible to hit now that cron fires for every null round?  I do not think it is.

actors/migration/nv3/reward.go
```diff
@@ -5,11 +5,36 @@ import (
...
 func (m *rewardMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid) (cid.Cid, error) {
-	return head, nil
+	var st reward.State
+	if err := store.Get(ctx, head, &st); err != nil {
+		return cid.Undef, err
+	}
+
+	// The baseline function initial value and growth rate are changed.
+	// As an approximation to working out what the baseline value would be at the migration epoch
+	// had the baseline parameters been this way all along, this just sets the immediately value to the
```
ZenGround0, line 24: nit: "the immediately value" => "the immediately successive value" or something like that

actors/migration/nv3/reward.go
```diff
@@ -5,11 +5,36 @@ import (
...
+	// new (higher) initial value, essentially restarting its calculation.
+	// The baseline and realized cumsums, and effective network time, are not changed.
+	// This boils down to a step change in the baseline function, as if it had been defined piecewise.
+	// This will be a bit annoying for external analytical calculations of the baseline function.
+	newBaselineInitialValue := reward.BaselineInitialValueV3
+
+	if st.ThisEpochBaselinePower.GreaterThan(newBaselineInitialValue) {
+		return cid.Undef, xerrors.Errorf("unexpected baseline power %v higher than new initial value %v",
```
ZenGround0, line 32: Do we really want to crash the migration for this reason?  What would we do in this case, change the new baseline function?  Does anything go immediately wrong if this value falls below the current baseline.  I am very suspicious that no code will break.

actors/builtin/reward/reward_actor.go
anorth: Probably not, but I don't want to touch it here. Maybe in v2.

actors/migration/nv3/reward.go
anorth: Well, if this doesn't hold, we must have a serious misunderstanding about how the baseline function works. But ok, perhaps a sanity check can go in the state check later instead.

ZenGround0 (approved):

PR #1180, opened 2020-09-23
add cc correction migration

### LINKED ISSUE #1177: Repair state of miners broken by upgrade/fault collision #1100

acruikshank:
closes #1177

### Motivation

The CC upgrade then fault problem has been fixed in master, but we would like to be able to correct miners that have encountered this problem in space race.

### Proposed Changes

1. Augment miner migration to detect state inconsistencies cause by the problem and correct them.
2. Fixe miner and power state of miners who are no longer working due to broken cron.

### DIFF
--- a/actors/migration/nv3/miner.go
+++ b/actors/migration/nv3/miner.go
@@ -83,7 +83,7 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 				return err
 			}
 
-			exqRoot, partitionActivePower, partitionFaultyPower, _, err := m.correctExpirationQueue(exq, sectors, part.Terminated, part.Faults, sectorSize)
+			exqRoot, stats, err := m.correctExpirationQueue(exq, sectors, part.Terminated, part.Faults, sectorSize)
 			if err != nil {
 				return err
 			}
@@ -98,12 +98,12 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 				alteredPartitions[uint64(partIdx)] = part
 			}
 
-			if !part.LivePower.Equals(partitionActivePower.Add(partitionFaultyPower)) {
-				part.LivePower = partitionActivePower.Add(partitionFaultyPower)
+			if !part.LivePower.Equals(stats.totalActivePower.Add(stats.totalFaultyPower)) {
+				part.LivePower = stats.totalActivePower.Add(stats.totalFaultyPower)
 				alteredPartitions[uint64(partIdx)] = part
 			}
-			if !part.FaultyPower.Equals(partitionFaultyPower) {
-				part.FaultyPower = partitionFaultyPower
+			if !part.FaultyPower.Equals(stats.totalFaultyPower) {
+				part.FaultyPower = stats.totalFaultyPower
 				alteredPartitions[uint64(partIdx)] = part
 			}
 			allFaultyPower = allFaultyPower.Add(part.FaultyPower)
@@ -269,11 +269,22 @@ func (m *minerMigrator) updateClaim(ctx context.Context, store adt.Store, st *mi
 	return nil
 }
 
+type expirationQueueStats struct {
+	// total of all active power in the expiration queue
+	totalActivePower miner.PowerPair
+	// total of all faulty power in the expiration queue
+	totalFaultyPower miner.PowerPair
+	// sector numbers of all sectors believed to have already expired in the active state
+	// when they should have been faulty.
+	terminatedAsNonFaults bitfield.BitField
+}
+
 // Updates the expiration queue by correcting any duplicate entries and their fallout.
 // If no changes need to be made cid.Undef will be returned.
+// Returns the new root of the expiration queue
 func (m *minerMigrator) correctExpirationQueue(exq miner.ExpirationQueue, sectors miner.Sectors,
 	allTerminated bitfield.BitField, allFaults bitfield.BitField, sectorSize abi.SectorSize,
-) (cid.Cid, miner.PowerPair, miner.PowerPair, bitfield.BitField, error) {
+) (cid.Cid, expirationQueueStats, error) {
 	// processed expired sectors includes all terminated and all sectors seen in earlier expiration sets
 	processedExpiredSectors := allTerminated
 	terminatedAsNonFaults := bitfield.New()
@@ -340,18 +351,18 @@ func (m *minerMigrator) correctExpirationQueue(exq miner.ExpirationQueue, sector
 		return nil
 	})
 	if err != nil {
-		return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+		return cid.Undef, expirationQueueStats{}, err
 	}
 
 	// if we didn't find any faults that needed to be erased, we're done
 	if len(alteredExpirationSets) == 0 {
-		return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, nil
+		return cid.Undef, expirationQueueStats{}, nil
 	}
 
 	// save expiration sets that have been altered
 	for epoch, set := range alteredExpirationSets { //nolint:nomaprange
 		if err := exq.Set(uint64(epoch), &set); err != nil {
-			return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+			return cid.Undef, expirationQueueStats{}, err
 		}
 	}
 
@@ -373,7 +384,7 @@ func (m *minerMigrator) correctExpirationQueue(exq miner.ExpirationQueue, sector
 		return nil
 	})
 	if err != nil {
-		return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+		return cid.Undef, expirationQueueStats{}, err
 	}
 
 	/*
@@ -384,16 +395,20 @@ func (m *minerMigrator) correctExpirationQueue(exq miner.ExpirationQueue, sector
 	// save expiration sets that have been altered
 	for epoch, set := range alteredExpirationSets { //nolint:nomaprange
 		if err := exq.Set(uint64(epoch), &set); err != nil {
-			return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+			return cid.Undef, expirationQueueStats{}, err
 		}
 	}
 
 	expirationQueueRoot, err := exq.Root()
 	if err != nil {
-		return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+		return cid.Undef, expirationQueueStats{}, err
 	}
 
-	return expirationQueueRoot, partitionActivePower, partitionFaultyPower, terminatedAsNonFaults, nil
+	return expirationQueueRoot, expirationQueueStats{
+		partitionActivePower,
+		partitionFaultyPower,
+		terminatedAsNonFaults,
+	}, nil
 }
 
 func correctExpirationSet(exs *miner.ExpirationSet, sectors miner.Sectors,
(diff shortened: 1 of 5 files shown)


actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+		if erasedFaults, err = bitfield.MultiMerge(erasedFaults, earlyDuplicates, onTimeDuplicates); err != nil {
+			return err
+		}
+
+		// record on time sectors
```
anorth, line 322: Record all sectors that would be terminated after this queue entry is processed.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+			if err != nil {
+				return err
+			}
+			terminatedAsNonFaults, err = bitfield.MergeBitFields(terminatedAsNonFaults, terminatedActive)
+			if err != nil {
+				return err
+			}
+		}
+
+		if erasedFaults, err = bitfield.MultiMerge(erasedFaults, earlyDuplicates, onTimeDuplicates); err != nil {
```
anorth, line 318: I don't think you know whether the power was in the faulty sum or not or not, but luckily this value appears unused and can be removed.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+	if len(alteredExpirationSets) == 0 {
+		return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, nil
+	}
+
+	// save expiration sets that have been altered
+	for epoch, set := range alteredExpirationSets {
+		if err := exq.Set(uint64(epoch), &set); err != nil {
+			return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+		}
+	}
```
anorth, line 339: Nit: we could save these during the iteration. This is the safe case of updating an AMT value for a fixed key. It would remove the need to keep track of which ones changed.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+	err = exq.ForEach(&exs, func(epoch int64) error {
+		modified, activePower, faultyPower, err := correctExpirationSet(&exs, sectors, allFaults, sectorSize)
+		if err != nil {
+			return err
+		}
+		partitionActivePower = partitionActivePower.Add(activePower)
+		partitionFaultyPower = partitionFaultyPower.Add(faultyPower)
+
+		if modified {
+			alteredExpirationSets[abi.ChainEpoch(epoch)] = exs
```
anorth, line 353: Ditto about deep copy the expiration set.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+		partition.Faults - foundFault = set of faults that have been deducted erroneously
+		= sectors cc upgraded and faulted for which cc upgrade expiration set has been processed
+	*/
+
+	// save expiration sets that have been altered
+	for epoch, set := range alteredExpirationSets {
+		if err := exq.Set(uint64(epoch), &set); err != nil {
+			return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+		}
+	}
```
anorth, line 372: Ditto about write them in the original iteration.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+		}
+
+		// record on time sectors
+		if processedExpiredSectors, err = bitfield.MultiMerge(processedExpiredSectors, exs.EarlySectors, exs.OnTimeSectors); err != nil {
+			return err
+		}
+		return nil
+	})
+
+	// if we didn't find any faults that needed to be erased, we're done
```
anorth, line 329: This terminology about "faults that needed to be erased" makes me nervous. I'm not sure every problem here is in faulty state. Or if it is, it's not necessary for everything above to be correct.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+
+	expirationQueueRoot, err := exq.Root()
+	if err != nil {
+		return cid.Undef, miner.PowerPair{}, miner.PowerPair{}, bitfield.BitField{}, err
+	}
+
+	return expirationQueueRoot, partitionActivePower, partitionFaultyPower, terminatedAsNonFaults, nil
+}
+
+func correctExpirationSet(exs *miner.ExpirationSet, sectors miner.Sectors,
```
anorth, line 382: `correctExpirationSetPower`

anorth (changes requested):

anorth:
Thanks @acruikshank .

We discussed testing this via porting the scenario test framework from master into this branch. That's still a good idea, but maybe a shorter path to getting _some_ test would be to reach the miner state for this by invoking the `State` methods that would be invoked by the actor calls in that scenario test. That's not as fragile as trying to reconstruct it directly by manipulating queues, etc. We can do that as a unit tests here with no additional infrastructure beyond `ipld.NewADTStore()`.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+			exs.EarlySectors, err = bitfield.SubtractBitField(exs.EarlySectors, earlyDuplicates)
+			if err != nil {
+				return err
+			}
+
+			terminatedActive, err := bitfield.IntersectBitField(earlyDuplicates, allTerminated)
+			if err != nil {
+				return err
+			}
+			terminatedAsNonFaults, err = bitfield.MergeBitFields(terminatedAsNonFaults, terminatedActive)
```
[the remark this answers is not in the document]
acruikshank: Yes, it is confusing. We started down the path of using this to compute the power that needs to be returned to the miner's claim. The miner's claim has been zeroed out, so we need to recompute all the power. That is why this is unused. I think I'll just remove it.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+		onTimeDuplicates, err := bitfield.IntersectBitField(exs.OnTimeSectors, processedExpiredSectors)
```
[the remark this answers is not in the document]
acruikshank: Agreed

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,426 @@
...
+
+	if !dealinesModified {
+		return false, nil
+	}
+
+	offset := st.ProvingPeriodStart % miner.WPoStProvingPeriod
+	st.ProvingPeriodStart = offset + ((epoch-offset)/miner.WPoStProvingPeriod)*miner.WPoStProvingPeriod
+	st.CurrentDeadline = uint64((epoch - st.ProvingPeriodStart) / miner.WPoStChallengeWindow)
+
+	return true, err
```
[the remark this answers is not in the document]
acruikshank: This is a good optimization. I pursued it a bit, and realized its a little dangerous given the way we exit out of of loops when the miner looks correct. I'm going to hold off on this until we see some performance problem.

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,464 @@
...
+
+	quantSpec := st.QuantSpecEveryDeadline()
+
+	deadlines, err := st.LoadDeadlines(adt.WrapStore(ctx, store))
+	if err != nil {
+		return false, err
+	}
+
+	deadlinesModified := false
+	err = deadlines.ForEach(adt.WrapStore(ctx, store), func(dlIdx uint64, deadline *miner.Deadline) error {
```
anorth, line 70: Nit: take adtstore as a parameter

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,464 @@
...
+			allFaultyPower = allFaultyPower.Add(part.FaultyPower)
+
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
+		if len(alteredPartitions) > 0 {
+			for partIdx, part := range alteredPartitions { // nolint:nomaprange
```
anorth, line 117: Nit: not clear why we don't just do this in the foreach

actors/migration/nv3/miner.go
```diff
@@ -1,27 +1,464 @@
...
+			modified = true
+			exs.OnTimeSectors, err = bitfield.SubtractBitField(exs.OnTimeSectors, onTimeDuplicates)
+			if err != nil {
+				return err
+			}
+		}
+
+		if modified {
+			expirationSetPowerSuspect = true
+			exs2, err := copyES(exs)
```
anorth, line 330: Now that we're not longer putting these in a map for `Set()`ing later, this might not be necessary.

anorth (approved):

actors/migration/nv3/miner.go
```diff
@@ -1,30 +1,472 @@
...
+	// Check for faults that need to be erased.
+	// Erased faults will be removed from bitfields and the power will be recomputed
+	// in the subsequent loop.
+	err := exq.ForEach(&exs, func(epoch int64) error { //nolint:nomaprange
+		// Detect sectors that are present in this expiration set as "early", but that
+		// have already terminated or duplicate a prior entry in the queue, and thus will
+		// be terminated before this entry is processed. The sector was rescheduled here
+		// upon fault, but the entry is stale and should not exist.
+		modified := false
+		earlyDuplicates, err := bitfield.IntersectBitField(exs.EarlySectors, processedExpiredSectors)
```
Stebalien, line 307: Wouldn't `bitfield.SubtractBitField(exs.EarlySectors, liveSectors)` be safer? That would handle completely removed/compacted out sectors.

Note: this code should work as-is assuming no compaction, which is probably safe.

Stebalien: Same below.

Stebalien (approved):

actors/migration/nv3/miner.go
```diff
@@ -1,30 +1,472 @@
...
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
 	"github.com/filecoin-project/specs-actors/actors/states"
+	"github.com/filecoin-project/specs-actors/actors/util/adt"
 )
 
 type minerMigrator struct {
 }
 
-func (m *minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, _ abi.ChainEpoch, _ address.Address, _ *states.Tree) (cid.Cid, error) {
+func (m *minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, priorEpoch abi.ChainEpoch, a addr.Address, tree *states.Tree) (cid.Cid, error) {
```
Stebalien, line 23: It's a bit strange that this _returns_ the new miner's state (without updating the state tree) but updates the power actor in the state tree.

actors/migration/nv3/miner.go
```diff
@@ -1,30 +1,472 @@
...
+		return false, nil
+	}
+
+	if err = st.SaveDeadlines(adt.WrapStore(ctx, store), deadlines); err != nil {
+		return false, err
+	}
+
+	for st.ProvingPeriodStart+miner.WPoStProvingPeriod <= epoch {
+		st.ProvingPeriodStart += miner.WPoStProvingPeriod
+	}
```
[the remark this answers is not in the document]
anorth: @Stebalien I requested this blunter but more obviously correct approach earlier.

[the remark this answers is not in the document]
Stebalien: I guess the catch up shouldn't be terrible. I just hate seeing math done this way.

anorth:
@acruikshank will you rebase this against master, or open a new PR?

acruikshank:
closed because this needs to go into the V2 release now (see #1198).

PR #1181, opened 2020-09-24
[v0.9] Remove penalties on recovery and on missed PoSt (FIP-0002)

anorth:
Honest miners are sometimes disproportionately penalized for operational failures and network congestion by the high undeclared-fault penalty (aka SP) for missed Window PoSts. This change reduces those costs without sacrificing much of the incentive to maintain reliable storage.

Summary:
* Remove fee incurred when a sector is successfully recovered.
* Remove fee incurred by sectors faulted when a Window PoSt is missed.
* Update Sector Fault Fee (FF) to 3.51 days of expected block reward, for ongoing faults.

As a result of this, declaring faults in advance confers no advantage compared with skipping them at Window PoSt. The DeclareFaults method remains, but may be removed in the future.

This change is a demonstration of upcoming [FIP-0002](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0002.md), which will describe more background and rationale.

Note that the new values are not directly tested in this PR. I intend to replicate on the `v2` line and change all the tests there. UPDATE: see #1183

FYI @Stebalien @magik6k @whyrusleeping @zixuanzh @nicola

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -299,6 +299,7 @@ type SubmitWindowedPoStParams struct {
 func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) *abi.EmptyValue {
 	currEpoch := rt.CurrEpoch()
 	store := adt.AsStore(rt)
+	networkVersion := rt.NetworkVersion()
 	var st State
 
 	if params.Deadline >= WPoStPeriodDeadlines {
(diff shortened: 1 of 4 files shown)


zixuanzh:
FIP-0002 Draft: https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0002.md

actors/builtin/miner/miner_actor.go
```diff
@@ -1630,33 +1646,47 @@ func handleProvingDeadline(rt Runtime) {
...
+			}
 		}
 		{
 			// Record faulty power for penalisation of ongoing faults, before popping expirations.
 			// This includes any power that was just faulted from missing a PoSt.
-			penaltyTarget := PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, deadline.FaultyPower.QA)
+			ongoingFaultyPower := deadline.FaultyPower.QA
+			if networkVersion >= network.Version3 {
+				// From network version 3, this *excludes* any power that was just faulted from missing a PoSt.
+				ongoingFaultyPower = previouslyFaultyPower
```
ZenGround0, line 1686: @lucaniz @irenegia I looked over this again and it appears to be correct.  `previouslyFaultPower` already accounts for `failedRecoveryPower` (we only handle failedRecoveryPower separately in the previous version because we used to charge this power SP).  Another way to put it, the value of `deadline.FaultyPower.QA` at line 1683 after processing deadline end = `previouslyFaultPower - newFaultyPower`

actors/builtin/miner/miner_actor.go
```diff
@@ -1630,33 +1646,47 @@ func handleProvingDeadline(rt Runtime) {
...
+				pledgeDelta = big.Sub(pledgeDelta, penaltyFromVesting)
+			}
 		}
 		{
 			// Record faulty power for penalisation of ongoing faults, before popping expirations.
 			// This includes any power that was just faulted from missing a PoSt.
-			penaltyTarget := PledgePenaltyForDeclaredFault(epochReward.ThisEpochRewardSmoothed, pwrTotal.QualityAdjPowerSmoothed, deadline.FaultyPower.QA)
+			ongoingFaultyPower := deadline.FaultyPower.QA
+			if networkVersion >= network.Version3 {
+				// From network version 3, this *excludes* any power that was just faulted from missing a PoSt.
```
ZenGround0, line 1685: @anorth might be worth calling out that this includes failedRecoveryPower since this confused a few of us earlier.

ZenGround0 (approved):
This looks good

PR #1183, opened 2020-09-24
[v2] Remove penalties on recovery and on missed PoSt

anorth:
This is #1181 adapted to the v2 code on master.

Honest miners are sometimes disproportionately penalized for operational failures and network congestion by the high undeclared-fault penalty (aka SP) for missed Window PoSts. This change reduces those costs without sacrificing much of the incentive to maintain reliable storage.

Summary:
* Remove fee incurred when a sector is successfully recovered.
* Remove fee incurred by sectors faulted when a Window PoSt is missed.
* Update Sector Fault Fee (FF) to 3.51 days of expected block reward, for ongoing faults.

As a result of this:
- Declaring faults in advance is irrational compared with skipping them at Window PoSt. The DeclareFaults method remains, but miners should avoid calling it. The rationality may change in the future. 
- There are no penalties charged at all during SubmitWindowPoSt, so a bunch of code was deleted.
- Submitting a PoSt with all sectors skipped is now actively bad, rather than just pointless, so is now rejected.

This change is a demonstration of upcoming FIP-0008, which will describe more background and rationale.

FYI @Stebalien @magik6k @whyrusleeping @zixuanzh @nicola

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -331,9 +331,9 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		// just skipping all sectors.
 		windowPoStProofType, err := info.SealProofType.RegisteredWindowPoStProof()
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to determine window PoSt type")
-		if len(params.Proofs) > 1 {
-			rt.Abortf(exitcode.ErrIllegalArgument, "expected at most one proof, got %d", len(params.Proofs))
-		} else if len(params.Proofs) == 1 && params.Proofs[0].PoStProof != windowPoStProofType {
+		if len(params.Proofs) != 1 {
+			rt.Abortf(exitcode.ErrIllegalArgument, "expected exactly one proof, got %d", len(params.Proofs))
+		} else if params.Proofs[0].PoStProof != windowPoStProofType {
 			rt.Abortf(exitcode.ErrIllegalArgument, "expected proof of type %s, got proof of type %s", params.Proofs[0], windowPoStProofType)
 		}
 
@@ -404,16 +404,13 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load proven sector info")
 
 		if len(sectorInfos) == 0 {
-			// Abort verification if all sectors are (now) faults.
+			// Abort verification if all sectors are (now) faults. There's nothing to prove.
 			// It's not rational for a miner to submit a Window PoSt marking *all* non-faulty sectors as skipped,
 			// since that will just cause them to pay a penalty at deadline end that would otherwise be zero
 			// if they had *not* declared them.
 			rt.Abortf(exitcode.ErrIllegalArgument, "cannot prove partitions with no active sectors")
 		}
-		if len(params.Proofs) == 0 {
-			// The miner _was_ supposed to prove something, but didn't.
-			rt.Abortf(exitcode.ErrIllegalArgument, "no proofs submitted in window PoSt for %d sectors", len(sectorInfos))
-		}
+
 		// Verify the proof.
 		// A failed verification doesn't immediately cause a penalty; the miner can try again.
 		//
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -316,12 +316,6 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams) 
 		rt.Abortf(exitcode.ErrIllegalArgument, "expected at most %d bytes of randomness, got %d", abi.RandomnessLength, len(params.ChainCommitRand))
 	}
 
-	// Get the total power/reward. We need these to compute penalties.
-	rewardStats := requestCurrentEpochBlockReward(rt)
-	pwrTotal := requestCurrentTotalPower(rt)
-
-	penaltyTotal := abi.NewTokenAmount(0)
-	pledgeDelta := abi.NewTokenAmount(0)
 	var postResult *PoStResult
 
 	var info *MinerInfo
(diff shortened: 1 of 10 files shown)


Stebalien:
> Declaring faults in advance is irrational compared with skipping them at Window PoSt. The DeclareFaults method remains, but miners should avoid calling it. The rationality may change in the future.

I thought the point here was that WindowPoSt samples sectors. The reduced fee for reporting early encourages miners to report sectors that may not be caught during the actual window PoSt.

actors/builtin/miner/miner_actor.go
```diff
@@ -397,49 +394,22 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
-				// The miner _was_ supposed to prove something, but didn't.
-				rt.Abortf(exitcode.ErrIllegalArgument, "no proofs submitted in window PoSt for %d sectors", len(sectorInfos))
-			}
-			// Verify the proof.
-			// A failed verification doesn't immediately cause a penalty; the miner can try again.
-			//
-			// This function aborts on failure.
-			verifyWindowedPost(rt, currDeadline.Challenge, sectorInfos, params.Proofs)
+		if len(sectorInfos) == 0 {
+			// Abort verification if all sectors are (now) faults.
```
irenegia, line 398: @anorth 
We could improve this saying that we abort verification when new faults (asks skipped faults) are larger than
Recovering + [PartitionSize - (AlreadyFaulty+Recovering)]/4.51.
Indeed, it is not rational to declare so many skipped faults. The miner loses less going for missing this deadline. 

CC @lucaniz @nicola

anorth: I'd prefer to avoid anything complicated here – I think that policy logic belongs in the node, not the actors.

The only reason this was previously allowed was because there was a point when a skipped fault was cheaper than a detected fault, so we needed to support skipping everything. It's not just the rationality - there's nothing to prove. I think of this as just rejecting proof of nothing.

irenegia: >I'd prefer to avoid anything complicated here – I think that policy logic belongs in the node, not the actors.

Agreed, I opened an issues in the lotus repo (#4025)

actors/builtin/miner/miner_test.go
```diff
@@ -1952,44 +1919,39 @@ func TestWindowPost(t *testing.T) {
...
 
-		// Now submit PoSt with all faults skipped
-		// Sectors should be penalized as an undeclared fault and unproven power won't be activated.
-		// Fee for skipped fault is undeclared fault fee, but it is split into the ongoing fault fee
-		// which is charged at next cron and the rest which is charged during submit PoSt.
-		undeclaredFee := actor.undeclaredFaultPenalty(infos)
-		declaredFee := actor.declaredFaultPenalty(infos)
-		faultFee := big.Sub(undeclaredFee, declaredFee)
-
+		// Now submit PoSt with all sectors skipped. Unproven power won't be activated.
```
acruikshank, line 1922: Maybe "PoSt fails to validate leaving power un-activated"? Note the `expectedPowerDelta` below could be anything and this test will still pass. `cfg` should probably just be `nil`.

acruikshank (approved):
Looks good. Some minor suggestions.

actors/builtin/miner/miner_actor.go
```diff
@@ -397,49 +394,22 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
-		totalPenaltyTarget := big.Add(undeclaredPenaltyTarget, declaredPenaltyTarget)
-		err = st.ApplyPenalty(totalPenaltyTarget)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to apply penalty")
-
-		// Pay penalty
-		vestingPenaltyTotal, balancePenaltyTotal, err := st.RepayPartialDebtInPriorityOrder(store, currEpoch, rt.CurrentBalance())
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unlock penalty for %v", undeclaredPenaltyPower)
-		penaltyTotal = big.Add(vestingPenaltyTotal, balancePenaltyTotal)
-		pledgeDelta = big.Sub(pledgeDelta, vestingPenaltyTotal)
+		if len(params.Proofs) == 0 {
```
ZenGround0, line 404: Instead could we take care of this earlier by changing line 325 to `if len(params.Proofs) != 1`?

ZenGround0: That should hit this case where the post is actually over active sectors.
And it should not impact correct use because we no longer allow this call to pass when skipping all sectors which seems to be the only valid reason to input 0 proofs.

actors/builtin/miner/miner_test.go
```diff
@@ -1884,52 +1852,51 @@ func TestWindowPost(t *testing.T) {
...
-			expectedPowerDelta: pwr.Neg(),
-			expectedPenalty:    faultFee,
+			expectedPowerDelta: pwrDelta.Neg(),
 		}
 		partitions = []miner.PoStPartition{
 			{Index: pIdx2, Skipped: bf(uint64(infos[1].SectorNumber))},
 		}
-		actor.submitWindowPoSt(rt, dlinfo, partitions, infos, cfg)
+		// Now all sectors are faulty so there's nothing to prove.
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
```
ZenGround0, line 1884: Another place where we could use ExpectAbortContainsMessage

actors/builtin/miner/miner_test.go
```diff
@@ -2235,25 +2187,20 @@ func TestDeadlineCron(t *testing.T) {
...
-		expectedFee := actor.undeclaredFaultPenalty(sectors)
-		// Add lots of funds to locked funds so expectedFee is taken from locked funds
-		actor.applyRewards(rt, expectedFee, big.Zero())
-
 		// introduce lots of fee debt
 		st = getState(rt)
 		feeDebt := big.Mul(big.NewInt(400), big.NewInt(1e18))
 		st.FeeDebt = feeDebt
 		rt.ReplaceState(st)
+		// Miner balance = IP, debt repayment covered by IP
```
ZenGround0, line 2195: nit: debt repayment covered by unlocked IP

ZenGround0 (approved):
This looks good

PR #1190, opened 2020-09-26
[v0.9] Tests for baseline function migration

anorth:
I wrote some tests for the migration of the baseline function. They pass. 

I don't think we should actually land this as-is, due to:
- code duplication
- very slow (>= 20s)
- not valuable after the migration happens

We can discuss adjustments that would make them better, but since I demonstrated what I set out to, I don't think it's urgent or important.

FYI @zixuanzh

### DIFF
--- a/actors/migration/nv3/reward_test.go
+++ b/actors/migration/nv3/reward_test.go
@@ -1,0 +1,70 @@
+package nv3
+
+import (
+	"context"
+	gbig "math/big"
+	"testing"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/network"
+	assert "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


actors/migration/nv3/reward_test.go
```diff
@@ -0,0 +1,70 @@
...
+		// With network version 0, after 1 year expect value to be 200% higher (allowing for small error).
+		multiplier := big.NewInt(3)
+		expectedValue := big.Mul(reward.BaselineInitialValueV0, multiplier)
+		diff := big.Sub(expectedValue, st.ThisEpochBaselinePower)
+		perrFrac := gbig.NewRat(1, 1).SetFrac(diff.Int, expectedValue.Int)
+		perr, _ := perrFrac.Float64()
+		assert.Less(t, perr, 1e-8)
+	}
+
+	// Upgrade to network version 1.
```
ZenGround0, line 47: I think this should be 3?  Maybe this is right but if so it's confusing to me.

ZenGround0 (approved):

PR #1192, opened 2020-09-26
Update miner tests to use NV=3 Window PoSt fee reductions

anorth:
More validation for the changes from #1181 (FIP-0002). I have already done this on `master` but wanted to check this has the same behaviour. There are many fewer tests on this branch, so it's not a big change.


actors/builtin/miner/miner_test.go
```diff
@@ -3109,7 +3059,7 @@ type cronConfig struct {
 	expectedEnrollment        abi.ChainEpoch
 	vestingPledgeDelta        abi.TokenAmount // nolint:structcheck,unused
 	detectedFaultsPowerDelta  *miner.PowerPair
-	detectedFaultsPenalty     abi.TokenAmount
+	//detectedFaultsPenalty     abi.TokenAmount
```
acruikshank, line 3062: remove?

acruikshank (approved):
LGTM

PR #1194, opened 2020-09-28
consume the right state tree when checking state invariants

Stebalien:


Stebalien:
Or was this supposed to be a pre-upgrade check?

ZenGround0 (approved):
Yeah this was just a mistake

Issue #1161, opened 2020-09-15
Audit and remove assertions


anorth:
Assertion failures result in a panic, which is undefined execution. In most cases, we probably want to define an exit code. There might be some that warrant the panic, but we should justify and document that decision in code.

PR #1195, opened 2020-09-29
fix AssertMsg panic with exit code

IPFSMain-Official:
proof validation failed, sector not found in sector set after cron.
This error occurred when I was testing in the 2k environment.

The problem is that `st.AddInitialPledgeRequirement(totalPledge)` in `miner_actor.go:ConfirmSectorProofsValid` panics in `AssertMsg`, but the message execution result is `Exit Code: 0`.
**The AssertMsg operation in specs-actor seems to cause the message not to execute correctly, but ExitCode returns 0. This does not capture the correct results on the outside.**

Whether need to consider returning the correct ExitCode in AssertMsg, along with the correct error message.

### DIFF
--- a/actors/util/assert.go
+++ b/actors/util/assert.go
@@ -1,12 +1,20 @@
 package util
 
-import "fmt"
+import (
+	"fmt"
+	"github.com/filecoin-project/go-state-types/exitcode"
+)
+
+type abort struct {
+	code exitcode.ExitCode
+	msg  string
+}
 
 // Indicates a condition that should never happen. If encountered, execution will halt and the
 // resulting state is undefined.
 func AssertMsg(b bool, format string, a ...interface{}) {
 	if !b {
-		panic(fmt.Sprintf(format, a...))
+		panic(abort{exitcode.ErrForbidden, fmt.Sprintf(format, a...)})
 	}
 }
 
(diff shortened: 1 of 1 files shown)


anorth:
Thanks. Can you give any more detail about how you reached the state where the sector was not found in sector set after cron. Is this working from `master` (`release/v0.9` has known issues here)?

From the actors' point of view, an assertion failure is an undefined execution and _there should be no exit code_. The message and block containing it cannot be processed. However, Lotus have declined to implement those semantics so we're left having to commit undefined state transitions (probably programmer error) to the chain.

See also #1161.

IPFSMain-Official:
Here is my error in this situation
```text
spec actors failure: negative initial pledge requirement -858040375071868 after adding -858775806479408 to prior 735431407540
vmctx send failed: to: t01000, method: 17: ret: [], err: spec actors failure: negative initial pledge requirement -858040375071868 after adding -858775806479408 to prior 735431407540 (RetCode=1)
```
This error should be in 2k environment to calculate the power may have some problems. The main problem is that when I use `lotus state search-msg` to look for the corresponding `CommitMsg` when I meet this error, the result of its execution is `Exit Code: 0`, but the message did not execute successfully. This can be confusing and difficult to expose. Therefore, I think `AssertMsg` should also send out the error information and the right ExitCode.

IPFSMain-Official:
This change should result in a fork, so consider adding `network version` as well. This is only a suggested change, Please do not merge it.

PR #1196, opened 2020-09-29
Extract miner partition state checks for re-use

anorth:
Extracted from the great checks that @Stebalien wrote in the unit test. I made slight refactors, and removed a check that required an on-time sector's expiration queue epoch to match it's declared epoch (it could instead have been moved up for CC replacement).

Part of #1165. There's more work to do here, which I'll chain off this PR.

### DIFF
--- a/actors/builtin/miner/testing.go
+++ b/actors/builtin/miner/testing.go
@@ -1,0 +1,432 @@
+package miner
+
+import (
+	"github.com/filecoin-project/go-bitfield"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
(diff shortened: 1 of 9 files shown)


actors/builtin/miner/testing.go
```diff
@@ -0,0 +1,416 @@
...
+		return nil, nil, err
+	}
+
+	// Live contains active sectors
+	if err = requireContainsAll(live, active, acc, "live does not contain active"); err != nil {
+		return nil, nil, err
+	}
+
+	// Active contains no faults
+	if err = requireContainsNone(live, partition.Terminated, acc, "live includes faults"); err != nil {
```
acruikshank, line 213: This is the same as two tests up. Should it be `active` and `partition.Faults`?

acruikshank (approved):
Look good except for the duplicated test.

PR #1198, opened 2020-09-30
cc upgrade correction migration [v2]

acruikshank:
Integrates changes made in  #1180 into the V2 migration. This is a migration to correct issues miners experienced in spacerace related to cc upgrading and then faulting. This PR is huge, but the vast majority of it is uninteresting:

1. `miner_corrections.go` is just the `miner` migration reviewed in #1180 with one important change. Where the old migration loaded power actor state from the state tree to add claims and cron tasks to it directly, this version simply adds the claims and crons to the powerUpdates data structure that the power actor migration will later incorporate into state.
2. I've made significant modifications to the migration infrastructure to allow migrations to take in more information (`MigrationInfo`), pass information to each other (`PowerUpdates`) and control the ordering (power actor now needs to come at the end).
3. I had to create a V0 scenario test vm because V0.9 didn't have one yet. Future migrations will simply be able to use each other's VMs prior to and after the migration. This accounts for most of the additions in this PR and doesn't need much review (although suggestions about a different way to do this are welcome).
4. I had to modify the existing test vm to use states.Actors and add a constructor that starts with a given state root and epoch, so it will work with migrations.
5. I added `correct_cc_upgrade_then_fault_scenario_test.go` from #1180 and modified it slightly so it will use the V0 VM prior to the migration and V2 VM after. It works.


actors/migration/test/correct_cc_upgrade_then_fault_scenario_test.go
```diff
@@ -0,0 +1,424 @@
...
+	v, _ = vm0.AdvanceByDeadlineTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
+
+	// miner should have no power yet
+	assert.Equal(t, uint64(0), vm0.MinerPower(t, v, minerAddrs.IDAddress).Raw.Uint64())
+
+	// Prove commit sector after max seal duration
+	v, err := v.WithEpoch(proveTime)
+	require.NoError(t, err)
+
+	for i := uint64(0); i < numSectors; i += 200 {
```
anorth, line 83: I think I know, but comment what's going on here.

actors/migration/test/correct_cc_upgrade_then_fault_scenario_test.go
```diff
@@ -0,0 +1,424 @@
...
+	vm0 "github.com/filecoin-project/specs-actors/v2/actors/migration/vm"
+	vm2 "github.com/filecoin-project/specs-actors/v2/support/vm"
+)
+
+func TestMigrationCorrectsCCThenFaultIssue(t *testing.T) {
+	ctx := context.Background()
+	v := vm0.NewVMWithSingletons(ctx, t)
+	addrs := vm0.CreateAccounts(ctx, t, v, 2, big.Mul(big.NewInt(100_000), vm0.FIL), 93837778)
+	worker, unverifiedClient := addrs[0], addrs[1]
+	numSectors := uint64(2349)
```
anorth, line 36: Leave a note about why this many sectors.

actors/migration/test/correct_cc_upgrade_then_fault_scenario_test.go
```diff
@@ -0,0 +1,424 @@
...
+			Skipped: bitfield.New(),
+		}},
+		Proofs: []proof.PoStProof{{
+			PoStProof: abi.RegisteredPoStProof_StackedDrgWindow32GiBV1,
+		}},
+		ChainCommitEpoch: dlInfo.Challenge,
+		ChainCommitRand:  []byte("not really random"),
+	})
+
+	fmt.Println(vm0.MinerPower(t, v, minerAddrs.IDAddress))
```
anorth, line 228: Remove prints

actors/migration/test/correct_cc_upgrade_then_fault_scenario_test.go
```diff
@@ -0,0 +1,424 @@
...
+
+		v, err = v.WithEpoch(dlInfo.Last())
+		require.NoError(t, err)
+		vm2.ApplyOk(t, v2, builtin.SystemActorAddr, builtin.CronActorAddr, big.Zero(), builtin.MethodsCron.EpochTick, nil)
+
+		fmt.Println(vm2.MinerPower(t, v2, minerAddrs.IDAddress))
+	}
+
+	// miner still has power
+	assert.Equal(t, numSectors*32<<30, vm2.MinerPower(t, v2, minerAddrs.IDAddress).Raw.Uint64())
```
anorth, line 374: Please add a call to `states.CheckStateInvariants`

actors/migration/top.go
```diff
@@ -6,22 +6,35 @@ import (
...
+	claims map[address.Address]power0.Claim
+	crons  map[abi.ChainEpoch][]power0.CronEvent
+}
+
+type MigrationInfo struct {
+	address      address.Address // actor's address
+	balance      abi.TokenAmount // actor's balance
+	priorEpoch   abi.ChainEpoch  // epoch of last state transition prior to migration
+	powerUpdates *PowerUpdates   // used to update power and add cron tasks to power actor state
+}
```
anorth, line 32: 👍

anorth (commented):
This basically looks good.

I think it would be much better to land the v0 vm in the `release/v0.9` branch, so it is left behind. I like the pattern of migrations being able to use the vm from the before and after branches. Consider also landing the VM changes to use `states.Actor` separately, but no big deal.

anorth (approved):
Remove `TestActor` from `gen.go` too.

PR #1199, opened 2020-09-30
State test for the miner state fix migration.

anorth:
Constructs miner state corresponding to the broken case and then runs the migration to verify correctness.

This doesn't work yet: the migration being invoked doesn't yet repair the state.

Replaces #1188

### DIFF
--- a/actors/migration/miner_test.go
+++ b/actors/migration/miner_test.go
@@ -1,0 +1,492 @@
+package migration
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	addr "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-bitfield"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	builtin0 "github.com/filecoin-project/specs-actors/actors/builtin"
+	miner0 "github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	power0 "github.com/filecoin-project/specs-actors/actors/builtin/power"
+	states0 "github.com/filecoin-project/specs-actors/actors/states"
+	adt0 "github.com/filecoin-project/specs-actors/actors/util/adt"
+	cid "github.com/ipfs/go-cid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	miner2 "github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/v2/support/ipld"
+	tutil "github.com/filecoin-project/specs-actors/v2/support/testing"
+)
+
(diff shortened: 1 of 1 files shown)


actors/migration/miner_test.go
```diff
@@ -0,0 +1,492 @@
...
+		},
+	})
+
+	phead, err = store.Put(ctx, pst)
+	require.NoError(t, err)
+	setActorState(t, tree, builtin0.StoragePowerActorAddr, builtin0.StoragePowerActorCodeID, phead)
+
+	mhead, err = store.Put(ctx, st)
+	require.NoError(t, err)
+	setActorState(t, tree, maddr, builtin0.StorageMinerActorCodeID, mhead)
```
acruikshank, line 179: It would be a nice cross-check of `CheckStateInvariants` to run it here and assert that it's finding the problem.

acruikshank: I guess that's only possible if we back port it to v0.9.

actors/migration/miner_test.go
```diff
@@ -0,0 +1,492 @@
...
+		_, deadline, partition := loadDeadlineAndPartition2(t, &st, store, dlIdx, pIdx)
+		assertDeadlineExpirationQueue2(t, store, deadline, pIdx, map[abi.ChainEpoch][]uint64{
+			advancedExpiration: {pIdx},
+			faultExpiration:    {pIdx},
+			expectedExpiration: {pIdx},
+		})
+
+		assertPartitionExpirationQueue2(t, store, partition, map[abi.ChainEpoch]*miner2.ExpirationSet{
+			// The sector is  scheduled here, with faulty power.
+			advancedExpiration: {
```
acruikshank, line 222: This describes the case where the migration is run before the advancedExpiration expires. Given the relatively short window between upgrade and expiry, we're unlikely to find this in the wild. If the advanced expiration has expired, we also expect inaccuracies in partition and power table that need to be corrected, but this is still a good demonstration of the problem and evidence that the migration is working.

actors/migration/miner_test.go
```diff
@@ -0,0 +1,492 @@
...
+	assert.Equal(t, uint64(2), dlIdx)
+	assert.Equal(t, uint64(0), pIdx)
+	_, deadline, partition := loadDeadlineAndPartition0(t, st, store, dlIdx, pIdx)
+	quant := st.QuantSpecForDeadline(dlIdx)
+	expectedExpiration := quant.QuantizeUp(initialExpiration)
+
+	assertDeadlineExpirationQueue0(t, store, deadline, pIdx, map[abi.ChainEpoch][]uint64{
+		expectedExpiration: {pIdx},
+	})
+	assertPartitionExpirationQueue0(t, store, partition, map[abi.ChainEpoch]*miner0.ExpirationSet{
```
acruikshank, line 92: ❤️

acruikshank (approved):
This looks great. It would be great to expand the test to cover migrations after the advanced expiration set has expired, but this is valuable on its own.

PR #1202, opened 2020-09-30
Extract miner deadline state checks for re-use

anorth:
Continues #1165.


actors/builtin/miner/testing.go
```diff
@@ -51,6 +51,7 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 
 	// Check deadlines
 	if err := deadlines.ForEach(store, func(dlIdx uint64, dl *Deadline) error {
+		acc := acc.WithPrefix("deadline %d: ", dlIdx) // Shadow
```
acruikshank, line 54: 👍

acruikshank (approved):
Looks good.

PR #1203, opened 2020-10-01
Threadsafe

### LINKED ISSUE #1201: State migration is not threadsafe.
Lotus needs to be able to call the migration function multiple times concurrently. This function needs to be thread-safe (it doesn't need to be pure, caching would actually be quite nice, it just needs to be safe).

This is unsafe:

https://github.com/filecoin-project/specs-actors/blob/e061eb250c49c651e63f598d50fb4efa67dfd62d/actors/migration/top.go#L123

We should probably also drop the rest of the shared state (shared lock, etc) and create some form of "context" for the specific migration.

ZenGround0:
closes #1201 
Thanks for the context idea @stebalien and partial implementation @acruikshank


actors/migration/top.go
```diff
@@ -23,11 +23,15 @@ import (
...
 var (
 	defaultMaxWorkers = 16
-	sem               *semaphore.Weighted
-	actOutMu          = &sync.Mutex{}
-	powerUpdateMu     = &sync.Mutex{}
 )
 
+// SyncCtx traverses the migration to coordinate safe multithreaded state processing
+type SyncContext struct {
+	sem           *semaphore.Weighted
```
Stebalien, line 30: no need for the pointers.

actors/migration/top.go
```diff
@@ -164,17 +173,17 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
 		}
 
 		// Hand off migration of one actor, blocking if we are out of worker goroutines
-		if err := sem.Acquire(ctx, 1); err != nil {
+		if err := syncCtx.sem.Acquire(ctx, 1); err != nil {
 			return err
 		}
-		go migrateOneActor(ctx, store, addr, *actorIn, actorsOut, priorEpoch, powerUpdates, transferCh, errCh)
+		go migrateOneActor(ctx, store, addr, *actorIn, actorsOut, priorEpoch, powerUpdates, syncCtx, transferCh, errCh)
```
Stebalien, line 179: If you buffer the channels (buffer size = number of workers), you should be able to manage the semaphore entirely in this function.

```suggestion
		go func() {
		    migrateOneActor(ctx, store, addr, *actorIn, actorsOut, priorEpoch, powerUpdates, syncCtx, transferCh, errCh)
		    sem.Release(ctx, 1)
		}()
```

Stebalien (commented):
Digging into this a bit, I'm really not a fan using mutexes this way. What if we:

1. Managed the semaphore entirely inside the main loop (see my comment).
2. Attached locks to the _specific_ data structures (i.e., thread-safe state-tree, etc.).

Alternatively, funnel all update events back to some main event loop that builds up the new state tree (this is the "go" way to do things").

A sync-context makes sense for semaphores, but is kind of sketchy for locks.

Stebalien:
I'm sorry, I'm going to give some conflicting advice.  My concern with the current approach is that it makes the upgrade functions have side effects, when before they were nicely structured as side-effect free actor conversions.

What if we just returned everything to the top? That is, we return a "migration" result that includes:

* The new actor head (result of the state migration).
* Any new cron events.
* Any changes to the actor balance.
* Any updates to power claims.

That:

* Makes it easier to reason about the control flow. Basically, all migrations will be independent, and all results will be processed in one place a the very top (e.g., funnled through a single channel).
* Makes caching possible in the future (as actor upgrades will be independent (mostly).

E.g., something like (non-functional, but hopefully fleshed out enough so I don't keep giving conflicting advice):

```diff
diff --git a/actors/migration/miner.go b/actors/migration/miner.go
index 6ce42d38..2f4d5ff5 100644
--- a/actors/migration/miner.go
+++ b/actors/migration/miner.go
@@ -19,52 +19,51 @@ import (
 type minerMigrator struct {
 }
 
-func (m *minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, info MigrationInfo) (cid.Cid, abi.TokenAmount, error) {
-	transfer := big.Zero()
+func (m *minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, info MigrationInfo) (*StateMigrationResult, error) {
 	// first correct issues with miners due problems in old code
-	head, err := m.CorrectState(ctx, store, head, info.priorEpoch, info.address, info.syncCtx)
+	result, err := m.CorrectState(ctx, store, head, info.priorEpoch)
 	if err != nil {
-		return cid.Undef, big.Zero(), err
+		return nil, err
 	}
 
 	var inState miner0.State
-	if err := store.Get(ctx, head, &inState); err != nil {
-		return cid.Undef, big.Zero(), err
+	if err := store.Get(ctx, result.NewHead, &inState); err != nil {
+		return nil, err
 	}
 
 	infoCid, err := m.migrateInfo(ctx, store, inState.Info)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("info: %w", err)
+		return nil, xerrors.Errorf("info: %w", err)
 	}
 
 	vestingFunds, err := m.migrateVestingFunds(ctx, store, inState.VestingFunds)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("vesting funds: %w", err)
+		return nil, xerrors.Errorf("vesting funds: %w", err)
 	}
 
 	precommitsRoot, err := m.migratePreCommittedSectors(ctx, store, inState.PreCommittedSectors)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("precommitted sectors: %w", err)
+		return nil, xerrors.Errorf("precommitted sectors: %w", err)
 	}
 
 	precommitExpiryRoot, err := m.migrateBitfieldQueue(ctx, store, inState.PreCommittedSectorsExpiry)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("precommit expiry queue: %w", err)
+		return nil, xerrors.Errorf("precommit expiry queue: %w", err)
 	}
 
 	allocatedSectors, err := m.migrateAllocatedSectors(ctx, store, inState.AllocatedSectors)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("allocated sectors: %w", err)
+		return nil, xerrors.Errorf("allocated sectors: %w", err)
 	}
 
 	sectorsRoot, err := m.migrateSectors(ctx, store, inState.Sectors)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("sectors: %w", err)
+		return nil, xerrors.Errorf("sectors: %w", err)
 	}
 
 	deadlinesRoot, err := m.migrateDeadlines(ctx, store, inState.Deadlines)
 	if err != nil {
-		return cid.Undef, big.Zero(), xerrors.Errorf("deadlines: %w", err)
+		return nil, xerrors.Errorf("deadlines: %w", err)
 	}
 
 	// To preserve v2 Balance Invariant that actor balance > initial pledge + pre commit deposit + locked funds
@@ -77,7 +76,7 @@ func (m *minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore,
 	availableBalance := big.Sub(info.balance, minerLiabilities)
 	if availableBalance.LessThan(big.Zero()) {
 		debt = availableBalance.Neg() // debt must always be positive
-		transfer = debt
+		result.Transfer = debt
 	}
 
 	outState := miner2.State{
@@ -98,7 +97,12 @@ func (m *minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore,
 	}
 
 	c, err := store.Put(ctx, &outState)
-	return c, transfer, err
+	if err != nil {
+		return nil, err
+	}
+
+	result.NewHead = c
+	return result, nil
 }
 
 func (m *minerMigrator) migrateInfo(ctx context.Context, store cbor.IpldStore, c cid.Cid) (cid.Cid, error) {
diff --git a/actors/migration/miner_corrections.go b/actors/migration/miner_corrections.go
index e633abfb..bf2da2d3 100644
--- a/actors/migration/miner_corrections.go
+++ b/actors/migration/miner_corrections.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 
-	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
@@ -12,56 +11,64 @@ import (
 	cbor "github.com/ipfs/go-ipld-cbor"
 
 	miner "github.com/filecoin-project/specs-actors/actors/builtin/miner"
-	power0 "github.com/filecoin-project/specs-actors/actors/builtin/power"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
 )
 
 func (m *minerMigrator) CorrectState(ctx context.Context, store cbor.IpldStore, head cid.Cid,
-	priorEpoch abi.ChainEpoch, a addr.Address, syncCtx *SyncContext,
-) (cid.Cid, error) {
-
+	priorEpoch abi.ChainEpoch) (*StateMigrationResult, error) {
+	result := new(StateMigrationResult)
 	epoch := priorEpoch + 1
 	var st miner.State
 	if err := store.Get(ctx, head, &st); err != nil {
-		return cid.Undef, err
+		return nil, err
 	}
 
 	adtStore := adt.WrapStore(ctx, store)
 
 	sectors, err := miner.LoadSectors(adtStore, st.Sectors)
 	if err != nil {
-		return cid.Undef, err
+		return nil, err
 	}
 
 	info, err := st.GetInfo(adtStore)
 	if err != nil {
-		return cid.Undef, err
+		return nil, err
 	}
 
-	powerClaimSuspect, err := m.correctForCCUpgradeThenFaultIssue(ctx, store, &st, sectors, epoch, info.SectorSize)
+	powerClaimSuspect, err := m.correctForCCUpgradeThenFaultIssue(adtStore, &st, sectors, epoch, info.SectorSize)
 	if err != nil {
-		return cid.Undef, err
+		return nil, err
 	}
 
 	if powerClaimSuspect {
-		err := m.updatePowerState(ctx, adtStore, syncCtx, &st, a, epoch)
+		result.ClaimUpdate, err = m.computeClaim(adtStore, &st)
 		if err != nil {
-			return cid.Undef, err
+			return nil, err
 		}
+
+		evt, err := m.computeProvingPeriodCron(adtStore, &st, epoch)
+		if err != nil {
+			return nil, err
+		}
+		result.CronEvents = append(result.CronEvents, evt)
 	}
 
-	newHead, err := store.Put(ctx, &st)
-	return newHead, err
+	result.NewHead, err = store.Put(ctx, &st)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
 }
 
 func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
-	ctx context.Context, store cbor.IpldStore, st *miner.State, sectors miner.Sectors, epoch abi.ChainEpoch,
+	store adt.Store, st *miner.State, sectors miner.Sectors, epoch abi.ChainEpoch,
 	sectorSize abi.SectorSize,
 ) (bool, error) {
 
 	quantSpec := st.QuantSpecEveryDeadline()
 
-	deadlines, err := st.LoadDeadlines(adt.WrapStore(ctx, store))
+	deadlines, err := st.LoadDeadlines(store)
 	if err != nil {
 		return false, err
 	}
@@ -78,8 +85,8 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 	}
 
 	deadlinesModified := false
-	err = deadlines.ForEach(adt.WrapStore(ctx, store), func(dlIdx uint64, deadline *miner.Deadline) error {
-		partitions, err := adt.AsArray(adt.WrapStore(ctx, store), deadline.Partitions)
+	err = deadlines.ForEach(store, func(dlIdx uint64, deadline *miner.Deadline) error {
+		partitions, err := adt.AsArray(store, deadline.Partitions)
 		if err != nil {
 			return err
 		}
@@ -88,7 +95,7 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 		allFaultyPower := miner.NewPowerPairZero()
 		var part miner.Partition
 		err = partitions.ForEach(&part, func(partIdx int64) error {
-			exq, err := miner.LoadExpirationQueue(adt.WrapStore(ctx, store), part.ExpirationsEpochs, quantSpec)
+			exq, err := miner.LoadExpirationQueue(store, part.ExpirationsEpochs, quantSpec)
 			if err != nil {
 				return err
 			}
@@ -132,7 +139,7 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 		// if we've failed to update at last proving period, expect post submissions to contain bits it shouldn't
 		if missedProvingPeriodCron {
 			deadline.PostSubmissions = bitfield.New()
-			if err := deadlines.UpdateDeadline(adt.WrapStore(ctx, store), dlIdx, deadline); err != nil {
+			if err := deadlines.UpdateDeadline(store, dlIdx, deadline); err != nil {
 				return err
 			}
 			deadlinesModified = true
@@ -153,7 +160,7 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 
 			deadline.FaultyPower = allFaultyPower
 
-			if err := deadlines.UpdateDeadline(adt.WrapStore(ctx, store), dlIdx, deadline); err != nil {
+			if err := deadlines.UpdateDeadline(store, dlIdx, deadline); err != nil {
 				return err
 			}
 			deadlinesModified = true
@@ -168,56 +175,34 @@ func (m *minerMigrator) correctForCCUpgradeThenFaultIssue(
 		return false, nil
 	}
 
-	if err = st.SaveDeadlines(adt.WrapStore(ctx, store), deadlines); err != nil {
+	if err = st.SaveDeadlines(store, deadlines); err != nil {
 		return false, err
 	}
 
 	return true, err
 }
 
-func (m *minerMigrator) updatePowerState(ctx context.Context, store adt.Store, syncCtx *SyncContext, st *miner.State,
-	a addr.Address, epoch abi.ChainEpoch,
-) error {
-	err := m.updateClaim(ctx, store, st, syncCtx, a)
-	if err != nil {
-		return err
-	}
-
-	err = m.updateProvingPeriodCron(a, st, epoch, store, syncCtx)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
-
-func (m *minerMigrator) updateProvingPeriodCron(a addr.Address, st *miner.State, epoch abi.ChainEpoch, store adt.Store, syncCtx *SyncContext) error {
+func (m *minerMigrator) computeProvingPeriodCron(store adt.Store, st *miner.State, epoch abi.ChainEpoch) (*StateMigrationCronEvent, error) {
 	var buf bytes.Buffer
 	payload := &miner.CronEventPayload{
 		EventType: miner.CronEventProvingDeadline,
 	}
 	err := payload.MarshalCBOR(&buf)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	dlInfo := st.DeadlineInfo(epoch)
-	u := cronUpdate{
-		epoch: dlInfo.Last(),
-		event: power0.CronEvent{
-			MinerAddr:       a,
-			CallbackPayload: buf.Bytes(),
-		},
-	}
-	syncCtx.powerUpdateCh <- u
-
-	return nil
+	return &StateMigrationCronEvent{
+		Epoch:   dlInfo.Last(),
+		Payload: buf.Bytes(),
+	}, nil
 }
 
-func (m *minerMigrator) updateClaim(ctx context.Context, store adt.Store, st *miner.State, syncCtx *SyncContext, a addr.Address) error {
+func (m *minerMigrator) computeClaim(store adt.Store, st *miner.State) (*power.Claim, error) {
 	deadlines, err := st.LoadDeadlines(store)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	activePower := miner.NewPowerPairZero()
@@ -234,17 +219,12 @@ func (m *minerMigrator) updateClaim(ctx context.Context, store adt.Store, st *mi
 		})
 	})
 	if err != nil {
-		return err
+		return nil, err
 	}
-	u := claimUpdate{
-		addr: a,
-		claim: power0.Claim{
-			RawBytePower:    activePower.Raw,
-			QualityAdjPower: activePower.QA,
-		},
-	}
-	syncCtx.powerUpdateCh <- u
-	return nil
+	return &power.Claim{
+		RawBytePower:    activePower.Raw,
+		QualityAdjPower: activePower.QA,
+	}, nil
 }
 
 type expirationQueueStats struct {
diff --git a/actors/migration/top.go b/actors/migration/top.go
index a8d4df17..c482b1ae 100644
--- a/actors/migration/top.go
+++ b/actors/migration/top.go
@@ -17,6 +17,7 @@ import (
 	states0 "github.com/filecoin-project/specs-actors/actors/states"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
 	"github.com/filecoin-project/specs-actors/v2/actors/states"
 )
 
@@ -90,6 +91,18 @@ type StateMigration interface {
 	MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, info MigrationInfo) (newHead cid.Cid, balanceChange abi.TokenAmount, err error)
 }
 
+type StateMigrationResult struct {
+	NewHead     cid.Cid
+	ClaimUpdate *power.Claim
+	Transfer    abi.TokenAmount
+	CronEvents  []*StateMigrationCronEvent
+}
+
+type StateMigrationCronEvent struct {
+	Epoch   abi.ChainEpoch
+	Payload []byte
+}
+
 type ActorMigration struct {
 	OutCodeCID     cid.Cid
 	StateMigration StateMigration
@@ -192,8 +205,11 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
 			return err
 		}
 		go func(actorIn states0.Actor) {
-			migrateOneActor(ctx, store, addr, actorIn, priorEpoch, syncCtx)
-			sem.Release(1)
+			defer sem.Release(1)
+			result, err := migrateOneActor(ctx, store, addr, actorIn, priorEpoch)
+			if err != nil {
+				// TODO write to an result channel.
+			}
 		}(*actorIn) // race conditions appear unless we dereference outside the closure
 
 		// Read from err, actorsOut and transfer channels without blocking.
@@ -298,40 +314,41 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
 	return actorsOut.Flush()
 }
 
+type migrationResult struct {
+	states.Actor
+	StateMigrationResult
+}
+
 func migrateOneActor(ctx context.Context, store cbor.IpldStore, addr address.Address, actorIn states0.Actor,
-	priorEpoch abi.ChainEpoch, syncCtx *SyncContext,
-) {
+	priorEpoch abi.ChainEpoch) (*migrationResult, error) {
 	// This will be migrated at the end
 	if _, found := deferredMigrations[actorIn.Code]; found {
-		return
+		return nil, nil
 	}
 
 	migration := migrations[actorIn.Code]
 	codeOut := migration.OutCodeCID
-	headOut, transfer, err := migration.StateMigration.MigrateState(ctx, store, actorIn.Head, MigrationInfo{
+	result, err := migration.StateMigration.MigrateState(ctx, store, actorIn.Head, MigrationInfo{
 		address:    addr,
 		balance:    actorIn.Balance,
 		priorEpoch: priorEpoch,
-		syncCtx:    syncCtx,
 	})
 
 	if err != nil {
 		err = xerrors.Errorf("state migration error on %s actor at addr %s: %w", builtin.ActorNameByCode(codeOut), addr, err)
-		syncCtx.errCh <- err
-		return
+		return nil, err
 	}
 
 	// set up new state root with the migrated state
-	actorOut := states.Actor{
-		Code:       codeOut,
-		Head:       headOut,
-		CallSeqNum: actorIn.CallSeqNum,
-		Balance:    big.Add(actorIn.Balance, transfer),
-	}
-	syncCtx.actorsOutCh <- actorUpdate{addr: addr, actorOut: actorOut}
-	if transfer.GreaterThan(big.Zero()) {
-		syncCtx.transferCh <- transfer
-	}
+	return &migrationResult{
+		states.Actor{
+			Code:       codeOut,
+			Head:       result.NewHead,
+			CallSeqNum: actorIn.CallSeqNum,
+			Balance:    big.Add(actorIn.Balance, transfer),
+		},
+		result,
+	}, nil
 }
 
 func accumulateResults(syncCtx *SyncContext, transferAcc big.Int, powerUpdatesAcc *PowerUpdates, actorsOut *states.Tree) error {
```

actors/migration/top.go
```diff
@@ -249,11 +328,32 @@ func migrateOneActor(ctx context.Context, store cbor.IpldStore, addr address.Add
...
+		case transfer := <-syncCtx.transferCh:
+			transferAcc = big.Add(transferAcc, transfer)
+		case pwrUpdate := <-syncCtx.powerUpdateCh:
+			pwrUpdate.Apply(powerUpdatesAcc)
+		case actorUpdate := <-syncCtx.actorsOutCh:
+			if err := actorsOut.SetActor(actorUpdate.addr, &actorUpdate.actorOut); err != nil {
+				return err
+			}
+		default:
+			break READLOOP
```
Stebalien, line 353: It would be nice to accumulate results in a different goroutine, rather than start/stop like this. I think this works, but it's hard to reason about.

ZenGround0: I hear you.  One advantage of the current approach is the simplicity of error handling in the main goroutine.  I played with reading the error channel in the main loop and sending directly there from the worker goroutines which is nice.  But an annoying thing is that the processing goroutine can also error, so I can't think of a better way forward than having two error channels and routing all errors through the processing goroutine which is starting to make this change feel not worth it.  Am I missing any clean patterns for dealing with this?

Stebalien: The standard pattern here is to create a fan-out/fan-in pipeline.

1. Have one goroutine generating work, sending it to an "input" channel.
2. Have N (workers) goroutines processing work, sending results to an "output" channel.
3. Back on the main goroutine, process results off the "output" channel.

There are a couple of ways to do this, but the errgroup library makes this pretty simple:



```go
package main

import (
	"context"
	"sync"

	"golang.org/x/sync/errgroup"
)

const workers = 10

type job struct{}
type result struct{}

func getInput(ctx context.Context) (job, error)           { panic("unimplemented") }
func doWork(ctx context.Context, inp job) (result, error) { panic("unimplemented") }
func applyResult(ctx context.Context, res result) error   { panic("unimplemented") }

func process(ctx context.Context) (int, error) {
	grp, ctx := errgroup.WithContext(ctx)

	input := make(chan job)
	output := make(chan result)

	// Pipeline stage 1: generate jobs (fan out)

	grp.Go(func() error {
		// signal to the workers that there is no new input.
		defer close(input)

		for {
			nextInput, err := getInput(ctx)
			if err != nil {
				return err
			}

			select {
			case input <- nextInput:
			case <-ctx.Done():
				return ctx.Err()
			}
		}
	})

	// Pipeline stage 2: process jobs (map).

	var workerWg sync.WaitGroup
	for i := 0; i < workers; i++ {
		workerWg.Add(1)
		grp.Go(func() error {
			defer workerWg.Done()

			for inp := range input {
				res, err := doWork(ctx, inp)
				if err != nil {
					return err
				}
				select {
				case output <- res:
				case <-ctx.Done():
					return ctx.Err()
				}
			}
			return nil
		})
	}
	// standard pattern for closing an output channel once workers are done.
	grp.Go(func() error {
		workerWg.Wait()
		close(output)
		return nil
	})

	// Pipeline stage 3: apply results (fan in).

	// This doesn't _have_ to go in a separate goroutine, but it can
	// simplify error handling.
	var res int
	grp.Go(func() error {
		for res := range output {
			if err := applyResult(ctx, res); err != nil {
				return err
			}
		}
		return nil
	})

	if err := grp.Wait(); err != nil {
		return 0, err
	}

	return res, nil
}
```

ZenGround0: Thank you for the detailed explanation

actors/migration/test/correct_cc_upgrade_then_fault_scenario_test.go
```diff
@@ -305,7 +305,7 @@ func TestMigrationCorrectsCCThenFaultIssue(t *testing.T) {
 	// migrate miner
 	//
 
-	nextRoot, err := migration.MigrateStateTree(ctx, v.Store(), v.StateRoot(), v.GetEpoch(), migration.DefaultConfig()) 
+	nextRoot, err := migration.MigrateStateTree(ctx, v.Store(), v.StateRoot(), v.GetEpoch(), migration.Config{MaxWorkers: 1}) // v.Store() is not threadsafe
```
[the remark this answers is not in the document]
ZenGround0: Probably won't get to this for our next migration but agree we want this.  Filed here #1214

actors/migration/power.go
```diff
@@ -20,33 +20,34 @@ import (
...
-func (m *powerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, info MigrationInfo) (cid.Cid, abi.TokenAmount, error) {
+func (m *powerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, info MigrationInfo) (*StateMigrationResult, error) {
 	var inState power0.State
 	if err := store.Get(ctx, head, &inState); err != nil {
-		return cid.Undef, big.Zero(), err
+		return nil, err
 	}
 
-	cronEventsRoot, err := m.updateCronEvents(ctx, store, inState.CronEventQueue, info.powerUpdates)
+	cronEventsRoot, err := m.updateCronEvents(ctx, store, inState.CronEventQueue, m.powerUpdates)
```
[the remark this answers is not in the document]
ZenGround0: While I am planning to leave it like this for now in the interest of time I generally agree.  I think that there is a more general refactoring which satisfies ordering constraints between actors without having to special case certain migration functions that I'd like to explore in the future.

actors/migration/top.go
```diff
@@ -119,54 +158,110 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
-	if err != nil {
+
+	if err := grp.Wait(); err != nil {
 		return cid.Undef, err
 	}
 
 	// Migrate Power actor
 	pm := migrations[builtin0.StoragePowerActorCodeID].StateMigration.(*powerMigrator)
 	pm.actorsIn = actorsIn
+	pm.powerUpdates = powerUpdates
```
Stebalien, line 244: :+1:

actors/migration/top.go
```diff
@@ -19,30 +22,66 @@ import (
...
+	Apply(*PowerUpdates)
+}
+
+type claimUpdate struct {
+	addr  address.Address
+	claim power0.Claim
+}
+
+// Overwrite claim
+func (c claimUpdate) Apply(p *PowerUpdates) {
```
Stebalien, line 55: This is an unusual pattern in go (although I see why you're using it). I'd at least call it `x.ApplyTo(y)`, usually `x.Apply(y)` applies `y` to `x`.

actors/migration/top.go
```diff
@@ -119,54 +158,110 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
-	// Set new state root actors as we go
-	err = actorsIn.ForEach(func(addr address.Address, actorIn *states.Actor) error {
-		transfer, err := migrateOneActor(ctx, store, addr, *actorIn, actorsOut, priorEpoch, powerUpdates)
-		if err != nil {
-			return err
+	// Setup synchronization
+	grp, ctx := errgroup.WithContext(ctx)
+	inputCh := make(chan *migrationInput)
+	resultCh := make(chan *migrationResult)
+	var workerWg sync.WaitGroup
```
Stebalien, line 172: uber nit: I'd move this variable down right above the workers, to make it clear it's only used there.

Stebalien (approved):
This looks correct to me. Make sure to test it with the race detector enabled to double check.

ZenGround0:
The existing scenario test passes with the `-race` flag set.  Merging for now, will shoot for writing a test with more than one worker thread before next upgrade and at least will play with manual migrations using this code to get more coverage.

PR #1204, opened 2020-10-01
Remove concurrency for now

ZenGround0:
Remove parallelism for thread safety
We'll bring it back after it is fixed up according to review comments on #1203


Stebalien (approved):
Yeah, it's best to take our time with this and get it right. Thanks.

anorth:
Thank you! 🙌  for a tough call.

PR #1205, opened 2020-10-01
Migration Fixes

ZenGround0:
1. The proving period invariant being asserted is not actually an invariant because when we create miners their proving period starts are in the future.  We can add back the deadline invariant assertion here upon request since they should still always hold but would prefer if we test for them in the invariant checks instead.

2. Validation checks uncovered that we are not zeroing out recovering power.


anorth (approved):
Thanks. Could you please follow up to pass the epoch through the state checks and restore the proving period and deadline checks there?

PR #1207, opened 2020-10-01
Multisig purges approvals from removed signers

### LINKED ISSUE #1041: Swapping a signer in `MultiSigActor` allows for a signer to approve a transaction one more time
### Description

A MultiSigActor allows the signers (if they agree) to swap a signer’s address for another address.

The addresses are not modified for pending transactions. Hence, if there is an existing transaction which has the
user’s first address in the Approved array. They may call Approve() again after the address has been swapped
which will add the new address to Approved.

A signer may repeat this process multiple times giving them the potential to be the only signer for a transaction
if their SwapSigner() transactions are approved.

### Recommendations

Consider iterating through existing PendingTransactions and swapping the addresses or otherwise prevent
the double approving of a transaction.

## Notes
Concrete example: A, B, C have a 3/3 multisig.  A wants to approve Tx.  A proposes and approves Tx.  A proposes swapping A for A'.  B and C approve swap.  A' approves Tx. A' proposes swapping A' for A''.  B and C approve swap.  A'' approves Tx.  With 3 approvals Tx is executed even though B and C never approved.

Can we implement the recommendation given desired performance and gas targets for a multisig user?  What is the expected size of concurrent pending transactions? Does it not matter because the swapping party pays the burden of the gas covering security?  Feels like recommendation has acceptable trade offs -- swapping should be relatively rare.  If we accept the tradeoff then iterating and swapping should be an easy fix.

No action alternative is to communicate the security policy that approvers of SwapAddress must not sign off on swap and consider the operation safe until all transactions containing previous address are canceled (requires cooperation of all proposers so possibly all addresses in multisig which constrains secure behavior)

anorth:
### Motivation
When a multisig's signers are changed, especially through `SwapSigner`, if both a removed and added address are controlled by the same external entity, that entity might get to approve a pending transaction twice: once with a removed address and once with a new one. Whether or not this is a problem depends whether the desired behaviour treats the addresses as entities themselves, or as a handle to (possibly shared) other entities.

### Change
To avoid potential mis-use, this PR purges all approvals of pending transactions by removed signers.
- If a transaction's only approval (by its proposer) is removed, the transaction is deleted.
- If a transaction's proposer is removed but another signer has approved, the next approving signer becomes the "proposer", meaning they have the ability to cancel the transaction.

This cleans up an odd case in the old semantics where a pending transaction could be left un-cancellable if its proposer was removed as a signer.

Promoting the second approver to proposer was an arbitrary policy choice. We could just as easily delete all pending transactions proposed by a removed signer, regardless of other approvals, if desired.

Closes #1041.

### DIFF
--- a/actors/builtin/multisig/multisig_actor.go
+++ b/actors/builtin/multisig/multisig_actor.go
@@ -338,6 +338,7 @@ func (a Actor) RemoveSigner(rt runtime.Runtime, params *RemoveSignerParams) *abi
 	resolvedOldSigner, err := builtin.ResolveToIDAddr(rt, params.Signer)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve address %v", params.Signer)
 
+	store := adt.AsStore(rt)
 	var st State
 	rt.StateTransaction(&st, func() {
 		isSigner := isSigner(resolvedOldSigner, st.Signers)
@@ -370,6 +371,10 @@ func (a Actor) RemoveSigner(rt runtime.Runtime, params *RemoveSignerParams) *abi
 			}
 			st.NumApprovalsThreshold = st.NumApprovalsThreshold - 1
 		}
+
+		err := st.PurgeApprovals(store, resolvedOldSigner)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to purge approvals of removed signer")
+
 		st.Signers = newSigners
 	})
 
@@ -392,6 +397,7 @@ func (a Actor) SwapSigner(rt runtime.Runtime, params *SwapSignerParams) *abi.Emp
 	toResolved, err := builtin.ResolveToIDAddr(rt, params.To)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to resolve to address %v", params.To)
 
+	store := adt.AsStore(rt)
 	var st State
 	rt.StateTransaction(&st, func() {
 		fromIsSigner := isSigner(fromResolved, st.Signers)
@@ -412,6 +418,9 @@ func (a Actor) SwapSigner(rt runtime.Runtime, params *SwapSignerParams) *abi.Emp
 		}
 		newSigners = append(newSigners, toResolved)
 		st.Signers = newSigners
+
+		err := st.PurgeApprovals(store, fromResolved)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to purge approvals of removed signer")
 	})
 
 	return nil
--- a/actors/builtin/multisig/multisig_state.go
+++ b/actors/builtin/multisig/multisig_state.go
@@ -58,6 +58,62 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 	return locked
 }
 
+// Iterates all pending transactions and removes an address from each list of approvals, if present.
+// If an approval list becomes empty, the pending transaction is deleted.
+func (st *State) PurgeApprovals(store adt.Store, addr address.Address) error {
+	txns, err := adt.AsMap(store, st.PendingTxns)
+	if err != nil {
+		return xerrors.Errorf("failed to load transactions: %w", err)
+	}
+
+	// Identify the transactions that need updating.
+	var txnIdsToPurge []string // For stable iteration
+	txnsToPurge := map[string]Transaction{} // Values are not pointers, we need copies
+	var txn Transaction
+	if err = txns.ForEach(&txn, func(txid string) error {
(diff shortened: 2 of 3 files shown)


actors/builtin/multisig/multisig_state.go
```diff
@@ -58,6 +58,63 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
...
+		} else {
+			if err := txns.Delete(StringKey(txid)); err != nil {
+				return err
+			}
+		}
+	}
+
+	if newTxns, err := txns.Root(); err != nil {
+		return err
+	} else {
```
aarshkshah1992, line 112: can skip this else.

actors/builtin/multisig/multisig_state.go
```diff
@@ -58,6 +58,63 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
 	return locked
 }
 
+// Iterates all pending transactions and removes an address from each list of approvals, if present.
+// If an approval list becomes empty, the pending transaction is deleted.
+func (st *State) PurgeApprovals(store adt.Store, addr address.Address) error {
```
aarshkshah1992, line 63: We should wrap all the myriad of different errors down here to add information about what went wrong.

actors/builtin/multisig/multisig_state.go
```diff
@@ -58,6 +58,63 @@ func (st *State) AmountLocked(elapsedEpoch abi.ChainEpoch) abi.TokenAmount {
...
+
+	// Identify the transactions that need updating.
+	var txnIdsToPurge []string // For stable iteration
+	txnsToPurge := map[string]*Transaction{}
+	var txn Transaction
+	if err = txns.ForEach(&txn, func(txid string) error {
+		for _, approver := range txn.Approved {
+			if approver == addr {
+				txnIdsToPurge = append(txnIdsToPurge, txid)
+				cpy := txn
```
aarshkshah1992, line 77: If we are anyways copying this here, might as well do:

```
txnsToPurge := map[string]Transaction{}
txnsToPurge[txid] = txn
```

aarshkshah1992 (approved):
LGTM modulo some minor comments.

actors/builtin/multisig/multisig_actor.go
```diff
@@ -505,7 +514,7 @@ func getTransaction(rt runtime.Runtime, ptx *adt.Map, txnID TxnID, proposalHash
 	if checkHash {
 		calculatedHash, err := ComputeProposalHash(&txn, rt.HashBlake2b)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to compute proposal hash for %v", txnID)
-		if proposalHash != nil && !bytes.Equal(proposalHash, calculatedHash[:]) {
+		if len(proposalHash) > 0 && !bytes.Equal(proposalHash, calculatedHash[:]) {
```
ZenGround0, line 517: I think this is making the method robust against ambiguous decoding of proposal hash into either `nil` or `[]byte{}` 👍 .  Out of curiosity did you actually run into this case from serialized params?  I would have expected cbor gen to have 1. the same encoding of both values 2. a consistent decoding of the serialization of both possible values.  In both cases picking one

actors/builtin/multisig/multisig_test.go
```diff
@@ -1443,6 +1357,65 @@ func TestRemoveSigner(t *testing.T) {
...
+			Value:    big.Zero(),
+			Method:   builtin.MethodSend,
+			Params:   nil,
+			Approved: []addr.Address{bob},
+		})
+	})
+
+	t.Run("remove signer deletes solo proposals", func(t *testing.T) {
+		rt := builder.Build(t)
+		rt.AddIDAddress(anneNonID, anne)
```
ZenGround0, line 1403: nit: this and the above appear unnecessary because you call `removeSigner` on `anne` not `anneNonID`

ZenGround0 (approved):
Looks good

actors/builtin/multisig/multisig_actor.go
anorth: Yes this is true. I did not run into it directly, but traced through here while testing. The cbor-gen does look to produce a `nil`. Unit tests will have to consistently use nil too.

I'll revert for now.

actors/builtin/multisig/multisig_state.go
anorth: It's necessary for scope of `newTxns`.

actors/builtin/multisig/multisig_test.go
anorth: Thanks, I changed this one to use `anneNonID`.

PR #1210, opened 2020-10-02
Skip miner for miners who's proving periods haven't started

Stebalien:
The migration logic assumes that proving periods are never set in the future, and miners that haven't started their first proving period will not have experienced any faults that need correcting.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -1018,6 +1018,25 @@ func (st *State) AdvanceDeadline(store adt.Store, currEpoch abi.ChainEpoch) (*Ad
 	// we still use dlInfo.Last().
 	dlInfo := st.DeadlineInfo(currEpoch)
 
+	// Return early if the proving period hasn't started. While actors v2
+	// sets the proving period start into the past so this case can never
+	// happen, v1:
+	//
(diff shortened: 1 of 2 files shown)


Stebalien:
Depends on https://github.com/filecoin-project/specs-actors/pull/1209

Stebalien:
So, the question is, is this OK? I think it is, but I'm not sure if there's an edge-case I'm missing.

ZenGround0:
I'm not sure if its impossible to precommit, provecommit, cc upgrade, and fault before the proving period starts but likely it is.  It also should be impossible to expire precommits before the proving period starts.  I'll take a look tomorrow.

Even if it is technically possible this should probably be good enough since we are not expecting cc upgrades right now on testnet.  If there is a theoretical loophole we could also ask miners to hold off on cc upgrades a day or so before the launch.

Stebalien:
> fault before the proving period starts but likely it is.

You would have had to have submitted an explicit fault, which lotus won't do until you start proving, I think. But it may technically be possible.

Note: this isn't the only fix. We just need to make sure we don't (a) change the deadline/proving period or (b) reschedule cron in this case.

ZenGround0 (approved):
Should be fine.  If I have time I want to double check this for safety

PR #1211, opened 2020-10-02
use larger CI machines

Stebalien:
We're running out of ram.


ZenGround0 (approved):

Stebalien:
#1212 seems to fix this. We can re-open this PR if that's not the case.

anorth:
We're still on the edge, let's upgrade.

PR #1212, opened 2020-10-02
Remove invalid market test

anorth:
The market tests included a case that called functions with descending current epoch, which is not a case we generally handle. It expected an assertion failure: we'll attempt to remove the assertion in #1161.

- Removed test case
- Removed parallelism from the cases in `TestCronTickDealSlashing`, which may have been contributing to CI memory pressure (#1211)
- Tweaked error messages in state checks

It's a bit worrying that this case passed at all, rather than merely being flaky.

### DIFF
--- a/actors/builtin/market/market_test.go
+++ b/actors/builtin/market/market_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"strings"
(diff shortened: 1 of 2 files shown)


Stebalien (approved):
Fix looks correct, but I'm not familiar enough with this code to understand what was going on.

PR #1213, opened 2020-10-02
Replace some miner actor assertions with aborts.

anorth:
This is just part of an effort to remove ~all assertions (#1161), but a useful increment.

I think this subsumes #1195.

### DIFF
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -9,7 +9,6 @@ import (
 	mh "github.com/multiformats/go-multihash"
 
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
-	. "github.com/filecoin-project/specs-actors/v2/actors/util"
 )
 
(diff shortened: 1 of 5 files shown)


actors/builtin/miner/policy.go
```diff
@@ -164,7 +163,6 @@ func QualityForWeight(size abi.SectorSize, duration abi.ChainEpoch, dealWeight,
 	sectorSpaceTime := big.Mul(big.NewIntUnsigned(uint64(size)), big.NewInt(int64(duration)))
 	// totalDealSpaceTime = dealWeight + verifiedWeight
 	totalDealSpaceTime := big.Add(dealWeight, verifiedWeight)
-	Assert(sectorSpaceTime.GreaterThanEqual(totalDealSpaceTime))
```
ZenGround0, line 167: If this is an invalid state then perhaps we should throw an error, or validate the inputs before this function is called to be sure the invariant holds.

ZenGround0 (approved):

ZenGround0 (approved):

PR #1215, opened 2020-10-04
Baseline migration

ZenGround0:
The PR is motivated by the cryptoecon effort to fix some outstanding reward logic issues introduced in the 94000 upgrade.


davidad (commented):
This looks perfect so far!

ZenGround0:
@anorth these things are decoupled from the raw math and worth your review (especially the first bullet)
- Moving SimpleTotal and BaselineTotal on chain
- Deferring reward actor migration to order after power actor
- Refactoring math: moving functions into math package, renaming Precision to Precision128

It will help to read this commit by commit

actors/util/math/exp.go
```diff
@@ -0,0 +1,29 @@
...
+		return one
+	}
+	if n == 1 {
+		return base
+	}
+
+	// Recurse
+	if n < 0 {
+		inverseBase := big.Div(big.Lsh(one, Precision128), base) // Q.256 / Q.128 => Q.128
+		return ExpBySquaring(inverseBase, n)
```
anorth, line 20: This should be `-n` to prevent infinite regress, since `n` is already negative.

It doesn't look like we ever exercise this, so adding an error return and refusing negative exponents would also be ok.

anorth (approved):
Looks good as far as I can judge, after one fix.

actors/migration/reward.go
```diff
@@ -3,35 +3,153 @@ package migration
...
+	if err := store.Get(ctx, outPowerAct.Head, &outPowerSt); err != nil {
+		return cid.Undef, big.Zero(), xerrors.Errorf("failed to read migrated power actor state %w", err)
+	}
+	thisEpochRawBytePower := outPowerSt.ThisEpochRawBytePower
+
+	one := big.Lsh(big.NewInt(1), math2.Precision128) // Q.128
+	// Set baseline to the value it would be at the end of the prior epoch if
+	// current parameters had been in place at network start
+	outThisEpochBaselinePower := math2.ExpBySquaring(reward2.BaselineExponent, int64(migInfo.priorEpoch)) // Q.128
+	outThisEpochBaselinePower = big.Mul(reward2.BaselineInitialValue, outThisEpochBaselinePower)          // Q.0 * Q.128 => Q.128
```
davidad, line 67: Is it safe to assume that the `reward2.Baseline...` constants are the new ones (e.g. BaselineInitialValue is 2 888 888 880 000 000 000)?

actors/migration/reward.go
```diff
@@ -3,35 +3,153 @@ package migration
...
+	if !effNetworkTime.Int.IsInt64() {
+		return cid.Undef, big.Zero(), xerrors.Errorf("new effective network time chain epoch out of bounds %v", effNetworkTime)
+	}
+	outEffectiveNetworkTime := abi.ChainEpoch(effNetworkTime.Int.Int64())
+
+	// Set effective baseline power given the new effective network time and
+	// baseline function.
+	outEffectiveBaselinePower := math2.ExpBySquaring(reward2.BaselineExponent, int64(outEffectiveNetworkTime)) // Q.128
+	outEffectiveBaselinePower128 := big.Mul(reward2.BaselineInitialValue, outEffectiveBaselinePower)           // Q.0 * Q.128 => Q.128
+	outEffectiveBaselinePower = big.Rsh(outEffectiveBaselinePower128, math2.Precision128)                      // Q.128 => Q.0
```
davidad, line 90: ✅

actors/migration/reward.go
```diff
@@ -3,35 +3,153 @@ package migration
...
+
+	// Set cumsum baseline given new effective network time and baseline function
+	outCumSumBaseline := big.Sub(outEffectiveBaselinePower128, big.Lsh(reward2.BaselineInitialValue, math2.Precision128)) // Q.128 - Q.128 => Q.128
+	outCumSumBaseline = big.Mul(outCumSumBaseline, oneOverGTau)                                                           // Q.128 * Q.128 => Q.256
+	outCumSumBaseline = big.Rsh(outCumSumBaseline, 2*math2.Precision128)                                                  // Q.256 => Q.0
```
davidad, line 95: ✅

actors/migration/reward.go
```diff
@@ -3,35 +3,153 @@ package migration
...
+	// had been in place at network start
+	acc1 := big.Mul(reward2.BaselineInitialValue, oneOverGTau) // Q.0 * Q.128 => Q.128
+	acc1 = big.Add(acc1, big.Lsh(inState.CumsumRealized, math2.Precision128))
+	rawBytePower256 := big.Lsh(thisEpochRawBytePower, 2*math2.Precision128) // Q.0 => Q.256
+	acc1 = big.Div(rawBytePower256, acc1)                                   // Q.256 / Q.128 => Q.128
+	acc1 = big.Add(acc1, one)
+	acc1 = math2.Ln(acc1)
+	acc1 = big.Lsh(acc1, math2.Precision128)                         // Q.128 => Q.256
+	acc1 = big.Div(acc1, big.Lsh(big.NewInt(6), math2.Precision128)) // Q.256 / Q.128 => Q.128
+	acc1 = big.NewFromGo(math2.ExpNeg(acc1.Int))
```
davidad, line 110: I'm not sure what `NewFromGo` is doing here - is it just converting between two copies of the bigint library?

actors/migration/reward.go
```diff
@@ -3,35 +3,153 @@ package migration
...
+	acc3 = big.Mul(acc3, reward2.ExpLamSubOne) // Q.128 * Q.128 => Q.256
+	acc3 = big.Rsh(acc3, math2.Precision128)   // Q.256 => Q.128
+	outRewardSmoothedPosition := big.Add(acc1, acc3)
+
+	// Filter velocity precision is less important. Hard coded value of 31 x 10^51 is sufficient
+	outRewardSmoothedVelocity := big.Mul(
+		big.NewInt(31),
+		big.Exp(big.NewInt(10), big.NewInt(51)),
+	) // Q.128
+	outThisEpochRewardSmoothed := smoothing2.NewEstimate(outRewardSmoothedPosition, outRewardSmoothedVelocity)
```
davidad, line 130: ✅

actors/util/math/ln.go
```diff
@@ -0,0 +1,79 @@
...
+// Output is in Q.128 format.
+func lnBetweenOneAndTwo(x big.Int) big.Int {
+	// ln is approximated by rational function
+	// polynomials of the rational function are evaluated using Horner's method
+	num := Polyval(lnNumCoef, x.Int)     // Q.128
+	denom := Polyval(lnDenomCoef, x.Int) // Q.128
+
+	num = num.Lsh(num, Precision128)          // Q.128 => Q.256
+	return big.NewFromGo(num.Div(num, denom)) // Q.256 / Q.128 => Q.128
+}
```
davidad, line 79: I am assuming this is copied from an existing implementation, so I am not reviewing it today.

davidad (changes requested):
I have reviewed all the math. Everything looks great, except for one missing digit and a missed data dependency I had failed to recognize in the spec.

actors/migration/reward.go
ZenGround0: Yes they are the new ones and that is the value of BaselineInitialValue.  For completeness
`reward2.BaselineExponent = big.MustFromString("340282591298641078465964189926313473653") // Q.128`
and `reward2.Lambda` and `reward2.ExpLamSubOne` have not changed from `reward0` values

actors/migration/reward.go
ZenGround0: Yes its from an inconsistency between using raw go big.Int's and the actors wrapped big.Int type in expneg and polyval.  I came close to cleaning this up but it fell just below the threshold of feeling worth it.

actors/util/math/ln.go
ZenGround0: Yeah you reviewed this code already [here](https://github.com/filecoin-project/specs-actors/pull/762/files#diff-b03a88e903f66a868c96a8495135ea5aR178)

davidad (approved):
All issues have been fixed, looks great to me now. Thank you @ZenGround0!

Issue #1216, opened 2020-10-05
correctForCCUpgradeThenFaultIssue can set negative deadlines


arajasek:
Strictly looking at the code [here](https://github.com/filecoin-project/specs-actors/blob/bdd876b3bbe9f0558b7ef97190750103adbdd463/actors/migration/miner_corrections.go#L80-L84), it certainly seems like we can try to run `expectedDeadlline := uint64((epoch - st.ProvingPeriodStart) / miner.WPoStChallengeWindow)` with `epoch <  st.ProvingPeriodStart`, which would lead to bad things.

I don't know if some higher-level structure would prevent this from happening in the live network, but I ran into this in a devnet.

ZenGround0:
@arajasek @Stebalien's recent change introduced [this](https://github.com/filecoin-project/specs-actors/blob/bdd876b3bbe9f0558b7ef97190750103adbdd463/actors/migration/miner_corrections.go#L35) check which should always prevent this from running when `epoch < st.ProvingPeriodStart`

arajasek:
Nice, thanks! I was running a couple commits behind `master`. Probably safe to close this.

PR #1217, opened 2020-10-05
Fix negative deadline in the correction migration

arajasek:
_May_ be a correct fix for  #1216, I make no promises.


arajasek:
Nvm, this can't be right

PR #1218, opened 2020-10-06
add miner info invariant checks

acruikshank:
further work towards #1165

### Motivation

We want to run invariant checks on miner state after tests that we couldn't run on chain for performance reasons. This PR adds tests invariants hold for miner info.

### Proposed Changes

1 Add `CheckMinerInfo`.
2. Call it from `CheckStateInvariants` so it will be invoked on almost all the miner tests.

### DIFF
--- a/actors/builtin/miner/testing.go
+++ b/actors/builtin/miner/testing.go
@@ -591,6 +591,8 @@ func CheckMinerInfo(info *MinerInfo) (*builtin.MessageAccumulator, error) {
 	}
 
 	if info.PendingOwnerAddress != nil {
+		acc.Require(info.PendingOwnerAddress.Protocol() == addr.ID,
+			"pending owner address %v is not an ID address", info.PendingOwnerAddress)
 		acc.Require(*info.PendingOwnerAddress != info.Owner,
 			"pending owner address %v is same as existing owner %v", info.PendingOwnerAddress, info.Owner)
 	}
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/testing.go
```diff
@@ -570,6 +574,45 @@ func CheckEarlyTerminationQueue(earlyQ BitfieldQueue, terminated bitfield.BitFie
...
+		acc.Require(*info.PendingOwnerAddress != info.Owner,
+			"pending owner address %v is same as existing owner %v", info.PendingOwnerAddress, info.Owner)
+	}
+
+	sealProofInfo, found := abi.SealProofInfos[info.SealProofType]
+	acc.Require(found, "miner has unrecognized seal proof type %d", info.SealProofType)
+	if found {
+		acc.Require(sealProofInfo.SectorSize == info.SectorSize,
+			"sector size %d is wrong for seal proof type %d: %d", info.SectorSize, info.SealProofType, sealProofInfo.SectorSize)
+		acc.Require(sealProofInfo.SectorSize == info.SectorSize,
```
ZenGround0, line 603: Looks like a duplicate

actors/builtin/miner/testing.go
```diff
@@ -570,6 +574,45 @@ func CheckEarlyTerminationQueue(earlyQ BitfieldQueue, terminated bitfield.BitFie
...
+			"sector size %d is wrong for seal proof type %d: %d", info.SectorSize, info.SealProofType, sealProofInfo.SectorSize)
+	}
+	sealProofPolicy, found := builtin.SealProofPolicies[info.SealProofType]
```
ZenGround0, line 606: Should we `acc.Require(found)`?

actors/builtin/miner/testing.go
```diff
@@ -570,6 +574,45 @@ func CheckEarlyTerminationQueue(earlyQ BitfieldQueue, terminated bitfield.BitFie
...
+
+	if info.PendingWorkerKey != nil {
+		acc.Require(info.PendingWorkerKey.NewWorker.Protocol() == addr.ID,
+			"pending worker address %v is not an ID address", info.PendingWorkerKey.NewWorker)
+		acc.Require(info.PendingWorkerKey.NewWorker != info.Worker,
+			"pending worker key %v is same as existing worker %v", info.PendingWorkerKey.NewWorker, info.Worker)
+	}
+
+	if info.PendingOwnerAddress != nil {
+		acc.Require(*info.PendingOwnerAddress != info.Owner,
```
anorth, line 594: Check it's an ID address too

anorth (approved):

PR #1219, opened 2020-10-06
Multiply miner penalty by 3x

ZenGround0:
As part of red team / cryptoecon requirements for upcoming upgrade the miner penalty as computed in the node and passed on to the reward actor is scaled up by a factor of three before being passed on to the miner actor for deduction.


arajasek:
Sanity-checking the obvious: This won't need a version check since it's debuted in actors v2, yeah?

arajasek (approved):

anorth (approved):

Issue #1166, opened 2020-09-17
Implement cross-actor consistency checks


anorth:
Following #1165, implement consistency checks between actor states, such as consistency between the power, market, and miner actors' sector, power and deal state.

ZenGround0:
@acruikshank do you have ideas for what remains after your miner-market and miner-power cross check PRs?

acruikshank:
I'm pretty sure everything else gets pretty deep into cryptoeconomics. Like we might be able to put some bounds on a miner's initial pledge based on current reward actor state, or bound reward actors state values based on power actor stats, but I can't explain how to do that. I don't think there are any more simple relationships like the ones between miners and power or miners and market.

ZenGround0:
Closing this, we can open a new issue to advance to the next level of paranoia

PR #1221, opened 2020-10-06
Finish miner state checks

acruikshank:
continued work on #1166

### Motivation

Many invariants of miner state remain untested by our slow checks. This PR adds several more, and should complete coverage.

### Proposed Changes

1. Check miner balances (balance, locked funds, precommit deposit, initial pledge, fee debt) for consistency.
2. Assert sum of vesting table is locked funds and ensure locked fund entries are quantized.
3. Check precommit data structure for consistency.
4. Check that sector numbers of all on chain sectors have been allocated.
5. Check that deadline index is less than 48.
6. Modify test calls to pass in actor balance.

### DIFF
--- a/actors/builtin/miner/testing.go
+++ b/actors/builtin/miner/testing.go
@@ -43,6 +43,11 @@ func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount) (
 		return nil, acc, err
 	}
 
+	var allocatedSectors bitfield.BitField
+	if err := store.Get(store.Context(), st.AllocatedSectors, &allocatedSectors); err != nil {
+		return nil, acc, err
+	}
+
 	allSectors := map[abi.SectorNumber]*SectorOnChainInfo{}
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/testing.go
```diff
@@ -56,6 +79,9 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 	faultyPower := NewPowerPairZero()
 
 	// Check deadlines
+	acc.Require(st.CurrentDeadline < WPoStPeriodDeadlines,
+		"current deadline index is greater than deadlines per period(%d): %d", WPoStPeriodDeadlines, st.CurrentDeadline)
```
anorth, line 83: As follow-up, I think we should also accept epoch as a parameter and make dependent state checks like:
- proving period start
- expiration queues have been processed
- no faults are too old
etc

anorth (approved):

PR #1222, opened 2020-10-07
Direct construct filter to prevent double left shift

ZenGround0:
Fixing a bug.  NewEstimate does a left shift on constructor inputs so we should just directly set fields.  Thanks @arajasek for reporting, @davidad for finding the fix.


davidad (approved):
LGTM, thanks!

Issue #1223, opened 2020-10-07
Multisig does not validate positive InitialBalance


acruikshank:
The multisig `LockBalance` method does not validate that the supplied balance is non-negative and will store a negative `InitialBalance` if one is supplied. 

The consequences of this are small. It means that all transactions can be executed immediately so long as the actor has sufficient balance. It also means the multisig will be unable to call `LockBalance` until the original lock duration expires. The exact same effect could be achieved by locking zero balance, so locking a negative balance is almost certainly a mistake on the multisig owners part.

PR #1224, opened 2020-10-07
multisig invariant tests

acruikshank:
work towards #1165

### Motivation

We want to have tests for all actor state invariants, and we have yet to do this for Multisig. This PR adds a check that multisig state is internally consistent and runs that check in all unit tests.

The invariants tested are:

* `len(st.Signers) <= SignersMax`
* `len(st.Signers) > 0`
* `NextTxnID > max(keys(PendingTxns))`
* `txn.Approved ⊆ st.Signers` for all txns.

Some almost-invariants that cannot be asserted:

* `st.InitialBalance >= 0`. see #1223
* `txn.To is ID`. The multisig actor does not enforce this. There might be good reason for this since a fork between the proposal and execution causing the txn to be sent to the wrong account.
* `txn.Value is non-negative`. The multisig does not enforce this. It will cause the send to be invalid though.
* `len(txn.Approved) < st.NumApprovalsThreshold`. Usually the transaction approvals hitting the threshold will cause it to be executed and removed, but reducing the threshold can put the actor in a state where txns are approved but not yet sent. This is noted in #71 and addressed in #524. This PR adds a test for #524.

### Proposed Changes

1. Add `multisig.CheckStateInvariants` that deep checks multisig state.
2. Add a call to the above in every multisig unit test that is expected to call a multisig message successfully.
3. Add a test for re-approving after lowering a threshold to eliminate a resolved TODO.

### DIFF
--- a/actors/builtin/multisig/testing.go
+++ b/actors/builtin/multisig/testing.go
@@ -8,31 +8,6 @@ import (
 	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
 )
 
-/*
-len(Signers) > 0
-NumApprovals >= len(Signers)
-NextTxnID > max(key(PendingTxns))
-
-//
-InitialBalance >= 0
-UnlockDuration >= 0
-
-type Transaction struct {
-	To     addr.Address
-	Value  abi.TokenAmount
-	Method abi.MethodNum
-	Params []byte
-
-	// This address at index 0 is the transaction proposer, order of this slice must be preserved.
-	Approved []addr.Address
-}
-
-To is ID address? NOPE
-Value is non-negative? NOPE
-len(approved) <= numapprovalsthreshold ??? (do we send transactions when we lower threshold?) NOPE
-approved subset of signers (because we purge)
-*/
-
 type StateSummary struct {
 	PendingTxns  uint64
 	NumApprovals uint64
--- a/actors/builtin/multisig/testing.go
+++ b/actors/builtin/multisig/testing.go
@@ -9,9 +9,9 @@ import (
 )
 
 type StateSummary struct {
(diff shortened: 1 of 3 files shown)


actors/builtin/multisig/testing.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+)
+
+type StateSummary struct {
+	PendingTxns  uint64
+	NumApprovals uint64
+	Signers      int
+}
```
anorth, line 15: Naming nit: name numbers/counts appropriately, leaving namespace for including the actual data later without an awkward name.

```suggestion
type StateSummary struct {
	PendingTxnCount  uint64
	NumApprovalsThreshold uint64
	SignerCount      int
}
```

actors/builtin/multisig/testing.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			return err
+		}
+		if txnID > maxTxnID {
+			maxTxnID = txnID
+		}
+
+		for _, approval := range txn.Approved {
+			_, found := signers[approval]
+			acc.Require(found, "approval %v for transaction %d is not in signers list", approval, txnID)
```
anorth, line 52: Note: the multisig enforces this now, but we are not migrating it out of the state of any previously existing multisigs. Great if this holds, but it's possible it won't in the testnet chain state.

Leave the check here tho.

anorth (approved):

ZenGround0 (approved):

PR #1225, opened 2020-10-07
Hook up invariants to main call

ZenGround0:
Adds cron, market and account invariant checks to the full state invariant check function.

@acruikshank I am planning to use the same `priorEpoch` value as an argument to the main state invariant check function.  This will be the epoch of the last set of state transitions that have run on the input state, the last such one being the cron from `priorEpoch`.  I just passed this in directly to the market invariant tests.  Please let me know if I got that wrong and there should be a +1.

### DIFF
--- a/actors/states/check.go
+++ b/actors/states/check.go
@@ -7,16 +7,22 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/account"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/cron"
 	init_ "github.com/filecoin-project/specs-actors/v2/actors/builtin/init"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/market"
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin/verifreg"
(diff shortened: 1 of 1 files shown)


anorth (approved):
LGTM modulo epoch confirmation

actors/states/check.go
```diff
@@ -90,6 +125,8 @@ func CheckStateInvariants(tree *Tree, expectedBalanceTotal abi.TokenAmount) (*bu
 	//
 	_ = initSummary
 	_ = verifregSummary
+	_ = cronSummary
+	_ = marketSummary
```
acruikshank, line 129: Hopefully I'll get into the cross actor checks soon.

actors/states/check.go
```diff
@@ -58,6 +83,16 @@ func CheckStateInvariants(tree *Tree, expectedBalanceTotal abi.TokenAmount) (*bu
...
+			var st market.State
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				return err
+			}
+			if summary, msgs, err := market.CheckStateInvariants(&st, tree.Store, actor.Balance, priorEpoch); err != nil {
+				return err
+			} else {
+				acc.WithPrefix("market: ").AddAll(msgs)
+				marketSummary = summary
+			}
```
acruikshank, line 95: PaymentChannel and Multisig are merged now, Power and Reward coming up.

acruikshank (approved):
Excited to see what happens when this is run on testnet.

PR #1229, opened 2020-10-08
power actor invariant tests

acruikshank:
work towards #1165 and #1166 

### Motivation

We would like to be able to check the internal consistency of the storage power actor's state. This PR adds `CheckStateInvariants` for power state and calls it for all unit tests that are not expected to fail. 

### Proposed Changes

1. Add `CheckStateInvariants`
2. Add test harness `checkState` and call it in all relevant unit tests.
3. Add a call to `CheckStateInvariants` to `states.CheckStateInvariants`.
4. Add `CheckMinersAgainstPower` to check cross actor state
5. Run `CheckStateInvariants` in scenario tests including correct_cc_upgrade migration test.
6. Fix issue with power migration where a miner's corrected claim would not update power state in a consistent way.
7. Change account invariant check to only require a public key if actor is non-singleton. (when I rebased in the inclusion of account to the main state check, it wasn't working in scenario tests without this fix).

### DIFF
--- a/actors/builtin/power/testing.go
+++ b/actors/builtin/power/testing.go
@@ -60,7 +60,6 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 	acc.Require(st.TotalQualityAdjPower.GreaterThanEqual(big.Zero()), "total qa power is negative %v", st.TotalQualityAdjPower)
 	acc.Require(st.TotalBytesCommitted.GreaterThanEqual(big.Zero()), "total raw power committed is negative %v", st.TotalBytesCommitted)
 	acc.Require(st.TotalQABytesCommitted.GreaterThanEqual(big.Zero()), "total qa power committed is negative %v", st.TotalQABytesCommitted)
-	acc.Require(st.TotalRawBytePower.GreaterThanEqual(big.Zero()), "total raw power is negative %v", st.TotalRawBytePower)
 	acc.Require(st.TotalRawBytePower.LessThanEqual(st.TotalQualityAdjPower),
 		"total raw power %v is greater than total quality adjusted power %v", st.TotalRawBytePower, st.TotalQualityAdjPower)
 	acc.Require(st.TotalBytesCommitted.LessThanEqual(st.TotalQABytesCommitted),
@@ -70,13 +69,15 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 	acc.Require(st.TotalQualityAdjPower.LessThanEqual(st.TotalQABytesCommitted),
 		"total qua power %v is greater than qa power committed %v", st.TotalQualityAdjPower, st.TotalQABytesCommitted)
 
-	crons, msgs, err := CheckCronInvariants(st, store)
-	acc.AddAll(msgs)
+	crons, err := CheckCronInvariants(st, store, acc)
 	if err != nil {
 		return nil, acc, err
 	}
 
-	claims, msgs, err := CheckClaimInvariants(st, store)
+	claims, err := CheckClaimInvariants(st, store, acc)
+	if err != nil {
+		return nil, acc, err
+	}
 
 	return &StateSummary{
 		Crons:  crons,
@@ -84,12 +85,10 @@ func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.M
 	}, acc, nil
 }
 
-func CheckCronInvariants(st *State, store adt.Store) (CronEventsByAddress, *builtin.MessageAccumulator, error) {
-	acc := &builtin.MessageAccumulator{}
-
+func CheckCronInvariants(st *State, store adt.Store, acc *builtin.MessageAccumulator) (CronEventsByAddress, error) {
 	queue, err := adt.AsMultimap(store, st.CronEventQueue)
 	if err != nil {
-		return nil, acc, err
+		return nil, err
 	}
 
 	byAddress := make(CronEventsByAddress)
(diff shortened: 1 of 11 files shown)


actors/builtin/power/testing.go
```diff
@@ -0,0 +1,201 @@
...
+				for _, p := range existingPayloads {
+					acc.Require(p.EventType != miner.CronEventProvingDeadline, "found duplicate miner proving cron for miner %v", event.MinerAddr)
+				}
+			}
+			byAddress[event.MinerAddr] = append(existingPayloads, payload)
+
+			return nil
+		})
+	})
+	acc.Require(err != nil, "error attempting to read through power actor cron tasks: %v", err)
```
anorth, line 127: Ok, I do kinda like this. We could propagate this pattern more widely.

actors/builtin/power/testing.go
```diff
@@ -0,0 +1,201 @@
...
+	claimsWithSufficientPowerCount := int64(0)
+	byAddress := make(ClaimsByAddress)
+	var claim Claim
+	err = claims.ForEach(&claim, func(key string) error {
+		addr, err := address.NewFromBytes([]byte(key))
+		if err != nil {
+			return err
+		}
+		byAddress[addr] = claim
+		claimCount += 1
```
anorth, line 154: You can just use `len(byAddress)`

actors/builtin/power/testing.go
```diff
@@ -0,0 +1,201 @@
...
+// Checks internal invariants of power state.
+func CheckStateInvariants(st *State, store adt.Store) (*StateSummary, *builtin.MessageAccumulator, error) {
+	acc := &builtin.MessageAccumulator{}
+
+	// basic invariants around recorded power
+	acc.Require(st.TotalRawBytePower.GreaterThanEqual(big.Zero()), "total raw power is negative %v", st.TotalRawBytePower)
+	acc.Require(st.TotalQualityAdjPower.GreaterThanEqual(big.Zero()), "total qa power is negative %v", st.TotalQualityAdjPower)
+	acc.Require(st.TotalBytesCommitted.GreaterThanEqual(big.Zero()), "total raw power committed is negative %v", st.TotalBytesCommitted)
+	acc.Require(st.TotalQABytesCommitted.GreaterThanEqual(big.Zero()), "total qa power committed is negative %v", st.TotalQABytesCommitted)
+	acc.Require(st.TotalRawBytePower.GreaterThanEqual(big.Zero()), "total raw power is negative %v", st.TotalRawBytePower)
```
anorth, line 63: This duplicates line 59

actors/builtin/power/testing.go
```diff
@@ -0,0 +1,201 @@
...
+	acc.Require(st.TotalQualityAdjPower.LessThanEqual(st.TotalQABytesCommitted),
+		"total qua power %v is greater than qa power committed %v", st.TotalQualityAdjPower, st.TotalQABytesCommitted)
+
+	crons, msgs, err := CheckCronInvariants(st, store)
+	acc.AddAll(msgs)
+	if err != nil {
+		return nil, acc, err
+	}
+
+	claims, msgs, err := CheckClaimInvariants(st, store)
```
anorth, line 79: Blocking: `msgs` and `err` are both ignored here.

It looks like I've been inconsistent to date, but I think that inner methods within each actor's CheckStateInvariants should receive the accumulator as a parameter, rather than return a new one, to avoid this easy-to-make mistake.

anorth (commented):
Looking good

actors/migration/power.go
```diff
@@ -101,19 +101,33 @@ func (m *powerMigrator) migrateCronEvents(ctx context.Context, store cbor.IpldSt
...
+			return nil
+		}
+
+		if found {
+			rawPower = big.Sub(rawPower, oldClaim.RawBytePower)
+			qaPower = big.Sub(qaPower, oldClaim.QualityAdjPower)
+		}
+
+		if err := st.AddToClaim(adt0.WrapStore(ctx, store), addr, rawPower, qaPower); err != nil {
+			return err
```
anorth, line 126: See my comment on #1237, but in any case don't include this fix in this otherwise validation-only PR.

actors/builtin/power/testing.go
```diff
@@ -0,0 +1,232 @@
...
+
+// First epoch in which a cron task may be stored.
+// Cron will iterate every epoch between this and the current epoch inclusively to find tasks to execute.
+FirstCronEpoch abi.ChainEpoch
+
+// Claimed power for each miner.
+Claims cid.Cid // Map, HAMT[address]Claim
+
+ProofValidationBatch *cid.Cid // Multimap, (HAMT[Address]AMT[SealVerifyInfo])
+*/
```
anorth, line 58: Remove this before merging.

actors/builtin/power/testing.go
```diff
@@ -0,0 +1,232 @@
...
+	acc.Require(st.TotalQualityAdjPower.Equals(qaPower),
+		"recorded qa power %v does not match qa power in claims %v", st.TotalQualityAdjPower, qaPower)
+
+	return byAddress, nil
+}
+
+func CheckProofValidationInvariants(st *State, store adt.Store, claims ClaimsByAddress, acc *builtin.MessageAccumulator) (ProofsByAddress, error) {
+	if st.ProofValidationBatch == nil {
+		return nil, nil
+	}
```
anorth, line 194: It is in fact an invariant that this queue is nil at the end of every epoch, because it is emptied by cron.

actors/migration/test/correct_cc_upgrade_then_fault_scenario_test.go
```diff
@@ -394,19 +396,13 @@ func TestMigrationCorrectsCCThenFaultIssue(t *testing.T) {
...
-
-	assert.True(t, acc.IsEmpty())
-	if !acc.IsEmpty() {
-		for _, m := range acc.Messages() {
-			fmt.Println(m)
-		}
-	}
+	acc, err = states.CheckStateInvariants(stateTree, totalBalance, v2.GetEpoch())
+	require.NoError(t, err)
+	assert.True(t, acc.IsEmpty(), strings.Join(acc.Messages(), "\n"))
```
anorth, line 405: Upon reverting the migration fix above, I guess this won't hold. It's important that the state checks are useful to us as human network operators, at least as much as they are useful for tests. After the upgrade, we need to be able to run the state checks and see the un-fixed mismatch. That means this test must tolerate non-empty errors.

I think change this to assert the expected error message/s are returned.

actors/states/check.go
```diff
@@ -158,3 +173,41 @@ func CheckStateInvariants(tree *Tree, expectedBalanceTotal abi.TokenAmount, prio
...
+		var payload miner.CronEventPayload
+		hasProvingPeriodCron := false
+		for _, event := range crons {
+			err := payload.UnmarshalCBOR(bytes.NewReader(event.Payload))
+			acc.Require(err == nil, "miner %v registered cron at epoch %d with wrong or corrupt payload",
+				addr, event.Epoch)
+
+			if payload.EventType == miner.CronEventProvingDeadline {
+				acc.Require(!hasProvingPeriodCron, "miner %v has duplicate proving period cron at epoch %d",
+					addr, event.Epoch)
```
anorth, line 206: Can you print _both_ epochs?

anorth (approved):
LGTM after reverting the fix that's in #1237.

actors/builtin/power/testing.go
acruikshank: Agreed, but the tests will sometimes run this before cron, so it might catch something.

actors/migration/power.go
acruikshank: reverted

PR #1230, opened 2020-10-09
Standalone invariant checks should not take in balance

ZenGround0:
While reworking ent to validate standalone state roots it became clear that the expected balance parameter is only useful in the context of validating a state after migrating it from a previous state root.  I've separated out these two intents into two endpoints.


ZenGround0:
I was mistaken, we can always sensibly check balance because it is always a fixed value (2bn) since filecoin fund transfers are a closed system.

PR #1231, opened 2020-10-09
Speed up state checks by removing Bitfield.IsSet

anorth:
Calls to `Bitfield.IsSet` accounted for ~85% of the time to validate state. It's slow for a freshly-loaded bitfield. I replaced them all with map lookups.

Also added a poorly-documented multisig invariant. Note that this invariant was actually broken for three genesis msigs, but is expected to be restored in the Refuel migration (today).


acruikshank (approved):
looks great.

PR #1237, opened 2020-10-09
migration fix to maintain state

acruikshank:
### Motivation

It's a bad idea to update the Claims HAMT without also updating power.State variables. This PR uses `AddToClaim` to ensure all state is kept up to date.

### DIFF
--- a/actors/migration/power.go
+++ b/actors/migration/power.go
@@ -40,12 +40,12 @@ func (m *powerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, 
 		return nil, xerrors.Errorf("cron events: %w", err)
 	}
 
-	claimsRoot, err := m.updateClaims(ctx, store, inState.Claims, m.powerUpdates)
-	if err != nil {
+	// operates directly on inState
+	if err := m.updateClaims(ctx, store, inState.Claims, m.powerUpdates, &inState); err != nil {
 		return nil, xerrors.Errorf("claims: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


ZenGround0 (approved):

actors/migration/power.go
```diff
@@ -40,12 +40,12 @@ func (m *powerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore,
 		return nil, xerrors.Errorf("cron events: %w", err)
 	}
 
-	claimsRoot, err := m.updateClaims(ctx, store, inState.Claims, m.powerUpdates)
-	if err != nil {
+	// operates directly on inState
+	if err := m.updateClaims(ctx, store, inState.Claims, m.powerUpdates, &inState); err != nil {
```
anorth, line 44: It's not redundant to pass `inState.Claims`.

Nit: put the `inState` parameter in this place rather than last.

anorth (approved):

anorth:
We have determined not to attempt to land this and distribute to the network before the upgrade is actually going to happen. In this case I think we should not fix it here, and leave the code reflecting the upgrade that actually happened.

anorth:
@acruikshank now that the v2 upgrade has passed, we want this as a _new_ state migration. I don't think we're sure yet whether that will be part of a v3 release, or just a point release on v2.

anorth:
On further thought, that will probably look a bit different to this PR, as it needs to detect the broken ones. So probably best to close this PR and open a new one.

acruikshank:
This PR missed the release window. The problems it attempted to avoid will be corrected with #1265.

PR #1238, opened 2020-10-10
make migrations threadsafe

Stebalien:
Ideally, these would just be pure functions. But this, at least, forces a copy which prevents us from sharing state across invocations.

### DIFF
--- a/actors/migration/account.go
+++ b/actors/migration/account.go
@@ -14,7 +14,7 @@ import (
 type accountMigrator struct {
 }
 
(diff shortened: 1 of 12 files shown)


Stebalien:
@ZenGround0 is coming out with a better fix

anorth:
This is very subtle, and needs documentation lest it be reverted or ignored in the future. But looks ok.

ZenGround0 (approved):

PR #1239, opened 2020-10-10
Fix MigrateStateTree to be threadsafe

ZenGround0:
Move migrations global to MigrateStateTree scope

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,6 +76,7 @@ jobs:
             make test-coverage
             mkdir -p /tmp/artifacts
             mv coverage.out /tmp/artifacts/coverage.out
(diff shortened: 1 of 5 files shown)


Stebalien (changes requested):
Parallel test.

anorth (approved):

ZenGround0:
The parallel test is a minimal migration but should be sufficient when used with the race detector (now run on the migration tests by default with `make test` and on CI).  A reassuring datapoint: if I regress the migrations map change and keep the test the race detector fails on `pm.powerUpdates = powerUpdates`.

Stebalien (approved):
:heart: Thank you @ZenGround0!

PR #1240, opened 2020-10-11
Feat/reward invariants

ZenGround0:
Adding a few simple reward actor invariants

@zixuanzh @davidad let me know if any of these are not actually invariants.  Also if you have other invariants that should hold over the reward actor's state fields please let me know and I will add them.  The more obscure and paranoid the better.  Also efficiency isn't a big concern since these checks should never be running on the critical path.  One specific question @acruikshank and I have been wondering about since we started looking at power actor state: are there any useful invariants we can check between the latest value of the alpha beta filter and the latest observation?

### DIFF
--- a/actors/builtin/reward/testing.go
+++ b/actors/builtin/reward/testing.go
@@ -1,0 +1,32 @@
+package reward
+
+import (
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/math"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+)
+
+type StateSummary struct{}
+
+var FIL = big.NewInt(1e18)
+var StorageMiningAllocationCheck = big.Mul(big.NewInt(1_100_000_000), FIL)
+
+func CheckStateInvariants(st *State, store adt.Store, priorEpoch abi.ChainEpoch, balance abi.TokenAmount) (*StateSummary, *builtin.MessageAccumulator, error) {
+	acc := &builtin.MessageAccumulator{}
+
+	acc.Require(big.Add(st.TotalStoragePowerReward, balance).Equals(StorageMiningAllocationCheck), "reward given + reward left != storage mining allocation")
+
+	acc.Require(st.Epoch == priorEpoch, "reward state epoch %d does not match priorEpoch %d", st.Epoch, priorEpoch)
+	acc.Require(st.EffectiveNetworkTime <= st.Epoch, "effective network time greater than state epoch")
+
+	acc.Require(st.CumsumRealized.LessThanEqual(st.CumsumBaseline), "cumsum realized > cumsum baseline")
+	acc.Require(st.CumsumRealized.GreaterThanEqual(big.Zero()), "cumsum realized < 0")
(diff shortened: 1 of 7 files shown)


actors/builtin/reward/testing.go
```diff
@@ -0,0 +1,32 @@
...
+
+type StateSummary struct{}
+
+var FIL = big.NewInt(1e18)
+var StorageMiningAllocationCheck = big.Mul(big.NewInt(1_100_000_000), FIL)
+
+func CheckStateInvariants(st *State, store adt.Store, priorEpoch abi.ChainEpoch, balance abi.TokenAmount) (*StateSummary, *builtin.MessageAccumulator, error) {
+	acc := &builtin.MessageAccumulator{}
+
+	acc.Require(big.Add(st.TotalStoragePowerReward, balance).Equals(StorageMiningAllocationCheck), "reward given + reward left != storage mining allocation")
```
anorth, line 19: I don't think we can make an `==` check on the balance; anyone could send funds there by accident. But can check `>=`.

actors/builtin/reward/testing.go
```diff
@@ -0,0 +1,32 @@
...
+
+	acc.Require(st.Epoch == priorEpoch, "reward state epoch %d does not match priorEpoch %d", st.Epoch, priorEpoch)
+	acc.Require(st.EffectiveNetworkTime <= st.Epoch, "effective network time greater than state epoch")
+
+	acc.Require(st.CumsumRealized.LessThanEqual(st.CumsumBaseline), "cumsum realized > cumsum baseline")
+	acc.Require(st.CumsumRealized.GreaterThanEqual(big.Zero()), "cumsum realized < 0")
+	acc.Require(st.EffectiveBaselinePower.LessThanEqual(st.ThisEpochBaselinePower), "effective baseline power > baseline power")
+
+	computedBaseline := math.ExpBySquaring(BaselineExponent, int64(st.Epoch))
+	acc.Require(st.ThisEpochBaselinePower.Equals(computedBaseline), "state baseline power does not match computed")
```
anorth, line 29: I expect this doesn't hold now, but do you expect it to after the upgrade? 

It probably won't hold immediately after an empty tipset.

actors/builtin/reward/testing.go
ZenGround0: Yes we observed this not holding on testnet as it turns out people sent 65 FIL to reward actor.

actors/builtin/reward/testing.go
ZenGround0: There is some predictable error between the two methods of computing the exponential so you're right it won't ever hold with equality.  I also wouldn't expect this on v0 state, only v2.  However our error is worse than the small predictable error because I was off by one while recomputing this value so I want to try `computedBaseline := math.ExpBySquaring(BaselineExponent, int64(priorEpoch))`.  Empty epochs should not affect this since we do one cron per epoch and always run `updateToNextEpoch` which keeps `ThisEpochBaselinePower` current.

davidad:
> One specific question @acruikshank and I have been wondering about since we started looking at power actor state: are there any useful invariants we can check between the latest value of the alpha beta filter and the latest observation?

Not really. In the event of a step change in the observation, the latest values of the observation and filter could diverge arbitrarily, so there is no hard invariant that holds for the latest values. The closest thing I can think of is an invariant that the latest filter value is closer (or equal) to the _previous_ observation than the _previous_ filter value (to the same previous observation).

anorth (approved):

PR #1242, opened 2020-10-12
Make miner state checks robust to errors.

anorth:
State validation checks are typically used in tests or by manual operation on network state. If there's an error because some part of state is unreadable (e.g. corrupted bitfield or HAMT node), then we want to know this but don't want it to prevent getting any other information about unrelated state. So treating these errors in the standard Go way by propagating them immediately isn't so helpful.

This change re-arranges the miner state checks to avoid returning errors at all. Errors from low-level functions are translated into messages and, in a few cases, dependent checks are skipped. Part of #1244.

If we like it, we should propagate this pattern to all the other state checks.

While doing this, I discovered that the checks for deadline and partition expiration queue consistency were ineffective because the expiration epochs weren't propagated from partitions. This revealed that ExtendSectorExpiration has a bug, failing to update the deadline-level expiration queue (#1243). I've carved out that specific message as expected in the relevant tests.

### DIFF
--- a/actors/builtin/miner/testing.go
+++ b/actors/builtin/miner/testing.go
@@ -19,92 +19,84 @@ type StateSummary struct {
 }
 
 // Checks internal invariants of init state.
-func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount) (*StateSummary, *builtin.MessageAccumulator, error) {
+func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount) (*StateSummary, *builtin.MessageAccumulator) {
 	acc := &builtin.MessageAccumulator{}
+	sectorSize := abi.SectorSize(0)
+	minerSummary := &StateSummary{
+		LivePower:     NewPowerPairZero(),
+		ActivePower:   NewPowerPairZero(),
+		FaultyPower:   NewPowerPairZero(),
+		SealProofType: 0,
+	}
 
 	// Load data from linked structures.
-	info, err := st.GetInfo(store)
-	if err != nil {
-		return nil, nil, err
-	}
-	msgs, err := CheckMinerInfo(info)
-	acc.AddAll(msgs)
-	if err != nil {
-		return nil, acc, err
+	if info, err := st.GetInfo(store); err != nil {
+		acc.Addf("error loading miner info: %v", err)
+		// Stop here, it's too hard to make other useful checks.
+		return minerSummary, acc
+	} else {
+		minerSummary.SealProofType = info.SealProofType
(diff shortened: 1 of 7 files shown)


actors/builtin/miner/testing.go
```diff
@@ -491,20 +502,18 @@ func CheckExpirationQueue(expQ ExpirationQueue, liveSectors map[abi.SectorNumber
...
-		if err := exp.EarlySectors.ForEach(func(n uint64) error {
+		err = exp.EarlySectors.ForEach(func(n uint64) error {
 			sno := abi.SectorNumber(n)
 			// Check sectors are present only once.
 			acc.Require(!seenSectors[sno], "sector %d in expiration queue twice", sno)
 			seenSectors[sno] = true
 
 			// Check early sectors are faulty
-			acc.Require(partitionFaultsMap[n], "sector %d expiring early but not faulty", sno)
+			acc.Require(partitionFaultsMap != nil && partitionFaultsMap[n], "sector %d expiring early but not faulty", sno)
```
acruikshank, line 516: I'd prefer `partitionFaultsMap == nil || partitionFaultsMap[n]` here. If we can't load the partition faults map, this could generate lots of redundant messages.

actors/builtin/miner/testing.go
```diff
@@ -514,42 +523,50 @@ func CheckExpirationQueue(expQ ExpirationQueue, liveSectors map[abi.SectorNumber
...
-		} else if len(missing) > 0 {
-			acc.Addf("faulty sectors missing from live: %v", missing)
+			acc.Addf("error merging all on-time and early bitfields: %v", err)
+		} else {
+			if allActive, err := bitfield.SubtractBitField(all, partitionFaults); err != nil {
+				acc.Addf("error computing active sectors: %v", err)
+			} else {
+				activeSectors, missing, err = selectSectorsMap(liveSectors, allActive)
+				if err != nil {
+					acc.Addf("error selecting active sectors: %v", err)
```
acruikshank, line 543: Elsewhere you've used the pattern of setting a datastructure as empty when there is an error loading or constructing it. If you did that for `all`, `allActive`, `allFaulty`, etc. you could avoid this nesting and I believe it all works so no redundant errors are produced.

acruikshank (approved):
Looks good.

actors/builtin/miner/testing.go
anorth: I went this way because those structures are bitfields and testing `Empty()` produces yet another error to handle. Maybe using `==` with the zero value would work? I'm a bit hesitant as it's otherwise a broken value.

PR #1247, opened 2020-10-13
add miner/market cross check

acruikshank:
work towards #1166

**This PR only affects tests.**

### Motivation

We have many consistency checks for miner state and many checks for market state. We would like to add checks that miner and market state are consistent between each other. This PR adds checks that activated deals are referenced within a sector being proving by the deal provider, and that the deal state times line up with those of the sector.

### Proposed Changes

1. Add summary info for deal states in `MarketSummary`
2. Add deal info for sectors in `MinerSummary`
3. Implement `CheckDealStatesAgainstSectors` to check consistency between the above.

### DIFF
--- a/actors/builtin/market/testing.go
+++ b/actors/builtin/market/testing.go
@@ -15,8 +15,17 @@ import (
 	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
 )
 
+type DealStats struct {
+	Provider         address.Address
+	StartEpoch       abi.ChainEpoch
+	EndEpoch         abi.ChainEpoch
+	SectorStartEpoch abi.ChainEpoch
+	LastUpdatedEpoch abi.ChainEpoch
+	SlashEpoch       abi.ChainEpoch
+}
+
 type StateSummary struct {
-	ProposalIDs          []abi.DealID
+	Deals                map[abi.DealID]*DealStats
 	PendingProposalCount uint64
 	DealStateCount       uint64
 	LockTableCount       uint64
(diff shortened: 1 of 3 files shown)


actors/builtin/market/testing.go
```diff
@@ -15,8 +15,17 @@ import (
 	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
 )
 
+type DealStats struct {
```
anorth, line 18: DealSummary, just for consistency with it being part of StateSummary?

actors/states/check.go
```diff
@@ -213,3 +214,40 @@ func CheckMinersAgainstPower(acc *builtin.MessageAccumulator, minerSummaries map
...
+			deal.SectorStartEpoch, sectorDeal.SectorExpiration, deal.Provider)
+
+		acc.Require(deal.LastUpdatedEpoch <= sectorDeal.SectorExpiration,
+			"deal state update at %d after sector expiration %d for miner %v",
+			deal.LastUpdatedEpoch, sectorDeal.SectorExpiration, deal.Provider)
+
+		acc.Require(deal.SlashEpoch <= sectorDeal.SectorExpiration,
+			"deal state slashed at %d after sector expiration %d for miner %v",
+			deal.SlashEpoch, sectorDeal.SectorExpiration, deal.Provider)
+	}
```
anorth, line 252: I was going to ask to check that all sector deals are in market state, but I realise some of them might have expired and been removed. Leave a comment there explaining why we can't check that.

actors/states/check.go
```diff
@@ -213,3 +214,40 @@ func CheckMinersAgainstPower(acc *builtin.MessageAccumulator, minerSummaries map
...
+		minerSummary, found := minerSummaries[deal.Provider]
+		if !found {
+			acc.Addf("provider %v for deal %d not found among miners", deal.Provider, dealID)
+			continue
+		}
+
+		sectorDeal, found := minerSummary.Deals[dealID]
+		if !found {
+			acc.Addf("deal %d not referenced in active sectors of miner %v", dealID, deal.Provider)
+			continue
```
anorth, line 234: If a sector terminates, the sector might be gone but the deal still remain for deferred clean-up. But in that case I think we can assert the deal's SlashEpoch is set.

anorth (approved):

actors/states/check.go
acruikshank: I added a comment above the loop.

actors/states/check.go
acruikshank: good call

PR #1249, opened 2020-10-13
Vest a portion of block reward immediately (FIP-0004)

ZenGround0:
- [x] Code should be correct
- [x] Unrelated test failures caused by those tests using applyRewards as a test state setup helper are now fixed.
- [x] Reward tests updated to test new behavior
- [x] Reward test exercising rewards in both versions to catch any unexpected balance table assumptions about not having consistent vesting specs applied.
- [x] Land #1251 and rebase on master
- [ ] Update PR title when FIP is published if the number changes

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1482,6 +1482,7 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *abi.
 	if params.Penalty.Sign() < 0 {
 		rt.Abortf(exitcode.ErrIllegalArgument, "cannot penalize a negative amount of funds")
 	}
+	nv := rt.NetworkVersion()
 
 	var st State
 	pledgeDeltaTotal := big.Zero()
@@ -1491,18 +1492,28 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *abi.
 		store := adt.AsStore(rt)
 		rt.ValidateImmediateCallerIs(builtin.RewardActorAddr)
 
+		rewardToLock := params.Reward
+		lockedRewardVestingSpec := &RewardVestingSpecV5
+		if nv >= network.Version6 {
+			// rewardToLock is 75% of params.Reward
+			rewardToLock = LockedRewardFromRewardV6(params.Reward)
+
+			// vesting spec adds back in delay
+			lockedRewardVestingSpec = &RewardVestingSpecV6
+		}
+
 		// This ensures the miner has sufficient funds to lock up amountToLock.
 		// This should always be true if reward actor sends reward funds with the message.
 		unlockedBalance, err := st.GetUnlockedBalance(rt.CurrentBalance())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to calculate unlocked balance")
-		if unlockedBalance.LessThan(params.Reward) {
+		if unlockedBalance.LessThan(rewardToLock) {
 			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", unlockedBalance, params.Reward)
 		}
 
-		newlyVested, err := st.AddLockedFunds(store, rt.CurrEpoch(), params.Reward, &RewardVestingSpec)
+		newlyVested, err := st.AddLockedFunds(store, rt.CurrEpoch(), rewardToLock, lockedRewardVestingSpec)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to lock funds in vesting table")
 		pledgeDeltaTotal = big.Sub(pledgeDeltaTotal, newlyVested)
-		pledgeDeltaTotal = big.Add(pledgeDeltaTotal, params.Reward)
+		pledgeDeltaTotal = big.Add(pledgeDeltaTotal, rewardToLock)
 
 		// If the miner incurred block mining penalties charge these to miner's fee debt
 		err = st.ApplyPenalty(params.Penalty)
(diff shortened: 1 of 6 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -1491,18 +1492,28 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *abi.
...
+			lockedRewardVestingSpec = &RewardVestingSpecV6
+		}
+
 		// This ensures the miner has sufficient funds to lock up amountToLock.
 		// This should always be true if reward actor sends reward funds with the message.
 		unlockedBalance, err := st.GetUnlockedBalance(rt.CurrentBalance())
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to calculate unlocked balance")
-		if unlockedBalance.LessThan(params.Reward) {
+		if unlockedBalance.LessThan(rewardToLock) {
 			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", unlockedBalance, params.Reward)
```
anorth, line 1510:
```suggestion
			rt.Abortf(exitcode.ErrInsufficientFunds, "insufficient funds to lock, available: %v, requested: %v", unlockedBalance, rewardToLock)
```

actors/builtin/miner/monies.go
```diff
@@ -169,3 +173,7 @@ func ConsensusFaultPenalty(thisEpochReward abi.TokenAmount) abi.TokenAmount {
 		big.NewInt(builtin.ExpectedLeadersPerEpoch),
 	)
 }
+
+func LockedRewardFromRewardV6(reward abi.TokenAmount) abi.TokenAmount {
+	return big.Div(big.Mul(reward, LockedRewardFactorNumV6), LockedRewardFactorDenomV6)
```
anorth, line 178: My preferred pattern here is to pass the network version as a parameter here rather than switch on it at the caller. In general, this would reduce the number of places that need to get the version comparison right, and statically find all the relevant places. (because this is introducing a new function, previously identity, we don't get the static win here).

actors/builtin/miner/miner_test.go
```diff
@@ -1306,7 +1306,9 @@ func TestCCUpgrade(t *testing.T) {
 		oldPower := miner.QAPowerForSector(actor.sectorSize, oldSector)
 		expectedFee := miner.PledgePenaltyForContinuedFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, oldPower)
 		expectedPowerDelta := miner.NewPowerPairZero()
+		rt.SetNetworkVersion(network.Version5)
```
anorth, line 1309: I think in general all unit tests should target the latest network version. If we have a version switch, it will generally be because the prior version is already live in the network, and so we're working to test that the new version will behave as expected. E.g. maybe there's some unexpected interaction between the change guarded by this version and other things that are being tested here.

Which means generally all these tests that are not specifically about the reward vesting should require no diff (because the runtime uses VersionMax). If they do require a diff, that diff will specifically highlight the behaviour that's changing.

Is there a diff here to pledge expectations or something? I think we should update expectations to match (hopefully we can do it just once in the deadlineCron or something).

actors/builtin/miner/miner_test.go
```diff
@@ -3925,14 +3952,15 @@ func TestApplyRewards(t *testing.T) {
...
 
 		for i := range vestingFunds.Funds {
 			vf := vestingFunds.Funds[i]
-			require.EqualValues(t, expectedOffset, int64(vf.Epoch)%int64(miner.RewardVestingSpec.Quantization))
+			require.EqualValues(t, expectedOffset, int64(vf.Epoch)%int64(miner.RewardVestingSpecV6.Quantization))
 		}
 
-		assert.Equal(t, amt, st.LockedFunds)
+		lockedAmt := miner.LockedRewardFromRewardV6(amt)
+		assert.Equal(t, lockedAmt, st.LockedFunds)
```
anorth, line 3963: Here is where we can have a test for both the old and new versions. Grandfather the old one by setting the network version to the prior one, and add a new one.

actors/builtin/miner/miner_test.go
ZenGround0: > I think in general all unit tests should target the latest network version.

agree

> Which means generally all these tests that are not specifically about the reward vesting should require no diff

agree on the *should*

> Is there a diff here to pledge expectations or something?

Yes.  The problem is that we use applyRewards as a testing setup function in a bunch of places in a fragile way, not as part of actually testing "what happens if the miners gets a reward message and then x and y happen".  There is a better way to do it as we've known for a while, however designing it and then this refactor seems non-trivial to me.  I didn't want fixing up our testing problems to block this change so I used the ugly approach you see in the diff.  The benefit of this approach is that it makes it really obvious that there is a problem with how we setup tests, because as you point out we shouldn't be using v5 in tests.

anorth: I have addressed this in #1251 and will rebase this PR upon that one, removing the noise.

actors/builtin/miner/monies.go
anorth: Done.

actors/builtin/miner/miner_test.go
anorth: Actually now that I've reverted the change to initial delay, there's not really anything to grandfather here. This is just an improvement on the old test.

anorth:
@ZenGround0 @acruikshank  I have applied my own suggestions. Please carefully review the commits I added here, then I think it's good to go (after #1251).

anorth (approved):

actors/builtin/miner/miner_test.go
```diff
@@ -5074,11 +5103,8 @@ func (h *actorHarness) applyRewards(rt *mock.Runtime, amt, penalty abi.TokenAmou
 	// We further assume the miner can pay the penalty.  If the miner
 	// goes into debt we can't rely on the harness call
 	// TODO unify those cases
-	nv := rt.NetworkVersion()
-	pledgeDelta := big.Subtract(amt, penalty)
-	if nv >= network.Version6 {
-		pledgeDelta = big.Sub(miner.LockedRewardFromRewardV6(amt), penalty)
-	}
+	lockAmt, _ := miner.LockedRewardFromReward(amt, rt.NetworkVersion())
```
ZenGround0, line 5106: Nice demonstration of some of the benefits of pushing the network version down

acruikshank (approved):
LGTM

PR #1251, opened 2020-10-14
Clean up vesting reward use in miner tests

anorth:
Miner tests almost all apply some rewards to a miner actor from which penalties will be paid, both because this is the usual, expected case and also makes the delta easy to explicitly test. However, they didn't previously support part of the penalty coming from unlocked balance (after the vesting funds are exhausted).

This PR:
- Introduces a `bigRewards` number to use as a constant value for rewards that are used only as a target for penalties, replacing a bunch of ad-hoc constants
- Adds `penaltyFromUnlocked` to the cron config, supporting some or all penalty being paid from unlocked funds

This is intended as a prefactor to #1249. Once rebased upon this PR, most of the delta to miner tests can disappear. Only that part related to the changed vesting schedule remains.

This PR changes tests only.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -41,7 +41,8 @@ var testMultiaddrs []abi.Multiaddrs
 
 // A balance for use in tests where the miner's low balance is not interesting.
 var bigBalance = big.Mul(big.NewInt(1_000_000), big.NewInt(1e18))
-var onePercentBigBalance = big.Div(bigBalance, big.NewInt(100))
+// A reward amount for use in tests where the vesting amount wants to be large enough to cover penalties.
+var bigRewards = big.Mul(big.NewInt(1_000), big.NewInt(1e18))
 
 // an expriration 1 greater than min expiration
 const defaultSectorExpiration = 190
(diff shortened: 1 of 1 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1307,7 +1308,7 @@ func TestCCUpgrade(t *testing.T) {
 		oldPower := miner.QAPowerForSector(actor.sectorSize, oldSector)
 		expectedFee := miner.PledgePenaltyForContinuedFault(actor.epochRewardSmooth, actor.epochQAPowerSmooth, oldPower)
 		expectedPowerDelta := miner.NewPowerPairZero()
-		actor.applyRewards(rt, expectedFee, big.Zero())
+		actor.applyRewards(rt, bigRewards, big.Zero())
```
ZenGround0, line 1311: Non-blocking -- I think the use of applyRewards to set up test state for penalization is wrong.  This PR solves the problem at hand but the brittleness from relying on the production reward function won't go away.

actors/builtin/miner/miner_test.go
```diff
@@ -5055,12 +5051,12 @@ func (h *actorHarness) applyRewards(rt *mock.Runtime, amt, penalty abi.TokenAmou
...
 	expectedEnrollment        abi.ChainEpoch
-	vestingPledgeDelta        abi.TokenAmount // nolint:structcheck,unused
 	detectedFaultsPowerDelta  *miner.PowerPair
 	expiredSectorsPowerDelta  *miner.PowerPair
 	expiredSectorsPledgeDelta abi.TokenAmount
-	continuedFaultsPenalty    abi.TokenAmount
-	repaidFeeDebt             abi.TokenAmount
+	continuedFaultsPenalty    abi.TokenAmount // Expected amount burnt to pay continued fault penalties.
+	repaidFeeDebt             abi.TokenAmount // Expected amount burnt to repay fee debt.
+	penaltyFromUnlocked       abi.TokenAmount // Expected reduction in unlocked balance from penalties exceeding vesting funds.
```
ZenGround0, line 5059: ❤️

ZenGround0 (approved):

acruikshank (approved):
Definitely an improvement.

PR #1259, opened 2020-10-19
Remove duplicate constant for epochs in year

zgfzgf:


actors/builtin/sector.go
```diff
@@ -13,7 +13,7 @@ type SealProofPolicy struct {
 }
 
 // For all Stacked DRG sectors, the max is 5 years
-const epochsPerYear = 1_051_200
+const epochsPerYear = 365 * EpochsInDay
```
anorth, line 16: This constant used to exist to resolve a circular dependency that has disappeared, so this change is good.

@Stebalien there's another `epochsInYear` in `reward_logic.go` due to you that makes the same calculation and carries a comment about how it's not exported because it's not a true year. However, we have a constant definition of "year" here. WDYT about just putting `EpochsInYear` in `builtin_network.go` and use it everywhere?

anorth:
@zgfzgf let's remove the `epochsInYear` and `daysInYear` from reward_logic.go too.

actors/builtin/sector.go
Stebalien: I'm fine with that, as long as we make it clear that this is a 365 day year, not a true year so nobody assumes `TODAY+EpochsInYear` will be the same day the next year.

anorth (approved):

PR #1262, opened 2020-10-19
optimize actor method register

zgfzgf:


ZenGround0:
Hey @zgfzgf thanks for filing a pull request. In order to get PRs accepted the basic process is to first file an issue explaining the change and why it is an improvement and then link the PR to this issue.

I don't have a sense for how much speedup this change brings but my instinct is that it is not that much. To motivate this PR you will need to measure the performance improvements and report them in an issue. If the speedup is a negligible part of message execution we won't want to make this change.

Another less obvious factor to consider is that keeping a visible map from method number to actor method name at the top of actor definitions is very nice for developing. If this is a measurable performance bottleneck we should explore how we can optimize without removing this nice property of the code. 

Please reopen once you have a motivating issue, for now closing this down.

zgfzgf:
fix #1263

PR #1265, opened 2020-10-19
add migration to recompute power totals from claims

### LINKED ISSUE #1256: Fix power totals that are inconsistent after v2.0 upgrade
The migration associated with the upgrade to v2.0 modified power claims without updating the associated total power values. The error can be seen by the state validation functions:
```
➜  ent git:(master) ✗ ./ent validate v2 bafy2bzacecmdndullr2s6q6fabq67o4u4ufczixpjryjcudovcizsunwq3zf2 148899
version: 1
Validation: bafy2bzacecmdndullr2s6q6fabq67o4u4ufczixpjryjcudovcizsunwq3zf2 -- with errors -- 4m4.573204494s
f04 power: sum of raw power in claims 649004467919781888 does not match recorded bytes committed 641469068058034176
f04 power: sum of qa power in claims 649241962431381504 does not match recorded qa power committed 641706562569633792
f04 power: claims with sufficient power 317 does not match MinerAboveMinPowerCount 311
f04 power: recorded raw power 640838429420027904 does not match raw power in claims 648386885982355456
f04 power: recorded qa power 641075923931627520 does not match qa power in claims 648624380493955072
```

The error is <1%, so we deemed this of sufficiently low impact not to make an emergency fix, but should resolve it at our next opportunity for a state migration. This might nor might not be associated with a new actors version: we don't need to change the state schema, only fix some values.

The problem in the migration is highlighted in #1237 (which did not land in time). We need a new migration to traverse claims and recalculate totals.

acruikshank:
closes #1256

### Motivation

The V2 migration restored miners that had entered a bad state due to the CC upgrade then fault issue. Unfortunately the process of restoring their power claim did not also update the power actors totals. This leaves the network with slightly incorrect power accounting. This PR recomputes the power actors cached values from the Claims table.

### Proposed Changes

1. Introduce a new `nv4` migration that operates entirely on v2 actors.
2. Introduce `nv4.PowerMigrator` that recomputes `TotalRawBytePower`, `TotalBytesCommitted`, `TotalQualityAdjPower`, `TotalQABytesCommitted`, and `MinerAboveMinPowerCount` from Claims.
3. Introduce `correct_power_accounting_test` which is just the test for the v2 correction with an additional migration and a test that all invariants pass. This is very long. I couldn't figure out how to share the code without putting the test code in production.
4. Add a few helpers in the VM to facilitate the new migration and checking invariants afterwards.

### DIFF
--- a/actors/migration/nv4/test/correct_power_accounting_test.go
+++ b/actors/migration/nv4/test/correct_power_accounting_test.go
@@ -1,0 +1,471 @@
+package test_test
+
+import (
+	"context"
+	"github.com/filecoin-project/specs-actors/v2/actors/migration/nv4"
+	"strings"
+	"testing"
+
+	"github.com/ipfs/go-cid"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/migration"
+	"github.com/filecoin-project/specs-actors/v2/actors/runtime"
+	"github.com/filecoin-project/specs-actors/v2/actors/states"
+
+	addr "github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-bitfield"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/crypto"
+	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/filecoin-project/specs-actors/actors/builtin/market"
+	miner0 "github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/actors/runtime/proof"
+	tutil "github.com/filecoin-project/specs-actors/support/testing"
+	vm0 "github.com/filecoin-project/specs-actors/support/vm"
+
+	exported2 "github.com/filecoin-project/specs-actors/v2/actors/builtin/exported"
+	miner2 "github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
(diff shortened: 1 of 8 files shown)


actors/migration/nv4/test/correct_power_accounting_test.go
```diff
@@ -0,0 +1,472 @@
...
+
+	// all state invariants hold
+	stateTree, err = v2nv4.GetStateTree()
+	require.NoError(t, err)
+	totalBalance, err = v2nv4.GetTotalActorBalance()
+	require.NoError(t, err)
+	msgs, err = states.CheckStateInvariants(stateTree, totalBalance, v2nv4.GetEpoch())
+	require.NoError(t, err)
+	assert.Equal(t, 0, len(msgs.Messages()), strings.Join(msgs.Messages(), "\n"))
+}
```
anorth, line 423: Honestly I don't think this approach to testing this is very good. It's a good representation of the specific scenario we're in, but we could also just construct that much more directly, and could also construct more scenarios to demonstrate robustness of the fix to other things that might go wrong before it is run.

We're only touching the power actor. I recommend something closer to a unit test that either constructs the power state directly (since we got there by migrating it directly, but wrongly) and then inserts into a state tree with just the power actor for invoking the migration (and validation? but that will need corresponding miner state).

anorth:
I think I said something confusing regarding migration packaging. What I intended was to repackage the existing migration as `nv4` (the version to which it migrated) in a separate PR, and then add a new `nv7` package with this new one.

actors/migration/nv7/power.go
```diff
@@ -0,0 +1,97 @@
...
+	cbor "github.com/ipfs/go-ipld-cbor"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	power "github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+)
+
+type claimsSummary struct {
+	committedRawBytes              abi.StoragePower
+	committedQAPower               abi.StoragePower
```
anorth, line 18: `committedQABytes`

anorth (approved):

Issue #1232, opened 2020-10-09
power.CreateMiner does not update MinerAboveMinPowerCount for test proofs


acruikshank:
When power actor updates claims, it increments/decrements the MinerAboveMinPowerCount if the power has crossed the ConsensusMinerMinPower threshold for the actor's seal proof type. Power actor does not do this when it creates the initial claim as it creates the miner. The initial claim has zero power, so this is only an issue if the miner has a seal proof type that has zero ConsensusMinerMinPower. Currently this is only true for the test proof type `RegisteredSealProof_StackedDrg2KiBV1` which is generally not permitted.

Potential solutions:

1. Update MinerAboveMinPowerCount in create miner if the value is above the threshold.
2. Choose a non-zero ConsensusMinerMinPower for RegisteredSealProof_StackedDrg2KiBV1.

anorth:
Thanks, let's do (1).

PR #1273, opened 2020-10-21
add minor version 7 fixes to keep power stats edge cases from breaking invariants

### LINKED ISSUE #1228: Power actor deleteClaim does not decrement MinerCount
StoragePowerActor stores the number of miners in its `MinerCount` field. This number can also be derived (inefficiently) by counting all entries in its `Claims` lookup. When a miner's cron errors, we delete the miner's claim from `Claims` to prevent the miner from taking further actions that might cause additional state corruption. We do not decrement the `MinerCount` in this case even though we have one less active miner.

This issue is low impact. `MinerCount` is not used in any actor calculation (`MinerAboveMinPowerCount` is updated appropriately), and it will only deviate from the number of claims if something unexpected goes wrong in a cron task (something we'd very much like to avoid). Checking that MinerCount matches the size of Claims would be a nice consistency check, but we can't rely on it without this.

acruikshank:
closes #1228 and #1232

### Motivation

We now have invariant tests for the power actor and these have exposed some edge cases that can break invariants:

1. Adding a miner whose zero power claim exceeds the `ConsensusMinerMinPower` does not update `MinerAboveMinPowerCount`, so the count of claims above the threshold does not match the count.
2. Deleting a miner's claim (currently only done when a miner cron fails) updates most power stats, but does not decrement `MinerCount` so the claims table size no longer matches `MinerCount`.

This PR addresses both issues

### Proposed Changes

1. Update `go-state-types` to get some new network versions.
2. Call `updateStatsForNewMiner` in `power.CreateMiner`. This is a new method on state that currently only deals with incrementing `MinerAboveMinPowerCount`.
3. Decrement `MinerCount` when we delete a miner.
4. Test both behaviors do nothing prior to network version 7, and operate as expected afterwards.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -2,11 +2,11 @@ package power
 
 import (
 	"bytes"
-
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/cbor"
 	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/filecoin-project/go-state-types/network"
 	power0 "github.com/filecoin-project/specs-actors/actors/builtin/power"
 	"github.com/ipfs/go-cid"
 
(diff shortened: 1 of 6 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -139,6 +139,12 @@ func (a Actor) CreateMiner(rt Runtime, params *CreateMinerParams) *CreateMinerRe
 
 		st.MinerCount += 1
 
+		// starting version 7, call addToClaim to ensure new claim updates all power stats
+		if rt.NetworkVersion() >= network.Version7 {
+			err := st.updateStatsForNewMiner(params.SealProofType)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed update power stats for new miner %v", addresses.IDAddress)
```
anorth, line 145: Note: In v3 I'd like to push this whole thing including `setClaim` down into a state method.

actors/builtin/power/power_test.go
```diff
@@ -758,6 +791,58 @@ func TestCron(t *testing.T) {
...
+			power := abi.NewStoragePower(0)
+			rt.ExpectSend(builtin.RewardActorAddr, builtin.MethodsReward.UpdateNetworkKPI, &power, big.Zero(), nil, exitcode.Ok)
+			rt.SetCaller(builtin.CronActorAddr, builtin.CronActorCodeID)
+			rt.Call(actor.Actor.OnEpochTickEnd, nil)
+			rt.Verify()
+
+			// expect cron failure was logged
+			rt.ExpectLogsContain("OnDeferredCronEvent failed for miner")
+
+			// miner count is still one
```
anorth, line 841: misleading comment

anorth (approved):

PR #1276, opened 2020-10-23
Remove vesting from PreCommitSector and ConfirmSectorProofsValid

### LINKED ISSUE #1258: Remove vesting from PreCommitSector and ConfirmSectorProofsValid
**Background**
Vesting is processed via `State.UnlockVestedFunds`, which is called from `PreCommitSector`, `ConfirmSectorProofsValid`, `WithDrawBalance` and the deadline cron. The original motivation for computing vesting on sector commitment is so that any vested funds could be used for pre-commit deposit or pledge.

The vesting table is quantized to 12-hr increments. Since vesting is processed in the deadline cron every 30 minutes, the calls during sector commitment almost always achieve nothing. The most it could achieve is to accelerate a release by 30 minutes.

Processing reward vesting is quite expensive as it loads and stores a sizeable array for the vesting table. For high-scale miners, committing many sectors per deadline (or even per-epoch), this represents a sizeable portion of total gas cost and chain bandwidth consumption.

**Proposal**
Remove vesting from `PreCommitSector` and `ConfirmSectorProofsValid`, leaving it to the deadline cron. If necessary, the miner's owner can trigger vesting via `WithdrawBalance` (possibly with an argument of zero).

acruikshank:
closes #1258

### Motivation

Many miner methods attempt to move funds from the vesting table to available balance. While this helps keep the available balance accurate, it can be expensive to run and means that most operations will fail to find any balance to unlock. This is particularly true for PreCommitSector and ConfirmSectorProofsValid, so this PR stops these methods from attempting to unlock funds starting network version 7.

### Proposed Changes

1. Have PreCommitSector only call `UnlockVestedFunds` if network version is less than 7.
2. Have ConfirmSectorProofsValid only call `UnlockVestedFunds` if network version is less than 7.
3. Modify miner tests and add tests to assert that the behavior is correct before and after version 7.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -547,7 +547,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *abi.E
 	newlyVested := big.Zero()
 	feeToBurn := abi.NewTokenAmount(0)
 	rt.StateTransaction(&st, func() {
-		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+		// Stop vesting funds as of version 7. Its computationally expensive and unlikely to release any funds.
+		if rt.NetworkVersion() < network.Version7 {
+			newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+		}
+
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
 		// available balance already accounts for fee debt so it is correct to call
 		// this before RepayDebts. We would have to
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -547,7 +547,11 @@ func (a Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *abi.E
 	newlyVested := big.Zero()
 	feeToBurn := abi.NewTokenAmount(0)
 	rt.StateTransaction(&st, func() {
-		newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+		// Stop vesting funds as of version 7. Its computationally expensive and unlikely to release any funds.
+		if rt.NetworkVersion() < network.Version7 {
+			newlyVested, err = st.UnlockVestedFunds(store, rt.CurrEpoch())
+		}
+
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to vest funds")
```
ZenGround0, line 555: nit: I think it is fine in practice to keep this out here because err will always be nil but this should go within the if statement.

ZenGround0 (approved):

actors/builtin/miner/miner_actor.go
acruikshank: good catch

PR #1278, opened 2020-10-24
fix the baselineExponent comments

steven004:


anorth (approved):
Thanks, I must have messed this up copying from the upgrade on v0.9.

PR #1281, opened 2020-10-27
Reject Window PoSt for already-proven partition

### LINKED ISSUE #1182: Reject rather than attempt to ignore Window PoSt for already-proven partitions
A check in `DeadlineStete.RecordProvenSectors` attempts to ignore any already-proven partition in a WindowPoSt submission. This won't be effective if there are more than one partitions posted. By ignoring it, the sector infos are not loaded for proof and thus will not match the sector infos that the miner would have loaded in order to compute the PoSt.

We should just reject a submission that touches an already-proven partition. If the PoSt was for a single partition, there's no loss (it's already proven). If it was for multiple, we can't process it correctly since we need to ignore any state change for the already-proven partition and it just gets too complicated to keep track of.

anorth:
Fixes #1182.

A truly distinguishing test for this would involved submitting PoSt for two partitions, only one of them being a duplicate. I satisfied myself with inspection of the abort messages instead.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -1932,7 +1932,7 @@ func TestWindowPost(t *testing.T) {
 		}
 
 		{
-			// From version 7, a duplicate is actively rejected.
+			// From version 7, a duplicate is explicitly rejected.
 			rt.SetNetworkVersion(network.Version7)
 			rt.ExpectValidateCallerAddr(append(actor.controlAddrs, actor.owner, actor.worker)...)
 			rt.ExpectGetRandomnessTickets(crypto.DomainSeparationTag_PoStChainCommit, dlinfo.Challenge, nil, commitRand)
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/miner_test.go
```diff
@@ -1921,16 +1925,29 @@ func TestWindowPost(t *testing.T) {
...
+			rt.ExpectValidateCallerAddr(append(actor.controlAddrs, actor.owner, actor.worker)...)
+			rt.ExpectGetRandomnessTickets(crypto.DomainSeparationTag_PoStChainCommit, dlinfo.Challenge, nil, commitRand)
+			rt.ExpectAbortContainsMessage(exitcode.ErrIllegalArgument, "no active sectors", func() {
+				rt.Call(actor.a.SubmitWindowedPoSt, &params)
+			})
+			rt.Reset()
+		}
+
+		{
+			// From version 7, a duplicate is actively rejected.
```
ZenGround0, line 1942: extreme nit: actively -> explicitly

ZenGround0 (approved):
LGTM
I would really like the test you describe for extra confidence before this makes it to mainnet but I am ok with starting lotus integration without it existing.

anorth:
New test case added.

Issue #993, opened 2020-08-18
Higher-level scripted actor scenario tests


anorth:
The unit and scenario tests we have so far require detailed control of progressing epochs and sending messages. They're good for exercising edge cases but cumbersome to build up complex state.

To complement this, we could implement an automated miner "agent" that embodies some policy ("commit one sector at a time indefinitely", "fail every 10th PoSt") and responds to changes in state with appropriate messages. This would resemble miner worker code in Filecoin nodes, but much simplified by not having to deal with a real blockchain (e.g. no re-orgs). We could then build larger scale tests by having one or more of these agents interact with a VM which simply advances epochs, receives messages, then advertises the new state as if they were all included at the next tipset. 

This could help us with:
- exercising boundaries and limits around sectors, partitions, deadlines
- benchmarking transaction throughput
- profiling actor code

anorth:
It would be handy if this agent thingy could be re-usable outside our tests, for automating chain validation / test vector tests, and maybe even real network tests. Avoiding dependencies on anything else from our test support infra would probably suffice.

PR #1289, opened 2020-10-29
First pass at an agent simulation testing framework

acruikshank:
work towards #993
 
### Motivation

We have pretty good unit testing to cover actor behavior, and now have some scenario tests to cover specific inter-actor processes. We have not had any way to run larger scale network processes without integrating the actor code into a test network. This PR leverages the scenario test VM to introduce a simulation framework that models miner behavior.

Currently the simulation can:

1. Create miners a configured rate.
2. Models these miner's behavior with a `MinerAgent` that precommits sectors randomly at a given rate, prove commits them within the correct window, and then PoSt them when necessary.
3. Pays block rewards
4. Runs cron each tick.

TODO:

1. Deals / CC upgrades
2. Per-miner configuration to simulate a mix of big and small miners
3. Disk backed store to support larger sims.
4. Gas or gas-like metrics
5. Multsig, Payment Channels, Verfifreg...?
6. Well definite test cases that run against invariants to be included in a slow test build phase?

### DIFF
--- a/support/agent/miner_agent.go
+++ b/support/agent/miner_agent.go
@@ -10,7 +10,6 @@ import (
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
-	"github.com/filecoin-project/go-state-types/cbor"
 	"github.com/ipfs/go-cid"
 	mh "github.com/multiformats/go-multihash"
 
@@ -72,6 +71,9 @@ func (ma *MinerAgent) Tick(v VMState) ([]Message, error) {
 
 		messages = append(messages, ma.createPrecommit(v.GetEpoch(), sectorNumber, sectorActivation))
 
+		// assume precommit succeeds and schedule prove commit
+		ma.operationSchedule.ScheduleOp(sectorActivation, proveCommitAction{sectorNumber})
+
 		ma.nextPrecommit += precommitDelay(ma.Config.PrecommitRate, ma.rnd)
 		ma.nextSectorNumber++
 	}
@@ -80,14 +82,18 @@ func (ma *MinerAgent) Tick(v VMState) ([]Message, error) {
 	for _, op := range ma.operationSchedule.PopOpsUntil(v.GetEpoch()) {
 		switch o := op.action.(type) {
 		case proveCommitAction:
-			messages = append(messages, ma.createProveCommit(o.sectorNumber))
+			messages = append(messages, ma.createProveCommit(v.GetEpoch(), o.sectorNumber))
 		case registerSectorAction:
 			err := ma.registerSector(v, o.sectorNumber)
 			if err != nil {
 				return nil, err
 			}
 		case proveDeadlineAction:
-			messages = append(messages, ma.proveDeadline(v, o.dlIdx))
+			msg, err := ma.proveDeadline(v, o.dlIdx)
+			if err != nil {
+				return nil, err
+			}
+			messages = append(messages, msg)
 		}
 	}
 
@@ -95,11 +101,11 @@ func (ma *MinerAgent) Tick(v VMState) ([]Message, error) {
 }
 
 // proove sectors in deadline
-func (ma *MinerAgent) proveDeadline(v VMState, dlIdx uint64) Message {
+func (ma *MinerAgent) proveDeadline(v VMState, dlIdx uint64) (Message, error) {
 	var partitions []miner.PoStPartition
 	for pIdx, bf := range ma.deadlines[dlIdx] {
 		if empty, err := bf.IsEmpty(); err != nil {
-			panic(err)
+			return Message{}, err
 		} else if !empty {
 			partitions = append(partitions, miner.PoStPartition{
 				Index:   uint64(pIdx),
@@ -123,16 +129,19 @@ func (ma *MinerAgent) proveDeadline(v VMState, dlIdx uint64) Message {
 		ChainCommitEpoch: v.GetEpoch() - 1,
 		ChainCommitRand:  []byte("not really random"),
 	}
+
+	// assume prove sectors succeeds and schedule its first PoSt (if necessary)
+	if err := ma.scheduleNextProof(v, dlIdx); err != nil {
+		return Message{}, err
+	}
+
 	return Message{
 		From:   ma.Worker,
 		To:     ma.RobustAddress,
 		Value:  big.Zero(),
 		Method: builtin.MethodsMiner.SubmitWindowedPoSt,
 		Params: &params,
-		ReturnHandler: func(v VMState, _ Message, _ cbor.Marshaler) error {
-			return ma.scheduleNextProof(v, dlIdx)
-		},
-	}
+	}, nil
 }
 
 // looks up sector deadline and partition so we can start adding it to PoSts
@@ -183,26 +192,21 @@ func (ma *MinerAgent) scheduleNextProof(v VMState, dlIdx uint64) error {
 }
 
 // create prove commit message
-func (ma *MinerAgent) createProveCommit(sectorNumber abi.SectorNumber) Message {
+func (ma *MinerAgent) createProveCommit(epoch abi.ChainEpoch, sectorNumber abi.SectorNumber) Message {
 	params := miner.ProveCommitSectorParams{
 		SectorNumber: sectorNumber,
 	}
 
-	return Message{
-		From:          ma.Worker,
-		To:            ma.RobustAddress,
-		Value:         big.Zero(),
-		Method:        builtin.MethodsMiner.ProveCommitSector,
-		Params:        &params,
-		ReturnHandler: ma.handleProveCommit,
-	}
-}
+	// register an op for next epoch (after batch prove) to schedule a post for the sector
+	ma.operationSchedule.ScheduleOp(epoch, registerSectorAction{sectorNumber: sectorNumber})
 
-// register an op for next epoch (after batch prove) to schedule a post for the sector
-func (ma *MinerAgent) handleProveCommit(v VMState, msg Message, _ cbor.Marshaler) error {
-	sn := msg.Params.(*miner.ProveCommitSectorParams).SectorNumber
-	ma.operationSchedule.ScheduleOp(v.GetEpoch(), registerSectorAction{sectorNumber: sn})
-	return nil
+	return Message{
+		From:   ma.Worker,
+		To:     ma.RobustAddress,
+		Value:  big.Zero(),
+		Method: builtin.MethodsMiner.ProveCommitSector,
+		Params: &params,
+	}
 }
 
 // create precommit message and activation trigger
@@ -214,16 +218,13 @@ func (ma *MinerAgent) createPrecommit(currentEpoch abi.ChainEpoch, sectorNumber 
 		SealRandEpoch: currentEpoch - 1,
 		Expiration:    ma.sectorExpiration(currentEpoch),
 	}
+
 	return Message{
 		From:   ma.Worker,
 		To:     ma.RobustAddress,
 		Value:  big.Zero(),
 		Method: builtin.MethodsMiner.PreCommitSector,
 		Params: &params,
-		ReturnHandler: func(_ VMState, _ Message, _ cbor.Marshaler) error {
-			ma.operationSchedule.ScheduleOp(sectorActivation, proveCommitAction{sectorNumber})
-			return nil
-		},
 	}
 }
 
--- a/support/agent/miner_agent.go
+++ b/support/agent/miner_agent.go
@@ -1,46 +1,338 @@
 package agent
 
 import (
-	"errors"
+	"container/heap"
+	"crypto/sha256"
+	"github.com/filecoin-project/go-bitfield"
+	mh "github.com/multiformats/go-multihash"
+	"math"
+	"math/rand"
+
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
-	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/cbor"
+	"github.com/ipfs/go-cid"
+
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
-	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/v2/actors/runtime/proof"
 	vm "github.com/filecoin-project/specs-actors/v2/support/vm"
 )
 
-type MinerAgent struct {
-	owner           address.Address
-	worker          address.Address
-	proofType       abi.RegisteredSealProof
-	startingBalance abi.TokenAmount
-	idAddress       address.Address
-	robustAddress   address.Address
-	actions         []interface{}
-	created         bool
+type MinerAgentConfig struct {
+	PrecommitRate   float64 // average number of precommits per epoch
+	ProofType       abi.RegisteredSealProof
+	StartingBalance abi.TokenAmount
 }
 
-func (ma *MinerAgent) Tick(v *vm.VM) error {
-	// create miner if it has not yet been created
-	if !ma.created {
-		ret, code := v.ApplyMessage(ma.owner, builtin.StoragePowerActorAddr, ma.startingBalance, builtin.MethodsPower.CreateMiner, &power.CreateMinerParams{
-			Owner:         ma.owner,
-			Worker:        ma.worker,
-			SealProofType: ma.proofType,
-			Peer:          abi.PeerID("not really a peer id"),
-		})
-		if code != exitcode.Ok {
-			return errors.New("Could not create miner")
-		}
+type MinerAgent struct {
+	Config        MinerAgentConfig
+	Owner         address.Address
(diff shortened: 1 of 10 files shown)


support/agent/miner_agent.go
```diff
@@ -0,0 +1,338 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	sum := sha256.Sum256(data)
+	hash, err := mh.Encode(sum[:], miner.SealedCIDPrefix.MhType)
+	if err != nil {
+		panic(err)
+	}
+	return cid.NewCidV1(miner.SealedCIDPrefix.Codec, hash)
```
anorth, line 265: Do we need these CIDs to be this "real"? Could we just use a Raw/Identity CID instead?

In any case I think this should use a `cid.V1Builder` rather than do the SHA and encoding manually. See example in `builtin/codes.go`

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/filecoin-project/go-state-types/exitcode"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
+	vm "github.com/filecoin-project/specs-actors/v2/support/vm"
+)
+
+type Sim struct {
```
anorth, line 24: Please add explanatory and motivating docs.

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+
+		blockMessages = append(blockMessages, msgs...)
+	}
+
+	// add at most 1 miner per epoch.
+	if len(s.Miners) < len(s.Accounts) && s.rnd.Float32() < s.Config.CreateMinerProbability {
+		addr := s.Accounts[len(s.Miners)]
+		blockMessages = append(blockMessages, s.createMiner(addr, MinerAgentConfig{
+			PrecommitRate:   2.5,
```
anorth, line 96: Please make configurable.

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+	var blockMessages []Message
+
+	// compute power table before state transition to create block rewards at the end
+	powerTable, err := ComputePowerTable(s.v, s.Miners)
+	if err != nil {
+		return err
+	}
+
+	// add all miner messages
+	for _, miner := range s.Miners {
```
anorth, line 83: I think this is unnecessarily specialised to the concept of miners. Arbitrary agents could interact with the system. We don't have any right now, but it doesn't seem to cost anything to change the terminology here to allow for that.

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// add at most 1 miner per epoch.
+	if len(s.Miners) < len(s.Accounts) && s.rnd.Float32() < s.Config.CreateMinerProbability {
+		addr := s.Accounts[len(s.Miners)]
+		blockMessages = append(blockMessages, s.createMiner(addr, MinerAgentConfig{
+			PrecommitRate:   2.5,
+			ProofType:       abi.RegisteredSealProof_StackedDrg32GiBV1,
+			StartingBalance: s.Config.AccountInitialBalance, // miner gets all account funds
+		}))
+	}
```
anorth, line 100: Rather than put this inside the sim, I think it could be modelled as just another agent that creates owner accounts and miners.

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+type MinerPowerTable struct {
+	addr    address.Address
+	qaPower abi.StoragePower
+}
+
+type PowerTable struct {
+	blockReward  abi.TokenAmount
+	totalQAPower abi.StoragePower
+	minerPower   []MinerPowerTable
+}
```
anorth, line 59: These structs confused me a lot when I first saw them. I mistook them for state, but I think they're just transient values. Perhaps push them down the file and don't export them.

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+				return pt, err
+			} else if sufficient {
+				pt.minerPower = append(pt.minerPower, MinerPowerTable{miner.IDAddress, claim.QualityAdjPower})
+			}
+		}
+	}
+	return pt, nil
+}
+
+func (s *Sim) WinCount(minerPower abi.StoragePower, totalPower abi.StoragePower) uint64 {
```
anorth, line 189: This needs documentation of the intent and algorithm.

support/agent/sim.go
```diff
@@ -0,0 +1,251 @@
...
+			if !ok {
+				return errors.Errorf("create miner return has wrong type: %v", ret)
+			}
+
+			params := msg.Params.(*power.CreateMinerParams)
+			if !ok {
+				return errors.Errorf("create miner params has wrong type: %v", msg.Params)
+			}
+
+			miner := NewMinerAgent(params.Owner, params.Worker, createMinerRet.IDAddress, createMinerRet.RobustAddress, s.rnd, cfg)
```
anorth, line 246: Ok, this use of a return handler seems more legitimate. I would love to find a way to avoid it, inspecting state instead, but it justifies the mechanism for now.

As noted above, I think this could be pulled out into a miner factor agent.

support/testing/address.go
```diff
@@ -1,46 +1,37 @@
...
-	"math/rand"
-	"testing"
-
 	addr "github.com/filecoin-project/go-address"
+	"github.com/stretchr/testify/require"
+	"math/rand"
 )
 
-func NewIDAddr(t testing.TB, id uint64) addr.Address {
+func NewIDAddr(t require.TestingT, id uint64) addr.Address {
```
anorth, line 9: Can we just use `*testing.T`? It's not clear why we can't do this the normal way.

anorth (commented):
This looks like a great start. There are a few things I think could be simpler and/or more general at the same time.

There are a few places where you've used randomness that I'm not sure is helpful. I didn't comment most of them, but my bias would be towards more determinism. Where simple determinism isn't enough, maybe I'd escalate to a dependency-injected policy function (which could involve randomness). Could we get rid of randomness from the miner agent entirely (I see we probably need it for elections)? No change necessary right now though.

support/agent/miner_agent.go
```diff
@@ -0,0 +1,338 @@
...
+		ChainCommitRand:  []byte("not really random"),
+	}
+	return Message{
+		From:   ma.Worker,
+		To:     ma.RobustAddress,
+		Value:  big.Zero(),
+		Method: builtin.MethodsMiner.SubmitWindowedPoSt,
+		Params: &params,
+		ReturnHandler: func(v *vm.VM, _ Message, _ cbor.Marshaler) error {
+			return ma.scheduleNextProof(v, dlIdx)
```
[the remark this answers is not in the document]
acruikshank: The callback mechanism (or something like it) is necessary for create miner because we need to capture its return value. For this an all other calls in miner agent it is unnecessary (so long as we assume that messages always succeed), so I'm moving all these handlers back into the main body and leaving the handler nil.

support/agent/miner_agent.go
```diff
@@ -0,0 +1,338 @@
...
+		SectorNumber: sectorNumber,
+	}
+
+	return Message{
+		From:          ma.Worker,
+		To:            ma.RobustAddress,
+		Value:         big.Zero(),
+		Method:        builtin.MethodsMiner.ProveCommitSector,
+		Params:        &params,
+		ReturnHandler: ma.handleProveCommit,
```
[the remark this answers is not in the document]
acruikshank: Done. See comment above.

support/agent/miner_agent.go
```diff
@@ -0,0 +1,338 @@
...
+	ma.operationSchedule.ScheduleOp(v.GetEpoch(), registerSectorAction{sectorNumber: sn})
+	return nil
+}
+
+// create precommit message and activation trigger
+func (ma *MinerAgent) createPrecommit(currentEpoch abi.ChainEpoch, sectorNumber abi.SectorNumber, sectorActivation abi.ChainEpoch) Message {
+	params := miner.PreCommitSectorParams{
+		SealProof:     ma.Config.ProofType,
+		SectorNumber:  sectorNumber,
+		SealedCID:     sectorSealCID(ma.rnd),
```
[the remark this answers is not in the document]
acruikshank: My hope is to get more entropy into these simulations than we have in unit or scenario tests. Using seeded RNGs is just a quick way to get an even distribution. The downside is that small changes could completely alter the simulation (i.e. it's easy to reproduce problems but harder to confirm they're fixed).

One partial solution would be to used nested RNGs (a tree of RNG's where each node is seeded with a number taken from the parent). This would reduce but not eliminate the chance that a small change alters the sequence that leads to a problem.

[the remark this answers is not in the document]
acruikshank: This was trivial to do per miner, so I went ahead and did it. Now changes in the random sequence of one miner will not affect the random sequences of other miners.

support/agent/miner_agent.go
```diff
@@ -0,0 +1,338 @@
...
+	// start precommits, for now assume we have enough pledge funds
+	ma.nextPrecommit -= 1.0
+	for ma.nextPrecommit < 1.0 {
+		// go ahead and choose when we're going to activate this sector
+		sectorActivation := ma.sectorActivation(v.GetEpoch())
+		sectorNumber := ma.nextSectorNumber
+
+		messages = append(messages, ma.createPrecommit(v.GetEpoch(), sectorNumber, sectorActivation))
+
+		ma.nextPrecommit += precommitDelay(ma.Config.PrecommitRate, ma.rnd)
```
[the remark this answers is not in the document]
acruikshank: Comments added at the top of the loop.

support/agent/miner_agent.go
```diff
@@ -0,0 +1,338 @@
...
+			ma.PreCommits++
+			ma.operationSchedule.ScheduleOp(sectorActivation, proveCommitAction{sectorNumber})
+			return nil
+		},
+	}
+}
+
+// create a random valid sector expiration
+func (ma *MinerAgent) sectorExpiration(currentEpoch abi.ChainEpoch) abi.ChainEpoch {
+	// these are precommit bounds. Prove commit is more lenient but contains this range.
```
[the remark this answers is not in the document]
acruikshank: I made similar assumptions, but I kept hitting errors. Here are the constraints on expiration:

**PreCommit**
1. [activation := precommitEpoch + MaxProveCommitDuration](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L555)
2. [expiration > activation](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L555)
3. [expiration >= MinSectorExpiration + activation](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L2002) (note this is stronger than 2.)
4. [expiration <= rt.CurrEpoch()+MaxSectorExpirationExtension](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L2008)
5. [expiration <= builtin.SealProofSectorMaximumLifetime(sealProof) + activation](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L2016)

**ProveCommit**
Does not validate expiration

**ConfirmSectorProofsValid**
1. [expiration >= rt.CurrEpoch(prove) + MinSectorExpiration](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L865) (skips sector but does not error)

Since rt.CurrEpoch() + MaxProveCommitDuration[params.SealProof] at precommit time is greater than or equal rt.CurrEpoch() at prove time, PreCommit (3) is a stronger constraint than ConfirmSectorProofsValid (1). I might have made an error in this analysis. I vaguely remember setting the ranges like this on purpose to avoid potential errors in ConfirmSectorProofsValid.

support/agent/sim.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+
+		if msg.ReturnHandler != nil {
+			if err := msg.ReturnHandler(s.v, msg, ret); err != nil {
+				return err
+			}
+		}
+	}
+
+	// apply block rewards
```
ZenGround0, line 131: While not impossible to see on mainnet (i.e. in the case of all tipset blocks containing identical messages) this sequencing is unlikely so will move simulations further away from the expected behavior of [interleaving rewards](https://github.com/filecoin-project/lotus/blob/2a294f048ac211d999ea6b044fe4412963901fe7/chain/stmgr/stmgr.go#L353) between blocks.  It is probably not a very important difference.

support/agent/miner_agent.go
acruikshank: We actually do. At least we need them to have the right Cid prefix because we validate it in PreCommit. I would like to use the `miner.SealedCIDPrefix.Sum` function, but it panics because it's not-implemented. This is the simplest solution I've found so far that works.

support/agent/sim.go
acruikshank: Hmm. I thought all rewards came at the end of the tipset. I guess if this is the case, shuffling the rewards in with the other message might actually be more realistic (except there could be a reward at zero and it is unlikely that the last message will be a reward).

support/agent/sim.go
acruikshank: I really like this idea. I'll try to get it in.

support/testing/address.go
acruikshank: I started off with the assumption this would be run from the command line outside the test runner. This is a trivial interface to implement. That assumption was bad, though, so I'll revert.

support/agent/miner_agent.go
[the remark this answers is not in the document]
anorth: This would benefit from explanatory comments [like this](https://github.com/filecoin-project/specs-actors/blob/74047a94726e66d895b4fb672f050f61b4360e8b/actors/builtin/miner/miner_actor.go#L553). Note that since expiration is validated at pre-commit, it conservatively assumes the latest possible activation, and then requires at least `MinSectorExpiration` additional lifetime from that point.

support/agent/sim.go
anorth: Please just leave a comment highlighting this difference from the real chain, so it's transparent if anyone runs into it later.

support/ipld/store.go
```diff
@@ -66,3 +66,31 @@ func (ss *SyncBlockStoreInMemory) Put(b block.Block) error {
 func NewSyncADTStore(ctx context.Context) adt.Store {
 	return adt.WrapStore(ctx, cbor.NewCborStore(NewSyncBlockStoreInMemory()))
 }
+
+type MetricsStore struct {
+	bs    cbor.IpldBlockstore
+	Puts  uint64
+	Reads uint64
```
anorth, line 73: Please name either as `{Puts, Gets}` or `{Reads, Writes}`.

support/vm/invocation_context.go
```diff
@@ -31,6 +31,23 @@ import (
...
+type StatsByCall map[MethodKey]*CallStats
+
+type MethodKey struct {
+	Code   cid.Cid
+	Method abi.MethodNum
+}
+
+type CallStats struct {
+	Reads    uint64
+	Puts     uint64
```
anorth, line 43: Ditto

support/vm/invocation_context.go
```diff
@@ -31,6 +31,23 @@ import (
...
+	Calls    uint64
+	SubStats StatsByCall
+
+	startReads uint64
+	startPuts  uint64
+}
```
anorth, line 49: Break this out from this large file, and so that the methods can be placed adjacent. Please document how `startReads` and `startPuts` are to be used.

`StatsByCall` too.

support/vm/invocation_context.go
```diff
@@ -316,13 +342,65 @@ func (ic *invocationContext) Send(toAddr address.Address, methodNum abi.MethodNu
...
+
 	err := ret.Into(out)
 	if err != nil {
 		ic.Abortf(exitcode.ErrSerialization, "failed to serialize send return value into output parameter")
 	}
 	return code
 }
 
+func (ic *invocationContext) mergeSubStat(code cid.Cid, methodNum abi.MethodNum, newStats *CallStats) {
+	if ic.stats.SubStats == nil {
```
anorth, line 356: I _think_ Go might let you implement `CallStats.MergeStats` to accept a `nil` receiver and still do this, which would be nice. Otherwise I'd prefer we init `ic.stats.SubStats` with a pointer to zeros, or use a non-pointer.

support/vm/testing.go
```diff
@@ -51,9 +51,13 @@ func init() {
 //
 
 // Creates a new VM and initializes all singleton actors plus a root verifier account.
-func NewVMWithSingletons(ctx context.Context, t *testing.T) *VM {
+func NewVMWithSingletons(ctx context.Context, t require.TestingT) *VM {
```
anorth, line 54: `*testing.T`

anorth (approved):
Since the stats wasn't in this PR when I first reviewed it, I would much prefer that it had been withheld for a subsequent PR
- This one was already very large (but coherent)
- It's hard to judge the stats approach, especially the coupling introduced, among the all other changes
- I have to chose between seeing any requested changes to this new code and unblocking you on the primary purpose of this change

PR #1291, opened 2020-11-02
Replace V1 seal proof types with V1_1 in a two-stage upgrade.

ZenGround0:
TODO
- [x] merge https://github.com/filecoin-project/go-state-types/pull/19 and update go-state-types dep

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -99,8 +99,7 @@ func (a Actor) Constructor(rt Runtime, params *ConstructorParams) *abi.EmptyValu
 	checkControlAddresses(rt, params.ControlAddrs)
 	checkPeerInfo(rt, params.PeerId, params.Multiaddrs)
 
-	_, ok := SupportedProofTypes[params.SealProofType]
-	if !ok {
+	if !CanPreCommitSealProof(params.SealProofType, rt.NetworkVersion()) {
 		rt.Abortf(exitcode.ErrIllegalArgument, "proof type %d not allowed for new miner actors", params.SealProofType)
 	}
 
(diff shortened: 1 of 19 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -511,13 +510,27 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+//	SealedCID       cid.Cid `checked:"true"` // CommR
+//	SealRandEpoch   abi.ChainEpoch
+//	DealIDs         []abi.DealID
+//	Expiration      abi.ChainEpoch
+//	ReplaceCapacity bool // Whether to replace a "committed capacity" no-deal sector (requires non-empty DealIDs)
+//	// The committed capacity sector to replace, and it's deadline/partition location
+//	ReplaceSectorDeadline  uint64
+//	ReplaceSectorPartition uint64
+//	ReplaceSectorNumber    abi.SectorNumber
+//}
```
acruikshank, line 525: 🙏

actors/builtin/miner/miner_actor.go
```diff
@@ -1998,9 +2030,23 @@ func validateReplaceSector(rt Runtime, st *State, store adt.Store, params *PreCo
...
+	if nv < network.Version7 {
+		if params.SealProof != replaceSector.SealProof {
+			rt.Abortf(exitcode.ErrIllegalArgument, "cannot replace sector %v seal proof %v with seal proof %v",
+				params.ReplaceSectorNumber, replaceSector.SealProof, params.SealProof)
+		}
+	} else {
+		// From network version 7, the new sector's seal type must have the same Window PoSt proof type as the one
+		// being replaced, rather than be exactly the same seal type.
+		// This permits replacing sectors with V1 seal types with V1_1 seal types.
+		replaceWPoStProof, err := replaceSector.SealProof.RegisteredWindowPoStProof()
```
acruikshank, line 2042: Consider extracting this logic since it's duplicated at least once.

actors/builtin/miner/policy.go
```diff
@@ -90,12 +91,57 @@ var SealedCIDPrefix = cid.Prefix{
...
+// From network version 8, sectors sealed with the V1 seal proof types cannot be committed.
+var PreCommitSealProofTypesV8 = map[abi.RegisteredSealProof]struct{}{
+	abi.RegisteredSealProof_StackedDrg32GiBV1_1: {},
+	abi.RegisteredSealProof_StackedDrg64GiBV1_1: {},
+}
+
+// Checks whether a seal proof type is supported for new miners and sectors.
+func CanPreCommitSealProof(s abi.RegisteredSealProof, nv network.Version) bool {
+	_, ok := PreCommitSealProofTypesV0[s]
+	if nv >= network.Version7 {
```
acruikshank, line 116: Is there a reason not to reverse the order and use else ifs/elses? The extra map hits aren't going to cost much, but it looks a little sloppy to my eye.

acruikshank (approved):
Looks good.

actors/builtin/miner/policy.go
anorth: My motivation for this pattern is only to maintain the pattern that every network version check takes the form `nv >= network.VersionX`, and to also maintain a pattern that the cases for different versions are presented in their historical order. With only one way to do it, there's less thought and chance of error. 

However, we could re-order these and maintain the first property (but not the second) and I think it would be fine.

anorth (approved):

actors/builtin/miner/miner_actor.go
ZenGround0: I'm going to do this one

ZenGround0: After poking around a bit I decided not to do this.  Might be worth coming back to when we clear away old network version checks in v3.

PR #1292, opened 2020-11-03
Remove error returns from state validation functions

### LINKED ISSUE #1244: Make state validation checks robust to errors
State validation checks are typically used in tests or by manual operation on network state. If there's an error because some part of state is unreadable (e.g. corrupted bitfield or HAMT node), then we want to know this but don't want it to prevent getting any other information about unrelated state. So treating these errors in the standard Go way by propagating them immediately isn't so helpful.

Re-arrange existing state checks to avoid returning errors at all. Translate errors from low-level functions into messages, skipping dependent checks where necessary.

anorth:
Instead of returning an error that prevents the validation completing at all, all errors are now accumulated as messages. In some cases this might lead to a bit of an overload, but I think that's better than getting nothing.

I was unable to run a full mainnet state validation with `ent` because Lotus has not yet upgraded to consume the changes that preceded this. It could be worth trying to break the ent->Lotus dependency some day.

Closes #1244.

### DIFF
--- a/actors/states/check.go
+++ b/actors/states/check.go
@@ -2,11 +2,13 @@ package states
 
 import (
 	"bytes"
+
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
-	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
 	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
 
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
 	"github.com/filecoin-project/specs-actors/v2/actors/builtin/account"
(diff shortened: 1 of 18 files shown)


actors/states/check.go
```diff
@@ -52,46 +54,33 @@ func CheckStateInvariants(tree *Tree, expectedBalanceTotal abi.TokenAmount, prio
 			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
 				return err
```
ZenGround0, line 55: Returning an error loudly like this seems fine given how messed up the tree must be to not have a required singleton.  However I don't understand why we don't follow the new pattern with this class of errors. Are the effects on other actors invariant checks (i.e. in this case not being able to resolve id addresses) a complication that is hard to handle if we keep going? Or is just being loud because it is a very bad tree the point?

ZenGround0 (approved):

actors/states/check.go
anorth: There is no principled reason. I didn't change it only because the motivation to change it wasn't very strong.

Issue #1028, opened 2020-08-24
Enhance mock and scenario runtime with detection of illegal state modifications


anorth:
Actors are prohibited from mutating ReadOnly state objects (for their own safety, since such mutation wouldn't persist, and is probably a bug). A method which does so should abort at the first instance after the illegal modification of either (a) opening a state transaction, or (b) returning. We should enforce this in the mock runtime and the scenario VM.

For inspiration, [here is how go-filecoin implemented it](https://github.com/filecoin-project/go-filecoin/blob/5a38f94e9351d6979723319d20effbea51a0556a/internal/pkg/vm/internal/vmcontext/actor_state_handle.go#L90).

PR #1293, opened 2020-11-04
Mock runtime detects illegal modification of read-only state

anorth:
As requested in #1028.

@acruikshank @ZenGround0 what do you think is the marginal value of adding the same thing to the scenario VM? It's not too hard, but may slightly impact performance. Are there other more valuable things to do first?


ZenGround0:
@anorth I say go for it. I think the value is fairly high from the point of view of this error case being detected in particular because adding to the scenario vm will soon put this check in the loop during all agent tests which we expect to do a better job of hitting pathological edge cases.  Slight increase to scenario vm runtime sounds ok to me given that our tests are already quite fast and that optimizing agent tests for speed is currently at a low priority.

acruikshank (approved):
Like the mockrt test.

PR #1294, opened 2020-11-05
Scenario VM detects illegal modification of read-only state

### LINKED ISSUE #1028: Enhance mock and scenario runtime with detection of illegal state modifications

anorth:
The same pattern as the mock runtime.

Closes #1028

### DIFF
--- a/support/ipld/cbor.go
+++ b/support/ipld/cbor.go
@@ -1,0 +1,24 @@
+package ipld
+
+import (
+	"bytes"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/cbor"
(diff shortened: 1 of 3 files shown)


acruikshank (approved):
looks good.

PR #1295, opened 2020-11-09
Replace all uses of Assert with explicit aborts or other mechanisms

### LINKED ISSUE #1161: Audit and remove assertions

anorth:
These were fairly straightforward. Some notes:
- For miner balance checks, removed inline checks and rely on invariant checks at the end of top-level methods instead

- [x] Wait for repackaging to `v3` before landing

Closes #1161.


actors/builtin/miner/miner_state.go
```diff
@@ -725,22 +724,18 @@ func (st *State) SaveVestingFunds(store adt.Store, funds *VestingFunds) error {
 //
 
 func (st *State) AddPreCommitDeposit(amount abi.TokenAmount) {
-	newTotal := big.Add(st.PreCommitDeposits, amount)
-	AssertMsg(newTotal.GreaterThanEqual(big.Zero()), "negative pre-commit deposit %s after adding %s to prior %s",
```
ZenGround0, line 729: Removing these makes me nervous. I guess it's ok as long as we are sure that every call modifying the balance allocations runs CheckBalanceInvariants.  I would prefer returning an error here as we do in `AddLockedFunds` so that we don't have to worry about missed checks allowing error states to survive across message calls.

ZenGround0 (approved):

actors/builtin/miner/miner_state.go
anorth: Done

PR #1296, opened 2020-11-09
Feat/migrate minerinfo prooftype

ZenGround0:
Migrate minerState's Info SealProofType field from V1 to V1_1 for 32 and 64 GiB sectors.  This change was requested by lotus to ease integration.  Added in a migration of power table claim seal proof types to fix the inconsistency there too which the invariant checks caught.

We need to measure migration runtime.  If this significantly increases migration runtime we should defer this to the upcoming migration where we can do ahead-of-time migration or other strategies to reduce burden of long runtimes.

### DIFF
--- a/actors/migration/nv7/miner.go
+++ b/actors/migration/nv7/miner.go
@@ -1,0 +1,42 @@
+package nv7
+
+import (
+	"context"
+
+	"github.com/filecoin-project/go-state-types/abi"
+
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/v2/actors/util/adt"
+	cid "github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+)
+
+type MinerMigrator struct{}
+
+func (m MinerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, head cid.Cid, info MigrationInfo) (*StateMigrationResult, error) {
+	var state miner.State
+	if err := store.Get(ctx, head, &state); err != nil {
+		return nil, err
+	}
+	adtStore := adt.WrapStore(ctx, store)
+
+	mInfo, err := state.GetInfo(adtStore)
+	if err != nil {
+		return nil, err
+	}
+	if mInfo.SealProofType == abi.RegisteredSealProof_StackedDrg32GiBV1 {
+		mInfo.SealProofType = abi.RegisteredSealProof_StackedDrg32GiBV1_1
+	} else if mInfo.SealProofType == abi.RegisteredSealProof_StackedDrg64GiBV1 {
+		mInfo.SealProofType = abi.RegisteredSealProof_StackedDrg64GiBV1_1
+	}
+	if err := state.SaveInfo(adtStore, mInfo); err != nil {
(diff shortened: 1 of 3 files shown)


ZenGround0:
Also note the complexity and added migration risk put this on the fence for being to risky to include whether or not runtime increase is ok.

actors/migration/nv7/miner.go
```diff
@@ -0,0 +1,42 @@
...
+
+	mInfo, err := state.GetInfo(adtStore)
+	if err != nil {
+		return nil, err
+	}
+	if mInfo.SealProofType == abi.RegisteredSealProof_StackedDrg32GiBV1 {
+		mInfo.SealProofType = abi.RegisteredSealProof_StackedDrg32GiBV1_1
+	} else if mInfo.SealProofType == abi.RegisteredSealProof_StackedDrg64GiBV1 {
+		mInfo.SealProofType = abi.RegisteredSealProof_StackedDrg64GiBV1_1
+	}
```
anorth, line 32: This should really cover the other sector sizes too, to handle a network that supports different sector sizes but also goes through this network version at an epoch after genesis.

actors/migration/nv7/power.go
```diff
@@ -87,11 +90,22 @@ func (a PowerMigrator) ComputeClaimsStats(ctx context.Context, store cbor.IpldSt
 			summary.rawPower = big.Add(summary.rawPower, claim.RawBytePower)
 			summary.qaPower = big.Add(summary.qaPower, claim.QualityAdjPower)
 		}
-		return nil
+		sealProofType := claim.SealProofType
+		if sealProofType == abi.RegisteredSealProof_StackedDrg32GiBV1 {
+			sealProofType = abi.RegisteredSealProof_StackedDrg32GiBV1_1
+		} else if sealProofType == abi.RegisteredSealProof_StackedDrg64GiBV1 {
+			sealProofType = abi.RegisteredSealProof_StackedDrg32GiBV1_1
+		}
```
anorth, line 98: Ditto

actors/migration/nv7/top.go
```diff
@@ -48,6 +48,38 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, stateRootIn cid
...
+			priorEpoch: priorEpoch,
+		})
+		if err != nil {
+			return err
+		}
+		minerActorOut := states.Actor{
+			Code:       builtin.StorageMinerActorCodeID,
+			Head:       minerResult.NewHead,
+			CallSeqNum: actorIn.CallSeqNum,
+			Balance:    actorIn.Balance,
```
anorth, line 76: Nit: It's probably worth us establishing some more re-usable pieces here to aid future migrations. This ignores the transfer from the minerResult, which we "know" is zero right now, but is error-prone to bake in. So something that does this generically from a `StateMigrationResult`.

anorth (commented):
If we do this, we must also add to the tests associated with the original change. When writing it, I considered the case of miner info = V1_1 and pre-commit or sector info being V1 as not possible. I'm not sure that, say, an in-flight pre-commit for V1 will be accepted.

However I think that overall we should try to avoid this migration. The actor code doesn't need it to be internally consistent, it's motivated by an ergonomic change to Lotus. But even that is probably short-lived, as I hope to remove the seal proof type from the miner info altogether in the next schema change since we're now demonstrating that it's not a fixed property of the miner.

PR #1297, opened 2020-11-10
Simulate faults and recoveries

acruikshank:
### Motivation

To better simulate all actor processes we need to simulate miner behavior when storage faults. This PR adds a mechanism to simulate and track faults which either result in declared fault messages or sectors being skipped when PoSts are submitted. Similarly, miners can recover their faults, so the simulation can now add a probability that a fault will be recovered resulting in a declare fault recovered message.

This PR also:

1. Introduces a `math.go` file with a new `RateIterator` to contain some of the stochastic machinery used by the simulation.
2. Introduces the `RateIterator` struct that will call a function some number of times each tick based on a rate and the Poisson distribution.

### DIFF
--- a/support/agent/math.go
+++ b/support/agent/math.go
@@ -16,12 +16,17 @@ type RateIterator struct {
 
 func NewRateIterator(rate float64, seed int64) *RateIterator {
 	rnd := rand.New(rand.NewSource(seed))
+	next := 1.0
+	if rate > 0.0 {
+		next -= math.Log(1.0-rnd.Float64()) / rate
+	}
+
 	return &RateIterator{
 		rnd:  rnd,
 		rate: rate,
 
 		// choose first event in next tick
-		nextOccurrence: 1.0 - math.Log(1.0-rnd.Float64())/rate,
+		nextOccurrence: next,
 	}
 }
 
@@ -29,6 +34,11 @@ func NewRateIterator(rate float64, seed int64) *RateIterator {
 // The function will be called `rate` times on average, but may be called zero or many times in any Tick.
 //
 func (ri *RateIterator) Tick(f func() error) error {
+	// wait until we have a positive rate before doing anything
+	if ri.rate <= 0.0 {
+		return nil
+	}
+
 	// next tick becomes this tick
 	ri.nextOccurrence -= 1.0
 
@@ -48,6 +58,9 @@ func (ri *RateIterator) Tick(f func() error) error {
 
 // convenience method for when rate depends on changing variables
 func (ri *RateIterator) TickWithRate(rate float64, f func() error) error {
+	if rate > 0.0 && ri.rate <= 0.0 {
+		ri.nextOccurrence -= math.Log(1.0-ri.rnd.Float64()) / rate
+	}
 	ri.rate = rate
 	return ri.Tick(f)
 }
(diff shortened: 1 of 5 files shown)


support/agent/miner_agent.go
```diff
@@ -156,34 +195,187 @@ func (ma *MinerAgent) Tick(v SimState) ([]message, error) {
...
+		return nil, errors.Errorf("could not check if %d in deadline %d partition %d is faulty",
+			recoveryNumber, recoveryDlInfo.Index, pIdx)
+	} else if !set {
+		return nil, errors.Errorf("recovery %d in deadline %d partition %d was not a fault",
+			recoveryNumber, recoveryDlInfo.Index, pIdx)
+	}
+
+	// If it's too late, schedule recovery rather than declaring it
+	if recoveryDlInfo.FaultCutoffPassed() {
+		ma.operationSchedule.ScheduleOp(recoveryDlInfo.Close, recoverSectorAction{
```
ZenGround0, line 342: Out of curiosity do we go through the operation schedule when we terminate or expire a sector to remove all outstanding operations on terminated sectors?

ZenGround0 (approved):
I'm missing a fair bit of context from not reading through the whole framework last review but the changes here look good.

support/agent/miner_agent.go
acruikshank: we do now.

support/agent/math.go
```diff
@@ -0,0 +1,131 @@
...
+
+import (
+	"math"
+	big2 "math/big"
+	"math/rand"
+
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+type RateIterator struct {
```
anorth, line 11: Brief documentation please

support/agent/miner_agent.go
```diff
@@ -115,75 +134,260 @@ func NewMinerAgent(owner address.Address, worker address.Address, idAddress addr
...
+		}
+		messages = append(messages, msgs...)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	// Recover sectors.
+	// Rate must be multiplied by the number of live sectors
+	recoveryRate := ma.Config.RecoveryRate * float64(len(ma.faultySectors))
```
anorth, line 212: Comment disagrees with code

support/agent/miner_agent.go
```diff
@@ -253,46 +471,149 @@ func (ma *MinerAgent) scheduleNextProof(v SimState, dlIdx uint64) error {
...
-// create precommit message and activation trigger
-func (ma *MinerAgent) createPrecommit(currentEpoch abi.ChainEpoch, sectorNumber abi.SectorNumber, sectorActivation abi.ChainEpoch) message {
-	params := miner.PreCommitSectorParams{
-		SealProof:     ma.Config.ProofType,
-		SectorNumber:  sectorNumber,
-		SealedCID:     sectorSealCID(ma.rnd),
-		SealRandEpoch: currentEpoch - 1,
-		Expiration:    ma.sectorExpiration(currentEpoch),
+func filterSlice(ns []uint64, toRemove map[uint64]bool) []uint64 {
+	var nextLive []uint64
```
anorth, line 574: Consider pre-allocating, maybe to `len(ns) - len(toRemove)`

support/agent/miner_agent.go
```diff
@@ -340,6 +656,45 @@ func sectorSealCID(rnd *rand.Rand) cid.Cid {
...
+	var err error
+	part.sectors, err = bitfield.SubtractBitField(part.sectors, newExpired)
+	if err != nil {
+		return err
+	}
+	part.faults, err = bitfield.SubtractBitField(part.faults, newExpired)
+	if err != nil {
+		return err
+	}
+	part.faults, err = bitfield.SubtractBitField(part.faults, newExpired)
```
anorth, line 683: This is duplicated 4 lines above

anorth (approved):

support/agent/miner_agent.go
acruikshank: We don't, but only `registerSector` and `delayedRecoveryMessage` operate on individual on-chain sectors and they both handle missing sectors gracefully.

PR #1298, opened 2020-11-11
Remove confusing plumbing of known-zero power for inactive sectors

anorth:
When new sectors are committed, the power delta is always zero, since they are activated on first Window PoSt. This plumbing was a legacy. Some of it remains for the case of moving sectors between partitions during compaction, but even there it's only for a consistency check, not actually functional.

- [x] Wait for repackaging to v3

cc @nicola


nicola:
Can't comment on the implementation, but if the idea is to avoid adding claims at ProveCommit, that's a good idea.

actors/builtin/miner/miner_actor.go
```diff
@@ -1501,11 +1498,12 @@ func (a Actor) CompactPartitions(rt Runtime, params *CompactPartitionsParams) *a
...
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load moved sectors")
 
-		newPower, err := deadline.AddSectors(store, info.WindowPoStPartitionSectors, true, sectors, info.SectorSize, quant)
+		proven := true
+		addedPower, err := deadline.AddSectors(store, info.WindowPoStPartitionSectors, proven, sectors, info.SectorSize, quant)
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to add back moved sectors")
 
-		if !removedPower.Equals(newPower) {
-			rt.Abortf(exitcode.ErrIllegalState, "power changed when compacting partitions: was %v, is now %v", removedPower, newPower)
+		if !removedPower.Equals(addedPower) {
```
ZenGround0, line 1505: The correctness of this hinges on the property of `Deadline::RemovePartitions` that it is an error to remove a partition with unproven power.  This property should be documented in RemovePartitions' function comment "Returns an error if any of the partitions contained faulty sectors, early terminations, or unproven sectors"

ZenGround0 (approved):
Great stuff

PR #1299, opened 2020-11-11
Update to v3

ZenGround0:
1. Update go.mod path
2. Update internal dependencies from v2 to v3
3. Move most v0 struct imports to v2 imports either replacing v0 aliases or introducing new aliases for types defined uniquely in v2


ZenGround0:
@anorth I'm very open to undoing the v0 => v2 alias changes.  It felt vaguely nicer to depend primarily on v2 and delegate v2's aliasing to v0 solely to v0 but no strong reason for this.

anorth:
I think I would avoid the v0 => v2 alias change, but also don't feel strongly about it. Aliasing the oldest matching version clearly shows what has/n't changed in the most recent version, and for how many versions something has been constant. There's some chance of faster building due to the already-resolved alias, too.

The thing that tips it for me is that leaving them is less work and a smaller diff, which can be done with a script.

anorth (approved):

PR #1301, opened 2020-11-12
Porep Gas charge reduced 4x past v7

ZenGround0:
Gas rebalancing [here](https://github.com/filecoin-project/FIPs/issues/31) puts PoRep gas charges 4x lower to account for speedup in default proofs implementation.  @Kubuxu has requested we include this in actors v2.3.x.  Iit is high impact and a simple change hence this PR.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -2,6 +2,7 @@ package power
 
 import (
 	"bytes"
+
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/cbor"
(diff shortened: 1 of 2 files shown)


Kubuxu (approved):

actors/builtin/power/power_actor.go
```diff
@@ -259,10 +260,15 @@ func (a Actor) UpdatePledgeTotal(rt Runtime, pledgeDelta *abi.TokenAmount) *abi.
 // This number is empirically determined
 const GasOnSubmitVerifySeal = 34721049
```
acruikshank, line 261: Not a big deal, but you may want to make this point to the constant in the V0 repo the way we do parameter types.

acruikshank (approved):
Looks good.

PR #1302, opened 2020-11-12
Simulate deals in actor code

acruikshank:
### Motivation

In order to create a more accurate and complete simulation, this PR adds `DealClientAgent`s that interact with `MinerAgent`s to create and publish deals. This PR does a few other things along the way:

1. It moves `MinerGenerator` into its own file.
2. It modifies the sim to allow agents to chose a `DealProvider` allowing clients to communicate to miners to make deals.
3. It better simulates the circulating supply function in the VM runtime. since this is now used when calculating deal collateral.

### DIFF
--- a/support/agent/deal_client_agent.go
+++ b/support/agent/deal_client_agent.go
@@ -1,0 +1,206 @@
+package agent
+
+import (
+	"crypto/sha256"
+	mh "github.com/multiformats/go-multihash"
+	"math/bits"
+	"math/rand"
+	"strconv"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/cbor"
+	"github.com/filecoin-project/go-state-types/crypto"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/market"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/v2/actors/builtin/reward"
+	"github.com/ipfs/go-cid"
+)
+
+type DealClientConfig struct {
+	DealRate         float64         // deals made per epoch
+	MinPieceSize     uint64          // minimum piece size (actual piece size be rounded up to power of 2)
+	MaxPieceSize     uint64          // maximum piece size
+	MinStoragePrice  abi.TokenAmount // minimum price per epoch a client will pay for storage (may be zero)
+	MaxStoragePrice  abi.TokenAmount // maximum price per epoch a client will pay for storage
+	MinMarketBalance abi.TokenAmount // balance below which client will top up funds in market actor
+	MaxMarketBalance abi.TokenAmount // balance to which client will top up funds in market actor
+}
+
+type DealClientAgent struct {
+	DealCount int
+
+	account    address.Address
+	config     DealClientConfig
+	dealEvents *RateIterator
+	rnd        *rand.Rand
+
+	// tracks funds expected to be locked for client deal payment
+	expectedMarketBalance abi.TokenAmount
+}
+
+func AddDealClientsForAccounts(s SimState, accounts []address.Address, seed int64, config DealClientConfig) []*DealClientAgent {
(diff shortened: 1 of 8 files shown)


anorth:
This includes a large number of commits from #1297. Could you please rebase it?

support/agent/deal_client_agent.go
```diff
@@ -0,0 +1,206 @@
...
+		pieceSize = 1 << bits.Len64(pieceSize)
+	}
+
+	providerCollateral, err := calculateProviderCollateral(s, abi.PaddedPieceSize(pieceSize))
+	if err != nil {
+		return err
+	}
+
+	// if provider does not have enough collateral, just skip this deal
+	if provider.AvailableCollateral().LessThan(providerCollateral) {
```
ZenGround0, line 129: Simulating failures is interesting so removing this check is definitely something we should configure later.  I'm mulling over what is simpler, removing this check and letting failures happen (less code) or keeping this check and having things succeed by default.  Probably the current approach is simpler since failure modes are somewhat simulation implementation dependent and correct protocol execution is not.  Just bringing this up as I'm curious to hear the thinking that got you here.

support/agent/miner_agent.go
```diff
@@ -415,6 +407,70 @@ func (ma *MinerAgent) submitPoStForDeadline(v SimState, dlIdx uint64) ([]message
...
+					size: params.Deals[idx].Proposal.PieceSize,
+					ends: params.Deals[idx].Proposal.EndEpoch,
+				})
+			}
+			return nil
+		},
+	}}
+}
+
+func (ma *MinerAgent) updateMarketBalance() []message {
```
ZenGround0, line 446: I am noting duplication with client agent and that there exists an opportunity to reduce code with a shared method though the current level of duplication does seem fine.

support/agent/miner_agent.go
```diff
@@ -480,6 +536,40 @@ func (ma *MinerAgent) scheduleSyncAndNextProof(v SimState, dlIdx uint64) error {
...
+	if err != nil {
+		panic(err)
+	}
+
+	// pieces are aligned so that each starts at the first multiple of its piece size >= the next empty slot.
+	// just stop when we find one that doesn't fit in the sector. Assume pieces can't have zero size
+	loc := uint64(0)
+	for _, piece := range ma.dealsPendingInclusion {
+		size := uint64(piece.size)
+		loc = (loc + size - 1) / size * size
```
ZenGround0, line 554: Another pair of parentheses would help greatly I spent a few minutes confused why we were dividing by the square of the size.

ZenGround0 (approved):
Looks great

support/agent/deal_client_agent.go
```diff
@@ -0,0 +1,206 @@
...
+// Create a proposal
+// Return false if provider if deal can't be performed because client or provider lacks funds
+func (dca *DealClientAgent) createDeal(s SimState, provider DealProvider) error {
+	pieceCid, err := dca.generatePieceCID()
+	if err != nil {
+		return err
+	}
+
+	pieceSize := dca.config.MinPieceSize +
+		uint64(dca.rnd.Int63n(int64(dca.config.MaxPieceSize-dca.config.MinPieceSize)))
```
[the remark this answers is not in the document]
acruikshank: The goal is to assume a uniform distribution over some range of file sizes and then compute the padded piece size from that. This will give an uneven distribution of padded piece sizes which should be somewhat realistic while also covering the range.

support/agent/deal_client_agent.go
acruikshank: I agree that at some point we should cover failures. We already support failures in ConfirmSectorProofsValid (which appear naturally in the upcoming CC upgrade PR when both faults and upgrades are allowed at a sufficient rate). This currently works because ConfirmSectorProofsValid occurs in cron, and doesn't cause the cron message to fail.

Some design will be required to support messages that fail. Specifically do we need a way to expect failure, or do we simply log failures and continue? Adding an expected exit code to messages wouldn't be hard, and might be a reasonable thing to do in this case. I think I'd prefer to save it for another PR though.

support/agent/miner_agent.go
acruikshank: I took a stab at this. It definitely can be done, but the additional interfaces and implementations will require many more lines of code that simply duplicating the methods and/or make configuration more complex. I'm going to hold off on this till I figure out a better way, or we want to add something that would cause more duplication.

PR #1303, opened 2020-11-14
Revert "Porep Gas charge reduced 4x past v7 (#1301)"

Kubuxu:
This reverts commit d84628a278c877bd7ff3a0eadca42fc06105b899.

Due to cryptoeconomic requirements we will not be able to ship this adjustment
in next network version.

@ZenGround0 sorry for making you write this on short time and now reverting it,
and also causing another release having to be pushed.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -2,7 +2,6 @@ package power
 
 import (
 	"bytes"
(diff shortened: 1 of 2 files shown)


ZenGround0 (approved):
No problem, thanks for the PR

PR #1304, opened 2020-11-17
Simulate cc upgrades

acruikshank:
### Motivation

In order to more fully simulate all filecoin processes, this PR gives miner agents the ability to perform committed capacity upgrades to replace sectors with no deals with sectors that have deals. It does this by tracking sectors that have no deals, and choosing one of them to specify as a replacement in a PreCommit if configured to do so.

In addition, there a some small improvements and bug fixes:

* The simulation now outputs all VM log messages which greatly aids debugging.
* The miner agent now schedules a PoSt at the end of the next proving period even if a deadline contains no sectors to prove. This prevents missed PoSts when sectors are added to the empty deadline.


support/agent/miner_agent.go
```diff
@@ -372,6 +417,11 @@ func (ma *MinerAgent) submitPoStForDeadline(v SimState, dlIdx uint64) ([]message
 		}
 	}
 
+	// schedule post-deadline state synchronization and next PoSt
+	if err := ma.scheduleSyncAndNextProof(v, dlIdx); err != nil {
```
acruikshank, line 421: Move this to earlier in the call so that it still gets called when PoSt contains no sectors.

support/agent/miner_agent.go
```diff
@@ -275,6 +317,9 @@ func (ma *MinerAgent) createFault(v SimState) ([]message, error) {
 	faultNumber, ma.liveSectors = PopRandom(ma.liveSectors, ma.rnd)
 	ma.faultySectors = append(ma.faultySectors, faultNumber)
 
+	// avoid trying to upgrade a faulty sector
+	ma.ccSectors = filterSlice(ma.ccSectors, map[uint64]bool{faultNumber: true})
```
ZenGround0, line 321: Should we add back on recovery?

support/agent/miner_agent.go
```diff
@@ -232,25 +240,59 @@ func (ma *MinerAgent) createPreCommit(currentEpoch abi.ChainEpoch) message {
...
+		dlInfo, pIdx, err := ma.dlInfoForSector(s, upgradeNumber)
+		if err != nil {
+			return message{}, err
+		}
+
+		params.ReplaceCapacity = true
+		params.ReplaceSectorNumber = abi.SectorNumber(upgradeNumber)
+		params.ReplaceSectorDeadline = dlInfo.Index
+		params.ReplaceSectorPartition = pIdx
+		ma.UpgradedSectors++
```
ZenGround0, line 267: Because this is updated at precommit the name is a bit confusing.  Either this should be updated on prove or the name changed to reflect that not all of these upgrades are successful.

support/agent/cases_test.go
```diff
@@ -240,6 +240,88 @@ func TestCreateDeals(t *testing.T) {
 	}
 }
 
+func TestCCUpgrades(t *testing.T) {
+	t.Skip("this is slow")
```
ZenGround0, line 244: Idea: not sure how to do it from go end but it would be awesome to create a make command that somehow runs these slow tests.  I think I remember we found a way to do this in gfc.

ZenGround0 (approved):

support/agent/miner_agent.go
acruikshank: Getting a few more sectors to upgrade usually won't affect the sim that much. I did attempt this just to show that a recovered sector could be upgraded, but it ended up requiring substantial additional state management in the miner agent, so I backed out for now.

support/agent/miner_agent.go
acruikshank: I delayed this stat update till the sector is registered (after CSPV).

Issue #1241, opened 2020-10-11
Add batched SectorPreCommit (FIP-0008)


anorth:
When a miner is going to commit many sectors in a small span of epochs, much gas usage could be saved by a batched invocation.

Note that in order to be effective, nodes will need to group pre-commit invocations, perhaps across a few epochs.

This is described in [FIP-0008](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0008.md) (discussion #25)

anorth:
I'm splitting this in half to separate pre-commit and prove-commit. We already have a form of prove-commit batching, though we currently verify proofs charge gas as if proof verifications were not in fact batched.

anorth:
The FIP is being drafted in https://github.com/filecoin-project/FIPs/pull/27.

anorth:
Update: this work needs to wait for a batched prove commit first.

BigLep:
@anorth @ZenGround0 : given the PRs are merged, I'm closing this issue.  Please reopen if there is more work.

PR #1306, opened 2020-11-23
Convert market actor VerifyDealsForActivation to batch

### LINKED ISSUE #981: Miner/Market: Clean up deal weight calculation during precommit

anorth:
As pre-work to batching sector pre-commits (#1241), this PR changes the market actor's `VerifyDealsForActivation` to accept deal information for multiple sectors in a single call.

Along the way, I looked to solve #981. I removed the fake sector start epoch from the method parameters, but left the check in place (using the current epoch as determined by the market actor). I think it is more helpful to check this here than not to, since this could prevent a miner losing a pre-commit deposit. It can't prevent all cases, but it can stop some.

`VerifyDealsForActivation` is only invoked internally, by the miner actor, so the changed API is expected to have little external impact.

- [x] Wait for #1299 to land first

FYI @Stebalien. Closes #981

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -13,7 +13,6 @@ import (
 	"github.com/filecoin-project/go-state-types/exitcode"
 	rtt "github.com/filecoin-project/go-state-types/rt"
 	market0 "github.com/filecoin-project/specs-actors/actors/builtin/market"
-	market2 "github.com/filecoin-project/specs-actors/v2/actors/builtin/market"
 	"github.com/ipfs/go-cid"
 	cbg "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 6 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -639,14 +665,22 @@ func deleteDealProposalAndState(dealId abi.DealID, states *DealMetaArray, propos
...
-	seenDealIDs := make(map[abi.DealID]struct{}, len(dealIDs))
+	return validateAndComputeDealWeight(proposals, dealIDs, minerAddr, sectorExpiry, currEpoch)
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Checks
+////////////////////////////////////////////////////////////////////////////////
 
+func validateAndComputeDealWeight(proposals *DealArray, dealIDs []abi.DealID, minerAddr addr.Address,
+	sectorExpiry abi.ChainEpoch, sectorActivation abi.ChainEpoch) (big.Int, big.Int, uint64, error) {
```
acruikshank, line 681: Consider subbing `minimumDealStart` for `sectorActivation` since this parameter can be either deal activation or current epoch depending on how it's called.

actors/builtin/miner/miner_actor.go
```diff
@@ -2222,29 +2231,38 @@ func requestUnsealedSectorCID(rt Runtime, proofType abi.RegisteredSealProof, dea
...
-		return market.VerifyDealsForActivationReturn{
-			DealWeight:         big.Zero(),
-			VerifiedDealWeight: big.Zero(),
+func requestDealWeights(rt Runtime, sectors []market.SectorDeals) *market.VerifyDealsForActivationReturn {
+	// Short-circuit if there are no deals in any of the sectors.
+	dealCount := 0
+	for _, sector := range sectors {
+		dealCount += len(sector.DealIDs)
+	}
+	if dealCount == 0 {
```
acruikshank, line 2240: Ensuring that we don't contact market actor when we have no deals at all, but not worrying about individual sectors that contain no deals only makes sense if sending a message costs a lot of gas but message size of an internal message costs much much less. I believe this is the case. 

If message size is an issue, it would be better to move this conditional outside this method, or to move the allocation of deals into market.SectorDeals inside.

acruikshank (approved):

actors/builtin/market/market_actor.go
anorth: I considered it but left as is. I initially misunderstood your proposal and thought you meant to type `maximum` [epoch at which the deal could start]. I eventually realised how to interpret it your way, but decided to just leave it meaning "proposed activation epoch". Then the question became whether to drop `sector` from the name, I left it there for consistency with `sectorExpiry`, where it makes good sense.

actors/builtin/miner/miner_actor.go
anorth: I considered dropping the optimization altogether, but didn't want to incur the significant cost of message dispatch in the very common case of no deals. I'm less concerned about optimizing beyond that unless profiling shows it to matter.

Issue #1309, opened 2020-11-26
Sectors sealed with V1 seal proof type cannot be extended


ethan-xiao:
ERROR: GasEstimateMessageGas error: estimating gas used: message execution failed: exit 18, reason: cannot extend expiration for sector 1049 with unsupported seal type 3 (RetCode=18)

anorth:
This is by design, motivated by security considerations for the V1 proof. @nicola can you add any more clarity?

deltazxm:
V1 proof do windowPOST have no security problem.I think V1 proof should be allowed extension to maxLifetime.Otherwise it is unfair to the miners of V1 proof sectors.

maobisheng:
Just extend the lifetime of sectors to maintain the power, I think it has noting to do with security.

coder-lb:
I also think V1 proof doesn't have security problems. If these sectors can't be extended, so much sealing machine resources will be wasted to seal sectors again. Filecoin should be a project with less energy consumed, then it will be more competitive in storage markets.
If the huge number of sectors can't be extended, there will be several side effects. If the miners seal sectors again, it will take much more time, and much more gas. More messages will make the chain congested, and other miners will pay more high gas fees, too.

steven004:
> This is by design, motivated by security considerations for the V1 proof. @nicola can you add any more clarity?

@anorth @nicola Would you please elaborate what kind of security concerns being considered?

nicola:
There is a risk that a year+ from now, with highly pipelined custom hardware (less likely to be software - but not impossible), one may able to avoid storing 100% of the sector and still succeed at WindowPoSt.

As we come closer to these sectors' expiration (most are a ~year from now), depending on (1) what percentage of power these sectors sum up to, (2) what is the state of the art of ASICs and Proof of Space software, the community should decide what to do with those sectors.

I would not recommend to allow for sector extension right now.

PR #1310, opened 2020-11-26
Fix: sector extension error #1309

ethan-xiao:


anorth:
This behaviour is by design, which we can discuss that in #1309. I am going to close this PR because even if we decided to change it, the change would have to land with a new network version upgrade (v9 or greater) in order to preserve chain history compatibility.

deltazxm:
V1 proof do windowPOST have no security problem.I think V1 proof should be allowed extension to maxLifetime.Otherwise it is unfair to the miners of V1 proof sectors.

KaptinLin:
I don't think it's unfair to V1 proof sectors as there are almost SP1 sectors.

PR #1314, opened 2020-12-01
Check consensus fault target is receiving actor

anorth:
The miner actor fails to check that the target of a reported consensus fault is the miner itself. This PR adds the missing check, and expands tests to cover this and also the case of an invalid fault (already correctly handled in the code).

This issue was reported by @zgfzgf under our responsible disclosure program. Thank you! It cannot be exploited in the network at present since the Lotus node enforces that the fault target is the receiving actor during the `VerifyConsensusFault` syscall (but it shouldn't have to).

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1628,8 +1628,6 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *abi.
 type ReportConsensusFaultParams = miner0.ReportConsensusFaultParams
 
 func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultParams) *abi.EmptyValue {
-	// Note: only the first reporter of any fault is rewarded.
-	// Subsequent invocations fail because the target miner has been removed.
 	rt.ValidateImmediateCallerType(builtin.CallerTypesSignable...)
 	reporter := rt.Caller()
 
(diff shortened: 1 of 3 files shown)


arajasek (approved):

actors/builtin/miner/miner_actor.go
```diff
@@ -1628,21 +1628,22 @@ func (a Actor) ApplyRewards(rt Runtime, params *builtin.ApplyRewardParams) *abi.
 type ReportConsensusFaultParams = miner0.ReportConsensusFaultParams
 
 func (a Actor) ReportConsensusFault(rt Runtime, params *ReportConsensusFaultParams) *abi.EmptyValue {
-	// Note: only the first reporter of any fault is rewarded.
```
Stebalien, line 1631: can we update this to explain why subsequent calls fail? That was tricky to figure out.

Stebalien (approved):

actors/runtime/types.go
```diff
@@ -2,13 +2,13 @@ package runtime
 
 import (
 	"github.com/filecoin-project/go-state-types/rt"
-	runtime2 "github.com/filecoin-project/specs-actors/v2/actors/runtime"
+	runtime0 "github.com/filecoin-project/specs-actors/actors/runtime"
```
ZenGround0, line 5: 👍

ZenGround0 (approved):

Kubuxu (approved):

PR #1315, opened 2020-12-01
various agent testing bug fixes and improvements

acruikshank:
### Motivation

In the process of using the agent testing framework to predict the benefits of market actor performance enhancements, I discovered some of its issues and limitations. This PR Addresses those

### Proposed Changes

1. Added a 500 epoch sanity check simulation. The agent testing framework was broken in master. Because the simulations are too slow to run in CI, actor changes are likely to break the simulation without anyone knowing. This ensures anything that obviously breaks the simulation will draw attention.
2. Fixed an issue with the rate iterator that caused it to fire too often (this fixed the issue mentioned above).
3. Fixed an issue where the simulation would fail if a client attempted to make a deal before any miners had been constructed.
4. Add ReadBytes and WriteBytes to per-call stats. This data was made available in the store in a previous update, so this just wires it to the call stats.

### DIFF
--- a/support/agent/cases_test.go
+++ b/support/agent/cases_test.go
@@ -62,6 +62,86 @@ func TestCreate20Miners(t *testing.T) {
 	}
 }
 
+func Test500Epochs(t *testing.T) {
+	ctx := context.Background()
+	initialBalance := big.Mul(big.NewInt(1e8), big.NewInt(1e18))
+	cumulativeStats := make(vm_test.StatsByCall)
+	minerCount := 10
+	clientCount := 9
+
+	// set up sim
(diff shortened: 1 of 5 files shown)


support/agent/cases_test.go
```diff
@@ -62,6 +62,86 @@ func TestCreate20Miners(t *testing.T) {
 	}
 }
 
+func Test500Epochs(t *testing.T) {
```
anorth, line 65: Please document the intention here of a sanity check that is fast enough to run during dev cycles and CI.

anorth (approved):

ZenGround0 (approved):

Issue #1305, opened 2020-11-18
PublishStorageDeals Optimizations


ZenGround0:
Anecdotally the PublishStorageDeals method is very expensive in terms of gas cost.  This is becoming especially apparent in light of more accurate gas assessments.

Some of this is unavoidable since storing deal metadata on chain is expensive.  However it is likely that logic specific optimizations will help bring down costs.  This issue is to profile and explore these optimizations.  This issue does not cover HAMT / AMT parameter tuning and performance improvements which will also have a positive impact on this work.

ZenGround0:
Notes from conversation with @Stebalien 

Gas trace used in investigation is on ipfs at: `QmTqQES1LANjysAfKsNrAQUT3Qmb2ofB1Dgc7rpeZXXMFH`

PublishStorageDeals (PSD) gas cost is dominated by many series of Ipld Puts without a single culprit. The biggest single gas charge is signature verification but this is less that 1/100 of gas charge on typical PSD messages.

-- Synergy with existing datastructures work -- 
We expect to see some small improvements from fixing the HAMT to avoid duplicate writes but probably not on the order of 40-50% max possible savings due to the access patterns. Need to measure this once we have improved HAMT.  Also removing duplicate writes on AMT/HAMT is needed to unlock the benefits of using existing PSD batching because otherwise we will incur duplicate writes for individual deals.

-- New Solutions -- 
1. PendingProposals stores the proposal data (empirically ~153 byte values) but it is not needed.  We should just use cid keys.
2. DealOpsByEpoch
Good ideas:
  a. The HAMT HAMT structure is not ideal, it should be a HAMT of AMTs because we can pack items sequentially: key is array index, value is deal ID.  While cbor array would save space favor AMT over cbor array so that adding new ids doesn't require loading the whole set of deal ids.
  b. We should tune the inner AMT's width based on size of objects
  c. Add deals to DealOpsByEpoch in batches so that we load fewer of the inner AMTs.  This will increase variance in distribution of number of deals processed per epoch by X with X being the batch size which seems ok.  Complication: all deals in batch must be within one day away from the DOBE key they are placed at which constrains which deals can be put in a batch.

Less good / Discarded ideas:
[...]

PR #1316, opened 2020-12-01
use Set rather than HAMT[]Proposal for pending deals

acruikshank:
work towards #1305

### Motivation

The market actor records the hash of all pending proposals to prevent publishing duplicate proposals. The `PendingProposals` data structure is a `HAMT[cid]Proposal`. This data structure is used only to check set inclusion so the proposal value is never used. This is inefficient so this PR replaces the proposal with nil. Simulation suggests that this change will reduce the total data written and total data read from the store by ~23%.

### Proposed Changes

1. Replace proposal value with boolean in `market.State.PendingProposals`.
2. Remove invariant test that asserts hash of proposal stored in `PendingProposals` matches its key.

### Discussion

It is unclear whether this modification requires a migration. Since we never unmarshal the values from PendingProposals, having different types in the HAMT doesn't cause problems. It is probable that leaving Proposals in the data structure will add to the store throughput required by PublishStorageDeal. Proposals in PendingProposals should eventually be eliminated as the proposals are activated or expired. Work determining the precise effects of not migrating PendingProposals is ongoing.


Stebalien:
> migration

I'd just do it, honestly. The migration should be very fast and will avoid a weird intermediate state.

Stebalien:
> Replace proposal value with boolean in market.State.PendingProposals.

Nit: we can probably just use nil as the value (although I don't think that really makes a difference in practice).

acruikshank:
@Stebalien I though I tried storing nil and it didn't work, but I tried it again and I was wrong. So nil it is.

Kubuxu:
Bool is a single byte in CBOR, so I guess it does not matter.

actors/builtin/market/testing.go
```diff
@@ -153,21 +153,14 @@ func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount, c
...
-			pcid, err := pendingProposal.Cid()
-			if err != nil {
-				return err
-			}
-			acc.Require(pcid.Equals(proposalCID), "pending proposal's key does not match its CID %v != %v", pcid, proposalCID)
-
-			_, found := proposalCids[pcid]
-			acc.Require(found, "pending proposal with cid %v not fond within proposals %v", pcid, pendingProposals)
+			_, found := proposalCids[proposalCID]
+			acc.Require(found, "pending proposal with cid %v not fond within proposals %v", proposalCID, pendingProposals)
```
ZenGround0, line 163: super nit: not your bug but "fond" to "found"

ZenGround0 (commented):
LGTM, tag me again when its fully ready, I'm guessing after writing a migration

acruikshank:
@ZenGround0 added version 8 data migration with test.

actors/migration/nv8/market.go
```diff
@@ -0,0 +1,78 @@
+package nv7
```
anorth, line 1: This migration needs to come with network version 9. Version 8 is already live. Please add a new package for nv9.

actors/migration/nv8/market.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+
+	"github.com/filecoin-project/go-state-types/big"
+	cid "github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+
+	market "github.com/filecoin-project/specs-actors/v2/actors/builtin/market"
+	adt2 "github.com/filecoin-project/specs-actors/v3/actors/util/adt"
+
+	adt3 "github.com/filecoin-project/specs-actors/v3/actors/util/adt"
```
anorth, line 13: Some confused import aliases here. It is tricky to decide what to use, since there is no actual schema change happening. But make sure aliases match the paths.



Alternatively, I think you _could_ just import `/v3` for everything here.

actors/migration/nv8/market.go
```diff
@@ -0,0 +1,78 @@
...
+	return &StateMigrationResult{
+		NewHead:  newHead,
+		Transfer: big.Zero(),
+	}, err
+}
+
+func (a MarketMigrator) MapPendingProposals(ctx context.Context, store cbor.IpldStore, pendingProposalsRoot cid.Cid) (cid.Cid, error) {
+	oldPendingProposals, err := adt2.AsMap(adt2.WrapStore(ctx, store), pendingProposalsRoot)
+	if err != nil {
+		return cid.Cid{}, err
```
anorth, line 53: Use `cid.Undef`

actors/migration/nv8/top.go
```diff
@@ -0,0 +1,79 @@
+package nv7
```
anorth, line 1: `nv9` in a new package

anorth (commented):
A great find, I'm looking forward to this!

Looking good except for some confusion about the migration version.

anorth:
@acruikshank there is an `adt.Set` data structure that already implements this interpretation of a HAMT. I think it's worth using that for clarity.

actors/migration/nv8/top.go
```diff
@@ -0,0 +1,79 @@
...
+
+type MigrationInfo struct {
+	address    address.Address // actor's address
+	balance    abi.TokenAmount // actor's balance
+	priorEpoch abi.ChainEpoch  // epoch of last state transition prior to migration
+}
+
+type StateMigrationResult struct {
+	NewHead  cid.Cid
+	Transfer abi.TokenAmount
```
ZenGround0, line 32: This was quite a specialized artifact of the v4 migration and I think we can retire it until we need it again if ever.  Keeping the wrapping struct is nice to make extension beyond just returning the head really easy.

actors/migration/nv8/test/modify_pending_proposals_test.go
```diff
@@ -0,0 +1,226 @@
...
+	// In the same epoch, trigger cron to validate prove commit
+	// This activates deals hitting the pending deals structure.
+	vm3.ApplyOk(t, v3, builtin.SystemActorAddr, builtin.CronActorAddr, big.Zero(), builtin.MethodsCron.EpochTick, nil)
+
+	// Run cron after deal start to expire pending deals
+	v3, err = v3.WithEpoch(dealStart + 1)
+	require.NoError(t, err)
+	vm3.ApplyOk(t, v3, builtin.SystemActorAddr, builtin.CronActorAddr, big.Zero(), builtin.MethodsCron.EpochTick, nil)
+
+	// assert that power total invariants are broken
```
ZenGround0, line 134: This looks like out of date copy paste that can be fixed here and whereever you lifted it from

ZenGround0 (approved):
LGTM

actors/migration/nv8/market.go
acruikshank: I fixed the path for adt2 and changed the alias for market to market2, and introduced a market3 in the hopes of establishing a pattern that will be helpful when schema changes are required.

Issue #1313, opened 2020-12-01
Support different bitwidth configuration for different H/AMTs


anorth:
We currently have global constants for the bitwidth of each HAMT and AMT. But depending on value size and cardinality, different values will be optimal for different collections. Adjust the code so that each logical collection can use a different configuration.

anorth:
HAMT is done. Doing same for the AMT depends on upgrading our AMT dependency, which requires implementing the migration in order for existing migration tests to work.

PR #1320, opened 2020-12-03
Make Map bitwidth configurable for each instance

anorth:
We're likely to tune some HAMT bitwidths to values other than `5`. This is a fairly mechanical refactor to make that possible. It's a little noisy, some of the noise pointing to poor encapsualtion.

Partial satisfaction of #1313.

### DIFF
--- a/actors/builtin/market/set_multimap.go
+++ b/actors/builtin/market/set_multimap.go
@@ -14,23 +14,26 @@ import (
 )
 
 type SetMultimap struct {
-	mp    *adt.Map
-	store adt.Store
+	mp            *adt.Map
+	store         adt.Store
+	innerBitwidth int
 }
 
 // Interprets a store as a HAMT-based map of HAMT-based sets with root `r`.
-func AsSetMultimap(s adt.Store, r cid.Cid) (*SetMultimap, error) {
-	m, err := adt.AsMap(s, r)
+// Both inner and outer HAMTs are interpreted with branching factor 2^bitwidth.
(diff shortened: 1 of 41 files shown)


actors/migration/nv4/miner.go
```diff
@@ -11,6 +11,7 @@ import (
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"golang.org/x/xerrors"
 
+	builtin2 "github.com/filecoin-project/specs-actors/v3/actors/builtin"
```
anorth, line 14: This alias and path don't match, but I've made it consistent with what's there now. We'll either fix these all in bulk or, more likely, delete the `nv4` and `nv7` migrations from actors v3.

actors/builtin/market/set_multimap.go
```diff
@@ -14,23 +14,26 @@ import (
...
-	store adt.Store
+	mp            *adt.Map
+	store         adt.Store
+	innerBitwidth int
 }
 
 // Interprets a store as a HAMT-based map of HAMT-based sets with root `r`.
-func AsSetMultimap(s adt.Store, r cid.Cid) (*SetMultimap, error) {
-	m, err := adt.AsMap(s, r)
+// Both inner and outer HAMTs are interpreted with branching factor 2^bitwidth.
```
ZenGround0, line 23: In general these two should be allowed to differ.  I think it makes sense to enable that in this PR with another parameter

ZenGround0 (approved):

actors/builtin/market/set_multimap.go
anorth: Ok

PR #1321, opened 2020-12-04
Market actor accepts deals published from miner control addresses

### LINKED ISSUE #1317: Allow publishing deals using control addresses
Currently PublishStorageDeals requires deals to be published from miner worker address:

https://github.com/filecoin-project/specs-actors/blob/51482ab9f7aeb57bc778ad607046b04f41792cfa/actors/builtin/market/market_actor.go#L184-L185

We could also allow control addresses here to give miners more flexibility

anorth:
The market actor currently requires a deal to be published by the worker address of the deal provider (miner). This change also allows the miner's control addresses, which gives more flexibility to miners. The change is straightforward, and I refactored tests a little to support exercising it.

Closes #1317


ZenGround0 (approved):
LGTM

PR #1323, opened 2020-12-07
Actor tree traversal for migration nv9

anorth:
For the upgrade to actors v3 (network version 9) we expect to migrate most actors in order to upgrade H/AMT data structures. This PR copies the full state migration setup from the nv4 migration, but then simplifies it a fair bit.

- Removed handling of balance transfers and power claim recalculations
- Removed some layers of wrapping for migration objects, inputs, and outputs.

### DIFF
--- a/actors/migration/nv9/top.go
+++ b/actors/migration/nv9/top.go
@@ -2,77 +2,206 @@ package nv9
 
 import (
 	"context"
+	"fmt"
+	"sync"
 
 	address "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
-	"github.com/ipfs/go-cid"
+	cid "github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/xerrors"
 
-	"github.com/filecoin-project/specs-actors/v3/actors/builtin"
-	"github.com/filecoin-project/specs-actors/v3/actors/states"
-	"github.com/filecoin-project/specs-actors/v3/actors/util/adt"
+	builtin2 "github.com/filecoin-project/specs-actors/v2/actors/builtin"
+	states2 "github.com/filecoin-project/specs-actors/v2/actors/states"
+
+	builtin3 "github.com/filecoin-project/specs-actors/v3/actors/builtin"
+	states3 "github.com/filecoin-project/specs-actors/v3/actors/states"
+	adt3 "github.com/filecoin-project/specs-actors/v3/actors/util/adt"
 )
 
 // Config parameterizes a state tree migration
-type Config struct{}
-
-func DefaultConfig() Config {
-	return Config{}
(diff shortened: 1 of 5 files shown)


actors/migration/nv9/top.go
```diff
@@ -2,77 +2,206 @@ package nv9
...
+	for i := 0; i < cfg.MaxWorkers; i++ {
+		workerWg.Add(1)
+		grp.Go(func() error {
+			defer workerWg.Done()
+			for input := range inputCh {
+				result, err := migrateOneActor(ctx, store, input, priorEpoch)
+				if err != nil {
+					return err
+				}
+				if result == nil { // Actor is deferred
```
ZenGround0, line 134: This appears to now be handled earlier in the pipeline by never sending the deferred input to the worker thread so this carry over from the previous solution should now never be reached.  It might be worth throwing an error in this case, but maybe it's better to just remove the if statement and let the result processing panic in the unexpected event that a migration returns a nil result.

actors/migration/nv9/top.go
```diff
@@ -2,77 +2,206 @@ package nv9
...
+	})
 	if err != nil {
-		return cid.Undef, err
+		return nil, xerrors.Errorf("state migration failed for %s actor, addr %s: %w", builtin2.ActorNameByCode(actorIn.Code), addr, err)
 	}
 
-	return actorsIn.Flush()
+	// Set up new actor record with the migrated state.
+	return &migrationResult{
+		addr,
```
ZenGround0, line 199: nit: `// unchanged` here too would make sense

ZenGround0 (approved):
Looking good with a few minor suggestions.

Forgot to mention this on the PR removing v2 migrations but we should reinstate a ported version of [this test](https://github.com/filecoin-project/specs-actors/pull/1324/files#diff-79c7357c63d480af38d07c6ec260ac4286691b03b8dd4db276d9575007a4cfcdL30) so that the race detector can help us catch issues.

actors/migration/nv9/top.go
```diff
@@ -2,77 +2,206 @@ package nv9
...
+	var deferredCodeIDs = map[cid.Cid]struct{}{
+		// None
+	}
+	if len(migrations)+len(deferredCodeIDs) != 11 {
+		panic(fmt.Sprintf("incomplete migration specification with %d code CIDs", len(migrations)))
+	}
+
+	// Load input and output state trees
+	adtStore := adt3.WrapStore(ctx, store)
+	actorsIn, err := states2.LoadTree(adtStore, stateRootIn) // FIXME actorsRootIn?
```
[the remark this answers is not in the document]
anorth: Changed. Another question is whether we should in fact take that real state root object here. I'd rather like to, but it's not critical here.

actors/migration/nv9/top.go
anorth: Removed the check

anorth:
Ported and added `parallel_migration_test.go`

PR #1327, opened 2020-12-10
Implement off-chain Window PoSt verification

Stebalien:
Proposal: https://github.com/filecoin-project/FIPs/issues/42

# TODO

* [x] Get the tests/lints passing.
* [x] Test (or retain tests) for SubmitWindowPoSt method
  * [x] Skipping all sectors should prevent submission
* [x] Test the DisputeWindowedPoSt method.
  * [x] With a correct post.
  * [x] With a bad post.
  * [x] After compaction.
  * [x] While the deadline is open.
  * [x] Test with a sector terminated after a post is submitted but before the end of the deadline.
* [ ] ~(?) Refactor RecordProvenSectors to simplify it.~
* [x] Break most of DisputeWindowedPoSt into a deadline-state method.
* [ ] Decide on the final penalties for falsifying proofs (@zixuanzh?)
* [x] Audit for and address remaining TODOs in the code.

# Changes

In addition to the changes below, 

## Change: SubmitWindowedPoSt

Proofs that do not recover faulty power are not checked on-chain. Instead,

* The proof is recorded in an AMT of "optimistic" proofs.
* The proof is not verified.
* The proof may be "disputed" by anyone for 1800 epochs after the challenge window ends.

As usual, skipped sectors are marked as faulty and "unproven" sectors are marked as "proven".

Notes:

* If a proof _recovers_ faulty power, it is checked on-chain immediately to prevent "recovering" faulty sectors with invalid proofs.
* "Unproven" sectors (sectors newly added to a partition but never proven with a window PoSt) do not force an on-chain proof verification. This is intentional.

## Change: Deadline End

As usual, partitions without proofs are marked as faulty, "new" (unproven) sectors are activated if proofs for them are present, etc.

Immediately after processing all proof related logic, and before terminating any sectors:

* We record a snapshot of the partition AMT so proofs can be properly disputed later (`PartitionsSnapshot`)
* We record a snapshot of the proofs AMT (`ProofsSnapshot`).

We then proceed to terminate sectors as usual.

## Change: CompactPartitions

As `CompactPartitions` can move sectors between partitions, it makes it difficult mark these sectors as faulty when disputed. While the disputer could submit bitfields describing the new locations of the sectors in question, this dispute could be defeated by repeatedly re-compacting.

To simplify the prevent this attack and simplify the disputer's job, partition compaction is now prohibited for WPoStProofDisputePeriod (currently 2x finality or 15 deadlines) epochs after the deadline's last challenge window. This leaves a window of 960 epochs (8 hours) for partition compaction.

Proofs may not be disputed after this time has passed.

**Questions**

* [x] Is the Finality blackout correct? We _can_ make it longer (but probably not shorter). 2x finality would be safer.
  * I've made it 2x finality which should be much safer.

## New: DisputeWindowedPoSt

This patch introduces a new method: `DisputeWindowedPoSt`. This method selects a proof from the ProofsSnapshot AMT and verifies it.

0. The epoch and deadline are checked to make sure that the window for disputing proofs is still open.
1. The proof is verified against the partition state snapshot from the end of the last challenge window. This represents the state that was (supposedly) proven.
2. If the proof validates, this method aborts with no change. Otherwise, we continue.
3. The active sectors in the partition snapshot are declared as faulty. Terminated/moved sectors are left alone. The faulty termination epoch is set as if the sectors became faulty at the end of the _previous_ challenge window for the deadline in question.
4. The target miner looses power based on the sectors marked faulty in step 3. If a sector moves, the power will be retained.
5. The disputed proof is removed from the proofs AMT so it can't be disputed twice.
6. The target miner is fined a penalty based on the active power in the partitions that were incorrectly prove. This "active power" is taken from the partition snapshot.

**Questions**

* [x] Should we allow the miner to specify a sector/partition mapping to work around the case where a sector can be moved due to compaction? I'm concerned that this could lead to large, expensive disputes.
  * We're limiting disputes to the a dispute window.
* [ ] What should the penalty be? (to be determined by FIP)
* [ ] What should the reporter be paid, if anything?
* [x] Do we really need to snapshot the sectors AMT? I was concerned about compaction deleting sectors. However, we could also just make the window in which miners can compact a deadline disjoint from the window where proofs can be disputed. ~This might require extending the dispute window a bit (e.g., 2x finality)~ *(done)*.
  * We're limiting disputes to a dispute window.


Stebalien:
It looks like this is causing us to kill (OOM?) the migration test. Not sure what that's about.

Stebalien:
Turns out "immutable" deadlines aren't quite so immutable because we still allow sector termination. This is moderately annoying to fix because we can terminate sectors from cron (e.g., when terminating too many). We can fix the issue for all practical purposes, but getting a fully correct fix could be annoying.

A simple fix is to snapshot the partitions/sectors along with the proofs, instead of at the end of the deadline. Honestly, this is the "better" approach in terms of correctness. Unfortunately, while it will only marginally increase gas costs (proofs need 2 extra CIDs) it'll increase the size of the state tree by quite a bit. On the other hand, I'm not sure how much that matters.

One of the significant upsides here is that the "WindowPoSt" objects are now completely self-contained.

Thoughts?

```go
type WindowedPoSt struct {
    ProvenPartitions bitfield.BitField
    PartitionsSnapshot cid.Cid // new
    SectorsSnapshot cid.Cid // new
    Proofs []proofs.PoStProof
}
```

Stebalien:
Ah, wait. I'm confusing myself.

We do two things:

1. Actually terminate sectors.
2. Process early terminations.

1 is triggered from TerminateSectors and at deadline end. This is the part that actually _marks_ sectors as dead.

2 can be triggered from cron, but just pays fines.

TL;DR: all we need to do is block manual calls to TerminateSectors during the "blackout" period. That's pretty reasonable.

Stebalien:
> TL;DR: all we need to do is block manual calls to TerminateSectors during the "blackout" period. That's pretty reasonable.

Done. This also fixes the incentive mismatch where a logical miner will wait until missing data is first challenged to terminate a sector.

Stebalien:
Next step: migration.

This should be fairly straight-forward.

1. If the PoSt is submitted before the upgrade, it'll be verified and the partitions will be marked proven.
2. If the PoSt is submitted after the upgrade, it'll be stored.

The tricky part is that the "all proven sectors have proofs in the proofs AMT" invariant won't be true during this period. I was hoping we could use this as a global invariant but life just isn't that simple.

Stebalien:
Hm. Looks like the migration isn't fully implemented yet anyways. @anorth, should I pick that up? Or just move on for now.

anorth:
Re compaction and finality, could we simplify to get these windows completely non-overlapping? E.g. challenge window of 18hrs, compaction window of 6 hrs? It would be more robust of course to allow complete overlapping, but it doesn't look like that can be done without being quite complex. So if we can't do that, I think it might be worth removing the possibility of collision between those two actions ("sectors that have moved partitions will not be marked faulty, the miner will still be fined"). Edit: I see you mentioned that idea later.

Re migration, correct that's a work in progress. It's probably better not to attempt it in this PR, but note for follow-up.

actors/builtin/miner/miner_actor.go
```diff
@@ -487,6 +467,186 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			//    You can run but you can't hide.
+
+			// Load sectors. We can't use the snapshot as a sector
+			// with an extended expiration could end up being
+			// scheduled to expire at the wrong time.
+			//
+			// Unfortunately, this means we'll need to load all
+			// sector infos twice (once for the proof, once to mark
+			// faulty). We can't use memoized power info because
+			// sectors may have been terminated.
```
anorth, line 602: If we somehow avoid needing to load the current sector info to handle extensions, we could possibly memoize some other information in order to avoid recalculating power. E.g. we could store the terminated power since last challenge window in each deadline/partition.

actors/builtin/miner/miner_actor.go
```diff
@@ -487,6 +467,186 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			sectors, err := LoadSectors(store, st.Sectors)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors array")
+			faultExpirationEpoch := targetDeadline.Last() + FaultMaxAge
+			powerDelta, err = dl.DeclareFaults(store, sectors, info.SectorSize, QuantSpecForDeadline(targetDeadline), faultExpirationEpoch, faults)
```
anorth, line 606: `DeclareFaults` is an odd method to call here. I would expect something more like what happens in deadline cron when a PoSt is missed, which eventually calls `partition.RecordMissedPost()`.

If `DeclareFaults` really does the right thing for both this and ahead-of-time declaration, let's rename it to `RecordFaults` or similar.

actors/builtin/miner/miner_actor.go
```diff
@@ -487,6 +467,186 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+		// TODO: move into deadline state function.
+		{
+			// Load the target state.
+			deadlines, err := st.LoadDeadlines(store)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadlines")
+
+			dl, err := deadlines.LoadDeadline(store, params.Deadline)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline")
+
+			proofs, err := adt.AsArray(store, dl.ProofsSnapshot)
```
anorth, line 512: Snapshot vs current state vars will be easily confused, so I suggest naming quite explicitly:
- `dlCurrent`
- `proofsSnapshot`
- etc

actors/builtin/miner/miner_actor.go
```diff
@@ -1465,9 +1630,9 @@ func (a Actor) CompactPartitions(rt Runtime, params *CompactPartitionsParams) *a
 		info := getMinerInfo(rt, &st)
 		rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
 
-		if !deadlineIsMutable(st.ProvingPeriodStart, params.Deadline, rt.CurrEpoch()) {
+		if !deadlineCanCompact(st.ProvingPeriodStart, params.Deadline, rt.CurrEpoch()) {
```
anorth, line 1633: From my discussion at the top level, I'd like to unify these again if it's reasonable, with a longer immutability period.

actors/builtin/miner/miner_actor.go
```diff
@@ -2157,8 +2322,10 @@ func verifyWindowedPost(rt Runtime, challengeEpoch abi.ChainEpoch, sectors []*Se
 	}
 
 	// Verify the PoSt Proof
-	if err = rt.VerifyPoSt(pvInfo); err != nil {
-		rt.Abortf(exitcode.ErrIllegalArgument, "invalid PoSt %+v: %s", pvInfo, err)
+	if err = rt.VerifyPoSt(pvInfo); err == nil {
+		rt.Abortf(exitcode.ErrIllegalArgument, "valid PoSt %+v", pvInfo)
+	} else {
+		rt.Log(rtt.WARN, "PoSt successfully challenged: %s", err)
```
anorth, line 2328: This will almost always be someone else's proof, so `WARN` is a bit loud IMO.

gen/gen.go
```diff
@@ -179,6 +179,8 @@ func main() {
 		miner.WorkerKeyChange{},
 		miner.VestingFunds{},
 		miner.VestingFund{},
+		miner.WindowedPoSt{},
+		miner.ChallengeWindowedPoStParams{},
```
anorth, line 183: Nit: move params down to 201

anorth (commented):
Thanks, this is looking good.

I trust that all the TODOs will either be resolved here, or have an issue filed, or be rewritten as commentary about alternative possibilities, before merging.

Stebalien:
> Re compaction and finality, could we simplify to get these windows completely non-overlapping? E.g. challenge window of 18hrs, compaction window of 6 hrs? It would be more robust of course to allow complete overlapping, but it doesn't look like that can be done without being quite complex. So if we can't do that, I think it might be worth removing the possibility of collision between those two actions ("sectors that have moved partitions will not be marked faulty, the miner will still be fined"). Edit: I see you mentioned that idea later.

Yes, we can do this. And it's probably the simplest approach (and lets us not snapshot sectors).

actors/builtin/miner/miner_actor.go
Stebalien: Unfortunately, we can't do this easily. That would prevent us from assigning sectors to partitions in that window which would mean opening even _more_ partitions, which would be very painful for small miners.

* Adding a sector to a partition doesn't affect window PoSt verification/fees/faults.
* Compaction does.

actors/builtin/miner/miner_actor.go
Stebalien: It does the same thing. `RecordMissedPost` doesn't call this because `RecordMissedPost` doesn't know which sectors are faulty (it just marks the entire partitions faulty).

But I agree on the rename.

actors/builtin/miner/miner_actor.go
```diff
@@ -351,10 +351,8 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 	}
 
 	partitionIndexes := bitfield.New()
-	if nv >= network.Version7 {
-		for _, partition := range params.Partitions {
-			partitionIndexes.Set(partition.Index)
-		}
+	for _, partition := range params.Partitions {
+		partitionIndexes.Set(partition.Index)
```
wadeAlexC, line 355: This is now unused in this method, but this step is repeated within `deadline.RecordProvenSectors`

Stebalien: Ah, thanks for the reminder. I wanted to avoid checking this via `IsSet` because that method is kind of slow.

actors/builtin/miner/miner_actor.go
Stebalien: If we make the "compaction" and "challenge" windows disjoint, we can get rid of the sectors snapshot. If we get rid of the sectors snapshot, then determining the power from the sectors is effectively "free" because we already need to load them for the proof verification.

actors/builtin/miner/deadline_state.go
```diff
@@ -880,14 +926,29 @@ func (p *PoStResult) PenaltyPower() PowerPair {
...
+	// First check to see if we're proving any already proven partitions.
+	// This is faster than checking one by one.
+	if alreadyProven, err := bitfield.IntersectBitField(dl.PartitionsPoSted, partitionIndexes); err != nil {
+		return nil, xerrors.Errorf("failed to check proven partitions: %w", err)
+	} else if empty, err := alreadyProven.IsEmpty(); err != nil {
+		return nil, xerrors.Errorf("failed to check proven intersection is empty: %w", err)
+	} else if !empty {
+		return nil, xc.ErrIllegalArgument.Wrapf("partition already proven: %v", alreadyProven)
+	}
+
```
wadeAlexC, line 951: By moving this check outside of the loop over `postPartitions`, it becomes possible for `postPartitions` to contain duplicate indices (though `post.skipped` may not be duplicate!).

I couldn't find any route to this actually messing up accounting, but there are so many options to think about. At a minimum, the tests needed to effectively cover this new behavior would be quite extensive.

I would recommend moving this back in the loop below, or otherwise explicitly disallowing the miner submitting duplicate indices.

actors/builtin/miner/miner_state.go
```diff
@@ -1118,6 +1118,9 @@ func (st *State) AdvanceDeadline(store adt.Store, currEpoch abi.ChainEpoch) (*Ad
 
 	// No live sectors in this deadline, nothing to do.
 	if deadline.LiveSectors == 0 {
+		// TODO: do we still need to clear the post submissions here? I
+		// think we're technically fine, but this is a strange
+		// edge-case.
```
wadeAlexC, line 1123: I think yes, they should be cleared. 

It's a strange edge case that can be directly reached by a miner, simply by terminating a deadline's sectors. Unclear how this may interact with `ChallengeWindowedPoSt`, but I don't think we want the old snapshots to stick around!

actors/builtin/miner/deadline_state.go
Stebalien: Hm. You're right. I think we at least need to make sure there aren't duplicates. Those aren't allowed anyways.

wadeAlexC:
Seems like `ExtendSectorExpiration` can be used between submission of a PoSt and its EOD cron. Since this modifies deadline/partition state, I feel like this should be disallowed similar to the `deadlineIsMutable` check in `TerminateSectors`.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +461,185 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+				if found, err := partitionsSnapshot.Get(partIdx, &partitionSnapshot); err != nil {
+					return err
+				} else if !found {
+					return exitcode.ErrIllegalState.Wrapf("failed to find partition when challenging proof %d", params.ProofIndex)
+				}
+
+				// Record sectors for proof verification
+				allSectors = append(allSectors, partitionSnapshot.Sectors)
+				allIgnored = append(allIgnored, partitionSnapshot.Faults)
+				allIgnored = append(allIgnored, partitionSnapshot.Terminated)
```
wadeAlexC, line 533: Because of the logic in `ConfirmSectorProofsValid`, `partitionSnapshot` should absolutely NOT have any `Unproven` sectors.

On the off chance it does, shouldn't they be ignored?

Stebalien:
> Seems like ExtendSectorExpiration can be used between submission of a PoSt and its EOD cron. Since this modifies deadline/partition state, I feel like this should be disallowed similar to the deadlineIsMutable check in TerminateSectors.

I think that may cause more problems than it solves. `TerminateSectors` and `CompactPartitions` are pretty rare and likely to be manually invoked. I'd expect `ExtendSectorExpiration` to be called from some form of automation.

In terms of impact, applying "extend sector expiration" before or after the challenge window starts should be indistinguishable from the perspective of off-chain proofs. Really, the only thing we care about is that the sectors to be proven during the challenge window shouldn't change.

Stebalien:
Note: same applies to CC upgrades. A sector can be made to expire at the end of a deadline due to a CC upgrade. However, that's fine because we'll process the expiration _after_ we finish processing everything else.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +461,185 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			//
+			// Unfortunately, this means we'll need to load all
+			// sector infos twice (once for the proof, once to mark
+			// faulty). We can't use memoized power info because
+			// sectors may have been terminated.
+			sectors, err := LoadSectors(store, st.Sectors)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors array")
+			faultExpirationEpoch := targetDeadline.Last() + FaultMaxAge
+			powerDelta, err = dlCurrent.RecordFaults(store, sectors, info.SectorSize, QuantSpecForDeadline(targetDeadline), faultExpirationEpoch, faults)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to declare faults")
```
wadeAlexC, line 600: If `faults` contains a partition index that no longer exists due to compaction, this will return an error and abort here.

I understand that the pause for `ChainFinality` in `CompactPartitions` exists to address this, but if this doesn't cut it, the fraudulent miner can get away without being penalized.

wadeAlexC: The same goes for the case where the partition in question no-longer contains the faulty sectors.

> 2. Successful challenges will always penalize, even if we can't find the target sectors and mark them faulty.

Is the above comment implemented somehow?

wadeAlexC:
If I understand correctly, on successful challenge, the miner is only penalized 1 proving-period worth?

Doesn't this mean it's worth it to submit fraudulent PoSts, because you gain block rewards on the increased power between the proving epoch and challenge epoch?

actors/builtin/miner/miner_actor.go
Stebalien: Ah, I see. I was thinking about sector numbers, but you're right about partition indices. I'm just going to make compaction exclusive with the challenge window.

Stebalien:
> Doesn't this mean it's worth it to submit fraudulent PoSts, because you gain block rewards on the increased power between the proving epoch and challenge epoch?

The penalty for one proving period is `3.51*ExpectedBlockReward`. You're right in that we may want to double that in this case, but that's a question for @zixuanzh. I'll add a TODO.

Stebalien:
I've now restricted challenges to the 900 epoch period following the wpost challenge window close (side note: these two "challenge windows" are going to get confusing). We may want to increase this to 2x finality.

Due to this change, I was able to remove the sectors AMT snapshot.

actors/builtin/miner/miner_actor.go
Stebalien: You're right, they should. It won't hurt much to just add them in.

actors/builtin/miner/miner_actor.go
```diff
@@ -487,6 +467,186 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+	epochReward := requestCurrentEpochBlockReward(rt)
+	pwrTotal := requestCurrentTotalPower(rt)
+
+	toBurn := abi.NewTokenAmount(0)
+	pledgeDelta := abi.NewTokenAmount(0)
+	powerDelta := NewPowerPairZero()
+	var st State
+	rt.StateTransaction(&st, func() {
+		if st.CurrentDeadline == params.Deadline {
+			// TODO: blank out the previous deadline as well?
```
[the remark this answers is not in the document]
Stebalien: I've now created non-overlapping windows:

```
                        (no sectors assigned)
                 |<--blackout period------------------>|
|<--compaction-->|<--prev window-->|<--target window-->|<---proof challenge window--->|...
  (8hrs)                 ^                                  (2x finality)
                 challenge epoch
```

actors/builtin/miner/miner_state.go
Stebalien: So, I don't think this can happen anymore (after this patch). Given this patch:

1. I can't "prove" no sectors. We reject those proofs.
2. I can't terminate sectors during the challenge window (or the previous deadline).

Given this, I shouldn't _need_ to touch this array, right?

wadeAlexC: I think it should be done for safety, regardless. If the snapshots aren't reset each EOD cron, they may be technically valid when submitted to `ChallengeWindowedPoSt`

Stebalien: I guess I can just check without spending too much time. That, and log an error.

Stebalien:
As pointed out by @wadeAlexC, sectors can be "indefinitely" extended under the current implementation by repeatedly submitting bad proofs. While this would likely be very expensive, it's still something we should avoid.

Proposal: whenever _recovering_ faulty sectors, require on-chain proof verification.

One nice side effect of this proposal is that, if chain throughput ever rate-limits proof _challenges_, that rate-limit should apply equally to proofs that restore power. That way, an attacker should not be able to restore power faster than it can be challenged (assuming the submitter is refunded gas somehow).

Stebalien:
The above has now been implemented.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +472,172 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+
+	requestUpdatePower(rt, powerDelta)
+	notifyPledgeChanged(rt, pledgeDelta)
+	burnFunds(rt, toBurn)
+	rt.StateReadonly(&st)
+
+	err := st.CheckBalanceInvariants(rt.CurrentBalance())
+	builtin.RequireNoErr(rt, err, ErrBalanceInvariantBroken, "balance invariants broken")
+
+	// TODO: Pay submitter some fee.
```
anorth, line 637: Remove TODO pending a spec.

actors/builtin/miner/policy.go
```diff
@@ -196,6 +196,11 @@ const DealLimitDenominator = 134217728 // PARAM_SPEC
 // for permissioned actor methods and winning block elections.
 const ConsensusFaultIneligibilityDuration = ChainFinality
 
+// PoStChallengePeriod is the period after a challenge window ends during which
+// PoSts submitted during that period may be challenged.
+// TODO: this name can easily be confused with WPoStChallengeWindow.
+const WPoStProofChallengePeriod = 2 * ChainFinality // PARAM_TODO
```
anorth, line 202: Please add an `init` assertion that this is less than the deadline window.

Edit: I see you added a test instead. We already have some assertions above. I think it makes sense to keep them together, and in this file. But centralising them in either place is ok.

actors/builtin/miner/testing.go
```diff
@@ -201,14 +201,30 @@ func CheckDeadlineStateInvariants(deadline *Deadline, store adt.Store, quant Qua
...
+		acc.Require(partition.UnprovenPower.IsZero(), "snapshot partition has unproven power")
+		if noUnproven, err := partition.Unproven.IsEmpty(); err != nil {
+			acc.Addf("error counting unproven: %v", err)
+		} else {
+			acc.Require(noUnproven, "snapshot partition has unproven sectors")
+		}
+
+		return nil
+	})
+	acc.RequireNoError(err, "error iterating partitions snapshot")
```
anorth, line 227: There must be some more invariants about the post submissions, and things that could be broken if compaction and termination blackout windows were not effective.

actors/migration/nv9/miner.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	builtin3 "github.com/filecoin-project/specs-actors/v3/actors/builtin"
 	miner3 "github.com/filecoin-project/specs-actors/v3/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/v3/actors/util/adt"
```
anorth, line 14: In migrations, alias each versioned package to avoid errors using the wrong one.

anorth (commented):
I've reviewed everything except tests and it looks good – great work and many thanks. I left a lot of comments, but they're mainly structure and clarity things to ensure highest confidence in a complex change.

I've nitpicked documentation and naming to ensure my and others' understanding is correct.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +473,177 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			//
+			// However, some of these sectors may have been
+			// terminated. That's fine, we'll skip them.
+			faultExpirationEpoch := targetDeadline.Last() + FaultMaxAge
+			powerDelta, err = dlCurrent.RecordFaults(store, sectors, info.SectorSize, QuantSpecForDeadline(targetDeadline), faultExpirationEpoch, disputedSectors)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to declare faults")
+
+			// Delete disputed proof so it can't be charged multiple times.
+			err = proofsSnapshot.Delete(params.PoStIndex)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to delete disputed proof")
```
wadeAlexC, line 608: I assume this also shifts the array. May be worth mentioning that if 2 posts are disputable, this causes a race condition where a disputer may need to wait for a previous dispute to be processed before submitting the next dispute.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +473,177 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			rt.Log(rtt.INFO, "successfully disputed: %s", err)
+
+			// Ok, now we record faults. This always works because
+			// we don't allow compaction/moving sectors during the
+			// challenge window.
+			//
+			// However, some of these sectors may have been
+			// terminated. That's fine, we'll skip them.
+			faultExpirationEpoch := targetDeadline.Last() + FaultMaxAge
+			powerDelta, err = dlCurrent.RecordFaults(store, sectors, info.SectorSize, QuantSpecForDeadline(targetDeadline), faultExpirationEpoch, disputedSectors)
```
wadeAlexC, line 603: `RecordFaults` iterates over each partition that had active sectors during the snapshot. It records faults for each of these partitions. 

If ANY of these partitions don't have any new faults (due to terminations, for example), the entire method will fail.

actors/builtin/miner/miner_actor.go
wadeAlexC: Reference: https://github.com/filecoin-project/specs-actors/blob/d5288c22c9f53a514e72260040cde81667a4e791/actors/builtin/miner/deadline_state.go#L785-L790

wadeAlexC (changes requested):
First pass over `SubmitWindowedPoSt` and `DisputeWindowedPoSt`.

Still working to cross-reference against `DeclareFaults`, `DeclareFaultsRecovered`, `TerminateSectors`, and `handleProvingDeadline`

actors/builtin/miner/miner_actor.go
wadeAlexC: Reference: https://github.com/filecoin-project/specs-actors/blob/d5288c22c9f53a514e72260040cde81667a4e791/actors/builtin/miner/deadline_state.go#L785-L790

actors/builtin/miner/deadlines.go
```diff
@@ -66,3 +67,26 @@ func deadlineIsMutable(provingPeriodStart abi.ChainEpoch, dlIdx uint64, currentE
...
+// Deadlines may not be compacted when:
+//
+// 1. The deadline is currently being challenged.
+// 2. The deadline is to be challenged next.
+// 3. Optimistically accepted posts from the deadline's last challenge window
+//    can currently be disputed.
+func deadlineAvailableForCompaction(provingPeriodStart abi.ChainEpoch, dlIdx uint64, currentEpoch abi.ChainEpoch) bool {
+	return deadlineIsMutable(provingPeriodStart, dlIdx, currentEpoch) &&
+		!deadlineAvailableForOptimisticPoStDispute(provingPeriodStart, dlIdx, currentEpoch)
+}
```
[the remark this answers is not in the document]
Stebalien: I started out with this as one big inline function, but I really want to define these constraints in terms of the other constraints if possible. It reduces the changes of weird off-by-one errors (e.g., `<` versus `<=`).

But I'll see if I can simplify.

actors/builtin/miner/miner_actor.go
Stebalien: It just makes a hole, nothing is shifted.

actors/builtin/miner/miner_actor.go
Stebalien: Unless I'm misreading the code, I think it fails if `IsEmpty` fails, not if it's empty. If it's empty, it just doesn't record the partition as having faults.

actors/builtin/miner/policy.go
Stebalien: I've moved these constraints into the init function.

actors/builtin/miner/testing.go
Stebalien: So, unfortunately, I don't know the current epoch from this function. However, I could just clear the partitions/proofs snapshots on compaction. That's probably the most correct approach.

actors/builtin/miner/miner_actor.go
wadeAlexC: Oh, whoops! You're right, I misread that.

wadeAlexC:
In `SubmitWindowedPoSt`, is `params.Proofs[0]` checked for a maximum length? If not, it should be.

wadeAlexC:
I think we can eliminate the need for snapshots in cron by storing more info in `OptimisticPoStSubmissions` during `SubmitWindowedPoSt`. The additional info would contain the sectors to load for the proof, as well as the partitions' active sectors / power at the time `SubmitWindowedPoSt` was called. This info would be hashed and stored in `OptimisticPoStSubmissions`.

When disputing a post, the disputer can calculate the info themselves and provide it as parameters when calling `DisputeWindowedPoSt`. The data they provide can be hashed and checked against the stored submission in `OptimisticPoStSubmissions`.

My idea is to replace the current `DisputeWindowedPoStParams` with the following:

```go
type PoStInfo struct {
  AllSectorNos, IgnoredSectorNos bitfield.BitField // Can be passed directly into sectors.LoadForProof
  DisputedSectors                PartitionSectorMap // A map of active sectors when SubmitWPoSt was called
  DisputedPower                  PowerPair // The total power of the partitions when SubmitWPoSt was called
  Proofs                         []proof.PoStProof // The proof itself
}

type DisputeWindowedPoStParams struct {
  Deadline    uint64
  PoStIndex   uint64
  DisputeInfo PoStInfo
}
```

Optionally, epoch reward and power total could also be stored in `PoStInfo`.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +475,158 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to determine post proof type")
+			rewardTarget := RewardForDisputedWindowPoSt(postProofType, disputedPower)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to compute reward for disputed window post")
+
+			// Compute the target penalty by adding the
+			// base penalty to the target reward. We don't
+			// take reward out of the penalty as the miner
+			// could end up receiving a substantial
+			// portion of their fee back as a reward.
+			penaltyTarget := big.Add(penaltyBase, rewardTarget)
```
Stebalien, line 590: This is _not_ how consensus faults work, but I believe it's the "more correct" approach.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +475,158 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			toReward = big.Min(toBurn, rewardTarget)
+			toBurn = big.Sub(toBurn, toReward)
+
+			pledgeDelta = penaltyFromVesting.Neg()
+		}
+	})
+
+	requestUpdatePower(rt, powerDelta)
+
+	if !toReward.IsZero() {
```
Stebalien, line 609: This check isn't present in consensus faults reports and we may want to change that.

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +475,158 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+		// Try to send the reward to the reporter.
+		code := rt.Send(reporter, builtin.MethodSend, nil, toReward, &builtin.Discard{})
+
+		// If we fail, log and burn the reward to make sure the balances remain correct.
+		if !code.IsSuccess() {
+			rt.Log(rtt.ERROR, "failed to send reward")
+			toBurn = big.Add(toBurn, toReward)
```
Stebalien, line 616: This "burn it" isn't present in consensus fault reporting, but it's necessary to ensure that pledge values remain consistent.

Alternatively, we _could_ just fail the message. Honestly, _not_ failing the message is a bit of a footgun as lotus will think "the message succeeded" and will just send it to the network instead of kicking it back to the user to ensure the user gets paid.

Stebalien:
> In SubmitWindowedPoSt, is params.Proofs[0] checked for a maximum length? If not, it should be.

This is done.

Stebalien:
> My idea is to replace the current DisputeWindowedPoStParams with the following:

Discussed out of band: We can do this, but it would require changing maximum bitfield sizes. They're currently sized to fit up to a partition's worth of sector numbers worst-case, not more. This is not a hard change, it's just a bit tricky to make sure we don't mess it up somewhere.

I've also filed a proposal for the next upgrade to make this impossible: https://github.com/filecoin-project/specs-actors/issues/1352

TL;DR: I'm going to leave things as they are unless we have a compelling reason to do otherwise.

actors/builtin/miner/testing.go
anorth: The epoch can be plumbed here if helpful (it was used in a prior migration/validation).

But I see what you mean, it's hard to assert more without knowing about the window. I'm nervous about clearing the snapshots: code will probably assume they exist, and if they can fail to exist then everywhere should check for and handle that, even if we think that epoch windows should prevent ever discovering them missing.

Ok to leave this for now.

actors/builtin/miner/deadlines.go
```diff
@@ -66,3 +66,14 @@ func deadlineIsMutable(provingPeriodStart abi.ChainEpoch, dlIdx uint64, currentE
 	// that deadline opens.
 	return currentEpoch < dlInfo.Open-WPoStChallengeWindow
 }
+
+func deadlineAvailableForChallenge(provingPeriodStart abi.ChainEpoch, dlIdx uint64, currentEpoch abi.ChainEpoch) bool {
+	dlInfo := NewDeadlineInfo(provingPeriodStart, dlIdx, currentEpoch).NextNotElapsed()
+
+	return !dlInfo.IsOpen() && currentEpoch < (dlInfo.Close-WPoStProvingPeriod)+WPoStProofChallengePeriod
```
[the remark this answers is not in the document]
Stebalien: I'll make this change, but I still feel like this is an awkward edge-case. For example, deadlines with _no_ sectors still have "dispute" windows.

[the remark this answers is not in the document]
Stebalien: And, immediately after the proving period starts, someone can still dispute deadlines before the proving period started.

actors/builtin/miner/miner_actor.go
anorth: I'd rather the report succeeded even if we can't pay the reward. Burning it sounds good.

actors/builtin/miner/policy.go
```diff
@@ -196,6 +224,10 @@ const DealLimitDenominator = 134217728 // PARAM_SPEC
 // for permissioned actor methods and winning block elections.
 const ConsensusFaultIneligibilityDuration = ChainFinality
 
+// WPoStDisputeWindow is the period after a challenge window ends during which
+// PoSts submitted during that period may be disputed.
+const WPoStDisputeWindow = 2 * ChainFinality // PARAM_TODO
```
anorth, line 229: 👍

actors/builtin/miner/miner_actor.go
```diff
@@ -481,6 +475,160 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
...
+			// This operation REMOVES the PoSt from the snapshot so
+			// it can't be disputed again. If this method fails,
+			// this operation must be rolled back.
+			partitions, proofs, err := dlCurrent.TakePoStProofs(store, params.PoStIndex)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load proof for dispute")
+
+			// Load the partition info we need for the dispute.
+			disputeInfo, err := dlCurrent.LoadPartitionsForDispute(store, partitions)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load partition info for dispute")
+			disputedPower = disputeInfo.DisputedPower
```
anorth, line 541: It's worth noting again that this includes power for sectors that were active at proof submission time but may have since expired, and that this is and must be only used for penalty calculations, not power accounting. Maybe worth a rename to `penalisedPower`.

actors/builtin/miner/miner_test.go
```diff
@@ -2280,6 +2327,105 @@ func TestWindowPost(t *testing.T) {
...
+		rt.ExpectSend(builtin.StoragePowerActorAddr, builtin.MethodsPower.CurrentTotalPower, nil, big.Zero(),
+			&power.CurrentTotalPowerReturn{
+				RawBytePower:            networkPower,
+				QualityAdjPower:         networkPower,
+				PledgeCollateral:        actor.networkPledge,
+				QualityAdjPowerSmoothed: actor.epochQAPowerSmooth,
+			},
+			exitcode.Ok)
+
+		rt.ExpectAbortContainsMessage(exitcode.ErrForbidden, "can only dispute window posts during the dispute window", func() {
```
anorth, line 2424: Nit: you should be able to call `actor.disputeWindowPoSt` inside `rt.ExpectAbortContainsMessage`, and I would expect that to encapsulate all the expectation set-up above that you had to replicate. `Reset()` the mock afterwards some some expectations won't have been reached.

actors/builtin/miner/miner_test.go
```diff
@@ -5074,31 +5336,164 @@ func (h *actorHarness) advancePastDeadlineEndWithCron(rt *mock.Runtime) {
...
+	receiver := rt.Receiver()
+	err = receiver.MarshalCBOR(&buf)
+	require.NoError(h.t, err)
+
+	rt.ExpectGetRandomnessBeacon(crypto.DomainSeparationTag_WindowedPoStChallengeSeed, deadline.Challenge, buf.Bytes(), abi.Randomness(challengeRand))
+
+	actorId, err := addr.IDFromAddress(h.receiver)
+	require.NoError(h.t, err)
+
+	// if not all sectors are skipped
```
anorth, line 5393: Stray comment?

anorth (approved):

actors/builtin/miner/deadline_state.go
```diff
@@ -54,12 +56,42 @@ type Deadline struct {
...
+	// Snapshot of partition state at the end of the previous challenge
+	// window for this deadline.
+	PartitionsSnapshot cid.Cid
+	// Snapshot of the proofs submitted by the end of the previous challenge
+	// window for this deadline.
+	//
+	// These proofs may be disputed via DisputeWindowedPoSt. Successfully
+	// disputed window PoSts are removed from the snapshot.
+	OptimisticPoStSubmissionsSnapshot cid.Cid
```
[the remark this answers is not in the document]
Stebalien: Yeah, that seems useful.

actors/builtin/miner/miner_actor.go
Stebalien: For something like this, I assume the submitter would just try again. Any reason we need it to succeed the first time?

actors/builtin/miner/miner_test.go
Stebalien: Actually, this is a bit of a hole in our tests that we should probably fix. I'm trying to test that the expectations are met, even on the failure case. Otherwise, we can get gas mismatches on failure cases.

actors/builtin/miner/miner_actor.go
anorth: Why would it succeed the second time? I don't know what might have gone wrong to cause this failure (low balance?) but it is likely to be permanent. I don't think we should allow inability to pay the reward to stop the fraud being reported at all.

Stebalien: Hm. I was thinking something like the actor that sent the message suddenly got deleted (not sure how this can happen...).

But this isn't going to fix the "low balance" issue. If the actor doesn't have enough funds for the reward, they'll fail on the next step (burning).

actors/builtin/miner/deadline_state.go
```diff
@@ -1076,6 +1204,79 @@ func (dl *Deadline) RescheduleSectorExpirations(
...
+		if err != nil {
+			return err
+		}
+
+		// Record disputed power for penalties.
+		//
+		// NOTE: This also includes power that was
+		// activated at the end of the last challenge
+		// window, and power from sectors that have since
+		// expired.
```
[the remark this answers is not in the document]
Stebalien: Where? I could put it on top of the `ParitionsSnapshot` field, but it seems kind of obvious that the partition snapshot includes power that's no longer active. I put it here because the fact that we're "disputing" no longer active power is interesting.

actors/builtin/miner/miner_test.go
anorth: Fair. I hope that some kind of golden file approach in the future will abstract this away. Your call in the meantime.

Stebalien: Yeah, I agree that golden files will make this kind of thing unnecessary.

actors/builtin/miner/monies.go
```diff
@@ -62,6 +62,11 @@ const ConsensusFaultFactor = 5
 var LockedRewardFactorNum = big.NewInt(75)
 var LockedRewardFactorDenom = big.NewInt(100)
 
+// Base reward for successfully disputing a window posts proofs.
+var BaseRewardForDisputedWindowPoSt = big.Lsh(big.NewInt(4), math.Precision128) // PARAM_SPEC
```
anorth, line 66: I assume the intent here is "4 FIL". This doesn't look right, there's no 128-bit fixed point estimation going on here. A form that is more transparently correct and used elsewhere (builtin/network.go, reward_logic.go) is:

> var BaseRewardForDisputedWindowPoSt = big.Mul(big.NewInt(4), builtin.TokenPrecision)

anorth (changes requested):

actors/builtin/miner/monies.go
Stebalien: :roll_eyes: thanks for that.

anorth (approved):

Issue #1329, opened 2020-12-13
Replace miner state seal proof type with PoSt proof type


anorth:
After we upgraded seal proof type with network versions 7 and 8 it became clear that the seal proof type isn't the right thing to store in miner actor state. The seal proof type is now only used to look-up Window PoSt proof type. We should store that directly instead. This will reduce friction for any similar future proof upgrades.

This would require a small migration, which is probably happening anyway.

anorth:
@Stebalien @dirkmc @magik6k I did some investigation into uses of this before removing it. Many are simple, just using the seal proof type to get to the window/winning PoSt proof type, which we can store in state directly instead.

Lotus locally overrides the seal proof type stored in state if it's an old one, which makes some things below "work". We'll be able to remove this.
https://github.com/filecoin-project/lotus/blob/239e180683055746c29fe3e75ecb39c76309d270/node/impl/full/state.go#L142-L150

Some harder cases in Lotus:
- The [sealer inspects miner state](https://github.com/filecoin-project/lotus/blob/2fd93a55ac166df66d70671094381c213760c9da/extern/storage-sealing/sealing.go#L443) to choose seal proof type. I think this should instead be inferred from the Window PoSt proof type we store in state. That PoSt -> Seal mapping will need to be in Lotus rather than actors, since it's not 1:1 and depends on epoch
- The storage provider adaptor for go-fil-markets [inspects state](https://github.com/filecoin-project/lotus/blob/b8bba15ed22c3303b2f4c593fde71a732f89f6e4/markets/storageadapter/provider.go#L170) to get the seal proof type to create piece commitments. This will have to get it from the node API some other way (possibly using the PoSt -> Seal mapping)
- The storage client similarly inspects state to choose a seal proof type to put in the deal proposal. It'll need to select it differently (maybe using that mapping again).

Stebalien:
Yeah, some form of `PoStProofType -> PreferredSealProofType` mapping should make this pretty easy.

PR #1330, opened 2020-12-13
Remove most v2-only code paths

anorth:
This PR removes almost all handling of network versions that strictly precede actors v3. This was prompted by noise I observed in #1327.

The one path I retained is allowing the creation of new miner actors with an old seal proof type prior to v7, so that we can test how those miners are subsequently handled. This can probably go away after #1329.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -8,7 +8,6 @@ import (
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/go-state-types/cbor"
 	"github.com/filecoin-project/go-state-types/exitcode"
-	"github.com/filecoin-project/go-state-types/network"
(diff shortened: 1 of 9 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -494,10 +491,8 @@ func (a Actor) processDeferredCronEvents(rt Runtime) {
 					continue
 				}
 
-				// Starting version 7, decrement miner count to keep stats consistent.
-				if rt.NetworkVersion() >= network.Version7 {
-					st.MinerCount--
-				}
+				// Decrement miner count to keep stats consistent.
+				st.MinerCount--
```
Stebalien, line 495: Did we ever fix the count in an upgrade?

Stebalien (approved):

actors/builtin/power/power_actor.go
anorth: Yes

PR #1333, opened 2020-12-17
Change miner.Sectors AMT bitwidth to 5. Move constants to state files.

anorth:
@ZenGround0 's benchmarks indicate that the best bitwidth for `miner.Sectors` is 5. I think they accidentally wrote 6 in #1328, which was easy to miss.

I also moved the constants consistently into the `_state.go` files, rather than `policy.go`.

### DIFF
--- a/actors/builtin/market/market_state.go
+++ b/actors/builtin/market/market_state.go
@@ -30,6 +30,10 @@ const (
 	ProviderCollateral
 )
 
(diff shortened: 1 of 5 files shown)


anorth:
I'll ensure @ZenGround0 reviews this after break, but would like to merge so we can build on top of it.

Stebalien (approved):

PR #1337, opened 2020-12-24
Rename and re-arrange private migration inputs

anorth:
This is a cosmetic PR, no functional change required.

- Changed naming so that migration of a single actor is "actor migration" rather than "state migration" (which is the top-level thing)
- Distinguished between the migration job (incl the function to be called) from the functional input/result. The job only exists to support parallelism.
- Un-exported and moved the internal things to the bottom of `top.go`.

### DIFF
--- a/actors/migration/nv9/init.go
+++ b/actors/migration/nv9/init.go
@@ -12,7 +12,7 @@ import (
 
 type initMigrator struct{}
 
-func (m initMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, in StateMigrationInput) (*StateMigrationResult, error) {
+func (m initMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, in actorMigrationInput) (*actorMigrationResult, error) {
 	var inState init2.State
 	if err := store.Get(ctx, in.head, &inState); err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


actors/migration/nv9/init.go
```diff
@@ -12,7 +12,7 @@ import (
 
 type initMigrator struct{}
 
-func (m initMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, in StateMigrationInput) (*StateMigrationResult, error) {
+func (m initMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, in actorMigrationInput) (*actorMigrationResult, error) {
```
Stebalien, line 15: nit: making MigrateState public but `actorMigrationResult` private is strange. Given that this is all private anyways, I'd just make the function private too.

actors/migration/nv9/top.go
```diff
@@ -168,37 +138,66 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, actorsRootIn ci
...
+type actorMigrationInput struct {
+	address    address.Address // actor's address
+	balance    abi.TokenAmount // actor's balance
+	head       cid.Cid         // actor's state head CID
+	priorEpoch abi.ChainEpoch  // epoch of last state transition prior to migration
+}
+
+type actorMigrationResult struct {
+	NewCodeCID cid.Cid
+	NewHead    cid.Cid
```
Stebalien, line 150: nit: no need to make these public, right?

Stebalien (approved):

actors/migration/nv9/init.go
anorth: Done

Issue #1285, opened 2020-10-28
Pre Migrations


ZenGround0:
We need to reduce migration times during upgrade epochs.  A nice idea from @Stebalien is to migrate actors in the background in the node during the hours leading up to the upgrade epoch.  Then when the actual migration runs it can read from a cache and skip migrating actors which have not changed since they were premigrated.

This will involve making MigrateState tree aware of caching and potentially writing another endpoint for premigration.  Since it is threadsafe we can likely just use two invocations of MigrateStateTree aware of caching.

ZenGround0:
@anorth @Stebalien some thoughts on premigrations

## Simple solution

A simple first solution is to do per actor caching. To do this we could add a new argument to `MigrateStateTree`.  This argument will be an interface representing a per actor cache. Conceptually (and maybe concretely) it is a map from actor address to a pair of cids: the actor's state Head before and after the migration. The migration code checks an actor address before migrating. If it is in the cache migration code checks the pre-migration head cid. If it matches that actor's head in the current tree then the migration code just writes the second cid from the cache entry into the actor's Head field in the new tree.  Actors not found in the cache or that have a stale pre-migration head in the cache run through a full migration and then write back the pre and post migration state heads to the cache.

With this cache approach the lotus caller can run a long pre-migration to completion and populate a cache and then during the migration itself run another migration using the cache to use previously migrated actor cids to speed up the migration.

I'm expecting the lotus caller to implement the cache and handle resource management and persistence. There are about 100,000 actors in the filecoin state tree currently.  ID addresses are around 4 bytes, cids around 38 bytes putting the cache at (38*2+4)*100000 = 8MB which should be reasonable to keep entirely in memory during the main migration.  To avoid lots of memory overhead we might want to adapt a strategy of persisting the cache after populating it at first and then only loading it to memory soon before the migration starts.

# Feasibility Evaluation

[...]

PR #1338, opened 2020-12-24
Add logging and non-zero queue capacity to parallel migrations.

anorth:
Adds a log callback to the migration entry point in order to log progress for the operator.

Add configuration for the internal queue sizes. These were previously zero, i.e. rendezvous points. The lack of buffering struck me as likely to be poor for throughput. From a rather unscientific test of migration 500k account actors, buffer sizes of 1000 and 10 improve throughput by ~15-20%. The impact will probably be lower with more complex actors.

This is preparation for #1285.

### DIFF
--- a/actors/migration/nv9/top.go
+++ b/actors/migration/nv9/top.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	address "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/rt"
 	cid "github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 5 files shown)


actors/migration/nv9/util.go
```diff
@@ -122,3 +123,22 @@ func migrateHAMTAMTRaw(ctx context.Context, store cbor.IpldStore, root cid.Cid,
 	}
 	return store.Put(ctx, outRootNodeOuter)
 }
+
+// Guards an unsychronized store with a mutex.
```
ZenGround0, line 127: I think this is redundant with support/ipld/store.go `NewSyncADTStore`

ZenGround0 (approved):

actors/migration/nv9/top.go
```diff
@@ -22,11 +25,28 @@ import (
 
 // Config parameterizes a state tree migration
 type Config struct {
```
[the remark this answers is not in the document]
anorth: Hmm. I often prefer omitting defaults to force the caller to think about the values. Of course, Go doesn't help since it won't complain hard enough if a new field is added. I'm going to leave it now, but open to restoring the defaults in a follow-up. You might argue that we have a better idea about what good values are than integrators do, which is probably true.

actors/migration/nv9/util.go
anorth: Ah, I needed this for my ad-hoc tests which I hacked into the scenario test VM (which does not use a sync store), but yes we don't need it for production.

PR #1345, opened 2021-01-05
Replace miner on-chain seal proof type with window/winning PoSt proof types

### LINKED ISSUE #1329: Replace miner state seal proof type with PoSt proof type

anorth:
The seal proof type isn't the right thing to store in miner state. Many miners have sectors committed with two different seal proof types (but sharing a common Window/Winning PoSt proof type). This PR replaces the seal proof type with the Window/Winning PoSt proof types. It was basically only used for looking up those PoSt proof types anyway.

Reviewers: start with the miner state and actor, then power state and actor. Most of the rest is just noise.

Notes:
- Actors does not use WinningPoSt. I've opted to store it for presumed convenience for nodes, but we could drop if if there were a reliable Window -> Winning PoSt mapping (currently very simple), and save some noise
- I thus selected the Window PoSt proof type for the power claim and consensus minimum power calculation
- There were some existing errors where a v2 package was imported but a v3 one should have been

I nearly forgot to implement the migration, which was easy because the miner info is loaded via CID; I was saved by `TestUpdatePendingDealsMigration` which attempts to publish another deal after migration. We need a more general migration test that uses the sim to build some state, migrates, then continues the sim to exercise the migrated state.

Closes #1329. This issue has some discussion about possible downstream integration challenges.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -102,11 +102,10 @@ type MinerInfo struct {
 	// Slice of byte arrays representing Libp2p multi-addresses used for establishing a connection with this miner.
 	Multiaddrs []abi.Multiaddrs
 
-	// The proof types used for PoSt for this miner.
+	// The proof type used for Window PoSt for this miner.
 	// A miner may commit sectors with different seal proof types (but compatible sector size and
 	// corresponding PoSt proof types).
 	WindowPoStProofType  abi.RegisteredPoStProof
-	WinningPoStProofType abi.RegisteredPoStProof
 
 	// Amount of space in each sector committed by this miner.
 	// This is computed from the proof type and represented here redundantly.
@@ -230,20 +229,11 @@ func ConstructState(store adt.Store, infoCid cid.Cid, periodStart abi.ChainEpoch
 	}, nil
 }
 
-func ConstructMinerInfo(owner addr.Address, worker addr.Address, controlAddrs []addr.Address, pid []byte,
-	multiAddrs [][]byte, windowPoStProofType, winningPoStProofType abi.RegisteredPoStProof) (*MinerInfo, error) {
+func ConstructMinerInfo(owner, worker addr.Address, controlAddrs []addr.Address, pid []byte, multiAddrs [][]byte,
+	windowPoStProofType abi.RegisteredPoStProof) (*MinerInfo, error) {
 	sectorSize, err := windowPoStProofType.SectorSize()
 	if err != nil {
 		return nil, xc.ErrIllegalArgument.Wrapf("invalid sector size: %w", err)
-	}
-
-	checkSectorSize, err := winningPoStProofType.SectorSize()
-	if err != nil {
-		return nil, xc.ErrIllegalArgument.Wrapf("invalid sector size: %w", err)
-	}
-	if sectorSize != checkSectorSize {
-		return nil, xc.ErrIllegalArgument.Wrapf("incompatible Window and Winning PoSt proof types %d, %d",
-			windowPoStProofType, winningPoStProofType)
 	}
 
 	partitionSectors, err := builtin.PoStProofWindowPoStPartitionSectors(windowPoStProofType)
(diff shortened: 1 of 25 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -102,8 +102,11 @@ type MinerInfo struct {
 	// Slice of byte arrays representing Libp2p multi-addresses used for establishing a connection with this miner.
 	Multiaddrs []abi.Multiaddrs
 
-	// The proof type used by this miner for sealing sectors.
-	SealProofType abi.RegisteredSealProof
+	// The proof types used for PoSt for this miner.
+	// A miner may commit sectors with different seal proof types (but compatible sector size and
+	// corresponding PoSt proof types).
+	WindowPoStProofType  abi.RegisteredPoStProof
+	WinningPoStProofType abi.RegisteredPoStProof
```
anorth, line 109: @nicola you have previously noted that changing schemas like this makes dynamic interpretation (e.g. of JSON versions) across versions by external tools more difficult. To retain semantic compatibility here, we would need to leave a dummy integer field here, then append these two new ones at the end of the struct. I don't that's a huge pain or cost, so am willing to do so. 

We won't always get off so easily though - is it worth it? Tooling in general will need to understand these schemas and use the right one. Feedback also welcome @frrist @iand @ribasushi @willscott

ribasushi: @rvagg would be the best person to weigh in from IPLD on the above

nicola: It does break my tooling that does not have any of the above, since they assumed that things like this would not happen.

However, with good engineering on libraries and clear schemas it shouldn't be a problem. The expert here is @willscott.

nicola:
I may have misunderstood the implementation, but:

There are constraints that depend on SealProofType and we can't remove it:
- Sealing takes as parameter the seal proof type
- During ProveCommit, the seal proof type must be used to correctly verify the inputs to the snark
- Some rules may depend on the proof type e.g. sector sealed with this proof should not be allowed, how would we do that?

ZenGround0 (changes requested):
Looking good. Blocking because I don't think it is safe to skip the power state migration because there should still be some `RegisteredSealProof_StackedDrg32GiBV1` (3) in the first field of the claims struct which is not a valid WindowPostProof type value (should be 8).

I don't see sufficient motivation for including winning post type in the miner info instead of creating a static map from window to winning post proof, but not blocking on this.

actors/builtin/miner/miner_state.go
rvagg: Array length helps differentiate these changes, we're adding +1 to this array and there's no previous form with this many - [v0 has 9, v2 has 11](https://specs.ipld.io/data-structures/filecoin/state.html#storagemineractor), this has 12, so this should be as straightforward as other changes that have been introduced.

anorth: Given that the array length identifies the schema (this time), I'm going to leave it in this natural form, with no dummy.

anorth:
@nicola the seal proof type remains in the `SectorPreCommitInfo` (for verifying seal) and `SectorOnChainInfo` (for unknown future use).

anorth:
@ZenGround0 the missing migration of claims is an oversight, thanks.

I will remove the Winning PoSt proof type.

actors/builtin/miner/miner_state.go
anorth: Actually, I'm going to remove the Winning PoSt proof type, which will revert this to 11 fields. I think I'll just rename the proof type, but leave it in the existing array position. Since they were the same serialized type, the schema remains the same. The values will be disjoint between v2 and v3. I think this is likely to be the easiest thing for schema-less consumers to use anyway.

ZenGround0 (approved):
LGTM

PR #1346, opened 2021-01-06
Add caching for premigration

ZenGround0:
The design proposed here is a hierarchical cache.  The `MigrationCache` stores all per actor address `MigrationAddressCache`s.  Each `MigrationAddressCache` maps cids to cids.  Neither cache needs to be threadsafe for use in a single migration.  Caches should only be used in one migration at a time -- the premigration should finish completely before using a populated cache in another migration.  If this is a problem for lotus integration caches can be made threadsafe or we can revisit this design.

The toplevel migration reads and writes back from address cache.  This PR tweaks the actorMigration interface to avoid making every actor migrator do its own caching.

TODO
~test with two migrations in serial.~
- [x] add miner deadline and sector amt root cids to miner actor caches

~partial sectors amt caching~

### DIFF
--- a/actors/migration/nv9/top.go
+++ b/actors/migration/nv9/top.go
@@ -47,15 +47,15 @@ type Logger interface {
 type MigrationCacheKey string
 
 func ActorHeadKey(addr address.Address, head cid.Cid) MigrationCacheKey {
-	return MigrationCacheKey(fmt.Sprintf("%s-h-%s", addr.String(), head.String()))
+	return MigrationCacheKey(addr.String() + "-h-" + head.String())
 }
 
 func DeadlineKey(dlCid cid.Cid) MigrationCacheKey {
-	return MigrationCacheKey(fmt.Sprintf("d-%s", dlCid.String()))
+	return MigrationCacheKey("d-" + dlCid.String())
 }
 
 func SectorsRootKey(sCid cid.Cid) MigrationCacheKey {
-	return MigrationCacheKey(fmt.Sprintf("s-%s", sCid))
+	return MigrationCacheKey("s-" + sCid.String())
 }
 
 // MigrationCache stores and loads cached data. Its implementation must be threadsafe
--- a/actors/migration/nv9/top.go
+++ b/actors/migration/nv9/top.go
@@ -129,7 +129,6 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, actorsRootIn ci
 				cache:          cache,
 				actorMigration: migrations[actorIn.Code],
 			}
-			fmt.Printf("creating job on %s\n", addr)
 			select {
 			case jobCh <- nextInput:
 			case <-ctx.Done():
@@ -152,7 +151,6 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, actorsRootIn ci
 		grp.Go(func() error {
 			defer workerWg.Done()
 			for job := range jobCh {
-				fmt.Printf("running job on %s\n", job.Address)
 				result, err := job.run(ctx, store, priorEpoch)
 				if err != nil {
 					return err
--- a/actors/migration/nv9/top.go
+++ b/actors/migration/nv9/top.go
@@ -44,9 +44,21 @@ type Logger interface {
 	Log(level rt.LogLevel, msg string, args ...interface{})
 }
 
+// MigrationCache ByAddress calls need not be threadsafe
+// It must be safe to operation on any two output PreMigrationAddressCache concurrently
+type MigrationCache interface {
+	ByAddress(addr address.Address) MigrationAddressCache
+}
+
+// MigrationAddressCache need not be threadsafe
+type MigrationAddressCache interface {
+	Write(oldCid, newCid cid.Cid) error
+	Read(oldCid cid.Cid) (bool, cid.Cid, error)
+}
+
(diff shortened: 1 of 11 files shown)


actors/migration/nv9/top.go
```diff
@@ -223,12 +238,14 @@ type actorMigration interface {
 	// Loads an actor's state from an input store and writes new state to an output store.
 	// Returns the new state head CID.
 	migrateState(ctx context.Context, store cbor.IpldStore, input actorMigrationInput) (result *actorMigrationResult, err error)
+	migratedCodeCID() cid.Cid
```
anorth, line 241: It's a bit annoying that this was needed only for the cached case. I suggest instead making it a member of `cachedMigrator`, set in the `migrations` map in the same way and place it's provided to the `nilMigrator`.

actors/migration/nv9/top.go
```diff
@@ -241,12 +258,19 @@ func (job *migrationJob) run(ctx context.Context, store cbor.IpldStore, priorEpo
 		balance:    job.Actor.Balance,
 		head:       job.Actor.Head,
 		priorEpoch: priorEpoch,
+		cache:      job.cache,
```
anorth, line 261: Since the cache is available here, the cached migrator isn't abstracting anything. If retaining this pattern of per-actor-instance caches (see comments elsewhere), just ditch the `cachedMigrator` and inline the cache lookup here. Use the computing map pattern to encapsulate the write at L269 too.

But I think it would be better again to make a singleton cache a member of and hence encapsulated by the `cachedMigrator`: do both the reading and the writing to the cache in the same method, in `cachedMigrator`.

actors/migration/nv9/top.go
```diff
@@ -44,9 +44,21 @@ type Logger interface {
...
+// It must be safe to operation on any two output PreMigrationAddressCache concurrently
+type MigrationCache interface {
+	ByAddress(addr address.Address) MigrationAddressCache
+}
+
+// MigrationAddressCache need not be threadsafe
+type MigrationAddressCache interface {
+	Write(oldCid, newCid cid.Cid) error
+	Read(oldCid cid.Cid) (bool, cid.Cid, error)
+}
```
anorth, line 57: I think this two-level approach causes unnecessary work (here, and later in Lotus), while not being fully general enough to solve the problem of cache key semantics.

That problem, to be clear, is that CID `Ba1` may be migrated to `Ba2` in one context (say as a cron queue entry) but to `Ba3` (or remain `Ba1`) in another (e.g. some piece of market actor state). We'll have this problem if the same data schema has two different interpretations (which is quite unlikely since the schema is ultimately just a heterogenous array). The two level arrangement solves that problem only if one of the contexts is top-level actor state. I don't think you've run into that problem with the simple caching behaviour currently in use, but we could do so later if we were unlucky, or implemented some kind of fully general cache interceptor across all state objects.

In this case I think the general solution is also simpler. That general solution is just an abstraction over a `map[[]byte]Cid`, where the context of a migration input can be encoded in the key. In the simplest case, both `Address` and `Cid` have cheap `[]byte` interpretations and can be used as keys directly. In more complex cases, path or context information can be prepended to the key.

Say we got unlucky and an AMT intermediate node has the same schema as some top-level actor type, and indeed could contain the same data but be migrated to different things in v3. To work around that, prefix keys with "actor/", "miner/sectors/" etc.

Coming back to the here and now, I don't think you need to do any of that key prefixing yet, but the current structure of representing context by the hierarchy of caches will be painful to extend if/when we do need more. And is just noisy now anyway. Instead, represent the context in hierarchy of the keys in a single cache.

The one thing you'll lose is that the cache will need to be synchronized. But that will be easy and unlikely to affect performance.

This will also make passing the cache around easier. There's just one, so it can be initialised as a member of `cachedMigrator` at the top, before the actor address is resolved.

anorth (commented):
Thanks Zen. This PR is going in the right direction, but presents a couple of nice talking points about design. If my suggestions make sense, go ahead and make them and I'm sure we'll be LGTM after that. Otherwise I'm happy to discuss them in person when we chat next.

actors/migration/nv9/top.go
ZenGround0: It's ugly but it's better in that the Migrator maintains the CID for itself so there is only one piece of logic responsible for the CID returned by both the cached and uncached code paths.

ZenGround0:
I'm taking out of draft because I now think more extensive testing and the partial sectors amt caching are better suited to other PRs.

@anorth I followed most of your suggestions.  It turned out using a single map with a RWLock increased migration runtime about 2x so I tried out a sync.Map which got us back to the performance of per address maps.

actors/migration/nv9/top.go
```diff
@@ -44,9 +44,30 @@ type Logger interface {
 	Log(level rt.LogLevel, msg string, args ...interface{})
 }
 
+type MigrationCacheKey string
+
+func ActorHeadKey(addr address.Address, head cid.Cid) MigrationCacheKey {
+	return MigrationCacheKey(fmt.Sprintf("%s-h-%s", addr.String(), head.String()))
```
anorth, line 50: Nit: it probably doesn't matter here, but this kind of string formatting is much more expensive than just concatenating them.

anorth (approved):

PR #1350, opened 2021-01-11
Migration on simulated network

ZenGround0:
Use the new simulation framework to 1) generate realistic entropy within a state for migrating 2) continue simulating after migration to demonstrate common case network operations continue working on the migrated state.
- [x] Simulate state and then migrate
- [x] Continue simulating after the migration

### DIFF
--- a/actors/util/adt/store.go
+++ b/actors/util/adt/store.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/filecoin-project/go-state-types/cbor"
 	"github.com/filecoin-project/go-state-types/exitcode"
+	adt2 "github.com/filecoin-project/specs-actors/v2/actors/util/adt"
 	cid "github.com/ipfs/go-cid"
 	ipldcbor "github.com/ipfs/go-ipld-cbor"
 
@@ -12,10 +13,11 @@ import (
 )
 
 // Store defines an interface required to back the ADTs in this package.
-type Store interface {
-	Context() context.Context
-	ipldcbor.IpldStore
-}
+// type Store interface {
+// 	Context() context.Context
+// 	ipldcbor.IpldStore
+// }
+type Store = adt2.Store
 
 // Adapts a vanilla IPLD store as an ADT store.
 func WrapStore(ctx context.Context, store ipldcbor.IpldStore) Store {
--- a/actors/util/adt/store.go
+++ b/actors/util/adt/store.go
@@ -12,11 +12,6 @@ import (
 	vmr "github.com/filecoin-project/specs-actors/v3/actors/runtime"
 )
 
-// Store defines an interface required to back the ADTs in this package.
-// type Store interface {
-// 	Context() context.Context
-// 	ipldcbor.IpldStore
-// }
 type Store = adt2.Store
 
 // Adapts a vanilla IPLD store as an ADT store.
--- a/support/agent/cases_test.go
+++ b/support/agent/cases_test.go
@@ -21,7 +21,7 @@ import (
 	"github.com/filecoin-project/specs-actors/v3/actors/builtin"
 	"github.com/filecoin-project/specs-actors/v3/actors/builtin/exported"
 	"github.com/filecoin-project/specs-actors/v3/actors/builtin/power"
-	"github.com/filecoin-project/specs-actors/v3/actors/migration/nv9"
+	"github.com/filecoin-project/specs-actors/v3/actors/migration/nv10"
 	"github.com/filecoin-project/specs-actors/v3/actors/states"
 	"github.com/filecoin-project/specs-actors/v3/actors/util/adt"
 	"github.com/filecoin-project/specs-actors/v3/support/agent"
@@ -231,7 +231,7 @@ func TestMigration(t *testing.T) {
 	sim := agent.NewSimWithVM(ctx, t, v, v2VMFactory, agent.ComputePowerTableV2, blkStore, newBlockStore, v2MinerFactory, agent.SimConfig{
 		Seed:             rnd.Int63(),
 		CheckpointEpochs: 1000,
-	})
+	}, agent.CreateMinerParamsV2)
 
 	// create miners
 	workerAccounts := vm_test2.CreateAccounts(ctx, t, v, minerCount, initialBalance, rnd.Int63())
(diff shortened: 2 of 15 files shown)


ZenGround0:
@Stebalien @anorth I'm not exactly sure who to tag for review here given recent team changes.  This PR only impacts testing frameworks so it might not be critical enough for either of you to review.

support/agent/cases_test.go
```diff
@@ -29,7 +37,7 @@ func TestCreate20Miners(t *testing.T) {
 	rnd := rand.New(rand.NewSource(42))
 
 	sim := agent.NewSim(ctx, t, newBlockStore, agent.SimConfig{Seed: rnd.Int63()})
-	accounts := vm_test.CreateAccounts(ctx, t, sim.GetVM(), minerCount, initialBalance, rnd.Int63())
+	accounts := vm_test.CreateAccounts(ctx, t, sim.GetVM().(*vm_test.VM), minerCount, initialBalance, rnd.Int63())
```
Stebalien, line 40: Can we make `sim.GetVM` return the concrete type?

support/agent/cases_test.go
```diff
@@ -177,7 +185,150 @@ func TestCommitPowerAndCheckInvariants(t *testing.T) {
...
+
+			fmt.Printf("Power at %d: raw: %v  cmtRaw: %v  cmtSecs: %d  cnsMnrs: %d avgWins: %.3f  msgs: %d\n",
+				epoch, pwrSt.TotalRawBytePower, pwrSt.TotalBytesCommitted, sectorCount.Uint64(),
+				pwrSt.MinerAboveMinPowerCount, float64(sim.WinCount)/float64(epoch), sim.MessageCount)
+		}
+	}
+}
+
+func TestMigration(t *testing.T) {
+	t.Skip("slow")
```
Stebalien, line 211: Can we bind this to an environment variable and then test it in a separate CI job?

support/agent/cases_test.go
```diff
@@ -177,7 +185,150 @@ func TestCommitPowerAndCheckInvariants(t *testing.T) {
...
+	ctx := context.Background()
```
Stebalien, line 212: generally a good idea to cancel this.

support/agent/cases_test.go
```diff
@@ -177,7 +185,150 @@ func TestCommitPowerAndCheckInvariants(t *testing.T) {
...
+	for _, ba := range exported.BuiltinActors() {
+		lookup[ba.Code()] = ba
+	}
+
+	v3, err := vm_test.NewVMAtEpoch(ctx, lookup, v2.Store(), nextRoot, priorEpoch+1)
+	require.NoError(t, err)
+
+	stateTree, err := v3.GetStateTree()
+	require.NoError(t, err)
+	totalBalance, err := v3.GetTotalActorBalance()
```
Stebalien, line 309: Shouldn't this be the balance from _before_ the migration?

support/agent/cases_test.go
```diff
@@ -177,7 +185,150 @@ func TestCommitPowerAndCheckInvariants(t *testing.T) {
...
+	require.NoError(t, err)
+	msgs, err := states.CheckStateInvariants(stateTree, totalBalance, priorEpoch)
+	require.NoError(t, err)
+	assert.Equal(t, 0, len(msgs.Messages()), strings.Join(msgs.Messages(), "\n"))
```
Stebalien, line 313: nit: `assert.Zero`.

support/agent/miner_state.go
```diff
@@ -0,0 +1,242 @@
+package agent
```
Stebalien, line 1: it is rather unfortunate that we have to create these shims inside actors, given that we've done the same thing inside lotus...

Stebalien: (just a note for future thinking)

Stebalien (commented):
Main comment is that the type assertions everywhere is a rather annoying. Is there any way to use a concrete type and/or get rid of the assertions?

actors/util/adt/store.go
```diff
@@ -5,17 +5,19 @@ import (
...
 
 // Store defines an interface required to back the ADTs in this package.
-type Store interface {
-	Context() context.Context
-	ipldcbor.IpldStore
-}
+// type Store interface {
+// 	Context() context.Context
+// 	ipldcbor.IpldStore
+// }
```
anorth, line 19: 👍 . You can delete the commented code.

Maybe even go back to alias the v2 store onto v0.

support/agent/cases_test.go
anorth: The test flags from go-filecoin (venus) worked well. Maybe time to replicate that here.

anorth (commented):
I skimmed. The shimming/handling of two different actor versions for the sim does at some noise and complexity. Another approach for testing across upgrades might be to preserve only the blockstore. That is, use a v2 VM and v2 sim to start, then upgrade, then use a v3 VM and sim, initialised with the same (upgraded) blockstore to keep going. Maybe at some point that would require abstracting over the simulator API.

support/agent/cases_test.go
ZenGround0: In general👍  
I want to get this overdue pr merged and do agent testing right so I've written up an issue to track this #1385

support/agent/cases_test.go
ZenGround0: No I think this should be balance after.  This call is checking that the v3 state is internally consistent post migration, it's not asserting the correctness of the migration transition.

We could have used v2 balance to do this because the migration function should not have changed total balance.

ZenGround0:
> That is, use a v2 VM and v2 sim to start, then upgrade, then use a v3 VM and sim, initialised with the same (upgraded) blockstore to keep going. Maybe at some point that would require abstracting over the simulator API.

FYI this was my original idea but I backed out of it when I realized I would need to do non-trivial transformations on internal simulation state to go between v2 and v3 sims.  This problem might not be so bad in the future if the sim framework does not add new functionality and stays stable between subsequent version changes.  However in an ideal world we *will* be adding simulation functionality so the current approach doesn't get in the way of improvement.

support/agent/cases_test.go
Stebalien: Then I don't think this is useful. GetTotalActorBalance just does the exact same thing as CheckStateInvariants.

support/agent/sim.go
```diff
@@ -200,7 +279,7 @@ func (s *Sim) AddDealProvider(d DealProvider) {
 }
 
 func (s *Sim) GetVM() *vm.VM {
-	return s.v
+	return s.v.(*vm.VM)
```
Stebalien, line 282: Why not just change the type of `s.v`? I assume this will never panic.

Stebalien (approved):

support/agent/cases_test.go
ZenGround0: I don't follow.  We need the output of `GetTotalActorBalance` as input to `CheckStateInvariants`.  Are you saying this    should be called inside CheckStateInvariants instead of outside?  If so I agree but it belongs in its own PR.

ZenGround0: To clarify: "This call is checking that the v3 state is internally consistent post migration, it's not asserting the correctness of the migration transition." refers to CheckStateInvariants, not GetTotalActorBalance(), my bad on the confusing wording.

PR #1356, opened 2021-01-19
Upgrade to HAMT and AMT v3.0.0

anorth:
This is mostly just plumbing through of the new errors and return values.

I added a `Pop()` method to Map and Array for the "get then delete" use case (twice). Have not yet updated the miner actor to use `SetIfAbsent` in pre-commit (will do so later).

FYI @arajasek

### DIFF
--- a/actors/builtin/power/power_state.go
+++ b/actors/builtin/power/power_state.go
@@ -89,9 +89,13 @@ func ConstructState(store adt.Store) (*State, error) {
 	if err != nil {
 		return nil, xerrors.Errorf("failed to create empty map: %w", err)
 	}
-	emptyCronQueueMMapCid, err := adt.MakeEmptyMultimap(store, CronQueueHamtBitwidth, CronQueueAmtBitwidth).Root()
+	emptyCronQueueMMap, err := adt.MakeEmptyMultimap(store, CronQueueHamtBitwidth, CronQueueAmtBitwidth)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to create empty multimap: %w", err)
+	}
+	emptyCronQueueMMapCid, err := emptyCronQueueMMap.Root()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get empty multimap CID: %w", err)
 	}
 
 	return &State{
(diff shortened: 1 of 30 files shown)


actors/builtin/power/power_state.go
```diff
@@ -89,10 +89,14 @@ func ConstructState(store adt.Store) (*State, error) {
...
 		return nil, xerrors.Errorf("failed to create empty map: %w", err)
 	}
-	emptyCronQueueMMapCid, err := adt.MakeEmptyMultimap(store, CronQueueHamtBitwidth, CronQueueAmtBitwidth).Root()
+	emptyCronQueueMMap, err := adt.MakeEmptyMultimap(store, CronQueueHamtBitwidth, CronQueueAmtBitwidth)
 	if err != nil {
 		return nil, xerrors.Errorf("failed to create empty multimap: %w", err)
 	}
+	emptyCronQueueMMapCid, err := emptyCronQueueMMap.Root()
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get empty multimap CID: %w", err)
```
Stebalien, line 98: nit: could provide a `StoreEmptyMultimap` function to mirror the others.

actors/util/adt/array.go
```diff
@@ -115,12 +128,24 @@ func (a *Array) Length() uint64 {
...
-	if err := a.root.Get(a.store.Context(), k, out); err == nil {
-		return true, nil
-	} else if _, nf := err.(*amt.ErrNotFound); nf {
-		return false, nil
+	if found, err := a.root.Get(a.store.Context(), k, out); err != nil {
+		return false, err
 	} else {
+		return found, nil
+	}
+}
```
Stebalien, line 136: Just return it?

actors/util/adt/map.go
```diff
@@ -86,34 +92,42 @@ func (m *Map) Put(k abi.Keyer, v cbor.Marshaler) error {
...
-// Get puts the value at `k` into `out`.
+// Get retrieves the value at `k` into `out`, if the `k` is present and `out` is non-nil.
+// Returns whether the key was found.
 func (m *Map) Get(k abi.Keyer, out cbor.Unmarshaler) (bool, error) {
-	if err := m.root.Find(m.store.Context(), k.Key(), out); err != nil {
-		if err == hamt.ErrNotFound {
-			return false, nil
-		}
+	if found, err := m.root.Find(m.store.Context(), k.Key(), out); err != nil {
 		return false, errors.Wrapf(err, "map get failed find in node %v with key %v", m.lastCid, k.Key())
```
Stebalien, line 99: nit: we can use this as an excuse to switch to `xerrors.Errorf` if you want.

Stebalien (approved):

actors/util/adt/array.go
Stebalien: Or wrap it.

actors/util/adt/map.go
anorth: Done, there was only one left untouched after this.

Stebalien (approved):

PR #1360, opened 2021-01-21
remove logic specific to network versions before actors v3

Stebalien:
Specifically, we no longer need to care about v0/v7/v8 supported seal type differences.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -57,7 +57,7 @@ func init() {
 	}
 
 	// permit 2KiB sectors in tests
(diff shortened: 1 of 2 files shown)


ribasushi:
@Stebalien does this change not preclude "syncing from genesis" ?

Stebalien:
Actually, I'm going to leave this. Otherwise, we can't create a miner with old sectors (for testing) and those miners exist on the network.

Stebalien:
> @Stebalien does this change not preclude "syncing from genesis" ?

No. We're using go major versions so lotus will use v0 and v2 until the v3 upgrade epoch.

anorth:
>  Otherwise, we can't create a miner with old sectors (for testing) and those miners exist on the network

Yes, that's why I left it behind in #1330

PR #1365, opened 2021-01-24
compute the hash of the proposal iff the proposal hash is present

### LINKED ISSUE #1364: skip calculating hash if not check, or remove param
in getting a msig transaction, if the `checkHash` is not set, the hash is calculated and gas charged even if the proposal hash is `nil`. Every use of this function sets `checkHash` to true, so this issue could be to remove that parameter to reduce confusion if not being planned to be used.

https://github.com/filecoin-project/specs-actors/blob/c7ea99141c44e83b7b3856d0dc5833ae2c5e50d4/actors/builtin/multisig/multisig_actor.go#L513

Stebalien:
Previously, we'd always compute the hash but only _compare_ it if it was specified. This simply increased the cost of messages that didn't specify the hash for no good reason.

fixes #1364


Stebalien:
@anorth please check this _thoroughly_. I believe this is the correct thing to do, but I'm not familiar with why the code was written the way it was in the first place.

ZenGround0:
Looks good.  Adding tests before merging.

droark (approved):
👍 

I'd prefer that the PR not be pulled for another day, in case I think of some reason why this shouldn't be pulled, but I'm pretty sure this is safe. I can't think of any reasons why a hash should be mandatory in this situation. The modified logic seems safe too.

ZenGround0 (approved):

PR #1367, opened 2021-01-26
remove the newtype from the migration cache

Stebalien:
This makes it easier to abstract over the migration cache in lotus without directly referring to _this_ implementation.


anorth (approved):
Makes sense. Thanks Go.

PR #1368, opened 2021-01-27
migration: add methods to the cache to clone/update

Stebalien:
This lets us:

1. Clone the cache before pre-migrating.
2. Update the original cache when done pre-migrating.

Otherwise, if we fail when pre-migrating, we may give a cache to the final migration that references blocks that were never actually written to disk.

(also, switch to referencing the cache by pointer)

### DIFF
--- a/actors/migration/nv10/util.go
+++ b/actors/migration/nv10/util.go
@@ -130,21 +130,19 @@ func migrateHAMTAMTRaw(ctx context.Context, store cbor.IpldStore, root cid.Cid, 
 }
 
 type MemMigrationCache struct {
-	MigrationMap *sync.Map
+	MigrationMap sync.Map
 }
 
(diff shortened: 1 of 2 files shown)


actors/migration/nv10/util.go
```diff
@@ -130,21 +130,19 @@ func migrateHAMTAMTRaw(ctx context.Context, store cbor.IpldStore, root cid.Cid,
 }
 
 type MemMigrationCache struct {
-	MigrationMap *sync.Map
+	MigrationMap sync.Map
```
Stebalien, line 133: Now that we're returning `MemMigrationCache` by pointer, we don't need to put the map behind a pointer.

Stebalien:
Only codecov: add tests, reduce test coverage.

anorth (approved):

PR #1371, opened 2021-01-28
allocate a new slice when combining options

Stebalien:
The existing code _should_ be fine as the "defaults" slice shouldn't have any extra capacity. But if, for some reason, it _did_, the append would not be thread-safe.

### DIFF
--- a/actors/migration/nv10/util.go
+++ b/actors/migration/nv10/util.go
@@ -25,7 +25,9 @@ func migrateHAMTRaw(ctx context.Context, store cbor.IpldStore, root cid.Cid, new
 		return cid.Undef, err
 	}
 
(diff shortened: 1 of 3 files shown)


anorth:
OMG Go I hate you.

Could we maybe abstract this into a function so it's easier to use reliably? I'd call it `append`, but I guess we need to use `appendSafe` or something.

anorth (approved):

Stebalien:
Note: I actually think the current code is safe because go _shouldn't_ over-allocate slice literals. I mostly submitted this patch when trying to figure out what's going wrong.

According to the spec: 

>  A slice literal describes the entire underlying array literal. Thus the length and capacity of a slice literal are the maximum element index plus one.

That means `append(slice, ...)` _must_ allocate a new slice.

So I think we can close this.

PR #1373, opened 2021-01-28
fix verified clients migration

Stebalien:
Don't replace verified clients with verifies. Luckily, migration verification logic catches this case.


arajasek (approved):

anorth:
😬  thanks. This shows up a risk in the laborious coding required for a migration.

Stebalien:
Yeah, I'm going to try to write a tool that looks two states and compares them. This shouldn't actually be too difficult given the abstractions we have in lotus.

Stebalien:
I.e., I can verify that the states are semantically equivalent.

PR #1379, opened 2021-02-19
fix: use accurate proof size limits

Stebalien:
Specifically, this fixes an issue where we could submit a window post for many (i.e., more than 10) partitions in a single window post for 2k sectors. This would lead to proof sizes over 1920 bytes (the old max).

Now, we _accurately_ compute the proof size limits from actual known constants.

Depends on https://github.com/filecoin-project/go-state-types/pull/23.

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -877,7 +877,7 @@ func TestCommitments(t *testing.T) {
 
 			// Too big at version 4
 			proveCommit := makeProveCommit(sectorNo)
-			proveCommit.Proof = make([]byte, 1920)
+			proveCommit.Proof = make([]byte, 192)
 			actor.proveCommitSectorAndConfirm(rt, precommit, proveCommit, proveCommitConf{
 				vestingPledgeDelta: &test.vestingPledgeDelta,
 			})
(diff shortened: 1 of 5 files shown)


Stebalien:
The alternative is to just over-estimate and pick a high number. Unfortunately, it would have to be pretty high for 2k sectors...

anorth (approved):
LGTM

Should this land after we update the gomod version on master to v4? Or are you intending to get this into v3?

Stebalien:
This needs to land in V3. At the moment, we have some issues on calibration net with 2k sectors because window post proofs are much larger.

Larger = 192*3000 = 562KiB because up to 3K 2-sector partitions (for 2k sectors) can be proved at once.

actors/builtin/miner/miner_test.go
```diff
@@ -877,7 +877,7 @@ func TestCommitments(t *testing.T) {
 
 			// Too big at version 4
 			proveCommit := makeProveCommit(sectorNo)
-			proveCommit.Proof = make([]byte, 1920)
+			proveCommit.Proof = make([]byte, 192)
```
ZenGround0, line 880: Not sure if its important to keep this case but this proof size is not too big for version 4 which had [a limit of 1024](https://github.com/filecoin-project/specs-actors/commit/c5e31f2282adc206030f5fbff1af0814c0765535#diff-25d448ee2846a665efe3de55eab372d83c4a4ef4b27108a0250b03439c6b48f0R69) without taking into account seal proof type.

ZenGround0 (approved):

actors/builtin/miner/miner_test.go
Stebalien: I'll try to improve the test (i.e., test with all sector sizes).

arajasek:
Merging this now to get it into Lotus 1.5.0 code-freeze.

PR #1381, opened 2021-02-23
Aggregate PoRep

### LINKED ISSUE #1257: Add b̶a̶t̶c̶h̶e̶d̶ aggregated SectorProveCommit endpoint
Sector prove-commits are already batched and processed at the end of each epoch via a cron call in the power actor. This might not be the best solution, though. It's complicated, can't reliably batch across epochs, and each ProveCommit invocation still mutates state etc.

It might be simpler and more effective to expose a ProveCommitBatch method, giving miners explicit control over batch size and direct exposure to the true gas cost.

ZenGround0:
Changes needed for introducing aggregate porep verification into the network.

- [x] Implement `ProveCommitAggregate`
- [x] VM tests of `ProveCommit` and `ProveCommitAggregate` to measure gas costs
- [x] Batch market `ComputeDataCommitment` call to save gas
- [x] `ComputeDataCommitment` testing
- [x] Check power claim during `ProveCommitAggreagate` to retain safety properties protecting against programmer error in deadline code
- [ ] ~Accurate gas table~
- [ ] ~Final min and max aggregate batch parameters and max proof size parameter~
- [x] Extend sector precommit expiry
- [x] AggregateProofType for upgradeability
- [x] Unit (and or more extensive VM) tests of `ProveCommitAggregate`

--edit--
I'm calling final Gas and param estimation out of scope because we'll need to merge this into lotus to start gas rebalancing.  We will do the final parameter setting in another PR.

Closes #1257.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -399,37 +399,45 @@ func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *abi.Empty
 	return nil
 }
 
-//type ComputeDataCommitmentParams struct {
-//	DealIDs    []abi.DealID
-//	SectorType abi.RegisteredSealProof
-//}
-type ComputeDataCommitmentParams = market0.ComputeDataCommitmentParams
+type SectorDataSpec struct {
+	DealIDs    []abi.DealID
+	SectorType abi.RegisteredSealProof
+}
 
-func (a Actor) ComputeDataCommitment(rt Runtime, params *ComputeDataCommitmentParams) *cbg.CborCid {
+type ComputeDataCommitmentParams struct {
+	Inputs []*SectorDataSpec
+}
+
+type ComputeDataCommitmentReturn struct {
+	CommDs []cbg.CborCid
+}
+
+func (a Actor) ComputeDataCommitment(rt Runtime, params *ComputeDataCommitmentParams) *ComputeDataCommitmentReturn {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 
 	var st State
 	rt.StateReadonly(&st)
 	proposals, err := AsDealProposalArray(adt.AsStore(rt), st.Proposals)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deal dealProposals")
+	commDs := make([]cbg.CborCid, len(params.Inputs))
+	for i, commInput := range params.Inputs {
+		pieces := make([]abi.PieceInfo, 0)
+		for _, dealID := range commInput.DealIDs {
+			deal, err := getDealProposal(proposals, dealID)
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get dealId %d", dealID)
 
-	pieces := make([]abi.PieceInfo, 0)
-	for _, dealID := range params.DealIDs {
-		deal, err := getDealProposal(proposals, dealID)
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to get dealId %d", dealID)
-
-		pieces = append(pieces, abi.PieceInfo{
-			PieceCID: deal.PieceCID,
-			Size:     deal.PieceSize,
-		})
+			pieces = append(pieces, abi.PieceInfo{
+				PieceCID: deal.PieceCID,
+				Size:     deal.PieceSize,
+			})
+		}
+		commD, err := rt.ComputeUnsealedSectorCID(commInput.SectorType, pieces)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to compute unsealed sectorCID: %s", err)
+		commDs[i] = (cbg.CborCid)(commD)
 	}
-
-	commd, err := rt.ComputeUnsealedSectorCID(params.SectorType, pieces)
-	if err != nil {
-		rt.Abortf(exitcode.ErrIllegalArgument, "failed to compute unsealed sector CID: %s", err)
+	return &ComputeDataCommitmentReturn{
+		CommDs: commDs,
 	}
-
-	return (*cbg.CborCid)(&commd)
 }
 
 //type OnMinerSectorsTerminateParams struct {
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -75,6 +75,7 @@ func (a Actor) Exports() []interface{} {
 		23:                        a.ChangeOwnerAddress,
 		24:                        a.DisputeWindowedPoSt,
 		25:                        a.PreCommitSectorBatch,
+		26:                        a.ProveCommitAggregate,
 	}
 }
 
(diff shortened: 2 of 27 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -403,33 +403,46 @@ func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *abi.Empty
 //	DealIDs    []abi.DealID
 //	SectorType abi.RegisteredSealProof
 //}
-type ComputeDataCommitmentParams = market0.ComputeDataCommitmentParams
+//type ComputeDataCommitmentParams = market0.ComputeDataCommitmentParams
```
anorth, line 406: Delete line

actors/builtin/market/market_actor.go
```diff
@@ -403,33 +403,46 @@ func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *abi.Empty
...
+	SectorType abi.RegisteredSealProof
+}
+
+type ComputeDataCommitmentParams struct {
+	Inputs []*SectorDataCommitmentInputs
+}
 
-func (a Actor) ComputeDataCommitment(rt Runtime, params *ComputeDataCommitmentParams) *cbg.CborCid {
+type ComputeDataCommitmentReturn struct {
+	CommDs []*cbg.CborCid
```
anorth, line 417: Can we avoid the pointer here? It was necessary when it was a bare return value, in order to match the exported method prototype, but I don't think is needed now.

actors/builtin/miner/miner_actor.go
```diff
@@ -902,85 +1025,75 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
...
 			// Check (and activate) storage deals associated to sector. Abort if checks failed.
 			// TODO: we should batch these calls...
 			// https://github.com/filecoin-project/specs-actors/issues/474
 			code := rt.Send(
 				builtin.StorageMarketActorAddr,
 				builtin.MethodsMarket.ActivateDeals,
 				&market.ActivateDealsParams{
-					DealIDs:      precommit.Info.DealIDs,
-					SectorExpiry: precommit.Info.Expiration,
+					DealIDs:      preCommit.Info.DealIDs,
```
anorth, line 1049: I suggest that changing capitalization style here isn't worthwhile, unless you're going to do so throughout the whole file (in another PR). We could debate the capitalisation but I'd end up agreeing with you so long as it was consistent everywhere.

anorth (commented):
Partial review so far.

This PR could have been a number of smaller ones. The data commitment changes are fine, they could land if in a separate PR.

actors/builtin/miner/policy.go
```diff
@@ -171,11 +173,11 @@ var MaxProveCommitDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
...
-	abi.RegisteredSealProof_StackedDrg32GiBV1_1:  builtin.EpochsInDay + PreCommitChallengeDelay, // PARAM_SPEC
-	abi.RegisteredSealProof_StackedDrg2KiBV1_1:   builtin.EpochsInDay + PreCommitChallengeDelay,
-	abi.RegisteredSealProof_StackedDrg8MiBV1_1:   builtin.EpochsInDay + PreCommitChallengeDelay,
-	abi.RegisteredSealProof_StackedDrg512MiBV1_1: builtin.EpochsInDay + PreCommitChallengeDelay,
-	abi.RegisteredSealProof_StackedDrg64GiBV1_1:  builtin.EpochsInDay + PreCommitChallengeDelay,
+	abi.RegisteredSealProof_StackedDrg32GiBV1_1:  6*builtin.EpochsInDay + PreCommitChallengeDelay, // PARAM_SPEC
+	abi.RegisteredSealProof_StackedDrg2KiBV1_1:   6*builtin.EpochsInDay + PreCommitChallengeDelay,
+	abi.RegisteredSealProof_StackedDrg8MiBV1_1:   6*builtin.EpochsInDay + PreCommitChallengeDelay,
+	abi.RegisteredSealProof_StackedDrg512MiBV1_1: 6*builtin.EpochsInDay + PreCommitChallengeDelay,
+	abi.RegisteredSealProof_StackedDrg64GiBV1_1:  6*builtin.EpochsInDay + PreCommitChallengeDelay,
```
anorth, line 180: I think this needs to be guarded by a network version check, else nodes will disagree about the handling of a late ProveCommitSector. Add a `MaxProveCommitDurationVxx` and abstract access to it through a function parameterised by network version (see examples above).

actors/builtin/miner/policy.go
```diff
@@ -306,3 +308,22 @@ func RewardForDisputedWindowPoSt(proofType abi.RegisteredPoStProof, disputedPowe
...
+// Return gas cost of verifying `batchSize` aggregate PoReps. Batch sizes round up.
+func AggregatePoRepVerifyGas(batchSize int) (int64, error) {
+	steps := aggregatePoRepBatchStepSize()
+	i := sort.SearchInts(steps, batchSize)
+	gasTable := aggregatePoRepBatchGasTable()
+	if i >= len(gasTable) {
+		return 0, xerrors.Errorf("batchSize %d unsupported by gas table", batchSize)
+	}
+	return gasTable[i], nil
+}
```
anorth, line 329: As noted in miner actor, I don't think this belongs in actors.

actors/builtin/power/policy.go
```diff
@@ -9,5 +9,4 @@ const ConsensusMinerMinMiners = 4 // PARAM_SPEC
 //
 // This limits the number of proof partitions we may need to load in the cron call path.
 // Onboarding 1EiB/year requires at least 32 prove-commits per epoch.
-const MaxMinerProveCommitsPerEpoch = 200 // PARAM_SPEC
-
+const MaxMinerProveCommitsPerEpoch = 1000 // PARAM_SPEC
```
anorth, line 12: This needs a version guard too.

Will this only apply to the old non-batch proof? I hope we're planning to deprecate that, in which case this will probably never be hit and the effort to parameterise it not worthwhile.

Edit: oh is this just for testing? Hmm, maybe you need to spread the proofs out over a few epochs (which is realistic anyway.

actors/builtin/power/power_actor.go
```diff
@@ -321,6 +322,20 @@ func (a Actor) CurrentTotalPower(rt Runtime, _ *abi.EmptyValue) *CurrentTotalPow
...
+// ErrForbidden indicates the claim is not found in claims table
+func (a Actor) CallerHasClaim(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
+	rt.ValidateImmediateCallerAcceptAny()
+	minerAddr := rt.Caller()
+
+	var st State
+	rt.StateReadonly(&st)
+
+	validateMinerHasClaim(rt, st, minerAddr)
+	return nil
```
anorth, line 336: I know you only need the boolean here, but it seems shortsighted not to return a useful value here, like some claim information. Otherwise we're as likely as not to need to add another method for that in the future.

actors/runtime/proof/verify.go
```diff
@@ -23,6 +25,25 @@ import (
 //}
 type SealVerifyInfo = proof0.SealVerifyInfo
 
+type AggregateSealVerifyInfo struct {
+	Number                abi.SectorNumber
+	DealIDs               []abi.DealID
```
anorth, line 30: @dignifiedquire This structure represents the info passed to proof verification. I'm curious as to why we need to pass deal ids here. Is this actually used, or are we plumbing useless data? AFAIK the proof verification routine can't use these to look up any information.

actors/test/committed_capacity_scenario_test.go
```diff
@@ -1,4 +1,4 @@
-package test_test
+package test
```
anorth, line 1: 🎉 

But also, some of this noise would have been better left for a separate PR.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/filecoin-project/go-amt-ipld/v3 v3.0.0
 	github.com/filecoin-project/go-bitfield v0.2.3
 	github.com/filecoin-project/go-hamt-ipld/v3 v3.0.1
-	github.com/filecoin-project/go-state-types v0.1.0
+	github.com/filecoin-project/go-state-types v0.1.1-0.20210504124540-e5e035a59744
```
anorth, line 10: This probably needs an upstream release.

support/mock/mockrt.go
```diff
@@ -660,6 +667,30 @@ func (rt *Runtime) BatchVerifySeals(vis map[addr.Address][]proof.SealVerifyInfo)
 	return nil, nil
 }
 
+func (rt *Runtime) ExpectAggregateVerifySeals(agg proof.AggregateSealVerifyProofAndInfos, err error) {
+	rt.expectAggregateVerifySeals = &expectAggregateVerifySeals{
+		agg.Infos, agg.Proof, err,
+	}
+}
```
anorth, line 674: Nit: this file is organised with all the expectation methods grouped together, and the interface methods grouped separately. Follow that grouping for now, and re-arrange everything in a separate PR if you wish.

support/mock/mockrt.go
```diff
@@ -660,6 +667,30 @@ func (rt *Runtime) BatchVerifySeals(vis map[addr.Address][]proof.SealVerifyInfo)
...
+		}
+		for i, expVI := range exp.inSVIs {
+			if agg.Infos[i].SealedCID != expVI.SealedCID {
+				rt.failTest("sealed cid does not match")
+			}
+			if agg.Infos[i].UnsealedCID != expVI.UnsealedCID {
+				rt.failTest("unsealed cid does not match")
+			}
+		}
+	}
```
anorth, line 690: ... else fail test now with unexpected call to `VerifyAggregateSeals`.

support/mock/mockrt.go
```diff
@@ -660,6 +667,30 @@ func (rt *Runtime) BatchVerifySeals(vis map[addr.Address][]proof.SealVerifyInfo)
...
+
+func (rt *Runtime) VerifyAggregateSeals(agg proof.AggregateSealVerifyProofAndInfos) error {
+	exp := rt.expectAggregateVerifySeals
+	if exp != nil {
+		if len(agg.Infos) != len(exp.inSVIs) {
+			rt.failTest("length mismatch, expected: %v, actual: %v", exp.inSVIs, agg.Infos)
+		}
+		for i, expVI := range exp.inSVIs {
+			if agg.Infos[i].SealedCID != expVI.SealedCID {
+				rt.failTest("sealed cid does not match")
```
anorth, line 684: Include the expected and actual in the message

support/mock/mockrt.go
```diff
@@ -938,10 +973,6 @@ func (rt *Runtime) Verify() {
 		rt.failTest("missing expected batch verify seals with %v", rt.expectBatchVerifySeals)
 	}
```
anorth, line 974: Insert failures if
- `expectAggregateVerifySeals` is not nil
- `gasCharged` is not empty

actors/test/commit_post_test.go
```diff
@@ -331,3 +333,307 @@ func TestCommitPoStFlow(t *testing.T) {
 		assert.True(t, networkStats.TotalPledgeCollateral.GreaterThan(big.Zero()))
 	})
 }
+
+func preCommitSectors(t *testing.T, v *vm.VM, batchSize int, worker, mAddr address.Address, sealProof abi.RegisteredSealProof) ([]abi.SectorNumber, []*miner.SectorPreCommitOnChainInfo) {
```
anorth, line 337: Nit: push the helpers to the bottom of the file, so the test cases come first and detail emerges top-down.

actors/test/commit_post_test.go
```diff
@@ -331,3 +333,307 @@ func TestCommitPoStFlow(t *testing.T) {
...
+	sectorNumberBase := 100
+	precommits := make([]*miner.SectorPreCommitOnChainInfo, 0)
```
anorth, line 339: If it's going to be empty, use `precommits := []*miner.SectorPreCommitOnChainInfo{}`

anorth (changes requested):

actors/builtin/power/policy.go
ZenGround0: Yeah this was just for testing that doesn't need to be merged.

actors/builtin/market/market_actor.go
```diff
@@ -403,33 +403,46 @@ func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *abi.Empty
...
-type ComputeDataCommitmentParams = market0.ComputeDataCommitmentParams
+//type ComputeDataCommitmentParams = market0.ComputeDataCommitmentParams
+type SectorDataCommitmentInputs struct {
+	DealIDs    []abi.DealID
+	SectorType abi.RegisteredSealProof
+}
+
+type ComputeDataCommitmentParams struct {
+	Inputs []*SectorDataCommitmentInputs
+}
```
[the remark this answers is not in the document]
ZenGround0: Going with `SectorDataSpec`

actors/builtin/miner/miner_actor.go
```diff
@@ -809,6 +810,114 @@ func (a Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *abi.E
...
+	builtin.RequireNoErr(rt, err, exitcode.ErrSerialization, "failed to marshal address for seal verification challenge")
+
+	for i, precommit := range precommits {
+		msd, ok := MaxProveCommitDuration[precommit.Info.SealProof]
+		if !ok {
+			rt.Abortf(exitcode.ErrIllegalState, "no max seal duration for proof type: %d", precommit.Info.SealProof)
+		}
+		proveCommitDue := precommit.PreCommitEpoch + msd
+		if rt.CurrEpoch() > proveCommitDue {
+			rt.Abortf(exitcode.ErrIllegalArgument, "commitment proof for %d too late at %d, due %d", precommit.Info.SectorNumber, rt.CurrEpoch(), proveCommitDue)
```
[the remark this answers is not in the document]
ZenGround0: > I assume we need the precommit information for all sectors in order to validate the proof. Could we note the sector as late, validate the proof, then drop the late sectors before the call to confirmSectorProofsValid?

I am unsure whether it's worth it for only between 0 and 30 minutes of buffer only after the first precommits have expired.  But I think this will be a clean change and therefore worth it.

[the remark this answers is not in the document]
anorth: Yes, I think we should adjust the precommit cleanup too, either by doing it less often (e.g. daily, on deadline index == 0) or subtracting a day or so from the epoch that is considered expired, or both.

actors/builtin/miner/miner_actor.go
```diff
@@ -809,6 +810,114 @@ func (a Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *abi.E
...
+	AggregateProof []byte
+}
+
+const MaxAggregatedSectors = 819
+const MinAggregatedSectors = 1
+const MaxAggregateProofSize = 192000
+
+func (a Actor) ProveCommitAggregate(rt Runtime, params *ProveCommitAggregateParams) *abi.EmptyValue {
+	rt.ValidateImmediateCallerAcceptAny()
+	checkClaimExists(rt)
```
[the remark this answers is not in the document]
anorth: Hmm, ok, I remember that. But this check will prevent us deprecating non-aggregate proofs, because new miners have no claim. It means Lotus and all clients need to maintain their implementation of single prove commit, rather than switch over and drop the legacy. Can we mitigate the scenario a different way?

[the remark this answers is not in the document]
ZenGround0: I think this is not an issue because new miners will have a claim unless they are removed in the case of deferredCronEvent error.  If new miners didn't have claims they wouldn't be able to do single prove commit either and we would have a bigger problem since this check is just duplicating the current state of single prove commit.

actors/runtime/proof/verify.go
cryptonemo: @magik6k will have the best insight here, as we don't directly use deal information in the aggregation API

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
ZenGround0: I'm going to leave this whole behavior change to a follow-up PR

actors/test/commit_post_test.go
```diff
@@ -331,3 +333,307 @@ func TestCommitPoStFlow(t *testing.T) {
...
+	proveTime := v.GetEpoch() + miner.MaxProveCommitDuration[sealProof]
+	v, _ = vm.AdvanceByDeadlineTillEpoch(t, v, minerAddrs.IDAddress, proveTime)
+
+	//
+	// prove and verify
+	//
+	v, err = v.WithEpoch(proveTime)
+	require.NoError(t, err)
+	for i := 0; i <= batchSize-1; i++ {
+		// Prove commit sector after max seal duration
```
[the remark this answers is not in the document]
ZenGround0: Got it. No, the intent is to be past the min seal duration, I'll update comment.

anorth (approved):

actors/runtime/proof/verify.go
anorth: No I'm pretty sure it's nothing to do with Lotus. This structure exists to encapsulate the data passed directly to proofs, via the ffi library. The non-aggregate version goes [directly to the VerifySeal FFI method](https://github.com/filecoin-project/filecoin-ffi/blob/master/proofs.go#L27). The data here was dictated to us by proofs team, though for dependency reasons the structure is defined in actors.

It looks like the deal IDs are indeed unused (not passed to Rust) and didn't need to be here, which we probably overlooked in a rush last year. It might not be worth the effort to remove from the non-aggregate version (which I hope we'll deprecate entirely soon), but @ZenGround0 let's not add them here until something forces us to.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
ZenGround0: This is a great point. I'm going to back out of checking the claim here.

actors/builtin/miner/miner_actor.go
```diff
@@ -74,6 +74,7 @@ func (a Actor) Exports() []interface{} {
 		22:                        a.RepayDebt,
 		23:                        a.ChangeOwnerAddress,
 		24:                        a.DisputeWindowedPoSt,
+		25:                        a.ProveCommitAggregate,
```
anorth, line 77: @ZenGround0 you'll need to rebase against the new method in #1407, now landed.

PR #1383, opened 2021-03-04
test: remove side-effects from runtime builder

Stebalien:
Otherwise, reusing the builder can have strange side-effects when running multiple tests.

### DIFF
--- a/support/mock/builder.go
+++ b/support/mock/builder.go
@@ -17,10 +17,9 @@ type RuntimeBuilder struct {
 }
 
 // Initializes a new builder with a receiving actor address.
-func NewBuilder(ctx context.Context, receiver addr.Address) RuntimeBuilder {
+func NewBuilder(receiver addr.Address) RuntimeBuilder {
 	var b RuntimeBuilder
 	b.add(func(rt *Runtime) {
-		rt.ctx = ctx
 		rt.receiver = receiver
 	})
 	return b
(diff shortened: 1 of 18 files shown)


support/mock/builder.go
```diff
@@ -13,16 +13,25 @@ import (
...
+		rt.receiver = receiver
+	})
+	return b
+}
+
+// Builds a new runtime object with the configured values.
+func (b RuntimeBuilder) Build(t testing.TB) *Runtime {
 	m := &Runtime{
-		ctx:               ctx,
+		ctx:               context.Background(),
```
acruikshank, line 32: This makes sense given the way the previous version behaved, but adding ctx as a parameter to `Build` could stave off some test problems in the future if this context is ever used.

acruikshank (approved):
Looks good. All the changes look seem to be big improvements.

support/mock/builder.go
Stebalien: Hm. I agree. That's the right place to put the context.

Stebalien: Actually, I'm just going to make this construct the context internally and return a cancel function. That's the "least easy to mess up" solution.

Stebalien: That also avoids passing a confusing context. If I pass it as a parameter, it _looks_ like canceling the context should cancel the build operation.

Stebalien: Actually, I just dropped it all together. Instead, I cancel the context when the test ends.

PR #1384, opened 2021-03-10
Replace sector extensiblility by lifetime setting

steven004:
This PR is to change the sector extension disability to the sector maximum lifetime limitation, currently the change just reflect on the V1 proof-type sectors as per the proposal in https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0014.md

What's more, this change also simplified the logic of the limitation of sector extension, could apply to further proof types too.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1171,6 +1171,10 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors in deadline %v partition %v", dlIdx, decl.Partition)
 				newSectors := make([]*SectorOnChainInfo, len(oldSectors))
 				for i, sector := range oldSectors {
+					if !CanExtendSealProofType(sector.SealProof, rt.NetworkVersion()) {
+						rt.Abortf(exitcode.ErrForbidden, "cannot extend expiration for sector %v with unsupported seal type %v",
+							sector.SectorNumber, sector.SealProof)
+					}
 					// This can happen if the sector should have already expired, but hasn't
 					// because the end of its deadline hasn't passed yet.
 					if sector.Expiration < currEpoch {
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -1171,10 +1171,6 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors in deadline %v partition %v", dlIdx, decl.Partition)
 				newSectors := make([]*SectorOnChainInfo, len(oldSectors))
 				for i, sector := range oldSectors {
-					if !CanExtendSealProofType(sector.SealProof) {
-						rt.Abortf(exitcode.ErrForbidden, "cannot extend expiration for sector %v with unsupported seal type %v",
-							sector.SectorNumber, sector.SealProof)
-					}
 					// This can happen if the sector should have already expired, but hasn't
 					// because the end of its deadline hasn't passed yet.
 					if sector.Expiration < currEpoch {
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -2124,7 +2124,7 @@ func validateExpiration(rt Runtime, activation, expiration abi.ChainEpoch, sealP
 	}
 
 	// total sector lifetime cannot exceed SectorMaximumLifetime for the sector's seal proof
-	maxLifetime, err := builtin.SealProofSectorMaximumLifetime(sealProof)
+	maxLifetime, err := builtin.SealProofSectorMaximumLifetime(sealProof, rt.NetworkVersion())
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "unrecognized seal proof type %d", sealProof)
 	if expiration-activation > maxLifetime {
 		rt.Abortf(exitcode.ErrIllegalArgument, "invalid expiration %d, total sector lifetime (%d) cannot exceed %d after activation %d",
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -3020,7 +3020,7 @@ func TestExtendSectorExpiration(t *testing.T) {
 		// and prove it once to activate it.
 		advanceAndSubmitPoSts(rt, actor, sector)
 
-		maxLifetime, err := builtin.SealProofSectorMaximumLifetime(sector.SealProof, network.Version0)
+		maxLifetime, err := builtin.SealProofSectorMaximumLifetime(sector.SealProof, network.VersionMax)
 		require.NoError(t, err)
 
 		st := getState(rt)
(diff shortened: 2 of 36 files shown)


steven004:
This is a candidate for https://github.com/filecoin-project/lotus/issues/5786

actors/builtin/miner/miner_actor.go
```diff
@@ -1171,10 +1171,6 @@ func (a Actor) ExtendSectorExpiration(rt Runtime, params *ExtendSectorExpiration
 				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors in deadline %v partition %v", dlIdx, decl.Partition)
 				newSectors := make([]*SectorOnChainInfo, len(oldSectors))
 				for i, sector := range oldSectors {
-					if !CanExtendSealProofType(sector.SealProof) {
```
ZenGround0, line 1174: This check and abort needs to stay on actors v3 for network version 10 to allow this code to sync network version 10 state changes.  The right path forward is to guard the new validation-free code path with a check that `rt.NetworkVersion >= network.Version(11)`

actors/builtin/miner/miner_test.go
```diff
@@ -3250,40 +3250,6 @@ func TestExtendSectorExpiration(t *testing.T) {
 		})
 		actor.checkState(rt)
 	})
-
-	t.Run("prevents extending old seal proof types", func(t *testing.T) {
```
ZenGround0, line 3254: We should keep this test, check that the old behavior applies to network version 10 and then check that v1 seal proof types *can* be extended up to max of 540 days in version 11.

We should also check that v1 seal proof types cannot be extended beyond 540 days.

actors/builtin/miner/policy.go
```diff
@@ -146,19 +146,6 @@ func CanPreCommitSealProof(s abi.RegisteredSealProof, nv network.Version) bool {
 	return ok
 }
 
-// List of proof types for which sector lifetime may be extended.
```
ZenGround0, line 149: These will need to stay to support the version 10 code path.  (See comment above)

actors/builtin/sector.go
```diff
@@ -7,43 +7,47 @@ import (
...
+// 	according to https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0014.md
+const EpochsIn540Days = stabi.ChainEpoch(540 * EpochsInDay)
+
+// For V1_1 Stacked DRG sectors, the max is 5 years
 const EpochsInFiveYears = stabi.ChainEpoch(5 * EpochsInYear)
 
 var SealProofPolicies = map[stabi.RegisteredSealProof]*SealProofPolicy{
 	stabi.RegisteredSealProof_StackedDrg2KiBV1: {
-		SectorMaxLifetime:      EpochsInFiveYears,
+		SectorMaxLifetime: EpochsIn540Days,
```
ZenGround0, line 22: We *might* be able to get away with this.  Without looking into this in depth I'm concerned that SectorMaxLifetime of v1 seal proofs somehow made its way into the code path of network version 10 v3 actors (current mainnet).  If that's the case this change could cause different behavior and prevent this code from syncing version 10 state transitions.  

It might be the case that the current extension ban in v10 has prevented this field from ever being read by network version 10 actors code, in which case this will work.  I will dig in deeper and provide a more definitive suggestion soon.

actors/builtin/miner/miner_actor.go
ZenGround0: Or alternatively guard the removed code path with `rt.NetworkVersion < network.Version(11)`

actors/builtin/miner/policy.go
ZenGround0: These names can change to indicate they are only in scope for network version 10 and lower.  i.e. `ExtensibleProofTypesBeforeNV11`, `CanExtendSealProofTypeBeforeNV11`

ZenGround0 (changes requested):

actors/builtin/sector.go
Stebalien: I agree this is probably fine, but I also agree that it's safer and easier to reason about to just have two maps and add a function that takes the current network version and returns the policy. I'd prefer the latter approach.

anorth: I concur

actors/builtin/miner/miner_actor.go
steven004: added rt.NetworkVersion() into the CanExtendSealProofType to solve this.

actors/builtin/sector.go
ZenGround0: @steven004 this comment is not yet resolved.  The recommendation after discussion here and on slack is that we have two seal proof policy maps and use a function taking the network version to switch between them whereever we read this.

steven004: Sorry for missing this. Now it should be solved. please check.

actors/builtin/miner/policy.go
```diff
@@ -147,16 +147,19 @@ func CanPreCommitSealProof(s abi.RegisteredSealProof, nv network.Version) bool {
...
 	abi.RegisteredSealProof_StackedDrg32GiBV1_1: {},
 	abi.RegisteredSealProof_StackedDrg64GiBV1_1: {},
 }
 
 // Checks whether a seal proof type is supported for new miners and sectors.
-func CanExtendSealProofType(s abi.RegisteredSealProof) bool {
-	_, ok := ExtensibleProofTypes[s]
-	return ok
+func CanExtendSealProofType(s abi.RegisteredSealProof, nv network.Version) bool {
+	if nv >= network.Version7 && nv <= network.Version10 {
```
ZenGround0, line 158: nit: since Version7 is never processed by actors v3 the `nv >= network.Version7` is redundant and can be removed.

actors/builtin/sector.go
```diff
@@ -2,48 +2,88 @@ package builtin
...
 )
 
 // Policy values associated with a seal proof type.
 type SealProofPolicy struct {
-	SectorMaxLifetime      stabi.ChainEpoch
+	SectorMaxLifetime stabi.ChainEpoch
 }
 
-// For all Stacked DRG sectors, the max is 5 years
+// For V1 Stacked DRG sectors, the max is 540 days since Network  Version 11
```
ZenGround0, line 14: nit: extra space between Network and Version

actors/builtin/miner/miner_test.go
```diff
@@ -3020,7 +3020,7 @@ func TestExtendSectorExpiration(t *testing.T) {
 		// and prove it once to activate it.
 		advanceAndSubmitPoSts(rt, actor, sector)
 
-		maxLifetime, err := builtin.SealProofSectorMaximumLifetime(sector.SealProof)
+		maxLifetime, err := builtin.SealProofSectorMaximumLifetime(sector.SealProof, network.Version0)
```
ZenGround0, line 3023: It should not matter in this case but typically unit tests adjust to cover the latest network version so best practice is to replace `network.Version0` with `network.VersionMax`

ZenGround0 (approved):
LGTM will merge to master if FIP-0014 becomes accepted

ZenGround0:
@steven004 since polling is showing strong support for this I am expecting FIP-0014 will pass. Would you mind rebasing in preparation for merge?

actors/builtin/miner/policy.go
steven004: I would like to keep this though it is redundant for better understanding. It matches the setting in different network versions.

steven004:
@ZenGround0 , to make the code cleaner and rebase on the latest master version since there are too many minor commits exist, I created a new PR #1388 to replace this one. #1388 is exactly the same as this one from code change point of view, please review and approve that one, as I am closing this. Thanks.

PR #1388, opened 2021-03-29
Replace sector extensiblility by lifetime setting

steven004:
This PR is to change the sector extension disability to the sector maximum lifetime limitation, currently the change just reflect on the V1 proof-type sectors as per the proposal in https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0014.md

What's more, this change also simplified the logic of the limitation of sector extension, could apply to further proof types too.

_Note: this PR is to replace #1384  using just one commit to make it cleaner._


steven004:
This is to support filecoin-project/lotus#5786

ZenGround0 (approved):

ZenGround0:
Speculatively merging under the assumption that FIP-0014 will be accepted.
This does not mean FIP-0014 is accepted yet, it is not.  While I expect it to be accepted I will revert this PR if I turn out to be wrong.

PR #1390, opened 2021-03-31
feat: Allow VerifiedClient DataCap to be topped up

### LINKED ISSUE #1325: Clients who apply for additional DataCap need to use a new address every time
In the current implementation of spec-actors we require Clients to treat DataCap as a "one time" credit card - per address you can only receive DataCap once. 

Practically, this introduces complexity for normal clients, who as a result need to manage which address of theres has DataCap, which addresses have already _had_ DataCap, and so on. 

Especially for web3 apps (like Slate) where the end-user is in control of their keys (but enabling Slate to handle the deal making) - we run into a scenario where the web3 dev needs to manage swapping out addresses for DataCap allocation / deal making unnecessarily.

placer14:
Implements [FIP-0012](https://github.com/filecoin-project/FIPs/blob/5ab54f271a87b3c759c21dc36948c033f105ac4d/FIPS/fip-0012.md) and fixes #1325.

Thoughts on these changes:
- The mock runtime `addVerifiedClient` function is now overloaded to support an optional `expectedAllowance` arg to be used in the cases where VerifiedClients will now have more than what was initially allotted. This makes the verification of an `expectedAllownace` a concern on the caller rather than letting the function reflect on its internal before/after state to verify correctness. I'd personally prefer the latter to keep the call simple, but there were errors when I attempted to access the runtime map for the "before" state of the Client's DataCap allowance. (An example of that is at https://gist.github.com/placer14/f2134a4064373607da44b48b4f338d69.)
- This change will mix all Verifier's allowances together for each Client address. If there were ever a need to know how DataCap is used on a per-Verifier basis, we would be eliminating that option here.
- The unit test which protects that a VerifiedClient can be referred to be its ID and non-ID addr is also protecting its ability to be topped-up. If it's preferred that these behaviors are protected by separate tests, I can, but felt the redundancy was unnecessary and left them combined.
- This change does not limit the DataCap upper bound and expects that DataCap boundaries are already protected where necessary.

### DIFF
--- a/actors/builtin/verifreg/verified_registry_test.go
+++ b/actors/builtin/verifreg/verified_registry_test.go
@@ -312,14 +312,14 @@ func TestAddVerifiedClient(t *testing.T) {
 
 		assert.EqualValues(t, clientAllowance, ac.getClientCap(rt, clientIdAddr))
 
-		// adding another verified client with the same ID address now fails
+		// adding another verified client with the same ID address increments
+		// the data cap which has already been granted
 		c2 := mkClientParams(clientIdAddr, clientAllowance)
 		verifier = mkVerifierParams(verifierAddr, allowance)
 		ac.addVerifier(rt, verifier.Address, verifier.Allowance)
 
-		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
-			ac.addVerifiedClient(rt, verifier.Address, c2.Address, c2.Allowance)
-		})
+		expectedAllowance := big.Add(c1.Allowance, c2.Allowance)
+		ac.addVerifiedClient(rt, verifier.Address, c2.Address, c2.Allowance, expectedAllowance)
 		ac.checkState(rt)
 	})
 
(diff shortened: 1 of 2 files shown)


actors/builtin/verifreg/verified_registry_test.go
```diff
@@ -825,7 +807,19 @@ func (h *verifRegActorTestHarness) generateAndAddVerifierAndVerifiedClient(rt *m
 	h.addVerifiedClient(rt, verifier.Address, client.Address, client.Allowance)
 }
 
-func (h *verifRegActorTestHarness) addVerifiedClient(rt *mock.Runtime, verifier, client address.Address, allowance verifreg.DataCap) {
+func (h *verifRegActorTestHarness) addVerifiedClient(rt *mock.Runtime, verifier, client address.Address, caps ...verifreg.DataCap) {
+	// caps[0] expects allowance to be set
+	// caps[1] (optional) expected allowance to be verified
+	// additional args are ignored
```
anorth, line 813: Fail the test rather than ignore an unexpected argument. Fail also if the slice has length zero.

Alternatively, I would be just as happy if the two args were explicit, and the usually-redundant expected cap was always explicitly passed.

anorth (approved):
Looks good to me, but please wait for approval from @ZenGround0

actors/builtin/verifreg/verified_registry_test.go
ZenGround0: I want to change this to two explict arguments.

ZenGround0 (commented):
Since your week of work is over @placer14 I'm taking this PR over.  I'm going to make some changes to testing but the code change looks good.

ZenGround0 (approved):

PR #1399, opened 2021-04-27
Feat/v5

ZenGround0:
Prepare actors for v5


ZenGround0:
`b134b8c` and `5a59ce3` have already passed review and landed on v4 before we applied the nv12 fix. There is no need to review those changes.

arajasek (approved):
LGTM

PR #1402, opened 2021-05-09
Batched miner state manipulations for sector number allocation

anorth:
This is broken out (1 of 2) of my work on #1241, batching pre-commits. During implementation I realised the essential similarity between allocating and masking sector numbers, so combined the state manipulation methods.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -309,7 +309,7 @@ const (
 )
 
 // Marks a set of sector numbers has having been allocated.
-// If allowCollision is false, fails if the set intersects with the sector numbers already allocated.
+// If policy is `DenyCollisions`, fails if the set intersects with the sector numbers already allocated.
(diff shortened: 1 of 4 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -301,54 +301,48 @@ func (st *State) QuantSpecForDeadline(dlIdx uint64) QuantSpec {
...
+	DenyCollisions  = CollisionPolicy(false)
+	AllowCollisions = CollisionPolicy(true)
+)
 
-func (st *State) MaskSectorNumbers(store adt.Store, sectorNos bitfield.BitField) error {
-	lastSectorNo, err := sectorNos.Last()
-	if err != nil {
-		return xc.ErrIllegalArgument.Wrapf("invalid mask bitfield: %w", err)
+// Marks a set of sector numbers has having been allocated.
+// If allowCollision is false, fails if the set intersects with the sector numbers already allocated.
```
ZenGround0, line 312:
```suggestion
// If policy is `DenyCollisions`, fails if the set intersects with the sector numbers already allocated.
```

ZenGround0 (approved):

PR #1404, opened 2021-05-10
Split out pre- and prove-commit tests to a new file

anorth:
Splits the sector pre-commit and prove-commit tests out of `miner_test.go` to `miner_commitments_test.go`, to make them a little more navigable. They both used shared harness and helpers that are for now still defined in `miner_test.go`.


anorth:
I did this with naive copy/paste, but with a bit of effort I can probably preserve most of the commit history on both files. I'll try that too.

anorth:
Ok, this branch now preserves attribution on both test files, following a method [documented here](https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904). You can't see it in the GitHub diff view (which squashes), but you can in the commit history.

It's a bit fragile though, relying on a merge commit in the history. I don't think it will survive squash-merging, or possibly even a rebase. I can recreate it if it's lost in a rebase against master prior to landing.

Since it won't survive a squash, we have to chose between:
- landing via merge: non-linear commit history, with some commits that don't build (due to duplication)
- landing via squash: lost attribution on one or the other of `miner_test.go` or `miner_commitments_test.go`

@ZenGround0 I'll take your preference here.

PR #1407, opened 2021-05-11
Implemented PreCommitSectorBatch method (FIP-0008)

anorth:
Implements a new `PreCommitSectorBatch` method, as specified by [FIP-0008](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0008.md) ([discussion](https://github.com/filecoin-project/FIPs/issues/25))

The non-batched `PreCommitSector` method remains, but should be considered deprecated. The implementation delegates internally to the batched method. This consumes less gas than before this change (due to removal of redundant checks in the batched method) but is slightly less efficient at runtime than a strictly-optimized singleton method could be. I expect it to be removed in the future.

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -660,7 +660,7 @@ type PreCommitSectorBatchParams struct {
 
 // Pledges the miner to seal and commit some new sectors.
 // The caller specifies sector numbers, sealed sector data CIDs, seal randomness epoch, expiration, and the IDs
-// of any storage marge deals contained in the sector data. The storage deal proposals must be already submitted
+// of any storage deals contained in the sector data. The storage deal proposals must be already submitted
 // to the storage market actor.
 // A pre-commitment may specify an existing committed-capacity sector that the committed sector will replace
 // when proven.
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -74,6 +74,7 @@ func (a Actor) Exports() []interface{} {
 		22:                        a.RepayDebt,
 		23:                        a.ChangeOwnerAddress,
 		24:                        a.DisputeWindowedPoSt,
+		25:                        a.PreCommitSectorBatch,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -715,93 +754,98 @@ func (a Actor) PreCommitSector(rt Runtime, params *PreCommitSectorParams) *abi.E
...
 	builtin.RequireNoErr(rt, err, ErrBalanceInvariantBroken, "balance invariants broken")
 	if needsCron {
-		newDlInfo := st.DeadlineInfo(rt.CurrEpoch())
+		newDlInfo := st.DeadlineInfo(currEpoch)
 		enrollCronEvent(rt, newDlInfo.Last(), &CronEventPayload{
 			EventType: CronEventProvingDeadline,
 		})
 	}
 
-	notifyPledgeChanged(rt, newlyVested.Neg())
```
ZenGround0, line 803: Nice

ZenGround0 (commented):
Still need to look at tests but LGTM so far

actors/builtin/miner/miner_test.go
```diff
@@ -4640,6 +4653,70 @@ func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.PreCommit
...
+	for i, sector := range params.Sectors {
+		sectorDeals[i] = market.SectorDeals{
+			SectorExpiry: sector.Expiration,
+			DealIDs:      sector.DealIDs,
+		}
+
+		if len(conf.sectorWeights) > i {
+			sectorWeights[i] = conf.sectorWeights[i]
+		} else {
+			sectorWeights[i] = market.SectorWeights{
```
ZenGround0, line 4678: suggestion: it would be nice to comment on the `sectorWeights` fields of the preCommitBatchConf that these weights are used for sectors 0 through `len(sectorWeights)` and not all need to be used.

actors/builtin/miner/miner_test.go
```diff
@@ -4640,6 +4653,70 @@ func (h *actorHarness) preCommitSector(rt *mock.Runtime, params *miner.PreCommit
...
+			Sectors: sectorWeights,
+		}
+		rt.ExpectSend(builtin.StorageMarketActorAddr, builtin.MethodsMarket.VerifyDealsForActivation, &vdParams, big.Zero(), &vdReturn, exitcode.Ok)
+	}
+	st := getState(rt)
+	if st.FeeDebt.GreaterThan(big.Zero()) {
+		rt.ExpectSend(builtin.BurntFundsActorAddr, builtin.MethodSend, nil, st.FeeDebt, nil, exitcode.Ok)
+	}
+
+	if first {
```
ZenGround0, line 4705: suggestions: if it's not too hard to do for this new method and it makes sense to you, you could add this param to the config.  I think that might be the place it belongs.

actors/builtin/miner/miner_commitment_test.go
```diff
@@ -326,20 +423,220 @@ func TestCommitments(t *testing.T) {
...
+			assert.True(t, totalDeposit.GreaterThan(big.Zero()))
+			assert.Equal(t, totalDeposit, st.PreCommitDeposits)
+
+			expirations := actor.collectPrecommitExpirations(rt, st)
+			expectedPrecommitExpiration := st.QuantSpecEveryDeadline().QuantizeUp(precommitEpoch + miner.MaxProveCommitDuration[actor.sealProofType] + 1)
+			assert.Equal(t, map[abi.ChainEpoch][]uint64{expectedPrecommitExpiration: sectorNoAsUints}, expirations)
+		})
+	}
+
+	t.Run("one bad apple ruins batch", func(t *testing.T) {
```
ZenGround0, line 589: 🍎

actors/builtin/miner/miner_commitment_test.go
```diff
@@ -397,27 +700,17 @@ func TestProveCommit(t *testing.T) {
...
-		// expect new onchain sector
-		sector := actor.getSector(rt, sectorNo)
 		sectorPower := miner.NewPowerPair(big.NewIntUnsigned(uint64(actor.sectorSize)), qaPower)
 
 		// expect deal weights to be transferred to on chain info
-		assert.Equal(t, onChainPrecommit.DealWeight, sector.DealWeight)
-		assert.Equal(t, onChainPrecommit.VerifiedDealWeight, sector.VerifiedDealWeight)
-
-		// expect activation epoch to be current epoch
-		assert.Equal(t, rt.Epoch(), sector.Activation)
```
ZenGround0, line 415: Why remove this check?

ZenGround0 (approved):

actors/builtin/miner/miner_actor.go
```diff
@@ -647,60 +648,97 @@ func (a Actor) DisputeWindowedPoSt(rt Runtime, params *DisputeWindowedPoStParams
 //}
 type PreCommitSectorParams = miner0.SectorPreCommitInfo
```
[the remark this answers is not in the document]
anorth: Actually, every version has aliased this `PreCommitSectorParams` onto `miner0.SectorPreCommitInfo`, e.g. [in v4](https://github.com/filecoin-project/specs-actors/blob/v4.0.0/actors/builtin/miner/miner_actor.go#L648). This is necessary to preserve type-compatibility in Lotus without extra effort. `PreCommitSectorParams` is the ugly special case here where a state object (which we usually redefine) was used as a parameter type.

I opened a discussion in Filecoin slack about how we might consider aliasing the state types too in the future.

actors/builtin/miner/miner_actor.go
```diff
@@ -643,65 +644,103 @@ func (a Actor) DisputeWindowedPoSt(rt Runtime, params *DisputeWindowedPoStParams
...
+		if set {
+			rt.Abortf(exitcode.ErrIllegalArgument, "duplicate sector number %d", precommit.SectorNumber)
+		}
+		sectorNumbers.Set(uint64(precommit.SectorNumber))
 
-	challengeEarliest := rt.CurrEpoch() - MaxPreCommitRandomnessLookback
-	if params.SealRandEpoch < challengeEarliest {
-		rt.Abortf(exitcode.ErrIllegalArgument, "seal challenge epoch %v too old, must be after %v", params.SealRandEpoch, challengeEarliest)
-	}
+		if !CanPreCommitSealProof(precommit.SealProof, nv) {
```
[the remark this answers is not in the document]
anorth: Agreed, I will follow up with that.

actors/builtin/miner/miner_actor.go
```diff
@@ -643,65 +644,103 @@ func (a Actor) DisputeWindowedPoSt(rt Runtime, params *DisputeWindowedPoStParams
...
+			rt.Abortf(exitcode.ErrIllegalArgument, "unsupported seal proof type %v at network version %v", precommit.SealProof, nv)
+		}
+		if precommit.SectorNumber > abi.MaxSectorNumber {
```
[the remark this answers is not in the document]
anorth: I would rather keep the policy logic here, so deleting the check in `AllocateSectorNumbers`.

actors/builtin/miner/miner_test.go
anorth: Easy, done.

actors/builtin/miner/miner_commitment_test.go
anorth: I moved it to L688 alongside the rest of basic sector state checks.

PR #1408, opened 2021-05-11
Add gas to test vm

### LINKED ISSUE #1266: Add gas accounting to scenario test VM
Gas is mostly a thing that happens outside the direct control of the actor code, in the VM, so the gas table etc are not part of actor code. However, gas usage is the critical resource that actor code execution consumes, so we should have good tooling to inspect, benchmark, and profile it. We expect to be putting quite significant effort into improving and optimising gas performance in the near term, so should be able to quantify our changes.

To this end, let's add gas tracking to the VM used for scenario tests so that we can replicate and analyse gas usage traces.

Scenario tests might then include an assertion about total gas usage, forcing explicit recognition and review of any changes.

ZenGround0:
Closes #1266 

I'm aiming for correctness to filecoin protocol / lotus vm now but it is not dangerous if we miss something in review.  All inconsistencies externalized in test vectors will register as failures when ran against lotus (which I will then debug and fix) 
before releasing them as conformance tests.


gen/gen.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/filecoin-project/specs-actors/v5/actors/builtin/system"
 	"github.com/filecoin-project/specs-actors/v5/actors/builtin/verifreg"
 	"github.com/filecoin-project/specs-actors/v5/actors/util/smoothing"
+	vm_test "github.com/filecoin-project/specs-actors/v5/support/vm"
```
anorth, line 19: This is weird. I see that the existing vm is all under this vm_test package, but I don't think that's helpful (we should reserve that for any testing _of_ the vm). Fixed in #1409.

support/agent/sim.go
```diff
@@ -441,7 +441,7 @@ type PowerTable struct {
 
 // VM interface allowing a simulation to operate over multiple VM versions
 type SimVM interface {
-	ApplyMessage(from, to address.Address, value abi.TokenAmount, method abi.MethodNum, params interface{}) (cbor.Marshaler, exitcode.ExitCode)
+	ApplyMessage(from, to address.Address, value abi.TokenAmount, method abi.MethodNum, params interface{}) (cbor.Marshaler, exitcode.ExitCode, int64)
```
anorth, line 444: This really deserves a struct return type now. At your option, though.

support/vm/invocation_context.go
```diff
@@ -33,6 +33,8 @@ import (
 
 var EmptyObjectCid cid.Cid
 
+const testGasLimit = 5_000_000_000
```
anorth, line 36: This should be a member of the `VM` struct, with this as `defaultGasLimit`.

support/vm/invocation_context.go
```diff
@@ -47,6 +49,10 @@ type invocationContext struct {
 	// Used for detecting modifications to state outside of transactions.
 	stateUsedObjs map[cbor.Marshaler]cid.Cid
 	stats         *CallStats
+	// Gas trackign fields
+	gasPrices    runtime.Pricelist
```
anorth, line 53: The fields in this struct are supposed to be specific to an inner call.

Prices don't change between inner calls, so move this to topLevelContext. I think the values below are intended to capture the whole top-level call, so move them too.

support/vm/invocation_context.go
```diff
@@ -109,6 +118,21 @@ func (ic *invocationContext) storeActor(actr *states.Actor) {
 	}
 }
 
+func (ic *invocationContext) chargeGas(gas runtime.GasCharge) {
```
anorth, line 121: Move/delegate this method to toplevel context too

support/vm/invocation_context.go
```diff
@@ -183,22 +214,31 @@ func (ic *invocationContext) StateTransaction(obj cbor.Er, f func()) {
...
 	return ic.Syscalls().ComputeUnsealedSectorCID(reg, pieces)
 }
 
 func (ic *invocationContext) VerifySeal(vi proof.SealVerifyInfo) error {
+	ic.chargeGas(ic.gasPrices.OnVerifySeal(vi))
 	return ic.Syscalls().VerifySeal(vi)
 }
 
 func (ic *invocationContext) BatchVerifySeals(vis map[address.Address][]proof.SealVerifyInfo) (map[address.Address][]bool, error) {
+	// no explicit gas charged
```
anorth, line 241: See https://github.com/filecoin-project/specs-actors/pull/1381#discussion_r629873262, I hope we can remove the special case.

support/vm/vm.go
```diff
@@ -64,6 +66,50 @@ type InternalMessage struct {
...
+		}
+	}
+
+	return &ChainMessage{
+		Version:    0,
+		From:       from,
+		To:         to,
+		Nonce:      nonce,
+		Value:      value,
+		GasLimit:   5_000_000_000,
```
anorth, line 105: Reference the default gas limit constant, or use a different value (since I don't think this value is actually checked).

support/vm/vm.go
```diff
@@ -64,6 +66,50 @@ type InternalMessage struct {
...
+		GasFeeCap:  big.Zero(),
+		GasPremium: big.Zero(),
```
anorth, line 107: I was guessing that the GasLimit above was set non-zero so as to better represent the serialized size of a chain message. If so, these should be non-zero too. If not, just set the unused GasLimit to zero to match

actors/runtime/runtime.go
```diff
@@ -231,3 +231,30 @@ type StateHandle interface {
 	// ```
 	StateTransaction(obj cbor.Er, f func())
 }
+
+type Pricelist interface {
```
anorth, line 235: I don't think this should be visible to actor code that imports `runtime`. Move it and `GasCharge` to the `vm` package.

support/vm/invocation_context.go
```diff
@@ -799,10 +853,7 @@ func (ic *invocationContext) replace(obj cbor.Marshaler) cid.Cid {
 	if !found {
 		ic.rt.Abortf(exitcode.ErrIllegalState, "failed to find actor %s for state", ic.msg.to)
 	}
-	c, err := ic.rt.store.Put(ic.rt.ctx, obj)
-	if err != nil {
-		ic.rt.Abortf(exitcode.ErrIllegalState, "could not save new state")
-	}
+	c := ic.StorePut(obj)
 	actr.Head = c
 	err = ic.rt.setActor(ic.rt.ctx, ic.msg.to, actr)
```
anorth, line 858: We need to charge gas for the write of the actor struct into the state tree.

anorth (commented):
Basically smooth, nice work.

support/vm/invocation_context.go
```diff
@@ -47,6 +49,10 @@ type invocationContext struct {
 	// Used for detecting modifications to state outside of transactions.
 	stateUsedObjs map[cbor.Marshaler]cid.Cid
 	stats         *CallStats
+	// Gas trackign fields
```
arajasek, line 52:
```suggestion
	// Gas tracking fields
```

arajasek:
Broadly looks good. Will look in a bit more detail after a rebase (it's currently a bit confusing to look at diff).

support/vm/vm.go
ZenGround0: Yes GasLimit is set non-zero for serialization accuracy purposes. We can't set it to zero because then we will hit SysErrOutOfGas immediately.  If we set GasFeeCap and GasPremium non-zero we will have to implement base fee burning and miner payments which I want to avoid.

support/vm/invocation_context.go
ZenGround0: This is intentionally not charging gas because as I understand the lotus vm code the filecoin protocol does not charge gas for reads and writes from the top level state tree.

ZenGround0:
Since we are all quite busy right now I went ahead and merged to unblock myself more quickly because existing review was thorough, this PR only impacts testing code, and gas calculation implementation errors will only become more apparent as we evaluate test vectors against lotus.

PR #1412, opened 2021-05-13
Feat/aggregate porep less harsh error

ZenGround0:
1. AggregateProveCommit does not fail if a subset of precommits exist on chain but are out of date (prove commit epoch greater than epoch due).  Instead all inputs are provided to the proof, but only the sectors not out of date are added to the state through confirmSectorsProofsValid.
2. PreCommitExpiries are now scheduled an additional 8 hours after the corresponding prove commit epoch is due. This provides much more time (8 hours vs 0-30 minutes) for aggregates including expired precommits to make it on chain.  This is a somewhat arbitrary choice, feedback on this would be great

TODO:
- [ ] land #1381 and rebase this on top before merging

### DIFF
--- a/actors/test/commit_post_test.go
+++ b/actors/test/commit_post_test.go
@@ -99,6 +99,10 @@ func TestCommitPoStFlow(t *testing.T) {
 	//
 
 	t.Run("missed prove commit results in precommit expiry", func(t *testing.T) {
+		// advance time to precommit expiry
+		expiryTime := proveTime + miner.PreCommitExpiryDelay
+		v, dlInfo := vm.AdvanceByDeadlineTillEpoch(t, v, minerAddrs.IDAddress, expiryTime)
+
 		// advanced one more deadline so precommit is late
 		tv, err := v.WithEpoch(dlInfo.Close)
 		require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


actors/test/commit_post_test.go
```diff
@@ -450,6 +454,80 @@ func TestMeasurePoRepGas(t *testing.T) {
...
+			{To: builtin.RewardActorAddr, Method: builtin.MethodsReward.ThisEpochReward},
+			{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.CurrentTotalPower},
+			{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.UpdatePledgeTotal},
+		},
+	}.Matches(t, v.LastInvocation())
+
+	balances := vm.GetMinerBalances(t, v, minerAddrs.IDAddress)
+	assert.True(t, balances.InitialPledge.GreaterThan(big.Zero()))
+	assert.True(t, balances.PreCommitDeposit.GreaterThan(big.Zero()))
+
```
anorth, line 528: I suggest rolling forward cron until the expired precommit is cleaned up, too

anorth (approved):
LGTM after clarifying language.

Out of curiosity, why did you elect to write a scenario test rather than a unit test for testing the ProveCommitAggregate with a expired sector?

ZenGround0:
> Out of curiosity, why did you elect to write a scenario test rather than a unit test for testing the ProveCommitAggregate with a expired sector?

With the upcoming test vectors project all scenario tests will now generate conformance tests so I now have a slight bias towards testing with scenario tests if it is not too hard to do.  In this case I was already warmed up writing similar tests from measuring aggregate porep gas so it seemed worth the extra effort.

PR #1421, opened 2021-05-19
Add check for allowed Window PoSt proof type at miner creation

anorth:
In #1345 we unintentionally lost a [check](https://github.com/filecoin-project/specs-actors/pull/1345/files#diff-d807293f55563b410c2d07719d2f38727016f41c6a17df8635500abc4070f74cL102) that prevented creation of new miners with proofs types that are not supported by the network. Such miners can't do any harm (they can never commit a sector) but it's still better to prevent their creation up front.


actors/builtin/miner/miner_actor.go
```diff
@@ -333,6 +337,10 @@ func (a Actor) SubmitWindowedPoSt(rt Runtime, params *SubmitWindowedPoStParams)
 	store := adt.AsStore(rt)
 	var st State
 
+	if !CanWindowPoStProof(params.Proofs[0].PoStProof) {
```
ZenGround0, line 340: params with 0 proofs should be valid up to this point so this can cause a panic. This check should go below the check that there is exactly 1 proof.

ZenGround0 (changes requested):

actors/builtin/miner/miner_actor.go
anorth: Thanks for catching that. It revealed a big hole in test coverage for SubmitWindowPost preconditions, which I have now filled in this PR.

ZenGround0 (approved):

PR #1422, opened 2021-05-20
Raise AddressedSectorsMax from 10,000 to 25,000.

### LINKED ISSUE #1415: Raise max partitions proven in SubmitWindowedPoSt
Currently, at most 4 partitions can be proven at once because _disputing_ this proof might need to touch each and every sector in the partitions.

We should consider raising this based on actual estimated gas usage:

1. Try raising this to 8, then maybe 16?
2. Determine gas savings for SubmitWindowedPoSt. This must be done with _on-chain_ data (ideally from a large miner) to get reasonable results:
    1. Take an existing sequence of SubmitWindowedPoSt messages (likely from the largest miner).
    2. Combine them (and concatenate proofs).
    3. No-op the actual proof verification (won't work).
    4. Run a gas estimation and compare.
3. Make sure these proofs can still be disputed.
    1. Locally dispute a window post proof to determine the cost.
    2. Multiply by 2 (8 partitions) and 4 (16 partitions) to make sure that gas still fits within a block (comfortably).

Once we have results, we can consider raising the max. Actually making this change should be trivial.

anorth:
This number bounds the amount of memory used or state loaded in a number of operations
such as declaring faults. A critical need is to limit the number of sectors that may
be addressed by a single WindowPoSt sumbission to ensure it can be safely disputed
later if fraudulent.

25000 sectors permits submitting (and disputing) up to 10 partitions at once
for both 32GiB (25,000/2,349 ~= 10.6) and 64GiB (25,000/2,300 ~= 10.8) sectors.

Empirically, disputing 10 partitions fits within the block gas limit >= 3 times over.

Closes #1415

### DIFF
--- a/actors/builtin/miner/miner_test.go
+++ b/actors/builtin/miner/miner_test.go
@@ -1352,12 +1352,9 @@ func TestWindowPost(t *testing.T) {
 
 		// Too many partitions.
 		rt.ExpectAbortContainsMessage(exitcode.ErrIllegalArgument, "too many partitions", func() {
-			partitions := []miner.PoStPartition{
-				{Index: pIdx, Skipped: bf()},
-				{Index: pIdx + 1, Skipped: bf()},
-				{Index: pIdx + 2, Skipped: bf()},
-				{Index: pIdx + 3, Skipped: bf()},
-				{Index: pIdx + 4, Skipped: bf()},
+			partitions := make([]miner.PoStPartition, 11)
+			for i, p := range partitions {
+				p.Index = pIdx + uint64(i)
 			}
 			params := miner.SubmitWindowedPoStParams{
 				Deadline:         dlInfo.Index,
--- a/actors/builtin/miner/policy.go
+++ b/actors/builtin/miner/policy.go
@@ -81,7 +81,7 @@ const DeclarationsMax = AddressedPartitionsMax
 
 // The maximum number of sector infos that can be loaded in a single invocation.
 // This limits the amount of state to be read in a single message execution.
-const AddressedSectorsMax = 10_000 // PARAM_SPEC
+const AddressedSectorsMax = 25_000 // PARAM_SPEC
 
 // Libp2p peer info limits.
 const (


ZenGround0 (approved):
> Empirically, disputing 10 partitions fits within the block gas limit >= 3 times over.

This reasoning makes me uneasy. There is no restriction on operations per block since each window post submission checks the limit independently.

However this PR doesn't make this worse as you could just as easily submit 8 wdpost messages with 4 proofs in a single block as 3 wdpost messages with 10 proofs.

Stebalien:
> This reasoning makes me uneasy. There is no restriction on operations per block since each window post submission checks the limit independently.

The issue is the dispute. I need to be able to submit a dispute somehow, even if I can't fit them all into a single block. That means each post needs to be independently disputable.


---

@anorth I'm slightly concerned about sector termination, but that's probably fine.

Stebalien (approved):

ZenGround0:
@anorth I think this is going to break your test [here](https://github.com/filecoin-project/specs-actors/pull/1421/files#diff-53806ced87bf190b7c5e9a8ef1a4cdbf772252bdfbc2a7dab9247c1589c3d86cR1353).  I'm planning to merge that PR first and then leave the fix for this PR.

ZenGround0:
> @anorth I'm slightly concerned about sector termination, but that's probably fine.

Suggestion: we could maintain current levels of safety without thinking more about this using two separate constants one for max sectors accessed by post one for max addressed by other methods.

anorth:
@Stebalien I infer your concern about terminating sectors to be the explicit `TerminateSectors` exported method. Sector termination does not happen in the `DisputeWindowPost` call tree. I do think this is ok since it is a directly-invoked method, where the caller will pay all gas. 

It may be that mismatches between gas cost and true computational cost make this a place to target chain validation attacks. It would be a good place for us to profile a workload targeting this (probably in Lotus, cc @magik6k).

PR #1429, opened 2021-05-20
Set ConsensusMinerMinPower to 10 TiB for all PoStProofPolicies

arajasek:
Lotus has been unintentionally overwriting this value to 10 TiB for all proof types since mainnet. We might as well reflect this behaviour in the actors code.

### DIFF
--- a/actors/builtin/power/power_test.go
+++ b/actors/builtin/power/power_test.go
@@ -97,11 +97,11 @@ func TestCreateMinerFailures(t *testing.T) {
 		rt, ac := basicPowerSetup(t)
 
 		createMinerParams := &power.CreateMinerParams{
-			Owner:                owner,
-			Worker:               owner,
-			WindowPoStProofType:  windowPoStProofType,
-			Peer:                 peer,
-			Multiaddrs:           mAddr,
+			Owner:               owner,
+			Worker:              owner,
+			WindowPoStProofType: windowPoStProofType,
+			Peer:                peer,
+			Multiaddrs:          mAddr,
 		}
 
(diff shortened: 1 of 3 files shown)


actors/builtin/power/power_test.go
```diff
@@ -441,15 +441,16 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
 
 		power64Unit, err := builtin.ConsensusMinerMinPower(abi.RegisteredPoStProof_StackedDrgWindow64GiBV1)
 		require.NoError(t, err)
-		assert.True(t, power64Unit.GreaterThan(powerUnit))
+
+		belowLimitUnit := big.Div(powerUnit, big.NewInt(2))
```
ZenGround0, line 445: Remove this subtest which is testing deprecated behavior

ZenGround0 (approved):

actors/builtin/power/power_test.go
arajasek: Is it deprecated? I agree that the test name is wrong / the test can perhaps be simplified, but the sequence of "assert well below minPower isn't aded" -> "assert 1 byte below minPower isn't added" -> "assert precisely at minPower is added" seems valuable to me (and a quick skim didn't reveal the same sequence being tested somewhere else).

ZenGround0: I was only skimming the other tests and they seemed to be testing something similar.  I trust your judgement.  If this test is sufficiently different then rename the subtest, remove any parts related to different proof types and fix the comment on line 438.

ZenGround0 (approved):

PR #1430, opened 2021-05-20
Set ConsensusMinerMinPower to 10 TiB for all PoStProofPolicies

arajasek:
Lotus has been unintentionally overwriting this value to 10 TiB for all proof types since mainnet. We might as well reflect this behaviour in the actors code.

### DIFF
--- a/actors/builtin/power/power_test.go
+++ b/actors/builtin/power/power_test.go
@@ -97,11 +97,11 @@ func TestCreateMinerFailures(t *testing.T) {
 		rt, ac := basicPowerSetup(t)
 
 		createMinerParams := &power.CreateMinerParams{
-			Owner:                owner,
-			Worker:               owner,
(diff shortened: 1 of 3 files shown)


actors/builtin/power/power_test.go
```diff
@@ -441,15 +441,16 @@ func TestPowerAndPledgeAccounting(t *testing.T) {
 
 		power64Unit, err := builtin.ConsensusMinerMinPower(abi.RegisteredPoStProof_StackedDrgWindow64GiBV1)
 		require.NoError(t, err)
-		assert.True(t, power64Unit.GreaterThan(powerUnit))
+
```
ZenGround0, line 444: Remove this subtest which is now testing deprecated behavior ("consensus power depends on proof type")

ZenGround0 (approved):

PR #1432, opened 2021-05-21
Add test vector generation into ApplyMessage calls

### LINKED ISSUE #1431: Test vector generation from scenario tests
Generate test vectors from scenario tests

- [x] 1. Introduce a test vector type that can generate valid test vectors conforming to [the schema](https://github.com/filecoin-project/test-vectors/blob/master/schema/schema.go)
- [x] 2. Build test vectors inside test vm ApplyMessage calls
- [x] 3. Serialize and write test vectors to file behind an environment variable flag
- [x] 4. Verify that test vectors correctly validate against lotus implementation

ZenGround0:
Closes #1431


.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - run:
           name: "Ensure we don't need to run 'make gen'"
           command: |
-            make gen && git diff --exit-code
+            make gen && go mod tidy && git diff --exit-code
```
ZenGround0, line 66: I don't understand why go.sum was changing in between but it was and causing failures on ci

go.sum
```diff
@@ -1,8 +1,33 @@
+github.com/AndreasBriese/bbloom v0.0.0-20180913140656-343706a395b7/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
+github.com/Kubuxu/go-os-helper v0.0.1/go.mod h1:N8B+I7vPCT80IcP58r50u4+gEEcsZETFUpAzWW2ep1Y=
```
ZenGround0, line 4: These testing dependencies are an unfortunate side effect of taking the path of least resistance when implementing car writing.  I would like to follow up be reimplementing stubs for unused dag service dependencies here to reduce them.

support/vm/invocation_context.go
```diff
@@ -62,6 +67,8 @@ type topLevelContext struct {
 	gasPrices    Pricelist
 	gasUsed      int64
 	gasAvailable int64
+	// Temporary field to workaround test-vector limitations
+	fakeSyscallsAccessed bool
```
ZenGround0, line 71: Both this and RandString's fragile dependence on outside defined schema are things I expect to cover in an upcoming issue about absorbing and moving forward the test-vector format into the specs-actors project.

ZenGround0:
This PR introduces three new make targets 

1. `make determinism-gen` -- by running scenario tests this creates a test-vector corpus underneath `test-vectors/determinism` which should not change without behavior changes and creates a digest of the corpus in test-vectors/determinism-check.
2. `make determinism-check` -- this creates the same test-vector corpus underneath `test-vectors/determinism` and regenerates the digest to make sure that state transitions in scenario tests match the recorded run.  This now runs on CI.
3. `make conformance-gen` -- this runs scenario tests and generates a subset of test-vectors from test state transitions that can serve as valid conformance tests across implementations.  It is a subset of the determinism corpus currently because the test-vector format cannot yet handle faking crypto syscalls.  The generated corpus is generated underneath `test-vectors/conformance`

The structure of the generated files is designed to help with debugging inconsistencies.  Vectors live in a directory structure that identifies the name of the test that generated them.  Filenames are of the format `<sha256 hash of json data>-<sending actor addr>-<receiving actor addr>-<actor name>-<executed method number>.json`.  All test-vectors are of the "message" class and cover the execution of a single message.

arajasek (approved):
Looks broadly good to me

support/vm/vector_gen.go
```diff
@@ -0,0 +1,407 @@
+package vm
```
anorth, line 1: I think we should loosen the coupling with the VM and move this to its own `support` package. You'll have to widen the public interface to the VM slightly, but I think that's fine. We should be able to use this in principle with a different VM.

support/vm/vector_gen.go
```diff
@@ -0,0 +1,407 @@
...
+// not supported -- noop
+func (a *adtBlockStoreForDAGService) HashOnRead(enabled bool) {}
+
+// Top level state tree
+
+const CurrentStateTreeVersion = 3
+
+type StateTreeVersion uint64
+
+type StateRoot struct {
```
anorth, line 384: This bit could possibly stay in the VM package.

anorth:
> The structure of the generated files is designed to help with debugging inconsistencies. Vectors live in a directory structure that identifies the name of the test that generated them. Filenames are of the format <sha256 hash of json data>-<sending actor addr>-<receiving actor addr>-<actor name>-<executed method number>.json. All test-vectors are of the "message" class and cover the execution of a single message.

This comment should really appear in the code somewhere.

go.sum
ZenGround0: Turns out we still need these even after implementing our own node.Getter

support/vm/vm.go
```diff
@@ -340,8 +343,23 @@ type MessageResult struct {
 	GasCharged int64
 }
 
-// ApplyMessage applies the message to the current state.
-func (vm *VM) ApplyMessage(from, to address.Address, value abi.TokenAmount, method abi.MethodNum, params interface{}) MessageResult {
+func (vm *VM) ApplyMessage(from, to address.Address, value abi.TokenAmount, method abi.MethodNum, params interface{}, info string) (MessageResult, error) {
```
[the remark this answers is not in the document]
ZenGround0: Returning errors instead of panicing seems like the way to go

I can add a testing helper `RequireApplyMessage(t, v, ...)` to make this much less annoying

support/vm/vector_gen.go
```diff
@@ -0,0 +1,407 @@
...
+		Meta: &metadata{
+			ID: tv.ID,
+			Gen: []generationData{
+				{Source: "specs-actors_test_auto_gen"},
+			},
+		},
+		CAR: tv.StartState,
+		Pre: &preconditions{
+			Variants: []variant{
+				{ID: defaultNetworkName, Epoch: int64(tv.Epoch), NetworkVersion: uint(tv.Version)},
```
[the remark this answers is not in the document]
ZenGround0: Excellent callout.  I did not exactly inherit this.  The main reasons for using this serialization type were to separate the complexity of what the actors generation vectors need from the full scope of the test vector format.  For example it is really nice to not have to think about having multiple test-vector variants, or any non-message class vectors at the actors generation level.

We can build good general purpose patterns for working with the full format to avoid relying on assumptions to make static shortcuts but redoing these here without taking on the full responsibility of the test-vector format would have been overkill.

My plan for the moment is to actually couple these types further into the specs-actors reference vm to prevent any external reliance on these internal shortcut assumptions.  Then proper decoupling can be done in one go along with introducing a generally/externally usable format: https://github.com/filecoin-project/specs-actors/issues/1455

support/vm/vector_gen.go
ZenGround0: I'm putting this off until the future where specs-actors takes full responsibility for the format: #1455.  Until then it is not intended for these vector types to be used externally so I will move them to internal types.

support/vm/vector_gen.go
```diff
@@ -0,0 +1,407 @@
...
+	StartStateTree cid.Cid
+	Message        *ChainMessage
+
+	Receipt      MessageResult
+	EndStateTree cid.Cid
+
+	// Runtime values
+	Epoch      abi.ChainEpoch
+	Version    network.Version
+	CircSupply abi.TokenAmount
```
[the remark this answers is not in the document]
ZenGround0: Opting to improve this as part of #1455

Issue #1435, opened 2021-05-26
Shore up Publish Storage Deals load balancing 


ZenGround0:
During recent review @wadeAlexC identified that PublishStorageDeals DealOpsByEpoch load balancing relies on hashing user input to determine the random offset from start epoch.  This gives a bit more control than ideal to the deal client who can grind based on previous epoch beacon and deal proposal data.

This is not currently a concern given typically network base fee and gas cost of PSD and the current high system capacity for new deal processing.  However this is something we should fix before v5 which brings with it the potential for much lower network base fee.

A straightforward fix is to include only the dealID, outside of the user's control, as input to the offset generation function.

arajasek:
Added to the HyperDrive milestone for now, but please move to v6 if that's the plan

ZenGround0:
We've pivoted to modifying PublishStorageDeals to sample Ticket randomness instead of beacon randomness at the current epoch rather than the previous epoch.  This increases the barrier of entry so that a deal client trying to bias processing epoch must win elections to produce block tickets.

nikkolasg:
Can you explain why in this case the deal client can grind on beacon randomness but not on chain randomness ? (i.e. only when he's not a block producer as well).

ZenGround0:
The hypothesis is that beacon randomness is available to all parties before messages are included in blocks and ticket randomness is only available to block producers before messages are included in blocks.

PR #1439, opened 2021-05-27
 Modify randomness logic for PublishStorageDeals

arajasek:
Easy improvement for #1435 

Tests are unhappy due to the change in runtime calls


ZenGround0:
looks good, just gotta fix some tests

BigLep:
TODO: we need to get signoff from Cryptolab

ZenGround0:
Speculatively merging because crytonetlab blocking is unlikely but we need confirmation before upgrade

ZenGround0 (approved):

PR #1440, opened 2021-05-28
Alleviate penalty

lyswifter:
this need [go-state-type](https://github.com/filecoin-project/go-state-types/pull/28) to be merge first.


ZenGround0:
Thanks @lyswifter please file an issue to explain the motivation so we can determine whether to include in v5 or wait until v6.

actors/builtin/miner/policy.go
```diff
@@ -191,7 +192,16 @@ const FaultDeclarationCutoff = WPoStChallengeLookback + 50 // PARAM_SPEC
 
 // The maximum age of a fault before the sector is terminated.
 // This bounds the time a miner can lose client's data before sacrificing pledge and deal collateral.
-var FaultMaxAge = WPoStProvingPeriod * 14 // PARAM_SPEC
+func GetWindowFaultMaxAge(nv network.Version) abi.ChainEpoch {
+	age := WPoStProvingPeriod * 14
+
+	//this will be <= v14 && < v15
+	if nv >= network.Version12 && nv < network.Version12 {
+		age = WPoStProvingPeriod * 140
```
ZenGround0, line 200: There are issues with this approach related to the way that faults are accounted for in the expiration queue.  The way this is written any faults initiated during the v14 window will maintain termination only in the far future instead of 14 days even after v15 begins.  However the intended behavior is that faults return to normal termination behavior after v15 starts up again.

This can be fixed to an extent with a somewhat complex migration.  For example v14 state could be migrated so that all existing faults are set to terminate 14 days after the start of v14.  To reduce termination load we would want to add a random offset.

It's possible there is some globally better option rather than writing termination events to a far future epoch in the expiration queue when trying to minimize complexity both here and in the v15 migration.  For example maybe adding some simple state in a v14 migration and then removing in v15 would be better?  The current approach is *probably* the best but let's not limit our thinking.

ZenGround0:
As of last implementers sync this direction of the FIP was deemphasized.  Closing this down for now.  Will reopen if we start down this path again.

PR #1445, opened 2021-06-01
Feat/market dobe rand

ZenGround0:
Another fix for #1435.  This avoids the problem of increasing wait time between deal start and first processing by up to another day by choosing between next and current day based on offset.  It should remove all grinding possibilities because processing epoch is now assigned predictably but unbiasedly.

Tests still need to be fixed -- currently market tests everywhere assume the ability to set processing epoch which makes this a long fix

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -1,15 +1,12 @@
 package market
 
 import (
-	"bytes"
-	"encoding/binary"
 	"sort"
 
 	addr "github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/go-state-types/cbor"
-	"github.com/filecoin-project/go-state-types/crypto"
 	"github.com/filecoin-project/go-state-types/exitcode"
 	rtt "github.com/filecoin-project/go-state-types/rt"
 	market0 "github.com/filecoin-project/specs-actors/actors/builtin/market"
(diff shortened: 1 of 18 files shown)


actors/builtin/market/market_actor.go
```diff
@@ -636,18 +633,15 @@ func (a Actor) CronTick(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
-func genRandNextEpoch(currEpoch abi.ChainEpoch, deal *DealProposal, rbF func(crypto.DomainSeparationTag, abi.ChainEpoch, []byte) abi.Randomness) (abi.ChainEpoch, error) {
-	buf := bytes.Buffer{}
-	if err := deal.MarshalCBOR(&buf); err != nil {
-		return epochUndefined, xerrors.Errorf("failed to marshal proposal: %w", err)
+func genRandNextEpoch(currEpoch abi.ChainEpoch, deal *DealProposal, dealID abi.DealID) (abi.ChainEpoch, error) {
+	offset := abi.ChainEpoch(uint64(dealID) % uint64(DealUpdatesInterval))
+	q := builtin.NewQuantSpec(DealUpdatesInterval, 0)
+	prevDay := q.QuantizeDown(deal.StartEpoch)
+	if prevDay+offset >= deal.StartEpoch {
+		return prevDay + offset, nil
```
wadeAlexC, line 641: Shouldn't `QuantizeDown` only ever return `StartEpoch`?

And because `StartEpoch + offset` will always be `>= StartEpoch`, the second part of this method should never be hit.

wadeAlexC: Nevermind - I didn't factor in the truncating division!

actors/builtin/market/market_actor.go
```diff
@@ -636,18 +633,15 @@ func (a Actor) CronTick(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
 	return nil
 }
 
-func genRandNextEpoch(currEpoch abi.ChainEpoch, deal *DealProposal, rbF func(crypto.DomainSeparationTag, abi.ChainEpoch, []byte) abi.Randomness) (abi.ChainEpoch, error) {
-	buf := bytes.Buffer{}
-	if err := deal.MarshalCBOR(&buf); err != nil {
-		return epochUndefined, xerrors.Errorf("failed to marshal proposal: %w", err)
+func genRandNextEpoch(currEpoch abi.ChainEpoch, deal *DealProposal, dealID abi.DealID) (abi.ChainEpoch, error) {
```
wadeAlexC, line 636: Can remove `currEpoch` parameter and `error` return

wadeAlexC:
LGTM!

arajasek (approved):
SGWM!

PR #1448, opened 2021-06-08
Fix/check max sector number

### LINKED ISSUE #1444: Miner.ProveCommitAggregate does not check sectors against abi.MaxSectorNumber
#### Description

`ProveCommitSector` checks the passed-in sector against `abi.MaxSectorNumber`:

https://github.com/filecoin-project/specs-actors/blob/a9f9f95f5e93bb7a27fd9fe8c0217e347fa1bef9/actors/builtin/miner/miner_actor.go#L975-L980

`ProveCommitAggregate` does not do this.

#### Recommendation

Add a check for consistency between the two methods. The easiest place to do this would be in `GetAllPrecommittedSectors`:

https://github.com/filecoin-project/specs-actors/blob/a9f9f95f5e93bb7a27fd9fe8c0217e347fa1bef9/actors/builtin/miner/miner_state.go#L387-L392

ZenGround0:
Closes #1444

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -385,6 +385,9 @@ func (st *State) GetAllPrecommittedSectors(store adt.Store, sectorNos bitfield.B
 	}
 
 	if err := sectorNos.ForEach(func(sectorNo uint64) error {
+		if sectorNo > abi.MaxSectorNumber {
+			return xc.ErrIllegalArgument.Wrapf("sector number greater than maximum")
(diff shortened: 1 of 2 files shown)


Stebalien (commented):
Nice catch!

Stebalien (approved):

anorth (commented):
If we're going to lean in to using the VM more for unit-style tests, we gotta work on decreasing the boilerplate for simple things like "get a miner" toward zero.

PR #1449, opened 2021-06-08
Fix/prove commit aggregate len check

### LINKED ISSUE #1442: ProveCommitAggregate: Check `len(precommits)` prior to index access
#### Description

https://github.com/filecoin-project/specs-actors/blob/a9f9f95f5e93bb7a27fd9fe8c0217e347fa1bef9/actors/builtin/miner/miner_actor.go#L945-L946

The above check can never be triggered. If `len(precommits) == 0`, the index access in the previous line will cause a panic.

#### Recommendation

Switch the order of the lines and/or move the length check to be above the loops over `precommits`: https://github.com/filecoin-project/specs-actors/blob/a9f9f95f5e93bb7a27fd9fe8c0217e347fa1bef9/actors/builtin/miner/miner_actor.go#L887-L888

ZenGround0:
Fixes #1442

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -942,8 +942,8 @@ func (a Actor) ProveCommitAggregate(rt Runtime, params *ProveCommitAggregatePara
 		svis = append(svis, svi)
 	}
 
-	sealProof := precommits[0].Info.SealProof
 	builtin.RequireState(rt, len(precommits) > 0, "bitfield non-empty but zero precommits read from state")
+	sealProof := precommits[0].Info.SealProof
 	err = rt.VerifyAggregateSeals(
 		proof.AggregateSealVerifyProofAndInfos{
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_actor.go
```diff
@@ -942,8 +942,8 @@ func (a Actor) ProveCommitAggregate(rt Runtime, params *ProveCommitAggregatePara
 		svis = append(svis, svi)
 	}
 
-	sealProof := precommits[0].Info.SealProof
 	builtin.RequireState(rt, len(precommits) > 0, "bitfield non-empty but zero precommits read from state")
+	sealProof := precommits[0].Info.SealProof
```
ZenGround0, line 946: Note this should be strictly impossible since GetAllPrecommittedSectors should error and cause an abort before reaching here so I can't test.

Stebalien (approved):

PR #1462, opened 2021-06-25
Test deterministic offset

ZenGround0:
Sanity checking that publishing storage deals assigns cron processing to deals as expected with new offset changes.


arajasek (approved):
LGTM!

PR #1468, opened 2021-08-13
Update to v6

ZenGround0:
Updating from v5 to v6 in preparation for 1) CC upgrade FIP 2) QoL improvements milestone


Stebalien (approved):
LGTM but why had a good point: we may save some hassle if we avoided updating actor codes for actors that haven't changed.

ZenGround0:
Let's try it out

PR #1473, opened 2021-08-24
Remove cc upgrade

ZenGround0:
As part of SnapDeals we get to remove the old cc upgrade code pathway.


arajasek (approved):

arajasek:
Would be great, but a little tricky, to get rid of the ReplaceCapacity field in the SectorPreCommitInfo in the second phase of this

anorth:
❤️

PR #1481, opened 2021-09-18
Change builtin actor code CIDs to fil/6

arajasek:


jennijuju (approved):

ZenGround0 (approved):
Note that we are explicitly moving away from @whyrusleeping's experimental idea of changing this value only for actors that change.  The value there, as I understand it, is to keep msig code from swirling, but v6 changes msig code anyway so we'll do the easy thing this release and revisit the question later.

PR #1487, opened 2021-09-21
Fip 0021

ZenGround0:
Implements FIP: https://github.com/filecoin-project/FIPs/blob/master/_fips/fip-publish-deals-ret.md

Still need tests


actors/builtin/market/market_actor.go
```diff
@@ -181,94 +180,160 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+		}
+
+		/*
+			drop deals with insufficient lock up to cover costs
+		*/
+		if _, ok := totalClientLockup[client]; !ok {
+			totalClientLockup[client] = abi.NewTokenAmount(0)
+		}
+		totalClientLockup[client] = big.Sum(totalClientLockup[client], deal.Proposal.ClientBalanceRequirement())
+		clientBalanceOk, err := msm.balanceCovered(client, totalClientLockup[client])
```
arajasek, line 225: I see that this is the same as the current checks (the deletions below), but I don't understand the address handling here. Based on an examination of the `AddBalance` method, it appears that the balanceTables only get populated with ID addresses.

But we aren't resolving addresses here (or in the `balanceCovered` method), so this is...gonna fail if the proposal has robust addresses for the client / provider?

I'm probably missing something.

actors/builtin/market/market_actor.go
```diff
@@ -181,94 +180,160 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+			drop duplicate deals
+		*/
+		// Normalise provider and client addresses in the proposal stored on chain.
+		// Must happen after signature verification and before taking cid.
+		deal.Proposal.Provider = provider
+		resolvedAddrs[deal.Proposal.Client] = client
+		deal.Proposal.Client = client
+
+		pcid, err := deal.Proposal.Cid()
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to take cid of proposal %d", di)
```
arajasek, line 249: I'm not sure that we'd want to, but we could just log and continue from errors such as this (and the many other unexpected errors).

actors/builtin/market/market_actor.go
```diff
@@ -181,94 +180,160 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+				&verifreg.UseBytesParams{
+					Address:  client,
+					DealSize: big.NewIntUnsigned(uint64(deal.Proposal.PieceSize)),
+				},
+				abi.NewTokenAmount(0),
+				&builtin.Discard{},
+			)
+			if code.IsError() {
+				rt.Log(rtt.INFO, "invalid deal %d: failed to acquire datacap exitcode: %d", code)
+				validInputBf.Unset(uint64(di))
```
arajasek, line 278: Why do this? Is there some risk that we've set this bit already?

actors/builtin/market/market_actor.go
```diff
@@ -181,94 +180,160 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+				builtin.MethodsVerifiedRegistry.UseBytes,
+				&verifreg.UseBytesParams{
+					Address:  client,
+					DealSize: big.NewIntUnsigned(uint64(deal.Proposal.PieceSize)),
+				},
+				abi.NewTokenAmount(0),
+				&builtin.Discard{},
+			)
+			if code.IsError() {
+				rt.Log(rtt.INFO, "invalid deal %d: failed to acquire datacap exitcode: %d", code)
```
arajasek, line 277: I think you're missing the deal index in the log here

actors/builtin/market/market_actor.go
```diff
@@ -181,94 +180,160 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
...
+		proposalCidLookup[pcid] = struct{}{}
+		validProposalCids = append(validProposalCids, pcid)
+		validDeals = append(validDeals, deal)
+		validInputBf.Set(uint64(di))
+	}
+
+	builtin.RequirePredicate(rt, len(validDeals) == len(validProposalCids), exitcode.ErrIllegalState,
+		"%d valid deals but %d valid proposal cids", len(validDeals), len(validProposalCids))
+
+	validDealCount, err := validInputBf.Count()
```
arajasek, line 293: Perform the same equality assertion above on this value too?

actors/builtin/market/market_actor.go
arajasek: EDIT: I just noticed that the `balanceCovered` method is new -- in that case i think there's a bug here

arajasek (commented):
Some qs

actors/builtin/market/market_actor.go
ZenGround0: hmm does the client address resolution on line 212 address your concerns?

actors/builtin/market/market_actor.go
ZenGround0: This is a good thing to raise.  As far as I can tell this and the many illegal state errors represent programming errors and I want to keep those aborting the whole thing.

actors/builtin/market/market_actor.go
ZenGround0: nice thank you!

actors/builtin/market/market_actor.go
ZenGround0: Nice this was leftover from a previous factoring of the solution

actors/builtin/market/market_actor.go
arajasek: oh, yeah, that does it...that line was probably invisible last night

actors/builtin/market/market_actor.go
arajasek: makes sense, i think i agree

arajasek (approved):

PR #1489, opened 2021-09-22
Fix #1486

### LINKED ISSUE #1486: Remove repeated calls to reward/power actors to get data inside OnDeferredCronEvent
They're already being computed in OnEpochTick, should just pass them in instead.

Original discussion here: https://github.com/filecoin-project/specs-actors/pull/1480#discussion_r713360361_

laudiacay:
Removing repeated calls to reward and power actors from OnDeferredCronEvent method... branched off some other PRs that hopefully will land on master before this one!


actors/builtin/power/power_actor.go
```diff
@@ -476,11 +473,22 @@ func (a Actor) processDeferredCronEvents(rt Runtime) {
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to flush events")
 	})
 	failedMinerCrons := make([]addr.Address, 0)
+
 	for _, event := range cronEvents {
+
+		var payload miner0.CronEventPayload
+		err := payload.UnmarshalCBOR(bytes.NewBuffer(event.CallbackPayload))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to unmarshal miner cron payload into expected structure")
```
ZenGround0, line 481: Given the impl of miner actor this should never error ✅ 

Thinking more about this through the lens of paranoia I'd like to see this done a little differently. 
1) Include Payload bytes directly in the DeferredCronEventParams (renamed from DeferredCronEventPayload).  So `EventType` becomes `EventPayload` and type changes to []bytes
2) do the unmarshalling directly in the miner actor OnDeferredCronEvents and fail with a serialization exit code in the (should be impossible) case where unmarshalling fails

This way the weird contract where the miner actor sends over arbitrary bytes and power actor doesn't care what they are is logically maintained because we just send the bytes back, but now with metadata attached.  And the main benefit is that if there is a programmer error somewhere in miner or power actor causing unmarshal to fail then that failure will propagate as error exitcode from the call to miner actor below and we will maintain separation of cron failures between miner actors.

ZenGround0: The more correct thing to do out of scope here or for v14 is for the miner actor to send over an event type int and the power actor to store that int instead of bytes of wrapped int in the cron event queue.  

It's not obvious doing this is actually worth the effort but I filed an issue for it: https://github.com/filecoin-project/specs-actors/issues/1490

actors/test/commit_post_test.go
```diff
@@ -76,15 +76,11 @@ func TestCommitPoStFlow(t *testing.T) {
 			SubInvocations: []vm.ExpectInvocation{
 				{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.OnEpochTickEnd, SubInvocations: []vm.ExpectInvocation{
 					{To: builtin.RewardActorAddr, Method: builtin.MethodsReward.ThisEpochReward},
-					{To: builtin.StoragePowerActorAddr, Method: builtin.MethodsPower.CurrentTotalPower},
```
ZenGround0, line 79: Love all the red

actors/builtin/shared.go
```diff
@@ -91,13 +92,43 @@ type MinerAddrs struct {
...
+}
+
+type DeferredCronEventPayload struct {
+	EventType               miner0.CronEventType
+	RewardSmoothed          smoothing.FilterEstimate
+	QualityAdjPowerSmoothed smoothing.FilterEstimate
+}
+
+// func (b *ConfirmSectorProofsParams) UnmarshalCBOR(r io.Reader) error {
+// 	var c bytes.Buffer
```
ZenGround0, line 114: remove me

actors/builtin/shared.go
```diff
@@ -91,13 +92,43 @@ type MinerAddrs struct {
...
 }
 
+func (b *DeferredCronEventPayload) UnmarshalCBOR(r io.Reader) error {
+	var c bytes.Buffer
+	_, err := c.ReadFrom(r)
+	*b = DeferredCronEventPayload{}
+	return err
+}
+
+func (b DeferredCronEventPayload) MarshalCBOR(w io.Writer) error {
```
ZenGround0, line 102: remove me

actors/builtin/shared.go
```diff
@@ -91,13 +92,43 @@ type MinerAddrs struct {
...
+	_, err := w.Write([]byte("aaaa"))
+	return err
+}
+
+type DeferredCronEventPayload struct {
+	EventType               miner0.CronEventType
```
ZenGround0, line 108: Changes discussed in previous comments will happen here: rename and adding full payload into field.

ZenGround0 (changes requested):
Looking good, there is one paranoid change I want to see happen before merging

actors/builtin/shared.go
ZenGround0: (and use gen to get the real code added to builtin/cbor_gen.go)

actors/test/commit_post_test.go
laudiacay: hee hee favorite color

actors/builtin/power/power_actor.go
laudiacay: yeehaw dooone

actors/builtin/miner/miner_actor.go
```diff
@@ -1065,8 +1065,8 @@ func (a Actor) ConfirmSectorProofsValid(rt Runtime, params *builtin.ConfirmSecto
 	precommittedSectors, err := st.FindPrecommittedSectors(store, params.Sectors...)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load pre-committed sectors")
 
-	confirmSectorProofsValid(rt, precommittedSectors, params.RewardStatsThisEpochBaselinePower,
-		params.RewardStatsThisEpochRewardSmoothed, params.PwrTotalQualityAdjPowerSmoothed)
+	confirmSectorProofsValid(rt, precommittedSectors, params.RewardBaselinePower,
```
ZenGround0, line 1068: nit: this now fits on one line so remove line break

actors/builtin/power/power_actor.go
```diff
@@ -487,11 +483,19 @@ func (a Actor) processDeferredCronEvents(rt Runtime) {
 		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to flush events")
 	})
 	failedMinerCrons := make([]addr.Address, 0)
+
 	for _, event := range cronEvents {
+
+		input := builtin.DeferredCronEventParams{
```
ZenGround0, line 489: nit: idiomatic name is "params"

actors/builtin/power/power_test.go
```diff
@@ -609,15 +609,24 @@ func TestCron(t *testing.T) {
 
 		rt.SetEpoch(1)
 		actor.enrollCronEvent(rt, miner1, 2, []byte{0x1, 0x3})
-		actor.enrollCronEvent(rt, miner2, 4, []byte{0x2, 0x3})
+		actor.enrollCronEvent(rt, miner2, 4, []byte{0x1, 0x3})
```
ZenGround0, line 612: checking ordering is an explict intent of this test iiuc.  The test checks that the first enrolled event gets processed then the next by setting two different byte values.  To maintain this testing property you should change this bad to 0x2

ZenGround0 (approved):

PR #1493, opened 2021-09-24
Divide once when calculating qa power

### LINKED ISSUE #1250: Miner Policy: Correct truncating division
`QualityForWeight` returns the average of weighted space time here:

https://github.com/filecoin-project/specs-actors/blob/6c71cf5bdcd11b4ae78b15604e49d2e306d7799d/actors/builtin/miner/policy.go#L184-L185

Dividing twice may introduce error in some cases. The return can be re-written to divide only once:

```go
return big.Div(scaledUpWeightedSumSpaceTime, big.Mul(sectorSpaceTime, builtin.QualityBaseMultiplier))
```

ZenGround0:
Fixes #1250

This is already well tested in miner/policy_test.go `TestQuality`.  The high value of SectorQualityPrecision seems to have kept any error from creeping in to the computation, at least in test cases.


ZenGround0:
From @Stebalien this might actually need a migration over partition state to be safe.  Investigating...

ZenGround0:
This needs a migration for safety.  Closing down for now and removing from v14 scope. Putting migration ideas in issue.

PR #1494, opened 2021-09-24
Fix #176 (swap to xerrors from pkg/errors)

### LINKED ISSUE #176: Switch from pkg/errors to xerrors

laudiacay:
unfortunately still needs to import pkg/errors because of old versions relying on it... but all done! 🌻 #176


actors/builtin/market/set_multimap.go
```diff
@@ -68,7 +67,7 @@ func (mm *SetMultimap) Put(epoch abi.ChainEpoch, v abi.DealID) error {
 
 	// Add to the set.
 	if err = set.Put(dealKey(v)); err != nil {
-		return errors.Wrapf(err, "failed to add key to set %v", epoch)
+		return xerrors.Errorf("failed to add key to set %w: %w", epoch, err)
```
ZenGround0, line 70: You might know something I don't know but I think two `%w`s  / a non error `%w` is wrong.  From [xerrors readme](https://pkg.go.dev/golang.org/x/xerrors#section-readme):

"It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v."

ZenGround0 (changes requested):
I think %w is being used incorrectly.  If I'm wrong about that this is good to go.

actors/builtin/market/set_multimap.go
laudiacay: oh lol ur right ill fix

actors/builtin/power/power_state.go
```diff
@@ -127,13 +126,13 @@ func (st *State) MinerNominalPowerMeetsConsensusMinimum(s adt.Store, miner addr.
 		return false, err
 	}
 	if !ok {
-		return false, errors.Errorf("no claim for actor %v", miner)
+		return false, xerrors.Errorf("no claim for actor %w", miner)
```
ZenGround0, line 129: Keep as %v, mineris addr.Address

support/agent/miner_agent.go
```diff
@@ -483,7 +483,7 @@ func (ma *MinerAgent) publishStorageDeals() []message {
 			// add returned deal ids to be included within sectors
 			publishReturn, ok := ret.(*market.PublishStorageDealsReturn)
 			if !ok {
-				return errors.Errorf("create miner return has wrong type: %v", ret)
+				return xerrors.Errorf("create miner return has wrong type: %w", ret)
```
ZenGround0, line 486: Keep as %v ret is not necessarily an error type

support/agent/miner_generator.go
```diff
@@ -69,13 +69,13 @@ func (mg *MinerGenerator) createMiner(owner address.Address, cfg MinerAgentConfi
 		ReturnHandler: func(s SimState, msg message, ret cbor.Marshaler) error {
 			createMinerRet, ok := ret.(*power.CreateMinerReturn)
 			if !ok {
-				return errors.Errorf("create miner return has wrong type: %v", ret)
+				return xerrors.Errorf("create miner return has wrong type: %w", ret)
```
ZenGround0, line 72: ditto

support/agent/miner_generator.go
```diff
@@ -69,13 +69,13 @@ func (mg *MinerGenerator) createMiner(owner address.Address, cfg MinerAgentConfi
...
 			}
 
 			var worker, owner address.Address
 			params, ok := msg.Params.(*power.CreateMinerParams)
 			if !ok {
-				return errors.Errorf("create miner params has wrong type: %v", msg.Params)
+				return xerrors.Errorf("create miner params has wrong type: %w", msg.Params)
```
ZenGround0, line 78: ditto

support/agent/sim.go
```diff
@@ -162,7 +161,7 @@ func (s *Sim) Tick() error {
 
 		// for now, assume everything should work
 		if result.Code != exitcode.Ok {
-			return errors.Errorf("exitcode %d: message failed: %v\n%s\n", result.Code, msg, strings.Join(s.v.GetLogs(), "\n"))
+			return xerrors.Errorf("exitcode %d: message failed: %w\n%s\n", result.Code, msg, strings.Join(s.v.GetLogs(), "\n"))
```
ZenGround0, line 164: ditto

support/agent/sim.go
```diff
@@ -378,7 +377,7 @@ func computeCircSupply(v SimVM) error {
 	if err != nil {
 		return err
 	} else if !found {
-		return errors.Errorf("burnt actor not found at %v", builtin.BurntFundsActorAddr)
+		return xerrors.Errorf("burnt actor not found at %w", builtin.BurntFundsActorAddr)
```
ZenGround0, line 380: ditto

ZenGround0 (changes requested):

actors/builtin/market/set_multimap.go
Stebalien: (oops. Sorry, I'll review more carefully next time)

ZenGround0 (approved):
LGTM

PR #1496, opened 2021-09-24
fix #1248 changing the deal label to bytes not strings

### LINKED ISSUE #1248: DealProposal's Label field serializes as non-UTF8
The `Label` field in `DealProposal` is a string that is not guaranteed to be UTF8 as required in the CBOR spec. This causes issues for other client implementations (like us). It will take a LOT of effort for us to work around this bug. 

I would suggest somehow enforcing UTF-8 or switching the field to be arbitrary bytes instead.

laudiacay:
fix #1248


ZenGround0:
@laudiacay this is a good code change.  There is an associated data change to mainnet state that will need to also be made through an "actors state migration".  See this example PR for one example of how this looks: https://github.com/filecoin-project/specs-actors/pull/1265/files

ZenGround0:
This is also something that will need a Fillecoin Improvement Proposal (FIP) to signal the change and give time to implementers of external tooling around deal data.  This process takes place here: https://github.com/filecoin-project/FIPs.

actors/builtin/market/deal.go
```diff
@@ -20,33 +29,55 @@ var PieceCIDPrefix = market0.PieceCIDPrefix
...
-type DealProposal = market0.DealProposal
+type DealProposal struct {
+	PieceCID     cid.Cid `checked:"true"` // Checked in validateDeal, CommP
+	PieceSize    abi.PaddedPieceSize
+	VerifiedDeal bool
+	Client       addr.Address
+	Provider     addr.Address
+
+	// Label is an arbitrary client chosen label to apply to the deal
+	// TODO: Limit the size of this: https://github.com/filecoin-project/specs-actors/issues/897
```
ZenGround0, line 40: @laudiacay since we have a draft FIP started now is the perfect time to knock out this todo by enforcing a limit on this at the protocol level.  It could be added to the FIP which could increase in scope to "Deal Proposal Label improvements" or something like that.

To enforce a size all we need is validation code in the validate deal method that rejects proposals with too big labels with `ErrIllegalArgument`.  The tricky part is setting a good limit.  Investigating current labels in chain state will be informative (I can help with this).  Another helpful data point to look into is the cbor serialization limit on byte slices which I think is a few megabytes.

@whyrusleeping @ribasushi do you have insights into how the label is currently used / how you plan to use it?  Any recommendations on label size?

ribasushi: errrr @ZenGround0 this is already implemented...?
https://github.com/filecoin-project/specs-actors/blob/v5.0.4/actors/builtin/market/policy.go#L26-L27

laudiacay: okie dokie ill just take out the todo

actors/migration/nv14/test/deal_label_migration_test.go
```diff
@@ -0,0 +1,273 @@
...
+	require.NoError(t, err)
+	require.Equal(t, found, inStates)
+	dealpropcid, err := dealprop5.Cid()
+	require.NoError(t, err)
+	found, err = pendingProposals.Has(abi.CidKey(dealpropcid))
+	require.NoError(t, err)
+	require.Equal(t, found, inPendingProposals)
+}
+
+// there MUST be a way to consolidate these...
```
ZenGround0, line 258: You could create a function parameter that gets a proposal and returns the cid, returning cid.Undef if not found.  You could then get found info from this and the pending proposals key.

That said this is fine as is.  Either way remove this comment before merging.

actors/migration/nv14/test/deal_label_migration_test.go
```diff
@@ -0,0 +1,273 @@
...
+	pendingProposals, err = adt.AsSet(adtStore, market5State.PendingProposals, builtin.DefaultHamtBitwidth)
+	require.NoError(t, err)
+	// deal1 will just be in states and proposals and not pendingproposals
+	checkMarket6ProposalsEtcState(t, proposals, states, pendingProposals, deal1ID, true, true, false)
+	// deal2 will be in proposals pendingproposals and in states
+	checkMarket6ProposalsEtcState(t, proposals, states, pendingProposals, deal2ID, true, true, true)
+	// deal3 will be in proposals and pendingproposals but not in states
+	checkMarket6ProposalsEtcState(t, proposals, states, pendingProposals, deal3ID, true, false, true)
+
+	/// market.checkstateinvariants???
```
ZenGround0, line 184: Remove comment

Issue #1492, opened 2021-09-24
Add BatchBalancer fee to batch precommit


ZenGround0:
https://github.com/filecoin-project/FIPs/pull/174

PR #1497, opened 2021-09-24
BatchBalancer fee charged on precommit aggregate

ZenGround0:
WIP #1492


actors/builtin/miner/monies.go
```diff
@@ -211,15 +211,25 @@ func LockedRewardFromReward(reward abi.TokenAmount) (abi.TokenAmount, *VestSpec)
...
-func AggregateNetworkFee(aggregateSize int, baseFee abi.TokenAmount) abi.TokenAmount {
+func AggregateProveCommitNetworkFee(aggregateSize int, baseFee abi.TokenAmount) abi.TokenAmount {
+	return AggregateNetworkFee(aggregateSize, EstimatedSingleProveCommitGasUsage, baseFee)
+}
+
+func AggregatePreCommitNetworkFee(aggregateSize int, baseFee abi.TokenAmount) abi.TokenAmount {
+	return AggregateNetworkFee(aggregateSize, EstimatedSinglePreCommitGasUsage, baseFee)
+}
+
+func AggregateNetworkFee(aggregateSize int, gasUsage big.Int, baseFee abi.TokenAmount) abi.TokenAmount {
```
arajasek, line 230: Can we not export this? I don't think it's needed, and it's only too easy for something (Lotus) to misuse this since it shares a name with the v5 method

ZenGround0: no problem

actors/builtin/miner/monies.go
```diff
@@ -211,15 +211,25 @@ func LockedRewardFromReward(reward abi.TokenAmount) (abi.TokenAmount, *VestSpec)
 }
 
 var EstimatedSingleProofGasUsage = big.NewInt(65733297) // PARAM_SPEC
-var BatchDiscount = builtin.BigFrac{                    // PARAM_SPEC
+var EstimatedSingleProveCommitGasUsage = big.NewInt(49299973)
```
arajasek, line 214: 1) You should be able to nuke EstimatedSingleProofGasUsage above^
2) Out of curiosity, where do these numbers come from?

ZenGround0: They are rough estimates of the mean gas used per single PreCommit and ProveCommit.  Empirically measured from tests and validated with small samples of on chain gas values.  Wouldn't be surprised if they need tweaking down the line.

arajasek (approved):

PR #1502, opened 2021-10-01
Defensive programming: harden power actor against seal verify failures

ZenGround0:
While aborts from processBatchProofVerifies are all believed to be the result of programmer error or individual validator system error aborting from this code presents unnecessary risk if this is wrong.  This PR changes these aborts to errors which are then logged and ignored from top level power cron.

The impact is that now in case of programmer error a batch of seal proofs will fail to validate and propagate power to storage providers instead of aborting power cron and either halting the chain or taking out 1/60th of sps from deadline cron.

In the worst case where programmer error is global all prove commits (not prove commit aggregates) would be dropped until programmer error fixed.  In a more plausible state where programmer error is related to contents of the validation batch prove commits for one epoch would be dropped and then processing would continue as normal.

Reiterating that programmer error is not expected to exist but now defense against it is better.

### DIFF
--- a/actors/builtin/power/power_actor.go
+++ b/actors/builtin/power/power_actor.go
@@ -9,6 +9,7 @@ import (
 	"github.com/filecoin-project/go-state-types/cbor"
 	"github.com/filecoin-project/go-state-types/exitcode"
 	rtt "github.com/filecoin-project/go-state-types/rt"
+	xerrors "golang.org/x/xerrors"
 
 	power0 "github.com/filecoin-project/specs-actors/actors/builtin/power"
 	"github.com/ipfs/go-cid"
@@ -218,7 +219,9 @@ func (a Actor) OnEpochTickEnd(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
 	rewretcode := rt.Send(builtin.RewardActorAddr, builtin.MethodsReward.ThisEpochReward, nil, big.Zero(), &rewret)
 	builtin.RequireSuccess(rt, rewretcode, "failed to check epoch baseline power")
 
-	a.processBatchProofVerifies(rt, rewret)
+	if err := a.processBatchProofVerifies(rt, rewret); err != nil {
+		rt.Log(rtt.ERROR, "unexpected error processing batch proof verifies: %s. Skipping all verification for epoch %d", err, rt.CurrEpoch())
+	}
 	a.processDeferredCronEvents(rt, rewret)
 
 	var st State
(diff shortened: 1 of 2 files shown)


actors/builtin/power/power_actor.go
```diff
@@ -381,23 +395,34 @@ func (a Actor) processBatchProofVerifies(rt Runtime, rewret reward.ThisEpochRewa
...
+			}
 
 			verifies[a] = infos
 			return nil
 		})
-		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to iterate proof batch")
-
+		// Do not return immediately, all runs that get this far should wipe the ProofValidationBatchQueue.
+		// If we leave the validation batch then in the case of a repeating state error the queue
+		// will quickly fill up and repeated traversls will start ballooning cron execution time.
```
arajasek, line 407:
```suggestion
		// will quickly fill up and repeated traversals will start ballooning cron execution time.
```

arajasek (commented):
I'm a little unsure about this, only because I don't know how safe it is to absorb and ignore all of these cases. Skimming the various returns, it seems fine, and I assume it's fine, tho...

Can we come up with a system to scream bloody murder if we do wind up erroring? We could just get existing monitoring to pay attention for the log and alert if it's hit. Or something more sophisticated like having a "safety" mode that has the old behaviour / straight up halts if the log is hit.

actors/builtin/power/power_actor.go
```diff
@@ -381,23 +395,34 @@ func (a Actor) processBatchProofVerifies(rt Runtime, rewret reward.ThisEpochRewa
...
+		if err != nil {
+			stErr = xerrors.Errorf("failed to iterate proof batch: %w", err)
```
arajasek, line 409: I think you can simplify this by directly capturing the output of the `ForAll` in `stErr` if you're so inclined

ZenGround0:
> I'm a little unsure about this, only because I don't know how safe it is to absorb and ignore all of these cases.

The point of this PR is that it is safer than the alternative of aborting which will leave deadline cron in a bad state.  This bad state is totally avoided at the expense of killing prove commit verifies using the approach in this PR.  This is a good tradeoff.  A few storage Providers burning gas in the event of an emergency is much better than a significant portion of system power being frozen out of cron.

ZenGround0:
> Can we come up with a system to scream bloody murder if we do wind up erroring? 

Yes all actors log rt.ERROR values should be escalated with high severity, they all represent dangerous programmer error.

actors/builtin/power/power_actor.go
ZenGround0: I'm going to keep it this way to maintain a standard "wrap err and assign to stErr" pattern.

arajasek (approved):

Issue #1474, opened 2021-08-25
Remove miners with test sector types from state


ZenGround0:
From an earlier issue raised by @arajasek 

I noticed that [this PR ](https://github.com/filecoin-project/specs-actors/pull/1345) made possible creating miners with 2k / 8k / 512Mb post types. This is probably fine, since these miners can't add any storage, but it does have some impact on state -- the Power State's MinerAboveMinPowerCount for instance includes all 2 KiB miners since their min power is 0.

List of such miners:

```
f0466399
f0480175
f0480253
f0431713
f0431953
f0432178
f0480490
f0440705
f0481497
f0480393
f0486159
f0461500
f0431712
f0431939
f0431663
f0431716
f0453913
f0463860
f0463597
f0463917
f0481525
f0431711
f0480068
f0432231
f0466392
f0481115
f0481379
f0466398
f0431733
f0431725
f0431679
f0463769
f0480322
f0441531
f0481888
f0432245

Type: 5  Count:  13
Type: 6  Count:  3
Type: 7  Count:  20
```

We blocked further creation of these miners in v5 actors.  For v6 migration we should remove these miner actors from the state, returning any funds sent to them to their owner address.

ZenGround0:
Migration testing criteria:

* Miners with bad proof types need to go away
* Signature of miners with bad proof types in power actor need to change
	— claim is removed
	— total miners miners over conseusns …. is removed
* Market actor escrow table removes these bad miners 
* Owner address of these bad miners is given the funds belonging to the miner actor
* Owner address also gets funds put in the market actor escrow table

PR #1508, opened 2021-10-14
Migration to remove existing miners with invalid window post proof types

laudiacay:
see #1474


ZenGround0:
Random thought, I think we need to remove these miners from init actor state, not sure if this PR does that yet.

laudiacay:
It does not. Aaaaaaaaaaaaaagh. I’ll do this tomorrow.

Stebalien:
@ZenGround0 are you sure? I'd expect us to mimic the `DeleteActor` syscall, which _doesn't_ touch the init state. I don't think we currently garbage collect the init actor at all.

ZenGround0:
Ah in that case maybe not.

PR #1519, opened 2021-10-20
Actors v7

arajasek:


support/vm/testing.go
```diff
@@ -360,7 +359,7 @@ func AdvanceByDeadlineTillEpochWhileProving(t *testing.T, v *VM, minerIDAddress,
 				PoStProof: abi.RegisteredPoStProof_StackedDrgWindow32GiBV1,
 			}},
 			ChainCommitEpoch: dlInfo.Challenge,
-			ChainCommitRand:  []byte(vm.RandString),
```
arajasek, line 363: we might as well be using the v7 randstring here

actors/builtin/codes.go
```diff
@@ -35,17 +35,17 @@ func init() {
...
-		&CronActorCodeID:             {name: "fil/6/cron"},
-		&StoragePowerActorCodeID:     {name: "fil/6/storagepower"},
-		&StorageMinerActorCodeID:     {name: "fil/6/storageminer"},
-		&StorageMarketActorCodeID:    {name: "fil/6/storagemarket"},
-		&PaymentChannelActorCodeID:   {name: "fil/6/paymentchannel"},
-		&RewardActorCodeID:           {name: "fil/6/reward"},
-		&VerifiedRegistryActorCodeID: {name: "fil/6/verifiedregistry"},
-		&AccountActorCodeID:          {name: "fil/6/account", signer: true},
-		&MultisigActorCodeID:         {name: "fil/6/multisig", signer: true},
+		&SystemActorCodeID:           {name: "fil/7/system"},
```
arajasek, line 38: bumping all codes for now because i think that's better

Issue #677, opened 2020-07-15
Add a MustGet method to the adt.Array type


Stebalien:
We frequently assert that items in arrays are "found" (https://github.com/filecoin-project/specs-actors/pull/648#discussion_r454947033). We could simplify all `Get` calls like this by introducing a `MustGet` method.

PR #1523, opened 2021-10-20
Adding MustGet method to adt.Array in order to simplify code

laudiacay:
#677 
Add a mustget function so that we don’t have to do things like say `if err != nil {error out} if !found {error out}`

This is a draft because i want to work the new pattern into the code in at least a few places before we merge it :)

### DIFF
--- a/actors/util/adt/array.go
+++ b/actors/util/adt/array.go
@@ -134,6 +134,17 @@ func (a *Array) Get(k uint64, out cbor.Unmarshaler) (bool, error) {
 	}
 }
 
+// MustGet retrieves array element into the 'out' unmarshaler, returning an error if it 
+//  failed to get the element or if it did not find the element in the array.
+func (a *Array) MustGet(k uint64, out cbor.Unmarshaler) error {
+	if found, err := a.root.Get(a.store.Context(), k, out); err != nil {
+		return xerrors.Errorf("failed to get index %v in root %v: %w", k, a.root, err)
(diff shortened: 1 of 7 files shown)


actors/util/adt/array.go
```diff
@@ -134,6 +134,17 @@ func (a *Array) Get(k uint64, out cbor.Unmarshaler) (bool, error) {
 	}
 }
 
+// MustGet retrieves array element into the 'out' unmarshaler, returning an error if it 
+//  failed to get the element or if it did not find the element in the array.
+func (a *Array) MustGet(k uint64, out cbor.Unmarshaler) error {
+	if found, err := a.root.Get(a.store.Context(), k, out); err != nil {
+		return xerrors.Errorf("failed to get index %v in root %v: %w", k, a.root, err)
```
ZenGround0, line 141: nit: idiomatically use type specific printing directive if available. Cids have a String method so %s works for them.  You are already doing errors the right way with the w: `"failed to get index %d in root %s: %w", k, a.root, err)`

Stebalien: Also:

- The boring stuff: https://pkg.go.dev/fmt#hdr-Printing
- The fun stuff: https://pkg.go.dev/fmt#Formatter, https://pkg.go.dev/fmt#GoStringer, https://pkg.go.dev/fmt#Stringer

ZenGround0:
Archiving repo

PR #1528, opened 2021-10-27
Added Configs for actors

ortyomka:
#353
A requirement was to recalculate args after each change in runtime

I have 3 options for solution. 

1. One global config. It looks like antipattern.
2. Several small configs for each component. Better than big one, because an influence is divided between the policies.
3. Config for each actor. Several problems here. 
3.1. Inconsistency between actor configs
3.2. Values are not recalculated in runtime if some arg changes

I have chosen the second one, but open to your suggestion. 

P.S. I am new to Golang: any syntax and semantic suggestions are welcome


ZenGround0:
Apologies we never got this one landed  but we're now planning to archive repo.  Most of this work seems to be done in the current builtin-actors repo.

PR #1529, opened 2021-11-10
Implement FIP-0019

arajasek:
Implements FIP-0019. There are outstanding work items such as Snapshotting PoSts, testing, some refactoring, etc, but this gets the job done.


actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			continue
+		}
+
+		// TODO: confirm sector is healthy
+
+		// TODO: What's the impact of the following flow:
+		// 	- activating the deals
+		// - failing to verify the proof (or some other failure later)
+		// - allowing the message to succeed anyway ("skipping" the sector)
+		// Does that leave the deal uncommitable (since state doesn't get reverted)? is it better to just fail the msg, revert state, and allow for a retry later?
```
ZenGround0, line 2163: Yes once you successfully activate a batch of deals in a sector in the market state you need to fail or succeed this whole miner actor method atomically

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+		for k, updateWithSectorInfo := range prevalidatedUpdates {
+			err = rt.VerifyReplicaUpdate(
+				proof.ReplicaUpdateInfo{
+					NewSealedSectorCID:   updateWithSectorInfo.update.NewSealedSectorCID,
+					OldSealedSectorCID:   updateWithSectorInfo.sectorInfo.SealedCID,
+					NewUnsealedSectorCID: unsealedSectorCIDs[k],
+				})
+
+			if err != nil {
+				rt.Log(rtt.INFO, "failed to verify replica proof, skipping sector %d", updateWithSectorInfo.sectorInfo.SectorNumber)
```
ZenGround0, line 2216: As noted above this is going to leave the market actor in a bad state.  Your options are failing the whole message or moving this to above market actor side effects in the above loop.  The second option should be doable -- I don't think the VerifyReplicaUpdate and inputs require any state modification so you can take them out of the StateTransaction and put them above.

ZenGround0 (changes requested):
Still have more to go over and think more about but here are a few comments in the meantime.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+
+			// TODO: actual error / continue
+			if err != nil {
+				return
+			}
+
+			powerDelta = powerDelta.Add(partitionPowerDelta)
+			pledgeDelta = big.Add(pledgeDelta, partitionPledgeDelta) // expected to be zero, see note below.
+
+			err = partitions.Set(updateWithSectorInfo.update.Partition, &partition)
```
ZenGround0, line 2300: Right now this isn't persisted you need to flush partition and deadline (`partitions.Root()` => `deadline.UpdateDeadline` => `SaveDeadlines`)

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			newSectorInfo.ReplacedDayReward = updateWithSectorInfo.sectorInfo.ExpectedDayReward
+			newSectorInfo.ExpectedDayReward = ExpectedRewardForPower(rewRet.ThisEpochRewardSmoothed, powRet.QualityAdjPowerSmoothed, pwr, builtin.EpochsInDay)
+			newSectorInfo.ExpectedStoragePledge = ExpectedRewardForPower(rewRet.ThisEpochRewardSmoothed, powRet.QualityAdjPowerSmoothed, pwr, InitialPledgeProjectionPeriod)
+			newSectorInfo.ReplacedSectorAge = maxEpoch(0, rt.CurrEpoch()-updateWithSectorInfo.sectorInfo.Activation)
+
+			initialPledgeAtUpgrade := InitialPledgeForPower(pwr, rewRet.ThisEpochBaselinePower, rewRet.ThisEpochRewardSmoothed,
+				powRet.QualityAdjPowerSmoothed, rt.TotalFilCircSupply())
+
+			if initialPledgeAtUpgrade.GreaterThan(updateWithSectorInfo.sectorInfo.InitialPledge) {
+				deficit := big.Sub(initialPledgeAtUpgrade, updateWithSectorInfo.sectorInfo.InitialPledge)
```
ZenGround0, line 2248: This looks correct. I'm wondering if it might be cleaner to totally remove the old pledge and add back the new one.

support/mock/mockrt.go
```diff
@@ -690,6 +697,11 @@ func (rt *Runtime) VerifyAggregateSeals(agg proof.AggregateSealVerifyProofAndInf
 	return nil
 }
 
+func (rt *Runtime) VerifyReplicaUpdate(replicaInfo proof.ReplicaUpdateInfo) error {
+	//TODO
```
ZenGround0, line 701: You can drop this looks fine for now.  If creating errors becomes important in tests you can come back to it.

support/mock/mockrt.go
```diff
@@ -97,6 +98,12 @@ type expectAggregateVerifySeals struct {
 	err     error
 }
 
+type expectReplicaVerify struct {
+	inSVIs  []proof.ReplicaUpdateInfo
```
ZenGround0, line 102: nit: SVI stands for seal verify info so you can change this field name to better match the new data in it.

actors/builtin/miner/miner_actor.go
ZenGround0: Check out `ExtendSectorExpiration`

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			//	rt.Log(rtt.WARN, "precommit %d has lifetime %d less than minimum. ignoring", precommit.Info.SectorNumber, duration, MinSectorExpiration)
+			//	continue
+			//}
+
+			pwr := QAPowerForWeight(info.SectorSize, duration, newSectorInfo.DealWeight, newSectorInfo.VerifiedDealWeight)
+
+			newSectorInfo.ReplacedDayReward = updateWithSectorInfo.sectorInfo.ExpectedDayReward
+			newSectorInfo.ExpectedDayReward = ExpectedRewardForPower(rewRet.ThisEpochRewardSmoothed, powRet.QualityAdjPowerSmoothed, pwr, builtin.EpochsInDay)
+			newSectorInfo.ExpectedStoragePledge = ExpectedRewardForPower(rewRet.ThisEpochRewardSmoothed, powRet.QualityAdjPowerSmoothed, pwr, InitialPledgeProjectionPeriod)
+			newSectorInfo.ReplacedSectorAge = maxEpoch(0, rt.CurrEpoch()-updateWithSectorInfo.sectorInfo.Activation)
```
ZenGround0, line 2242: The new sector info looks more or less good but it is currently not being set.  You'll need to set a reference to grab a reference to the sectors AMT (I think your `sectors` variable should be fine) and modify it to set this sector info at the sectorID index in the AMT.  Then you'll need to flush this AMT at the very end before leaving the StateTransaction.

actors/runtime/proof/verify.go
```diff
@@ -28,6 +29,18 @@ type AggregateSealVerifyInfo = proof5.AggregateSealVerifyInfo
 
 type AggregateSealVerifyProofAndInfos = proof5.AggregateSealVerifyProofAndInfos
 
+///
+/// Replica
+///
+
+// Information needed to verify a replica update
+
+type ReplicaUpdateInfo struct {
```
ZenGround0, line 38: You will need a proof too.  Maybe some other things too but I'll check up on that and get back to you.

ZenGround0 (changes requested):

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+	// TODO: this method has some potential aborts -- are those all programmer error or is there scope to improve UX for bad input?
+	unsealedSectorCIDs := requestUnsealedSectorCIDs(rt, sectorsDataSpec...)
+
+	builtin.RequirePredicate(rt, len(unsealedSectorCIDs) == len(prevalidatedUpdates), exitcode.ErrIllegalState,
+		"unsealed sector cid request returned %d records, expected %d", len(unsealedSectorCIDs), len(prevalidatedUpdates))
+
+	succeededSectors := bitfield.New()
+	var st State
+	rt.StateTransaction(&st, func() {
+		for k, updateWithSectorInfo := range prevalidatedUpdates {
```
ZenGround0, line 2207: Something to consider is doing updates by iterating over deadlines and partitions rather than updates.  During the validation loop we could group updates by deadline and partition in a map and then iterate over it here.  Again take a look at `ExtendSectorExpiration` for how this looks.  We could save batch updaters some deadline and partition ipld write gas this way.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			continue
+		}
+
+		if len(update.Deals) <= 0 {
+			rt.Log(rtt.INFO, "must have deals to update, skipping sector %d", update.SectorID)
+			continue
+		}
+
+		// TODO: is there an upper limit on number of deals in a sector (beside the practical limit imposed by sector sizes and min deal size)
+		// TODO Confirm the sector is actually in the stated deadline? This happens later when we try to update partition, but maybe do here first?
```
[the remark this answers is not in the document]
arajasek: I don't really agree, I don't see why this should be an exception to the error handling principle stated [here](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0019.md#batching-and-error-handling). 

But it's a sufficiently odd error (with no real reason to happen) that I don't feel too strongly -- doing nothing is less work for me :P

actors/builtin/miner/miner_actor.go
arajasek: The reason not to do that is that it requires the data commitments to be all calculated before the more basic per-update checks (and i assume that isn't cheap).

I do think the optimal thing to do is to make 3 per-update passes -- 1) quick checks, 2) expensive checks / operations (proof verif & deal activation), and 3) actual miner / power state changes. CommD calculation can then be sandwiched between 1 and 2.

Feels a bit bloated, though. Tell me whether you agree and I'll implement that.

ZenGround0: I'm more in favor of just one set of checks and then the atomic updates.  It's good to handle failures in batch gracefully but you're already meeting my expectations for that.  The most important failures to handle without killing the batch are the epoch dependent ones that can't be caught during gas estimation, and you are already doing deal activation this way.

arajasek: That makes a lot of sense, yeah. Sounds good!

actors/builtin/miner/miner_actor.go
arajasek: lmao nice, i'm an amoeba brain

actors/builtin/miner/miner_actor.go
arajasek: I like that less, but seems like a preference thing.

actors/runtime/proof/verify.go
arajasek: I suppose I can be convinced that including a proof in here might be beneficial.

actors/builtin/miner/miner_actor.go
arajasek: Yeah, I have a TODO in there for that, i think it's the next (and hopefully last) improvement-tweak needed for the MVP.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+				found, err := partitions.Get(updateWithDetails.update.Partition, &partition)
+
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load deadline %v partition %v",
+					updateWithDetails.update.Deadline, updateWithDetails.update.Partition)
+
+				if !found {
+					rt.Abortf(exitcode.ErrNotFound, "no such deadline %v partition %v", dlIdx, updateWithDetails.update.Partition)
+				}
+
+				// TODO: Group by partition
```
arajasek, line 2306: I've been a keener since middle school, so I wanna do this too, but it'll be a future thing. 

or a never thing.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,253 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+	}
+
+	// Filters out the first wave of invalid updates with quick checks about sector status
+	var prevalidatedUpdates []*updateAndSectorInfo
+
+	var sectorsDeals []market.SectorDeals
+	var sectorsDataSpec []*market.SectorDataSpec
+	for _, update := range params.Updates {
+		// TODO: do we need an assertion that the same sector isn't upgraded twice?
+		// TODO: actual length
```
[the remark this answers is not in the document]
Kubuxu: AFAIK 192*16=3072 currently, we can do aggregation in future.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+func (a Actor) ProveReplicaUpdates(rt Runtime, params *ProveReplicaUpdatesParams) bitfield.BitField {
+	// Validate inputs
+
+	// TODO: check how this gets rendered and fix as needed
+	builtin.RequireParam(rt, len(params.Updates) <= ProveReplicaUpdatesMaxSize, "too many updates", len(params.Updates), ProveReplicaUpdatesMaxSize)
+
+	// TODO: Any more quick checks before we jump into it?
+	// see extendsectorexpirations for more checks to plagiarize
+
+	// Kinda sad to be doing (and pay this before a bunch of checks, but oh well?
```
ZenGround0, line 2095: nit: remove

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
ZenGround0: @arajasek you can bound it by an even number like 4096

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+
+	type updateAndSectorInfo struct {
+		update     *ReplicaUpdate
+		sectorInfo *SectorOnChainInfo
+	}
+
+	var sectorsDeals []market.SectorDeals
+	var sectorsDataSpec []*market.SectorDataSpec
+	// Filters out the first wave of invalid updates with quick checks about sector status
+	var prevalidatedUpdates []*updateAndSectorInfo
```
ZenGround0, line 2120: nit: there can only be one wave so just call it validatedUpdates.  The comment can go too.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+	sectorNumbers := bitfield.New()
+	for _, update := range params.Updates {
+		// Bitfied.IsSet() is fast when there are only locally-set values.
+		set, err := sectorNumbers.IsSet(uint64(update.SectorID))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error checking sector number")
+		if set {
+			rt.Log(rtt.INFO, "duplicate sector being updated %d, skipping", update.SectorID)
```
ZenGround0, line 2127: rtt.WARN for skips

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
ZenGround0: I was not being precise enough in the wording of the FIP.  As much as possible errors that can't be 100% eliminated by operators should be skipped.  Because deadlines and partitions are effectively immutable operators are going to filter these errors at gas estimation.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+
+		// TODO Confirm the sector is actually in the stated deadline? This happens later when we try to update partition, but maybe do here first?
+
+		sectorInfo, err := sectors.MustGet(update.SectorID)
+		if err != nil {
+			rt.Log(rtt.INFO, "failed to get sector, skipping sector %d", update.SectorID)
+			continue
+		}
+
+		// TODO: this will change to "was originally dealless, and is currently no active deals"
```
ZenGround0, line 2157: Remove TODO.  I don't think this future improvement even needs to be tracked in specs-actors but you could make the argument this is the right place.  If you think so then the right place to record is in an issue.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+		if len(sectorInfo.DealIDs) != 0 {
+			rt.Log(rtt.INFO, "cannot update sector with deals, skipping sector %d", update.SectorID)
+			continue
+		}
+
+		// TODO: confirm sector is healthy (happens implicitly later, but should be done ahead of time)
```
ZenGround0, line 2163: I don't know what this means.  Are you planning to make any changes based on this?

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			SectorType: sectorInfo.SealProof,
+			DealIDs:    update.Deals,
+		})
+	}
+
+	// Errors past this point cause the ProveReplicaUpdates call to fail (no more skipping sectors)
+
+	dealWeights := requestDealWeights(rt, sectorsDeals)
+
+	builtin.RequirePredicate(rt, len(dealWeights.Sectors) == len(prevalidatedUpdates), exitcode.ErrIllegalState,
```
ZenGround0, line 2197: supernit: remove newlines here

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+
+	rt.ValidateImmediateCallerIs(append(info.ControlAddresses, info.Owner, info.Worker)...)
+
+	sectors, err := LoadSectors(store, stReadOnly.Sectors)
+	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load sectors array")
+
+	powerDelta := NewPowerPairZero()
+	pledgeDelta := big.Zero()
+
+	type updateAndSectorInfo struct {
```
ZenGround0, line 2112: Internally scoped types like this are a nice pattern I wish I'd been using.

ZenGround0 (commented):
Update partway through this review:

Let's use rtt.WARN for logging update-skip-on-error events

Introducing TODOs to specs-actors code is usually the wrong thing to do.  To fixup your existing todos you should either 1) do the thing referenced 2) just delete the todo 3) file and issue in specs-actors repo about the todo if it is something we really should be doing

actors/builtin/market/market_actor.go
```diff
@@ -337,6 +337,7 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 	}
 }
 
+// TODO alias
```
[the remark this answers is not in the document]
arajasek: sure...i like leaving notes while driving through code (to reduce things getting missed), but happy to comply

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
arajasek: Going with <= 4096 for now, but @BlocksOnAChain can we get an AI to get final sign-offs and confirmation of this (after everything has been finalized)? We've had whoopsies here in the past.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
arajasek: Yup, sgtm.

actors/builtin/miner/miner_actor.go
arajasek: Not ready to remove just yet (will do before calling PR merge ready).

actors/builtin/miner/miner_actor.go
arajasek: Are you sure about that? Is that something everyone validating the chain needs to be seeing?

I have felt like VM logs (through actors) is too verbose.

actors/builtin/miner/miner_actor.go
arajasek: Oh, this was an open question -- should we confirming that the sector is not faulty at this stage? I'm leaning towards yes, but unsure.

It happens later on when we adjust the partitions, but Idk whether that's sufficient.

arajasek: just realized this is explicitly called necessary in the FIP, so we should do it? https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0019.md#spec

ZenGround0: We definitely need to make sure it's not terminated, faulty, or unproven before updating.  Whether we do that here or later is a tradeoff as talked about above.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+				newSectorInfo.SealedCID = updateWithDetails.update.NewSealedSectorCID
+				newSectorInfo.DealIDs = updateWithDetails.update.Deals
+
+				newSectorInfo.DealWeight = updateWithDetails.dealWeight.DealWeight
+				newSectorInfo.VerifiedDealWeight = updateWithDetails.dealWeight.VerifiedDealWeight
+
+				// compute initial pledge
+				duration := updateWithDetails.sectorInfo.Expiration - rt.CurrEpoch()
+
+				// TODO: Should we include this?
```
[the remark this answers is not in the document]
arajasek: Ack, thanks

actors/builtin/miner/miner_actor.go
ZenGround0: Note the current code is safe because this is being done in [ReplaceSectors](https://github.com/filecoin-project/specs-actors/pull/1529/files#diff-d807293f55563b410c2d07719d2f38727016f41c6a17df8635500abc4070f74cR2302) implicitly. But it would be cleaner to explicitly look for termination, faulty and unproven without relying on behavior of ReplaceSectors

The way to do this is to load the partition and check that `partition.ActiveSectors()` contains the sector number in question

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,288 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to load partitions for deadline %d", dlIdx)
+
+			quant := st.QuantSpecForDeadline(dlIdx)
+
+			for _, updateWithDetails := range declsByDeadline[dlIdx] {
+				err = rt.VerifyReplicaUpdate(
+					proof.ReplicaUpdateInfo{
+						NewSealedSectorCID:   updateWithDetails.update.NewSealedSectorCID,
+						OldSealedSectorCID:   updateWithDetails.sectorInfo.SealedCID,
+						NewUnsealedSectorCID: updateWithDetails.unsealedSectorCID,
```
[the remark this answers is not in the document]
arajasek: seems like a nice to have, but okay

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,287 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+				err = partitions.Set(updateWithDetails.update.Partition, &partition)
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save deadline %v partition %v",
+					updateWithDetails.update.Deadline,
+					updateWithDetails.update.Partition)
+
+				deadline.Partitions, err = partitions.Root()
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save partitions for deadline %d", updateWithDetails.update.Deadline)
+
+				err = deadlines.UpdateDeadline(store, updateWithDetails.update.Deadline, deadline)
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to save deadline %d", updateWithDetails.update.Deadline)
```
ZenGround0, line 2322: This should move to 2329 after the deadline specific update loop.  It's not wrong but it kills all of the benefits of grouping updates by deadline.  If you do it this way you should just iterate through updates directly.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,287 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+				partitionPowerDelta, partitionPledgeDelta, err := partition.ReplaceSectors(store,
+					[]*SectorOnChainInfo{updateWithDetails.sectorInfo},
+					[]*SectorOnChainInfo{newSectorInfo},
+					info.SectorSize,
+					quant)
+
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to replace sector at deadline %d partition %d", updateWithDetails.update.Deadline, updateWithDetails.update.Partition)
+
+				powerDelta = powerDelta.Add(partitionPowerDelta)
+				pledgeDelta = big.Add(pledgeDelta, partitionPledgeDelta) // expected to be zero, see note below.
```
ZenGround0, line 2311: This should be nonzero in the case the initial pledge has increased.  You should call `notifyPledgeChanged` with `pledgeDelta` at the end of the method to handle this correctly.

ZenGround0 (changes requested):
I think this is the last iteration of major review, this is looking close to ready.

One general style nit is that you can remove newlines between logically connected areas of the code, i.e. a sequence of checks or a call and its error handling.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,287 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+				//	continue
+				//}
+
+				pwr := QAPowerForWeight(info.SectorSize, duration, newSectorInfo.DealWeight, newSectorInfo.VerifiedDealWeight)
+
+				newSectorInfo.ReplacedDayReward = updateWithDetails.sectorInfo.ExpectedDayReward
+				newSectorInfo.ExpectedDayReward = ExpectedRewardForPower(rewRet.ThisEpochRewardSmoothed, powRet.QualityAdjPowerSmoothed, pwr, builtin.EpochsInDay)
+				newSectorInfo.ExpectedStoragePledge = ExpectedRewardForPower(rewRet.ThisEpochRewardSmoothed, powRet.QualityAdjPowerSmoothed, pwr, InitialPledgeProjectionPeriod)
+				newSectorInfo.ReplacedSectorAge = maxEpoch(0, rt.CurrEpoch()-updateWithDetails.sectorInfo.Activation)
+
```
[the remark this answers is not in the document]
arajasek: Done, thanks!

actors/builtin/miner/miner_actor.go
arajasek: Yes, most definitely

actors/builtin/miner/miner_actor.go
arajasek: Ah, yes, thank you

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,291 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
 	return nil
 }
 
+type ReplicaUpdate struct {
+	SectorID           abi.SectorNumber
+	Deadline           uint64
+	Partition          uint64
+	NewSealedSectorCID cid.Cid `checked:"true"`
```
arajasek, line 2075: What exactly am I promising by slapping `checked` on here? That I sanity checked the input to be defined and the right prefix?

test-vectors/determinism-check
```diff
@@ -1 +1 @@
-- 81b8e58b84a114ef13064115cecd3af249e588cd2ec7100f34528838df31c5c7
```
arajasek, line 1: I'm surprised that this changed at all -- does this make sense to you? Is the addition of the new method somehow noticeable by the vectors?

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,295 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+
+			quant := st.QuantSpecForDeadline(dlIdx)
+
+			for _, updateWithDetails := range declsByDeadline[dlIdx] {
+				updateProofType, err := updateWithDetails.sectorInfo.SealProof.RegisteredUpdateProof()
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "couldn't load update proof type")
+
+				err = rt.VerifyReplicaUpdate(
+					proof.ReplicaUpdateInfo{
+						UpdateProof:          updateProofType,
```
arajasek, line 2252: So most other places take this as a param to the method (see WindowPoSt), even though there's only one correct value, and it can be determined at runtime. 

Is there a reason for that pattern? Should the proof type be an input to this method (that we then verify)?

test-vectors/determinism-check
ZenGround0: Did this only change after the sector key addition?  If you don't remember go back and check the commit before you made that change.  I would expect sector key addition to change everything since on chain sector info format is different.  I wouldn't expect introducing a new method and no new tests without state changes to change this.

actors/builtin/miner/miner_actor.go
ZenGround0: Yes please include it.  This is so that proofs can do versioning easily.

ZenGround0: The ffi itself usually takes these from the runtime methods.

actors/builtin/miner/miner_actor.go
ZenGround0: [Pretty much see @magik6k's original PR and discussion](https://github.com/filecoin-project/specs-actors/pull/717)

actors/runtime/proof/verify.go
```diff
@@ -36,6 +37,7 @@ type AggregateSealVerifyProofAndInfos = proof5.AggregateSealVerifyProofAndInfos
 // Information needed to verify a replica update
 
 type ReplicaUpdateInfo struct {
+	UpdateProof          abi.RegisteredUpdateProof
```
ZenGround0, line 40: `UpdateProofType` would help avoid confusion with the proof itself

actors/builtin/miner/miner_actor.go
```diff
@@ -2144,6 +2135,22 @@ func (a Actor) ProveReplicaUpdates(rt Runtime, params *ProveReplicaUpdatesParams
...
+
+		if !update.NewSealedSectorCID.Defined() {
+			rt.Abortf(exitcode.ErrIllegalArgument, "new sealed CID undefined")
+		}
+
+		if update.NewSealedSectorCID.Prefix() != SealedCIDPrefix {
+			rt.Abortf(exitcode.ErrIllegalArgument, "new sealed CID had wrong prefix")
+		}
+
+		err = stReadOnly.CheckSectorHealth(store, update.Deadline, update.Partition, update.SectorID)
```
ZenGround0, line 2151: A subtle problem with this is that it misses unproven sectors in this check.  These will be disallowed later on when calling ReplaceSectors so this method isn't broken but you want to do this here too for error handling reasons.

ZenGround0: You'll want to make a new check sector health method extended to check for unproven sectors

ZenGround0 (approved):
This LGTM

actors/builtin/miner/miner_actor.go
arajasek: Gotcha, thanks!

test-vectors/determinism-check
arajasek: D'oh, forgot about the sector key. Confirmed this only changed after that was added, so is as expected.

actors/builtin/miner/miner_actor.go
arajasek: Will do -- should the provided proof type be validated that it's `sectorInfo.SealProof.RegisteredUpdateProof()`?

arajasek: After an offline discussion (that also cleared up some miscommunication), the decision is to include the proof type in the params to `ProveReplicaUpdates`. At the cost of one byte, our future selves get the ability to much more easily support multiple valid UpdateProof types at the same time.

actors/builtin/miner/miner_actor.go
arajasek: Nice catch, thanks!

actors/builtin/miner/miner_actor.go
```diff
@@ -2074,6 +2074,7 @@ type ReplicaUpdate struct {
 	Partition          uint64
 	NewSealedSectorCID cid.Cid `checked:"true"`
 	Deals              []abi.DealID
+	UpdateProofType    abi.RegisteredUpdateProof
```
ZenGround0, line 2077: Right now this seems to be ignored since we send the sealProofInfo.RegisteredUpdateProof type into the runtime verification call.  This is ok for now.  Slight preference to enforce that this correctly matches sealProofInfo.RegisteredUpdateProof when doing the validation step

ZenGround0 (approved):

actors/builtin/miner/miner_actor.go
arajasek: Unless I'm misunderstanding you, that validation happens on line 2250.

arajasek: Merging for now.

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+						Proof:                updateWithDetails.update.ReplicaProof,
+					})
+
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to verify replica proof for sector %d", updateWithDetails.sectorInfo.SectorNumber)
+
+				newSectorInfo := updateWithDetails.sectorInfo
+
+				newSectorInfo.SealedCID = updateWithDetails.update.NewSealedSectorCID
+				if newSectorInfo.SectorKeyCID == nil {
+					newSectorInfo.SectorKeyCID = &updateWithDetails.sectorInfo.SealedCID
```
Kubuxu, line 2267: updateWithDetails.sectorInfo is aliased with newSectorInfo which makes this code do the wrong thing

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+						UpdateProofType:      updateProofType,
+						NewSealedSectorCID:   updateWithDetails.update.NewSealedSectorCID,
+						OldSealedSectorCID:   updateWithDetails.sectorInfo.SealedCID,
+						NewUnsealedSectorCID: updateWithDetails.unsealedSectorCID,
+						Proof:                updateWithDetails.update.ReplicaProof,
+					})
+
+				builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "failed to verify replica proof for sector %d", updateWithDetails.sectorInfo.SectorNumber)
+
+				newSectorInfo := updateWithDetails.sectorInfo
```
Kubuxu, line 2263: solution is to make this a copy

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+
+		if !update.NewSealedSectorCID.Defined() {
+			rt.Abortf(exitcode.ErrIllegalArgument, "new sealed CID undefined")
+		}
+
+		if update.NewSealedSectorCID.Prefix() != SealedCIDPrefix {
+			rt.Abortf(exitcode.ErrIllegalArgument, "new sealed CID had wrong prefix")
+		}
+
+		err = stReadOnly.CheckSectorHealthExcludeUnproven(store, update.Deadline, update.Partition, update.SectorID)
```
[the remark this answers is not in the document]
jennijuju: and why not just call `CheckSectorProven`?

actors/builtin/miner/miner_state.go
```diff
@@ -692,6 +693,50 @@ func (st *State) CheckSectorHealth(store adt.Store, dlIdx, pIdx uint64, sector a
 	return nil
 }
 
+// Returns an error if the target sector cannot be found and/or is faulty/terminated/unproven.
+func (st *State) CheckSectorHealthExcludeUnproven(store adt.Store, dlIdx, pIdx uint64, sector abi.SectorNumber) error {
```
jennijuju, line 697: what's the difference between this and `CheckSectorProven` ?

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
jennijuju: if can be replaced by `CheckSectorProven`, we can have deadline and partition removed from PRU param too?

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalArgument, "sector is not healthy")
+
+		sectorInfo, err := sectors.MustGet(update.SectorID)
```
jennijuju, line 2155: should this be checked and done before `CheckSectorHealthExcludeUnproven`? if the sectors cannot be found, it implies the sector won't be proven, may save us one state read? idk

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+		if err != nil {
+			rt.Log(rtt.INFO, "failed to get sector, skipping sector %d", update.SectorID)
+			continue
+		}
+
+		if len(sectorInfo.DealIDs) != 0 {
```
jennijuju, line 2161: would it be better to check if `SectorKeyCID` is null or not? (in preparation of snap on snap - IIUC,  both CC sector and existing deal sectors will have `null` for `SectorKeyCID` - and they shouldn't be able to snapped.

Later on, when you can snap on snap deal sector, the DealID array wont be 0 length, but then you can get a non-null sectorkeycid -> can be updated again!

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			builtin.MethodsMarket.ActivateDeals,
+			&market.ActivateDealsParams{
+				DealIDs:      update.Deals,
+				SectorExpiry: sectorInfo.Expiration,
+			},
+			abi.NewTokenAmount(0),
+			&builtin.Discard{},
+		)
+
+		if code != exitcode.Ok {
```
jennijuju, line 2177: well.. annoying me again - but like...should we consider active the valid deals instead of skipping the whole sector? (similar to what was discussed for prove commit back then for  chocolate )

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+			SectorType: sectorInfo.SealProof,
+			DealIDs:    update.Deals,
+		})
+	}
+
+	builtin.RequireParam(rt, len(validatedUpdates) > 0, "no valid updates")
+
+	// Errors past this point cause the ProveReplicaUpdates call to fail (no more skipping sectors)
+
+	dealWeights := requestDealWeights(rt, sectorsDeals)
```
jennijuju, line 2198: okie so this is more of a q for my understanding

so seems like we can get the deal weights from activate deal earlier if we want, so we should save the value back then to save some gas cost here? cuz seems like `VerifyDealsForActivation` will cause some gas usage for loading the state again ?

actors/builtin/miner/policy.go
```diff
@@ -167,6 +167,10 @@ var MaxProveCommitDuration = map[abi.RegisteredSealProof]abi.ChainEpoch{
 // 32 sectors per epoch would support a single miner onboarding 1EiB of 32GiB sectors in 1 year.
 const PreCommitSectorBatchMaxSize = 256
 
+// The maximum number of sector replica updates in a single batch.
+// Same as PreCommitSectorBatchMaxSize for consistency
+const ProveReplicaUpdatesMaxSize = PreCommitSectorBatchMaxSize
```
jennijuju, line 172: I believe this should be re-evaluate and lowered according to our call today?

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
arajasek: Yes on soft failure!

`CheckSectorProven` doesn't seem to be doing all that much -- it seems to just check if the sector no is in the `Sectors` at all. I'm not sure whether that's enough. Maybe that + calling the existing `CheckSectorHealth` method might be enough?

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+}
+
+type ProveReplicaUpdatesParams struct {
+	Updates []ReplicaUpdate
+}
+
+func (a Actor) ProveReplicaUpdates(rt Runtime, params *ProveReplicaUpdatesParams) bitfield.BitField {
+	// Validate inputs
+
+	builtin.RequireParam(rt, len(params.Updates) <= ProveReplicaUpdatesMaxSize, "too many updates (%d > %d)", len(params.Updates), ProveReplicaUpdatesMaxSize)
```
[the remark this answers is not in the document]
arajasek: Will defer to whatever's in the audit followups doc.

actors/builtin/miner/miner_state.go
arajasek: it also checks that the sector isn't faulty

actors/builtin/miner/miner_actor.go
```diff
@@ -2067,6 +2068,297 @@ func (a Actor) RepayDebt(rt Runtime, _ *abi.EmptyValue) *abi.EmptyValue {
...
+		set, err := sectorNumbers.IsSet(uint64(update.SectorID))
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error checking sector number")
+		if set {
+			rt.Log(rtt.INFO, "duplicate sector being updated %d, skipping", update.SectorID)
+			continue
+		}
+
+		sectorNumbers.Set(uint64(update.SectorID))
+
+		if len(update.ReplicaProof) > 4096 {
```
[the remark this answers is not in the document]
arajasek: Will follow up from the audit notes

actors/builtin/miner/miner_actor.go
arajasek: i think you're right that it's less expensive than the health check, but i don't think it's worth changing

actors/builtin/miner/miner_actor.go
arajasek: Yeah, that _will_ become the change, but right now it's just "are there currently any deals"

We could also assert that the keycid is null -- if that isn't true but there are no deals, that's programmer error. Still good to catch

actors/builtin/miner/miner_actor.go
arajasek: Would be nice, but the Market Actor doesn't tell you which deals were successfully activated and which ones were failures, so it's impossible to selectively skip.

actors/builtin/miner/miner_actor.go
arajasek: Hmm, that's a great question. It seems like `VerifyDealsForActivation` and `ValidateDealsForActivation` do different things? But they both calculate weights, but in different ways? And the call to `ValidateDealsForActivation` in `ActivateDeals` explicitly ignores the resultant weight calculations?

@ZenGround0 pls help

PR #1537, opened 2021-12-03
Feat/alter burn funds

koalacxr:
this modified as follows:
1. modify burnFunds method, add the burn type, and change the log type to Info , query the burn type in logs.
2. modify multiAddrs parameter type of the ConstructMinerInfo method so that the type is used correctly.

### DIFF
--- a/actors/builtin/miner/burn_type.go
+++ b/actors/builtin/miner/burn_type.go
@@ -1,0 +1,16 @@
+package miner
+
+type BurnType string
+
+const (
+	BurnTypeDisputeWindowedPoSt      BurnType = "DisputeWindowedPoSt"
+	BurnTypePreCommitSectorBatch     BurnType = "PreCommitSectorBatch"
+	BurnTypeProveCommitAggregate     BurnType = "ProveCommitAggregate"
+	BurnTypeDeclareFaultsRecovered   BurnType = "DeclareFaultsRecovered"
+	BurnTypeApplyRewards             BurnType = "ApplyRewards"
+	BurnTypeReportConsensusFault     BurnType = "ReportConsensusFault"
+	BurnTypeWithdrawBalance          BurnType = "WithdrawBalance "
+	BurnTypeRepayDebt                BurnType = "RepayDebt"
+	BurnTypeProcessEarlyTerminations BurnType = "processEarlyTerminations"
+	BurnTypeHandleProvingDeadline    BurnType = "handleProvingDeadline "
+)
--- a/actors/builtin/miner/burn_type.go
+++ b/actors/builtin/miner/burn_type.go
@@ -1,16 +1,16 @@
 package miner
 
-type BurnType string
+type BurnMethod string
 
 const (
-	BurnTypeDisputeWindowedPoSt      BurnType = "DisputeWindowedPoSt"
-	BurnTypePreCommitSectorBatch     BurnType = "PreCommitSectorBatch"
-	BurnTypeProveCommitAggregate     BurnType = "ProveCommitAggregate"
-	BurnTypeDeclareFaultsRecovered   BurnType = "DeclareFaultsRecovered"
-	BurnTypeApplyRewards             BurnType = "ApplyRewards"
-	BurnTypeReportConsensusFault     BurnType = "ReportConsensusFault"
-	BurnTypeWithdrawBalance          BurnType = "WithdrawBalance "
-	BurnTypeRepayDebt                BurnType = "RepayDebt"
-	BurnTypeProcessEarlyTerminations BurnType = "processEarlyTerminations"
-	BurnTypeHandleProvingDeadline    BurnType = "handleProvingDeadline "
+	BurnMethodDisputeWindowedPoSt      BurnMethod = "DisputeWindowedPoSt"
+	BurnMethodPreCommitSectorBatch     BurnMethod = "PreCommitSectorBatch"
+	BurnMethodProveCommitAggregate     BurnMethod = "ProveCommitAggregate"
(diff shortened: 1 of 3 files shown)


koalacxr:
burnFunds method changed to make it easier to query what type of burn is consumed？

actors/builtin/miner/miner_actor.go
```diff
@@ -2937,9 +2937,9 @@ func resolveWorkerAddress(rt Runtime, raw addr.Address) addr.Address {
 	return resolved
 }
 
-func burnFunds(rt Runtime, amt abi.TokenAmount) {
+func burnFunds(rt Runtime, amt abi.TokenAmount, bt BurnType) {
 	if amt.GreaterThan(big.Zero()) {
-		rt.Log(rtt.DEBUG, "storage provder %s burning %s", rt.Receiver(), amt)
+		rt.Log(rtt.INFO, "storage provder %s burn type %s burning %s", rt.Receiver(), bt, amt)
```
ZenGround0, line 2942: Must be debug as this will be very noisy

actors/builtin/miner/burn_type.go
```diff
@@ -0,0 +1,16 @@
+package miner
+
+type BurnType string
```
ZenGround0, line 3: nit: change name to `BurnMethod`

actors/builtin/miner/miner_actor.go
ZenGround0:
```suggestion
		rt.Log(rtt.DEBUG, "storage provder %s burn type %s burning %s", rt.Receiver(), bt, amt)
```

actors/builtin/miner/burn_type.go
```diff
@@ -0,0 +1,16 @@
...
+const (
+	BurnTypeDisputeWindowedPoSt      BurnType = "DisputeWindowedPoSt"
+	BurnTypePreCommitSectorBatch     BurnType = "PreCommitSectorBatch"
+	BurnTypeProveCommitAggregate     BurnType = "ProveCommitAggregate"
+	BurnTypeDeclareFaultsRecovered   BurnType = "DeclareFaultsRecovered"
+	BurnTypeApplyRewards             BurnType = "ApplyRewards"
+	BurnTypeReportConsensusFault     BurnType = "ReportConsensusFault"
+	BurnTypeWithdrawBalance          BurnType = "WithdrawBalance "
+	BurnTypeRepayDebt                BurnType = "RepayDebt"
+	BurnTypeProcessEarlyTerminations BurnType = "processEarlyTerminations"
```
ZenGround0, line 14: nit: capitalize for consistency

ZenGround0 (changes requested):
I'll merge if you're good with logging being taken down to debug.

actors/builtin/miner/burn_type.go
koalacxr: agree with you

actors/builtin/miner/burn_type.go
koalacxr: agree with you

actors/builtin/miner/miner_actor.go
koalacxr: BurnFunds method logs only appear when burn occurs in Miner Actor. Generally, debug logs are not enabled in non-test environments, and burn information is a very important part of burnFunds method logs.

koalacxr: Because reward 、transfer and withdraw are important,  and burn is also important, need to check the account.

koalacxr: We wonder what type of method causes burn.

ZenGround0: No disagreement that it is important.  But it is not worth the tradeoff.  Burning is a common operation across all miner actors.  All lotus nodes will be overrun with log noise if we keep at info.  

The right thing to work on to make this better is an extension to actors logging that subscribes to logging at different levels per actor.

koalacxr: The burn method is important because you can filter the burn content by grep command on the log file.

ZenGround0: I understand but I'm not going to merge this with log level info

koalacxr: Don't get the burn method name in Lotus, only value.

ZenGround0: Right, and with the suggested change you will now get the burn method name if you turn on actors debug logs.

koalacxr: debug mode  enabled will be very noisy. Other debug logs are not required.

koalacxr: The server running Lotus enables info-level log. So burn log level also needs info-level. The debug mode is not suitable for the server running mode.

ZenGround0: There are many users of this code and while I would like to help you I am not going to make their lives worse because of your particular setup.  You don't want to use debug logs because they are noisy but we would cause the exact same situation for other users by logging the burn info of every miner at the INFO level.

If you want to keep working on getting burn events logged you can either (1) merge debug logs and modify your setup to filter only what you need from them (2) come up with a design for logging at different levels based on actor address.  I think (2) will be more work but I can talk about it with you in an issue if you want to go that route.

koalacxr: I think (2) feasible. thanks!   restore burnFunds log level to debug.

ZenGround0 (approved):

PR #1542, opened 2021-12-14
Remove vestiges of CC upgrade by replacement

anorth:
No functional change, the removed code was unreachable.

### DIFF
--- a/actors/builtin/miner/miner_state.go
+++ b/actors/builtin/miner/miner_state.go
@@ -655,48 +655,10 @@ func (st *State) PopEarlyTerminations(store adt.Store, maxPartitions, maxSectors
 	return result, !noEarlyTerminations, nil
 }
 
-// Returns an error if the target sector cannot be found and/or is faulty/terminated.
-func (st *State) CheckSectorHealth(store adt.Store, dlIdx, pIdx uint64, sector abi.SectorNumber) error {
-	dls, err := st.LoadDeadlines(store)
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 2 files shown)


actors/builtin/miner/miner_state.go
```diff
@@ -655,48 +655,10 @@ func (st *State) PopEarlyTerminations(store adt.Store, maxPartitions, maxSectors
...
-	}
-
-	return nil
-}
-
 // Returns an error if the target sector cannot be found, or some other bad state is reached.
 // Returns false if the target sector is faulty, terminated, or unproven
 // Returns true otherwise
-func (st *State) CheckSectorHealthExcludeUnproven(store adt.Store, dlIdx, pIdx uint64, sector abi.SectorNumber) (bool, error) {
+func (st *State) CheckSectorHealthExcludeUnproven(store adt.Store, dlIdx, pIdx uint64, sector abi.SectorNumber, requireProven bool) (bool, error) {
```
arajasek, line 661: Why add this if it's never being called with false?

arajasek (commented):
Thanks! Just one q?

actors/builtin/miner/miner_state.go
anorth: So that someone doesn't copy-paste this method in the future when they want a minor change in functionality.

I can remove this flag if @ZenGround0 concurs.

ZenGround0: LGTM

ZenGround0 (approved):

ZenGround0 (approved):

Issue #1521, opened 2021-10-20
Fix PoSt Dispute so snap deals upgrades can't be exploited 


ZenGround0:
PoSt disputing relies on immutable SealedCIDs in sector infos between submission of optimistic window post and dispute but the current snap deals construction mutates this field.  The consequence is that the PoSt proof covering the snap deals updated sector can always be successfully disputed until the updated sector is first PoSted with the new SealedCID.

A good way to fix this is to add a snapshot of sector infos to the DeadlineInfo.  Then during the dispute this sector info tree is read instead of the current one.  This is like the PartitionsSnapshot which exists for the same reason.

The migration should be simple.  The v7 deadline struct has a new field and it is populated with whatever the current sector info root from the miner state is.  Because SealedCid mutations only happen after the upgrade this should be fine.

ZenGround0:
## Description
The detailed description is above in the first issue comment.

## Done Criteria
* implementation in specs-actors and passing review
* a VM test verifying that a post cannot be successfully disputed immediately after a snap deals upgrade of a sector 

### Non-criteria
* the migration is tracked elsewhere


## Dependencies
* https://github.com/filecoin-project/specs-actors/issues/1511

## Ongoing work

Not yet started

arajasek:
We also need to decide what to do about sectors whose deadlines are currently open -- is that upgradeable?

BlocksOnAChain:
@Kubuxu @lucaniz, as we discussed on the protocol call - what is your opinion on: https://github.com/filecoin-project/specs-actors/issues/1521#issuecomment-975683414

ZenGround0:
> We also need to decide what to do about sectors whose deadlines are currently open -- is that upgradeable?

There shouldn't be any problems.  In particular post should be unaffected. Did you have other concerns about this?

arajasek:
@ZenGround0 The q (in my mind) is what happens if I Snap a sector whose deadline is currently open (a sector that i haven't posted for yet). Does the expected post become for the new sector cid or the old one?

The new sector probably makes more sense (and likely requires no code changes here), but might be something we need to explicitly handle in SP implementations like the lotus miner.

ZenGround0:
Ah I see this is a good point.  I think this actually does warrant actors code changes.  I think we should use a variation of the mutable deadlines idea that ProveCommitSector uses in order to avoid a lot of pain and suffering.  Updating sectors in the current and next deadlines should be an error.

[...]

PR #1544, opened 2021-12-18
FIP-0019: Snapshot sectors array for disputing PoSts

arajasek:
Addresses #1521


actors/builtin/miner/deadline_state.go
```diff
@@ -1261,6 +1281,9 @@ func (d *Deadline) IsLive() (bool, error) {
 		return true, nil
 	}
 
+	// TODO: Should the deadline be considered live if the sectors snapshot has changed?
```
ZenGround0, line 1284: No there's nothing worth checking in the sectors snapshots.  Partition and post snapshots cover edge cases where deadline is considered live without having live sectors.  The sectors snapshot is more information taken from miner state than information about this deadline.

ZenGround0: Another way to put it is that any changes to sectors in this deadline require that those sectors are live which is covered by the first check in this method.

ZenGround0 (approved):
LGTM.  A test that disputing correctly references the sectors snapshot and fails as expected would be a good addition to #1539

PR #1547, opened 2021-12-21
Feat/proof runtime types

ZenGround0:
Needed for snap deals integration


actors/builtin/miner/miner_actor.go
```diff
@@ -2068,7 +2068,7 @@ type ProveReplicaUpdatesParams struct {
 	Updates []ReplicaUpdate
 }
 
-func (a Actor) ProveReplicaUpdates(rt Runtime, params *ProveReplicaUpdatesParams) bitfield.BitField {
+func (a Actor) ProveReplicaUpdates(rt Runtime, params *ProveReplicaUpdatesParams) *bitfield.BitField {
```
ZenGround0, line 2071: lotus makes an assumption that all output values are pointers (shows up doing some reflect dereferencing) so I changed this to match that convention.

arajasek (commented):
Where is this actually getting used? I didn't see any of the FFI calls changing around WinningPoSt...

ZenGround0:
>  i'm surprised that the butterflynet was able to handle a PRU message if that's the case (unless it's just in the API or some such)

I am also surprised, I was assuming this would break butterfly.  Then again if it was launched with the feat/snap-deals branch of lotus then this change was getting pulled in.  feat/snap-deals updated actors v7 to point to here.

ZenGround0:
> Where is this actually getting used? I didn't see any of the FFI calls changing around WinningPoSt...

Yes no changes to the ffi call.  The problem is the layer above.  It needs to be able to select between sealed copy and updated replica when reading on disk files to start satisfying post challenges for generating both winning and window post.  This new sector info data type provides the info needed to select the correct file.

arajasek (approved):
That makes more sense! 2 followups:

- Can you link me to where the assumption about pointerness lives in Lotus? I think butterfly was updated before this commit existed, so we have some figuring out to do.
- Why is the Activation needed as part of the "figuring out how to PoSt"

ZenGround0:
1. Having trouble finding this with confidence.  I'll try changing it back and seeing where things fail again tomorrow.
2. Activation will tell us when the upgrade kicked in on chain. We need this because winning post needs to look at state a finality back and we want to make sure we only start generating winning posts off of the upgraded replica when the upgrade happened finality epochs ago.

ZenGround0:
> Can you link me to where the assumption about pointerness lives in Lotus? I think butterfly was updated before this commit existed, so we have some figuring out to do.

Here it is: https://github.com/filecoin-project/lotus/blob/master/chain/stmgr/utils.go#L37.  I think the butterfly mystery can be resolved by either 1) we built butterfly off of the feat/snap-deals branch which linked to a pointer return value, or 2) this message never succeeded on chain in butterfly.

PR #1548, opened 2021-12-23
NV15: Speed up the migration

arajasek:
The special Sectors migrating logic is as follows:

- For each miner, if the Head CID is found in the cache, use the cached value (same as any other actor)
- If not load the current state, do the quick steps like migrating MinerInfo
- Check the cache for the sectors AMT root -- if found, use the cached value and return back to miner migration
  - Note that this step will cover many hundreds of thousands of "empty" miners in the state tree
- If not found, check the cache to see if a previous sectors AMT has been cached **for this miner**. If not found, manually migrate all sectors, and insert the incoming and outgoing AMTs into the cache (specific to the miner)
- If the cache does have a previous sectors AMT cached for this miner, load it and its corresponding migrated AMT. Diff the previous AMT against the current one, mutating the previous migrated AMT appropriately. Update the cached AMT roots.

### DIFF
--- a/actors/migration/nv15/miner.go
+++ b/actors/migration/nv15/miner.go
@@ -3,79 +3,251 @@ package nv15
 import (
 	"context"
 
-	builtin7 "github.com/filecoin-project/specs-actors/v7/actors/builtin"
-	"github.com/filecoin-project/specs-actors/v7/actors/util/adt"
-	"golang.org/x/xerrors"
-
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-amt-ipld/v3"
+	"github.com/filecoin-project/go-state-types/abi"
 	miner6 "github.com/filecoin-project/specs-actors/v6/actors/builtin/miner"
+	builtin7 "github.com/filecoin-project/specs-actors/v7/actors/builtin"
 	miner7 "github.com/filecoin-project/specs-actors/v7/actors/builtin/miner"
-
-	cid "github.com/ipfs/go-cid"
+	"github.com/filecoin-project/specs-actors/v7/actors/util/adt"
+	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
+	"golang.org/x/xerrors"
 )
 
-type minerMigrator struct{}
+type minerMigrator struct {
+	emptyDeadlineV6  cid.Cid
+	emptyDeadlinesV6 cid.Cid
+	emptyDeadlineV7  cid.Cid
+	emptyDeadlinesV7 cid.Cid
+	emptySectorsV7   cid.Cid
+}
+
+func newMinerMigrator(ctx context.Context, store cbor.IpldStore) (*minerMigrator, error) {
+	if miner6.WPoStPeriodDeadlines != miner7.WPoStPeriodDeadlines {
+		return nil, xerrors.Errorf("unexpected WPoStPeriodDeadlines changed from %d to %d",
+			miner6.WPoStPeriodDeadlines, miner7.WPoStPeriodDeadlines)
+	}
+
+	ctxStore := adt.WrapStore(ctx, store)
+	edv6, err := miner6.ConstructDeadline(ctxStore)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadline v6: %w", err)
+	}
+
+	edv6cid, err := store.Put(ctx, edv6)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put empty deadline v6: %w", err)
+	}
+
+	edsv6 := miner6.ConstructDeadlines(edv6cid)
+	edsv6cid, err := store.Put(ctx, edsv6)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadlines v6 : %w", err)
+	}
+
+	edv7, err := miner7.ConstructDeadline(ctxStore)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadline v7: %w", err)
+	}
+
+	edv7cid, err := store.Put(ctx, edv7)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put empty deadline v7: %w", err)
+	}
+
+	edsv7 := miner7.ConstructDeadlines(edv7cid)
+	edsv7cid, err := store.Put(ctx, edsv7)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadlines v7 : %w", err)
+
+	}
+
+	essCid, err := adt.StoreEmptyArray(ctxStore, miner7.SectorsAmtBitwidth)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty sectors snapshot array: %w", err)
+	}
+
+	return &minerMigrator{
+		emptyDeadlineV6:  edv6cid,
+		emptyDeadlinesV6: edsv6cid,
+		emptyDeadlineV7:  edv7cid,
+		emptyDeadlinesV7: edsv7cid,
+		emptySectorsV7:   essCid,
+	}, nil
+}
 
 func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, in actorMigrationInput) (*actorMigrationResult, error) {
 	var inState miner6.State
 	if err := store.Get(ctx, in.head, &inState); err != nil {
-		return nil, err
+		return nil, xerrors.Errorf("getting inState: %w", err)
 	}
 
-	sectorsOut, err := migrateSectors(ctx, store, inState.Sectors)
+	outState := fromv6State(inState)
+	ctxStore := adt.WrapStore(ctx, store)
+
+	sectorsOut, err := migrateSectors(ctx, ctxStore, in.cache, in.address, inState.Sectors)
 	if err != nil {
-		return nil, err
+		return nil, xerrors.Errorf("failed to migrate sectors: %w", err)
 	}
 
-	outState := miner7.State{
-		Info:                       inState.Info,
-		PreCommitDeposits:          inState.PreCommitDeposits,
-		LockedFunds:                inState.LockedFunds,
-		VestingFunds:               inState.VestingFunds,
-		FeeDebt:                    inState.FeeDebt,
-		InitialPledge:              inState.InitialPledge,
-		PreCommittedSectors:        inState.PreCommittedSectors,
-		PreCommittedSectorsCleanUp: inState.PreCommittedSectorsCleanUp,
-		AllocatedSectors:           inState.AllocatedSectors,
-		Sectors:                    sectorsOut,
-		ProvingPeriodStart:         inState.ProvingPeriodStart,
-		CurrentDeadline:            inState.CurrentDeadline,
-		Deadlines:                  inState.Deadlines,
-		EarlyTerminations:          inState.EarlyTerminations,
-		DeadlineCronActive:         inState.DeadlineCronActive,
+	outState.Sectors = sectorsOut
+
+	deadlinesOut, err := m.migrateDeadlines(ctx, ctxStore, inState.Deadlines, sectorsOut)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to migrate deadlines: %w", err)
 	}
+
+	outState.Deadlines = deadlinesOut
 
 	newHead, err := store.Put(ctx, &outState)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to flush outState: %w", err)
+	}
+
 	return &actorMigrationResult{
 		newCodeCID: m.migratedCodeCID(),
 		newHead:    newHead,
-	}, err
+	}, nil
 }
 
 func (m minerMigrator) migratedCodeCID() cid.Cid {
 	return builtin7.StorageMinerActorCodeID
 }
 
-func migrateSectors(ctx context.Context, store cbor.IpldStore, inRoot cid.Cid) (cid.Cid, error) {
-	ctxStore := adt.WrapStore(ctx, store)
-	inArray, err := adt.AsArray(ctxStore, inRoot, miner6.SectorsAmtBitwidth)
-	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to read sectors array: %w", err)
+func migrateSectors(ctx context.Context, store adt.Store, cache MigrationCache, minerAddr address.Address, inRoot cid.Cid) (cid.Cid, error) {
+	return cache.Load(SectorsAmtKey(inRoot), func() (cid.Cid, error) {
+		inArray, err := adt.AsArray(store, inRoot, miner6.SectorsAmtBitwidth)
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to read sectors array: %w", err)
+		}
+
+		okIn, prevInRoot, err := cache.Read(MinerPrevSectorsInKey(minerAddr))
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to get previous inRoot from cache: %w", err)
+		}
+
+		okOut, prevOutRoot, err := cache.Read(MinerPrevSectorsOutKey(minerAddr))
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to get previous outRoot from cache: %w", err)
+		}
+
+		var outArray *adt.Array
+		if okIn && okOut {
+			// we have previous work, but the AMT has changed -- diff them
+			diffs, err := amt.Diff(ctx, store, store, prevInRoot, inRoot, amt.UseTreeBitWidth(miner7.SectorsAmtBitwidth))
+			if err != nil {
+				return cid.Undef, xerrors.Errorf("failed to diff old and new Sector AMTs: %w", err)
+			}
+
+			inSectors, err := miner6.LoadSectors(store, inRoot)
+			if err != nil {
+				return cid.Undef, xerrors.Errorf("failed to load inSectors: %w", err)
+			}
+
+			prevOutSectors, err := miner7.LoadSectors(store, prevOutRoot)
+			if err != nil {
+				return cid.Undef, xerrors.Errorf("failed to load prevOutSectors: %w", err)
+			}
+
+			for _, change := range diffs {
+				switch change.Type {
+				case amt.Remove:
+					if err := prevOutSectors.Delete(change.Key); err != nil {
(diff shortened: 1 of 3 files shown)


actors/builtin/miner/partition_state.go
```diff
@@ -56,10 +58,7 @@ const PartitionExpirationAmtBitwidth = 4
 const PartitionEarlyTerminationArrayAmtBitwidth = 3
 
 // Value type for a pair of raw and QA power.
-type PowerPair struct {
```
arajasek, line 59: Unrelated to this migration, sorry.

actors/migration/nv15/top.go
```diff
@@ -57,6 +53,20 @@ type MigrationCache interface {
 	Load(key string, loadFunc func() (cid.Cid, error)) (cid.Cid, error)
 }
 
+func CidKey(c cid.Cid) string {
```
arajasek, line 56: I'm just using one catch-all key for all migrated objects (unlike the ActorHeadKey etc we had before). Is that bad?

actors/migration/nv15/top.go
```diff
@@ -226,10 +238,9 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, actorsRootIn ci
 
 type actorMigrationInput struct {
 	address    address.Address // actor's address
-	balance    abi.TokenAmount // actor's balance
-	head       cid.Cid         // actor's state head CID
-	priorEpoch abi.ChainEpoch  // epoch of last state transition prior to migration
-	cache      MigrationCache  // cache of existing cid -> cid migrations for this actor
+	head       cid.Cid
```
arajasek, line 241: balance shouldn't have been in here, i don't think

actors/migration/nv15/top.go
```diff
@@ -246,7 +257,7 @@ type actorMigration interface {
 
 type migrationJob struct {
 	address.Address
-	states6.Actor
+	Actor states6.Actor
```
arajasek, line 260:
```suggestion
	states6.Actor
```
Stale, sorry.

actors/migration/nv15/top.go
```diff
@@ -57,6 +53,20 @@ type MigrationCache interface {
 	Load(key string, loadFunc func() (cid.Cid, error)) (cid.Cid, error)
 }
 
+func CidKey(c cid.Cid) string {
+	return c.String()
+}
+
+func MinerPrevSectorsInKey(m address.Address) string {
```
arajasek, line 60: we'll likely wanna persist these to allow for restarts after premigrations have run. maybe.

actors/migration/nv15/miner.go
```diff
@@ -19,63 +25,184 @@ type minerMigrator struct{}
...
+func migrateSectors(ctx context.Context, store adt.Store, cache MigrationCache, minerAddr address.Address, inRoot cid.Cid) (cid.Cid, error) {
+	return cache.Load(CidKey(inRoot), func() (cid.Cid, error) {
+		inArray, err := adt.AsArray(store, inRoot, miner6.SectorsAmtBitwidth)
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to read sectors array: %w", err)
+		}
+
+		okIn, prevInRoot, err := cache.Read(MinerPrevSectorsInKey(minerAddr))
+		if err != nil {
+			// TODO: Should we instead absorb errors like these and proceed cache-lessly?
```
arajasek, line 72: Opinions solicited

actors/migration/nv15/miner.go
```diff
@@ -19,63 +25,184 @@ type minerMigrator struct{}
...
 
 func (m minerMigrator) migratedCodeCID() cid.Cid {
 	return builtin7.StorageMinerActorCodeID
 }
 
-func migrateSectors(ctx context.Context, store cbor.IpldStore, inRoot cid.Cid) (cid.Cid, error) {
-	ctxStore := adt.WrapStore(ctx, store)
-	inArray, err := adt.AsArray(ctxStore, inRoot, miner6.SectorsAmtBitwidth)
+func migrateSectors(ctx context.Context, store adt.Store, cache MigrationCache, minerAddr address.Address, inRoot cid.Cid) (cid.Cid, error) {
+	return cache.Load(CidKey(inRoot), func() (cid.Cid, error) {
```
arajasek, line 64: The overwhelming majority of actors should be covered in the cache here:

- Even if this is the first premigration, the majority of miners are "empty" and so have identical sectors AMT roots
- If a premigration HAS already run, the non-empty miners will very likely be covered here because the sectors AMT doesn't change that much -- new sectors and extensions are the only things that can change it, i think (._..and CompactPartitions_)

ZenGround0: For peace of mind use the pattern we established last year to type cache entries: 
1. Types: https://github.com/filecoin-project/specs-actors/blob/a2aec9c6af0e55ff9f7da257b9b5ca4c3e88b8e9/actors/migration/nv10/top.go#L47
2. Usage: https://github.com/filecoin-project/specs-actors/blob/release/v3/actors/migration/nv10/miner.go#L40

Data collisions (the thing being hashed as an AMT somehow has the same bytes as the thing being hashed as miner state) really should be impossible but it would be nice not to think about this at all.

actors/migration/nv15/top.go
ZenGround0: Ah nice missed this call out.  Its probably fine but better to be safe.  More details in [this review comment](https://github.com/filecoin-project/specs-actors/pull/1548#discussion_r777055395)

actors/migration/nv15/top.go
ZenGround0: This is a lotus side thing but I agree

gen/gen.go
```diff
@@ -176,7 +176,7 @@ func main() {
 		miner.Deadline{},
 		miner.Partition{},
 		miner.ExpirationSet{},
-		miner.PowerPair{},
+		//miner.PowerPair{}, // Aliased
```
ZenGround0, line 179: nit: from v6

actors/migration/nv15/miner.go
ZenGround0: Either way is fine but I think we can rely on cache not erroring on read of a valid key.  If the cache is written right I expect the errors to be local system dependent only.

actors/migration/nv15/miner.go
```diff
@@ -19,63 +25,184 @@ type minerMigrator struct{}
...
+				return cid.Undef, xerrors.Errorf("failed to load prevOutSectors: %w", err)
+			}
+
+			for _, change := range diffs {
+				switch change.Type {
+				case amt.Remove:
+					if err := prevOutSectors.Delete(change.Key); err != nil {
+						return cid.Undef, xerrors.Errorf("failed to delete sector from prevOutSectors: %w", err)
+					}
+				case amt.Add:
```
ZenGround0, line 123: I think this needs to `fallthrough` otherwise you don't migrate sectors added to the amt in between.  https://go.dev/play/p/fPxSEXpevf7. This highlights the need for thorough testing.  A good place to draw inspiration is the deal label migration test: https://github.com/filecoin-project/specs-actors/pull/1551/files#diff-5f4089c6e3f8495e10bfa13e18ff8f0e34cb787c56256e0bfabc576f0253f66aR36

ZenGround0 (changes requested):
Other than the one bug this looks good.  It will be important to thoroughly test edge cases and the actors vm test framework works pretty well for this.

actors/migration/nv15/miner.go
```diff
@@ -19,63 +25,184 @@ type minerMigrator struct{}
...
+	outState := fromv6State(inState)
+	ctxStore := adt.WrapStore(ctx, store)
+
+	sectorsOut, err := migrateSectors(ctx, ctxStore, in.cache, in.address, inState.Sectors)
 	if err != nil {
-		return nil, err
+		return nil, xerrors.Errorf("failed to migrate sectors: %w", err)
 	}
 
-	outState := miner7.State{
```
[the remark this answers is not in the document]
laudiacay: honestly i found the old pattern hard to read- it was hard to see which fields were changing and which weren't? but I see this concern. valid.

laudiacay (changes requested):
I don't have any comments outside of what zen said other than "it would be cool to have a test for this". 

looks good and correct unless my lying eyes deceive me, and tests are good for catching code reviewers' lying eyes :P agree that the deal label migration code would be a good start

actors/builtin/miner/partition_state.go
arajasek: Ack

actors/migration/nv15/top.go
arajasek: Ack, will do!

actors/migration/nv15/miner.go
arajasek: Yessir!

actors/migration/nv15/miner.go
arajasek: sg, i think it's better to error -- doing something silently wrong is probably worse

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,201 @@ package nv15
...
+			if err != nil {
+				return cid.Undef, xerrors.Errorf("failed to construct new sectors array: %w", err)
+			}
+
+			var sectorInfo miner6.SectorOnChainInfo
+			if err = inArray.ForEach(&sectorInfo, func(k int64) error {
+				return outArray.Set(uint64(k), migrateSectorInfo(sectorInfo))
+			}); err != nil {
+				return cid.Undef, err
+			}
```
Stebalien, line 92: Followup: https://github.com/filecoin-project/go-amt-ipld/issues/61

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,201 @@ package nv15
...
+		return cid.Undef, xerrors.Errorf("unexpected WPoStPeriodDeadlines changed from %d to %d",
+			miner6.WPoStPeriodDeadlines, miner7.WPoStPeriodDeadlines)
 	}
 
-	var sectorInfo miner6.SectorOnChainInfo
-	if err = inArray.ForEach(&sectorInfo, func(k int64) error {
-		return outArray.Set(uint64(k), migrateSectorInfo(sectorInfo))
-	}); err != nil {
-		return cid.Undef, err
+	outDeadlines := miner7.Deadlines{Due: [miner7.WPoStPeriodDeadlines]cid.Cid{}}
```
Stebalien, line 167: Does this not work?

```suggestion
	var outDeadlines miner7.Deadlines
```

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,201 @@ package nv15
...
+
+	for i, c := range inDeadlines.Due {
+		outDlCid, err := cache.Load(DeadlineKey(c), func() (cid.Cid, error) {
```
Stebalien, line 170: This assumption is wrong. If the sectors CID has changed, we need to migrate the deadlines regardless.

We have two options:

1. Just migrate every time. Honestly, this is probably _fine_ as long as we're caching top-level actor migrations. If a miner is _active_, their deadlines will have changed significantly anyways. If a miner isn't active, the top-level actor cache will skip the miner.
2. Key by deadline CID _and_ sectors CID.

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,201 @@ package nv15
...
+		}
+
+		okOut, prevOutRoot, err := cache.Read(MinerPrevSectorsOutKey(minerAddr))
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to get previous outRoot from cache: %w", err)
+		}
+
+		if okIn != okOut {
+			return cid.Undef, xerrors.Errorf("cannot have only one of the inRoot and outRoot in the cache")
+		}
```
Stebalien, line 77: If we get an error here, we may want to just log it, assume _nothing_ is cached, and do the migration.

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,201 @@ package nv15
...
+			return cid.Undef, xerrors.Errorf("error writing new sectors AMT: %w", err)
+		}
+
+		if err = cache.Write(MinerPrevSectorsInKey(minerAddr), inRoot); err != nil {
+			return cid.Undef, xerrors.Errorf("failed to write previous sectors in key to cache: %w", err)
+		}
+
+		if err = cache.Write(MinerPrevSectorsOutKey(minerAddr), outRoot); err != nil {
+			return cid.Undef, xerrors.Errorf("failed to write previous sectors out key to cache: %w", err)
+		}
```
Stebalien, line 150: Again, consider logging an error and moving on.

actors/migration/nv15/top.go
```diff
@@ -57,6 +53,28 @@ type MigrationCache interface {
...
+	return addr.String() + "-h-" + head.String()
+}
+
+func SectorsAmtKey(sectorsAmt cid.Cid) string {
+	return "s-" + sectorsAmt.String()
+}
+
+func DeadlineKey(dl cid.Cid) string {
+	return "d-" + dl.String()
+}
```
Stebalien, line 66: Can we use `StringOfBase` here? Just to be clear.

actors/migration/nv15/top.go
```diff
@@ -57,6 +53,28 @@ type MigrationCache interface {
...
+
+func MinerPrevSectorsInKey(m address.Address) string {
+	// "i" for "in"
+	return "i-" + m.String()
```
Stebalien, line 70: nit: we can afford a few more characters.

actors/migration/nv15/top.go
Stebalien: I think we needed it for some migrations in the past.

Stebalien (changes requested):

actors/migration/nv15/miner.go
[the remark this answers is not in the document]
arajasek: I was lazy to do so because the cache as currently written can only have a CID as the `value` type :P

actors/migration/nv15/miner.go
arajasek: It does, I just copied the v3 migration

actors/migration/nv15/miner.go
arajasek: Yessir, you right -- I think I'm gonna migrate every time, with a special case for the empty deadline and deadlines objects which I think will be a sizeable difference.

actors/migration/nv15/miner.go
arajasek: Yeah, this was a convo between Zen and me upstream^, neither of us felt too strongly...I think I'll adopt your suggestion

actors/migration/nv15/top.go
arajasek: Looks like that would mean that the Key methods would have to return an error (or do something else about the error), which isn't great...

Stebalien: Panic? That can only happen if the encoding isn't supported.

Basically, `dl.String()` isn't guaranteed to a _specific_ base.

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,251 @@ package nv15
...
+				return cid.Undef, err
+			}
+		}
+
+		outRoot, err := outArray.Root()
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("error writing new sectors AMT: %w", err)
+		}
+
+		_ = cache.Write(MinerPrevSectorsInKey(minerAddr), inRoot)
```
arajasek, line 199: Silently dropping the error here, because I wasn't sure whether concurrent access of the logger would be safe...anyone know?

actors/migration/nv15/top.go
arajasek: Sure, I was too scared to do that :P

arajasek:
Rebased on top of latest master (revert of deal label PR)

arajasek:
Testing update: Ran on mainnet state with and without the cache, same result. 

Next steps are a unit test coming from @geoff-vball and maybe writing an unmigration, and testing that.

actors/migration/nv15/miner.go
Stebalien: Yes: https://github.com/filecoin-project/lotus/blob/dc26f3bd7453c515ca9aecc16fb9b3517784ef5e/chain/consensus/filcns/upgrades.go#L1205

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,251 @@ package nv15
...
+
+		_ = cache.Write(MinerPrevSectorsOutKey(minerAddr), outRoot)
+		return outRoot, nil
+	})
+}
+
+func (m minerMigrator) migrateDeadlines(ctx context.Context, store adt.Store, deadlines cid.Cid, sectors cid.Cid) (cid.Cid, error) {
+	if deadlines == m.emptyDeadlinesV6 && sectors == m.emptySectorsV7 {
```
Stebalien, line 207: Why do I care if the sectors are empty?

actors/migration/nv15/miner.go
```diff
@@ -3,79 +3,251 @@ package nv15
...
 	}
 
-	var sectorInfo miner6.SectorOnChainInfo
-	if err = inArray.ForEach(&sectorInfo, func(k int64) error {
-		return outArray.Set(uint64(k), migrateSectorInfo(sectorInfo))
-	}); err != nil {
-		return cid.Undef, err
+	var outDeadlines miner7.Deadlines
+	for i, c := range inDeadlines.Due {
+		if c == m.emptyDeadlineV6 && sectors == m.emptySectorsV7 {
```
Stebalien, line 219: I'm not sure if this is correct. Usually, when we hit a "dead" deadline, we skip it and won't record any snapshots: https://github.com/filecoin-project/specs-actors/blob/d46a6510c2ad20ca453f72e292adbe880a8bacb8/actors/builtin/miner/miner_state.go#L1149. That means that never-used deadlines shouldn't have sector snapshots.

This is actually a really interesting edge-case because we can run into a state where we keep a _really old_ copy of the sectors array and never update it (if a deadline "dies"). This isn't a problem for _partitions_ because we only snapshot the deadline's partitions (so, when the deadline "dies", the partitions array and the partitions snapshot array are identical).

actors/migration/nv15/miner.go
arajasek: Just being extrachecky -- i'm not really sure what it means to have an empty deadlines but non empty sectors, so i decided to say migrate it manually if that happens :P

Stebalien (approved):
We're going to address my comments in a new PR as we need to change behavior a bit.

arajasek:
Landing this, thanks Steb!

PR #1555, opened 2022-01-06
Use go-state-types v0.1.3

arajasek:


laudiacay (approved):
have you written a test for this?????

PR #1551, opened 2021-12-31
Pick/deal label

ZenGround0:
Migration converting deal proposal label to cbor bytes from cbor string

FIP draft: https://github.com/filecoin-project/FIPs/pull/188

Already reviewed before merging onto next

PR #1556, opened 2022-01-07
Revert #1551

laudiacay:
This isn't going into next release, we need to integrate it into lotus and that'll take a minute. from request of @arajasek in slack


arajasek (approved):

arajasek:
Background here is that we suspect that the tooling changes needed to accommodate this are prohibitively complex, so it's probably not getting into nv15 (v7 actors).

PR #1557, opened 2022-01-10
Zero out Sectors snapshot if no proofs have been submitted

arajasek:


actors/builtin/miner/deadline_state.go
```diff
@@ -1281,6 +1292,9 @@ func (d *Deadline) IsLive() (bool, error) {
 		return true, nil
 	}
 
+	// TODO: Should we also return true here if the SectorsSnapshot is non-empty?
```
arajasek, line 1295: Will remove if it's a "nah, don't bother"

actors/builtin/miner/deadline_state.go
```diff
@@ -1281,6 +1292,9 @@ func (d *Deadline) IsLive() (bool, error) {
...
+	// I don't think that can happen (without meeting an earlier return condition), but if it does it should get cleaned up?
```
Stebalien, line 1296: I think we're fine as-is.

actors/migration/nv15/miner.go
```diff
@@ -216,7 +216,8 @@ func (m minerMigrator) migrateDeadlines(ctx context.Context, store adt.Store, de
 
 	var outDeadlines miner7.Deadlines
 	for i, c := range inDeadlines.Due {
-		if c == m.emptyDeadlineV6 && sectors == m.emptySectorsV7 {
+		if c == m.emptyDeadlineV6 {
+			// empty deadlines don't get sectors snapshots
```
Stebalien, line 220: And dead deadlines? Or deadlines with no snapshotted proofs? I think we need a to check if there are snapshotted proofs here. If not, we skip.

actors/builtin/miner/deadline_state.go
arajasek: Ack

actors/migration/nv15/miner.go
arajasek: I convinced myself that anything except a fully empty deadline would be `Process`ed in 24 hours, at which point it would have its sectors zeroed out. 

Maybe that isn't true? And if not, shall we just check for `Live`ness here, and Snapshot sectors iff it's alive?

Stebalien: Deadlines can be _dead_ but non-empty.

Simper than checking liveness: snapshot if and only if there's a proof snapshot.

arajasek: Yeah, makes sense.

actors/migration/nv15/miner.go
```diff
@@ -216,26 +216,21 @@ func (m minerMigrator) migrateDeadlines(ctx context.Context, store adt.Store, de
 
 	var outDeadlines miner7.Deadlines
```
Stebalien, line 217: Several limes up, we can remove the `&& sectors == m.emptySectorsV7`. If the deadlines are empty, we don't need to do anything.

Stebalien (approved):

PR #1559, opened 2022-01-11
Add both the pre & post state-trees to conformance test CARs

Stebalien:
Before: Only the _pre_ state was included in the vector car files.
After: Both pre & post state is included. This makes it easier to debug issues.

This adds a few MiB (72->79MiB), but that really doesn't matter much.


support/vm/vector.go
```diff
@@ -78,12 +79,6 @@ func SetStartState(v *VM) Option {
 			return err
```
anorth, line 79: Rename this method `SetStartStateTree` for symmetry with end state.

support/vm/vector_gen.go
```diff
@@ -52,6 +53,17 @@ func (g *vectorGen) before(v *VM, name string) error {
...
 }
 
+func (g *vectorGen) fillCar(v *VM) error {
+	var err error
+	getter := nodeGetterFromStore(v.Store())
+	roots := []cid.Cid{g.vector.StartStateTree}
+	if g.conformance() {
+		roots = append(roots, g.vector.EndStateTree)
+	}
+	g.vector.State, err = encodeCAR(getter, roots...)
```
anorth, line 63: Rather than directly manipulate `vector.State`, it would be more consistent to follow the pattern of adding a `SetState()` option-style method. Other such method already take the VM as a parameter, so this logic could move there.

anorth (approved):
Functionally looks good. I recommend restructuring a little for consistency and abstraction.

PR #1560, opened 2022-01-11
Prototype implementation of FIL+ explicity subsidy

anorth:
This is an implementation sketch of the FIL+ explicit subsidy, in support of the FIP ([discussion](https://github.com/filecoin-project/FIPs/discussions/243)). The goal is to flesh out enough of the implementation to discover all the key challenges, and be able to test some of the assumptions and mechanisms.

This PR will probably never land, because we expect to migrate to Rust actors running in the FVM before it does. This PR can thus guide implementation over there after the FIP is accepted.

- [x] Remove QA power from miner and power actors
- [x] Account for verified space in market actor
- [ ] Implement reward split and claim/payment of verified space rewards
- [ ] Implement fault notification and handling by market actor
- [ ] Implement deal pledge in the market actor
- [ ] Implement fault penalties in market actor
- [ ] (Maybe) update all the tests

### DIFF
--- a/actors/builtin/cbor_gen.go
+++ b/actors/builtin/cbor_gen.go
@@ -158,8 +158,8 @@ func (t *ConfirmSectorProofsParams) MarshalCBOR(w io.Writer) error {
 		return err
 	}
 
-	// t.QualityAdjPowerSmoothed (smoothing.FilterEstimate) (struct)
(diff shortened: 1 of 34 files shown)


anorth:
If this is ever implemented it will be in the new FVM built-in actors.

PR #1561, opened 2022-01-12
fix: lotus returns SysErrForbidden, not ErrForbidden

Stebalien:


anorth (approved):
Thanks. The support VM was in error here.

PR #1567, opened 2022-01-13
Add both the pre & post state-trees to conformance test CARs

Stebalien:
Backport of #1559

This adds a few MiB (72->79MiB), but that really doesn't matter much.


Stebalien:
Merging as this just backports an approved patch (and doesn't affect consensus).

Issue #1554, opened 2022-01-04
Test the nv15 miner migration


arajasek:
nv15 is expected to have migrations for the miner, market, and verifreg. The miner has the most involved migration, relying on some new logic, and so needs some thorough testing. 

- Create an interesting v6 state tree
  - a large number of miners with different owner/worker addresses, with a large spread of sectors (many of them with 0 sectors, a good number with < 100 sectors spread across a few deadlines, and a small number with 10^6 - 10^7 sectors).
  - _(optional)_ throw in some accounts / multisigs for good measure
  - seed some of the sectors with deals created by different clients
- Migrate this tree, populating the cache. Assert correctness*
- Modify this tree by adding new miners, adding new sectors to existing miners, modifying sectors in the existing miners (at least one CC upgrade, and some extensions), deleting sectors from existing miners (needs both termination and compact partitions), changing miner info (worker address, for eg).
- Advance this tree a few hundred epochs, migrate using the populated cache. Assert correctness.
- _(optional)_ Assert that unchanged details (actor states, deadlines, sector AMTs, etc) get read and used from the cache
- Repeat the modification and assertion 1-2 more times.

*Assert correctness can be done a few different ways:
- The easiest is probably just to load every miner: every sector / deadline, and assert that the diff is as expected -- all fields are the same, except new fields which are null
- Also assert invariants such as:
  - Total power is the same
  - Total number of actors in the state tree is the same
  - Each individual miner's power is the same, etc.
- One more involved way is to write the unmigration for the nv15 migration. This shouldn't be _too_ much more work, but does mean that you can "unmigrate" the output, and check that its root is exactly the same as the input, which is quite nice.

ZenGround0:
Only things to add are (1) take a look at `CheckStateInvariants` which you can call from tests and should check most things and (2) run migration and check state invariants on main net state, either with ent or adding a new code path into lotus that checks invariants

PR #1568, opened 2022-01-17
Draft: Migration test for miner actor

geoff-vball:
Closes issue #1554

### DIFF
--- a/actors/migration/nv15/test/miner_migration_test.go
+++ b/actors/migration/nv15/test/miner_migration_test.go
@@ -2,9 +2,14 @@ package test
 
 import (
 	"context"
+
+	"github.com/filecoin-project/specs-actors/v7/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v7/actors/builtin/verifreg"
+
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
 	"github.com/filecoin-project/go-state-types/rt"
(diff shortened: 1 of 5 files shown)


arajasek (commented):
Rebased onto master and moved out of `actors/test` to remove from the determinism corpus.

Everything here looks generally good. There's basically unlimited scope for additional cases here, but I think we can land this if we add assertions that:

- network stats are unchanged
- every miner's stats are unchanged (lots can be done here, but at least power and # of active / faulty sectors)
- every single sector has null SectorKey
- the verified registry actor's `RemoveDataCapProposalIDs` is empty

actors/migration/nv15/test/miner_migration_test.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	for i, minerInfo := range minerInfos {
+		vm6Util.ProveCommitSectors(t, v, minerInfo.WorkerAddress, minerInfo.MinerAddress, precommitInfo[i], addDeals)
+	}
+
+	return append(minersToProve, minerInfos...), v
+}
+
+func TestMinerMigration(t *testing.T) {
```
arajasek, line 147:
```suggestion
func TestNv15Migration(t *testing.T) {
```

arajasek (approved):

anorth:
Any migration test should be invoking `states.CheckStateInvariants` after the migration to check a huge range of state invariants. And if there are new invariants in this network version, add them to the `CheckStateInvariants` for the relevant actor, or to the top-level for cross-actor invariants.

PR #1571, opened 2022-01-28
feat: implement fip-0029, support beneficiary

hunjixin:
This is an implementation for FIP-0029 PR #1571 to provide Beneficiary address support. In this implementation, version and upgrading is not considered, that is, directly put the code change into the current version. The back-compliance could be done once the next version content is determined.


actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+		info := getMinerInfo(rt, &st)
+
+		// This is a ChangeBeneficiary proposal when the caller is Owner
+		if rt.Caller() == info.Owner {
+			rt.ValidateImmediateCallerIs(info.Owner)
+			if params.NewBeneficiary != info.Owner {
+				// No need to check others when the newBeneficiary is set back to owner
+				checkNewBeneficialParam(rt, params)
+			}
+			if info.Beneficiary != info.Owner && info.BeneficiaryInfo.Effective(rt.CurrEpoch()) {
```
anorth, line 3168: Comment: "If there is a current beneficiary, store the new proposal and set next approver to that current beneficiary".

If you take my suggestion to add approval flags rather than a strict sequence, this case will merge with the one below.

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+				info.PendingBeneficiaryInfo = &PendingBeneficiaryChange{
+					NewBeneficiary: params.NewBeneficiary,
+					NewQuota:       params.NewQuota,
+					NewExpiration:  params.NewExpiration,
+					NextApprover:   info.Beneficiary,
+				}
+			} else if params.NewBeneficiary != info.Owner {
```
anorth, line 3175: "Else (no current beneficiary), if proposing a beneficiary other than the owner store the proposal and set next approver to the new beneficiary. If proposing the owner, it doesn't need approval (because the owner is calling)."

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+				info.PendingBeneficiaryInfo = &PendingBeneficiaryChange{
+					NewBeneficiary: params.NewBeneficiary,
+					NewQuota:       params.NewQuota,
+					NewExpiration:  params.NewExpiration,
+					NextApprover:   params.NewBeneficiary,
+				}
+			} else {
```
anorth, line 3182: "Else (no current beneficiary and not proposing a new one) immediately set the beneficiary info to the owner.

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+				info.Beneficiary = params.NewBeneficiary
```
anorth, line 3183: `info.Beneficiary = info.Owner` just to be clear

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+				info.BeneficiaryInfo = BeneficiaryInfo{
+					Quota:      big.Zero(),
+					Expiration: 0,
+					UsedQuota:  big.Zero(),
+				}
+				info.PendingBeneficiaryInfo = nil
+			}
+		} else {
+			// Non-owner calls ChangeBeneficiary is to approve a pending proposal
```
anorth, line 3192: "If the caller is not the owner, they must be approving an existing proposal made by the owner"

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+			builtin.RequireState(rt, info.PendingBeneficiaryInfo != nil, "No changeBeneficiary proposal exists")
```
anorth, line 3193: I think `ErrForbidden` is better here

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+	}
+}
+
+// ChangeBeneficiary proposal/approve beneficiary change
+func (a Actor) ChangeBeneficiary(rt Runtime, params *ChangeBeneficiaryParams) *abi.EmptyValue {
+	if params == nil {
+		rt.Abortf(exitcode.ErrIllegalArgument, "empty ChangeBeneficiary Paramenters")
+	}
+	if params.NewBeneficiary.Protocol() != addr.ID {
+		rt.Abortf(exitcode.ErrIllegalArgument, "NewBeneficiary address must be an ID address")
```
anorth, line 3154: Perhaps slightly friendlier, we could resolve the address to an ID address here, and then store that. Providing the ID address might be hard for some callers (like web UIs)

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+		} else {
+			// Non-owner calls ChangeBeneficiary is to approve a pending proposal
+			builtin.RequireState(rt, info.PendingBeneficiaryInfo != nil, "No changeBeneficiary proposal exists")
+			rt.ValidateImmediateCallerIs(info.PendingBeneficiaryInfo.NextApprover)
+
+			builtin.RequireParam(rt, info.PendingBeneficiaryInfo.NewBeneficiary == params.NewBeneficiary, "new beneficiary address must be equal expect %s, but got %s", info.PendingBeneficiaryInfo.NewBeneficiary, params.NewBeneficiary)
+			builtin.RequireParam(rt, info.PendingBeneficiaryInfo.NewQuota.Equals(params.NewQuota), "new beneficiary quota must be equal expect %s, but got %s", info.PendingBeneficiaryInfo.NewQuota, params.NewQuota)
+			builtin.RequireParam(rt, info.PendingBeneficiaryInfo.NewExpiration == params.NewExpiration, "new beneficiary expiredate must be equal expect %s, but got %s", info.PendingBeneficiaryInfo.NewExpiration, params.NewExpiration)
+
+			if params.NewBeneficiary != rt.Caller() && params.NewBeneficiary != info.Owner {
```
anorth, line 3200: I think `params.NewBeneficiary == info.Owner` is impossible, so this check is confusing.

Explain this better. "If the caller is not the new beneficiary, the caller must be the old beneficiary, and so the proposal still needs approval by the new one". Given that explanation, should we check the caller really is the old beneficiary? This may get simpler if the approval chain is more flexible.

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+				// We still need the NewBeneficiary to approve this proposal
+				info.PendingBeneficiaryInfo.NextApprover = params.NewBeneficiary
+			} else {
+				// All approved, Set BeneficiaryInfo
+				if params.NewBeneficiary == info.Owner {
+					// The beneficiary set back to owner
+					info.BeneficiaryInfo = BeneficiaryInfo{
+						Quota:      big.Zero(),
```
anorth, line 3208: If we enforce these zeros when the proposal is made, then we can drop this `if` branch and unify with the ones below.

anorth (commented):
Given the representation of the beneficiary address always being non-nil, and equal to the owner address by default, we must update the beneficiary address when the owner address is changed, if they are equal. Also it's hard to think about various cases when a proposed beneficiary change is active. I think we should delete any proposal when the owner changes. Please fix ChangeOwnerAddress to do this.

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+						Expiration: 0,
+						UsedQuota:  big.Zero(),
+					}
+				} else if params.NewBeneficiary == info.Beneficiary {
```
anorth, line 3212: You could simplify the cases below to (a) always set the quota and expiration according to the pending change, and then if new != old, reset the used quota.

actors/builtin/miner/miner_test.go
```diff
@@ -3582,11 +3867,17 @@ func newHarness(t testing.TB, provingPeriodOffset abi.ChainEpoch) *actorHarness
...
+		receiver:    receiver,
+		owner:       owner,
+		worker:      worker,
+		key:         workerKey,
+		beneficiary: owner,
+		beneficiaryInfo: miner.BeneficiaryInfo{
+			Quota:      abi.TokenAmount{},
+			Expiration: 0,
+			UsedQuota:  abi.TokenAmount{},
+		},
```
anorth, line 3880: I see you need the beneficiary address in the harness in order to set mock expectations, but I don't think you need the beneficiaryInfo.

actors/builtin/miner/miner_test.go
```diff
@@ -3797,6 +4123,22 @@ func (h *actorHarness) checkState(rt *mock.Runtime) {
...
+	param := &miner.ChangeBeneficiaryParams{
+		NewBeneficiary: newBeneficiaryAddr,
+		NewQuota:       Quota,
+		NewExpiration:  expireDate,
+	}
+
+	rt.ExpectValidateCallerAddr(callerId)
+	rt.SetCaller(callerId, builtin.AccountActorCodeID)
+	rt.Call(h.a.ChangeBeneficiary, param)
+	rt.Verify()
```
anorth, line 4139: Set `h.beneficiary` to the new value here.

actors/builtin/miner/miner_test.go
```diff
@@ -3617,6 +3908,41 @@ func (h *actorHarness) setProofType(proof abi.RegisteredSealProof) {
 	require.NoError(h.t, err)
 }
 
+func (h *actorHarness) setBeneficiary(rt *mock.Runtime, beneficiaryAddr addr.Address, beneficiary miner.BeneficiaryInfo) {
```
anorth, line 3911: I don't think you need this. Instead, update the harness benficiary address after successfully changing it through the method calls.

actors/builtin/miner/miner_test.go
```diff
@@ -3617,6 +3908,41 @@ func (h *actorHarness) setProofType(proof abi.RegisteredSealProof) {
...
+	require.Equal(h.t, found, true)
+	h.beneficiary = beneficiaryIdAddr
+	h.beneficiaryInfo = miner.BeneficiaryInfo{
+		Quota:      beneficiary.Quota,
+		Expiration: beneficiary.Expiration,
+	}
+}
+
+func (h *actorHarness) changeOwnerToNewBeneficiary(rt *mock.Runtime, beneficiaryAddr addr.Address, beneficiary miner.BeneficiaryInfo) {
+	beneficiaryIdAddr, found := rt.GetIdAddr(beneficiaryAddr)
```
anorth, line 3922: Remove this resolution if we get the actor to do it instead. Or just always call this with the right ID address.

actors/builtin/miner/miner_test.go
```diff
@@ -3617,6 +3908,41 @@ func (h *actorHarness) setProofType(proof abi.RegisteredSealProof) {
...
+	require.Equal(h.t, found, true)
+	h.beneficiary = beneficiaryIdAddr
```
anorth, line 3924: Do this after the successful call.

actors/builtin/miner/miner_test.go
```diff
@@ -3617,6 +3908,41 @@ func (h *actorHarness) setProofType(proof abi.RegisteredSealProof) {
...
+	h.beneficiaryInfo = miner.BeneficiaryInfo{
+		Quota:      beneficiary.Quota,
+		Expiration: beneficiary.Expiration,
+	}
+}
+
+func (h *actorHarness) changeOwnerToNewBeneficiary(rt *mock.Runtime, beneficiaryAddr addr.Address, beneficiary miner.BeneficiaryInfo) {
```
anorth, line 3921: This method name is a bit confusing. How about "proposeAndApproveInitialBeneficiary"?

support/mock/mockrt.go
```diff
@@ -50,7 +50,7 @@ type Runtime struct {
 
 	// Actor state
 	state   cid.Cid
-	balance abi.TokenAmount
+	balance map[addr.Address]abi.TokenAmount
```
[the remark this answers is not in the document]
hunjixin: what I want to verify here is that if I withdraw to another address, I want to verify the amount withdrawn. Is there any way to do this?

actors/builtin/miner/miner_actor.go
```diff
@@ -1983,13 +1984,14 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *abi.T
 	newlyVested := big.Zero()
 	feeToBurn := big.Zero()
 	availableBalance := big.Zero()
+	amountWithdrawn := big.Zero()
+
 	rt.StateTransaction(&st, func() {
 		var err error
 		info = getMinerInfo(rt, &st)
-		// Only the owner is allowed to withdraw the balance as it belongs to/is controlled by the owner
+		// Only the owner/beneficiary is allowed to withdraw the balance as it belongs to/is controlled by the owner
```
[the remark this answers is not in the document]
hunjixin: only beneficiary is allowed to withdraw, has fixed

actors/builtin/miner/miner_test.go
```diff
@@ -2240,6 +2240,80 @@ func TestWithdrawBalance(t *testing.T) {
...
+
+		actor.changeOwnerToNewBeneficiary(rt, benecifiaryId, miner.BeneficiaryInfo{
+			Quota:      bigBalance,
+			Expiration: periodOffset,
+			UsedQuota:  abi.TokenAmount{},
+		})
+		// withdraw 1% of balance
+		actor.withdrawFunds(rt, onePercentBalance, onePercentBalance, big.Zero())
+		actor.checkState(rt)
+		assert.True(t, rt.BalanceOf(benecifiaryId).Equals(onePercentBalance), "beneficiary address can not withdraw expect balance expect %s but got %s", onePercentBalance, rt.BalanceOf(benecifiaryId))
```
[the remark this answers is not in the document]
hunjixin: I see it, there is no need to check here, it has already been in ```withdrawFunds```

actors/builtin/miner/miner_state.go
```diff
@@ -85,6 +85,37 @@ type State struct {
...
+	// Return 0 when the usedQuota > Quota for safe
+	// This could happen while there is a race when setting beneficialInfo.newQuota lower
+	return big.Max(big.Sub(beneficiaryInfo.Quota, beneficiaryInfo.UsedQuota), big.NewInt(0))
+}
+
+type PendingBeneficiaryChange struct {
+	NewBeneficiary addr.Address
+	NewQuota       abi.TokenAmount
+	NewExpiration  abi.ChainEpoch
+	NextApprover   addr.Address
```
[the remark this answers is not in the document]
hunjixin: Great ideas!  looks more clear in this way

actors/builtin/miner/miner_state.go
```diff
@@ -85,6 +85,37 @@ type State struct {
...
+
+func (beneficiaryInfo BeneficiaryInfo) IsUsedUp() bool {
+	return beneficiaryInfo.UsedQuota.GreaterThanEqual(beneficiaryInfo.Quota)
+}
+
+func (beneficiaryInfo BeneficiaryInfo) IsExpire(cur abi.ChainEpoch) bool {
+	return beneficiaryInfo.Expiration < cur
+}
+
+func (beneficiaryInfo BeneficiaryInfo) Available() abi.TokenAmount {
```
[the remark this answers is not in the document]
hunjixin: LGTM

actors/builtin/miner/miner_state.go
```diff
@@ -100,6 +131,10 @@ type MinerInfo struct {
 
 	PendingWorkerKey *WorkerKeyChange
 
+	Beneficiary            addr.Address
+	BeneficiaryInfo        BeneficiaryInfo
```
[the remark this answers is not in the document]
hunjixin: use BeneficiaryTerm

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3125,114 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+	}
+	if params.NewBeneficiary.Protocol() != addr.ID {
+		rt.Abortf(exitcode.ErrIllegalArgument, "NewBeneficiary address must be an ID address")
+	}
+
+	var st State
+	rt.StateTransaction(&st, func() {
+		info := getMinerInfo(rt, &st)
+
+		// This is a ChangeBeneficiary proposal when the caller is Owner
```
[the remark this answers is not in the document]
hunjixin: After rewriting the logic you mentioned above, it becomes more clear here, no need to consider various relationships between current beneficiary, newbenneficiary and owner.

steven004:
@anorth , thanks for your comments, which are thoroughly considered and make things even simpler. Please review again. 
And, please also see if a new branch (instead of master) needed to do this merge.

actors/builtin/miner/miner_test.go
```diff
@@ -2616,6 +2679,253 @@ func TestChangeMultiAddrs(t *testing.T) {
...
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, firstBeneficiary, beneficiaryChange{firstBeneficiaryId, quota, expiration + 1}, &firstBeneficiary)
+		})
+
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, firstBeneficiary, beneficiaryChange{firstBeneficiaryId, big.Add(quota, big.NewInt(1)), expiration}, &firstBeneficiary)
+		})
+
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, firstBeneficiary, beneficiaryChange{secondBeneficiaryId, big.Add(quota, big.NewInt(1)), expiration}, &firstBeneficiary)
```
anorth, line 2898: Set the quota back so that this tests only the different address.

actors/builtin/miner/miner_test.go
```diff
@@ -2616,6 +2679,253 @@ func TestChangeMultiAddrs(t *testing.T) {
...
+		})
+	})
+
+	t.Run("failure change beneficiary has wrong params", func(t *testing.T) {
+		rt, actor := setupFunc()
+		actor.constructAndVerify(rt)
+
+		rt.SetEpoch(periodOffset)
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, actor.owner, beneficiaryChange{firstBeneficiary, abi.NewTokenAmount(10), periodOffset - 10}, nil)
```
anorth, line 2908: Should this be using `firstBeneficiaryId`? I'm not quite sure. Could you document which check each of these is verifying?

actors/builtin/miner/miner_test.go
```diff
@@ -2616,6 +2679,253 @@ func TestChangeMultiAddrs(t *testing.T) {
...
+			actor.changeBeneficiary(rt, firstBeneficiary, beneficiaryChange{firstBeneficiaryId, quota, expiration + 1}, &firstBeneficiary)
+		})
+
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, firstBeneficiary, beneficiaryChange{firstBeneficiaryId, big.Add(quota, big.NewInt(1)), expiration}, &firstBeneficiary)
+		})
+
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, firstBeneficiary, beneficiaryChange{secondBeneficiaryId, big.Add(quota, big.NewInt(1)), expiration}, &firstBeneficiary)
+		})
```
anorth, line 2899: For completeness, show a successful call with the correct parameters here, demonstrating that it would have succeeded. 

(If you want, you can collapse this with other tests, e.g. by moving these failure cases up to "successfully change beneficiary to another address". Your call, both ways are good for me).

actors/builtin/miner/miner_test.go
```diff
@@ -2616,6 +2679,253 @@ func TestChangeMultiAddrs(t *testing.T) {
...
+			actor.changeBeneficiary(rt, actor.owner, beneficiaryChange{firstBeneficiaryId, abi.NewTokenAmount(0), periodOffset + 10}, nil)
+		})
+
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, actor.owner, beneficiaryChange{actor.owner, abi.NewTokenAmount(20), 0}, nil)
+		})
+
+		rt.ExpectAbort(exitcode.ErrIllegalArgument, func() {
+			actor.changeBeneficiary(rt, actor.owner, beneficiaryChange{actor.owner, abi.NewTokenAmount(0), 1}, nil)
+		})
```
anorth, line 2925: Similarly, show what a successful call looks like

anorth (commented):
Thanks, this basically looks good to me. My comments and suggestions now are relatively minor.

I'd like to defer to @ZenGround0 for final approval though, and discussion about when and where to merge.

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3130,84 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+			}
+		} else {
+			// Non-owner calls ChangeBeneficiary is to approve a pending proposal
+			builtin.RequirePredicate(rt, info.PendingBeneficiaryTerm != nil, exitcode.ErrForbidden, "No changeBeneficiary proposal exists")
+
+			builtin.RequireParam(rt, info.PendingBeneficiaryTerm.NewBeneficiary == newBeneficiary, "new beneficiary address must be equal expect %s, but got %s", info.PendingBeneficiaryTerm.NewBeneficiary, params.NewBeneficiary)
+			builtin.RequireParam(rt, info.PendingBeneficiaryTerm.NewQuota.Equals(params.NewQuota), "new beneficiary quota must be equal expect %s, but got %s", info.PendingBeneficiaryTerm.NewQuota, params.NewQuota)
+			builtin.RequireParam(rt, info.PendingBeneficiaryTerm.NewExpiration == params.NewExpiration, "new beneficiary expiredate must be equal expect %s, but got %s", info.PendingBeneficiaryTerm.NewExpiration, params.NewExpiration)
+		}
+
```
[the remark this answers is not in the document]
hunjixin: have checked pending benenficiary term before

actors/builtin/miner/miner_test.go
```diff
@@ -2616,6 +2679,253 @@ func TestChangeMultiAddrs(t *testing.T) {
...
+		quota := abi.NewTokenAmount(100)
+		expiration := periodOffset * 2
+		actor.changeBeneficiary(rt, actor.owner, beneficiaryChange{firstBeneficiaryId, quota, expiration}, nil)
+		// assert change has been made in state
+		info := actor.getInfo(rt)
+		require.EqualValues(t, expiration, info.PendingBeneficiaryTerm.NewExpiration)
+		require.EqualValues(t, quota, info.PendingBeneficiaryTerm.NewQuota)
+		require.EqualValues(t, firstBeneficiaryId, info.PendingBeneficiaryTerm.NewBeneficiary)
+
+		rt.SetEpoch(periodOffset + 10)
```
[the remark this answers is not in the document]
hunjixin: you means  not necessary to check pending beneficiary?

actors/builtin/miner/miner_test.go
```diff
@@ -2616,6 +2679,253 @@ func TestChangeMultiAddrs(t *testing.T) {
...
+
+		//used up quota
+		rt.SetEpoch(periodOffset + 1)
+		actor.withdrawFunds(rt, quota, quota, big.Zero())
+		info := actor.getInfo(rt)
+		require.EqualValues(t, true, info.BeneficiaryTerm.IsUsedUp())
+
+		//increase quota
+		increaseQuota := abi.NewTokenAmount(100)
+		actor.changeBeneficiary(rt, actor.owner, beneficiaryChange{firstBeneficiaryId, increaseQuota, expiration}, &firstBeneficiaryId)
```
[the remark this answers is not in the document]
hunjixin: i would set  a epoch after expiration epoch

actors/builtin/miner/miner_test.go
hunjixin: both id/address are okay, but here have a dup test item,  would remove one. and  add one test for beneficiary address

hunjixin: comment for each case

actors/builtin/miner/miner_test.go
```diff
@@ -2240,6 +2240,69 @@ func TestWithdrawBalance(t *testing.T) {
...
+		actor.proposeAndApproveInitialBeneficiary(rt, beneficiaryId, miner.BeneficiaryTerm{
+			Quota:      onePercentBalance,
+			Expiration: periodOffset,
+			UsedQuota:  big.Zero(),
+		})
+		// withdraw 1% of balance
+		actor.withdrawFunds(rt, big.Add(onePercentBalance, onePercentBalance), onePercentBalance, big.Zero())
+		actor.checkState(rt)
+	})
+
```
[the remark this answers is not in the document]
hunjixin: add expect abort to test used up quot and add another test for expiration

actors/builtin/miner/miner_actor.go
```diff
@@ -2014,14 +2024,26 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *abi.T
...
-	amountWithdrawn := big.Min(availableBalance, params.AmountRequested)
-	builtin.RequireState(rt, amountWithdrawn.GreaterThanEqual(big.Zero()), "negative amount to withdraw: %v", amountWithdrawn)
-	builtin.RequireState(rt, amountWithdrawn.LessThanEqual(availableBalance), "amount to withdraw %v < available %v", amountWithdrawn, availableBalance)
+		// To get the actual amount to be withdrawed
+		amountWithdrawn = big.Min(availableBalance, params.AmountRequested)
+		builtin.RequireState(rt, amountWithdrawn.GreaterThanEqual(big.Zero()), "negative amount to withdraw: %v", amountWithdrawn)
+		builtin.RequireState(rt, amountWithdrawn.LessThanEqual(availableBalance), "amount to withdraw %v < available %v", amountWithdrawn, availableBalance)
+		if info.Beneficiary != info.Owner {
+			remainingQuota := info.BeneficiaryTerm.Available(rt.CurrEpoch())
+			builtin.RequirePredicate(rt, remainingQuota.GreaterThan(big.Zero()), exitcode.ErrForbidden,
```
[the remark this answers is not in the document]
steven004: I would like to go fail-fast principle. And from saving network resource point of view, I would suggest both places use `>` since transferring 0 doing nothing but taking computing resource on each node.

actors/builtin/miner/miner_actor.go
```diff
@@ -3108,3 +3130,84 @@ func checkPeerInfo(rt Runtime, peerID abi.PeerID, multiaddrs []abi.Multiaddrs) {
...
+
+	var st State
+	rt.StateTransaction(&st, func() {
+		info := getMinerInfo(rt, &st)
+		if rt.Caller() == info.Owner {
+			// This is a ChangeBeneficiary proposal when the caller is Owner
+			if newBeneficiary != info.Owner {
+				if params.NewExpiration <= rt.CurrEpoch() {
+					rt.Abortf(exitcode.ErrIllegalArgument, "new beneficial expiration (%d)  must bigger than current epoch (%d)", params.NewExpiration, rt.CurrEpoch())
+				}
```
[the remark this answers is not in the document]
steven004: Good point. We can leave some judgement to users when there is no harm to the network, and gas is saved generally.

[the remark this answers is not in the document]
steven004: and yeah, I'll update the FIP to make it consistent with the implementation.

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
anorth: I see that, but I'm trying to make this more robust to future changes. Someone could change the code before to introduce a code path where it's not checked.

actors/builtin/miner/miner_test.go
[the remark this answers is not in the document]
anorth: I just mean I don't expect advancing the epoch to be necessary.

actors/builtin/miner/miner_actor.go
```diff
@@ -2014,14 +2024,26 @@ func (a Actor) WithdrawBalance(rt Runtime, params *WithdrawBalanceParams) *abi.T
...
 		feeToBurn = RepayDebtsOrAbort(rt, &st)
-	})
 
-	amountWithdrawn := big.Min(availableBalance, params.AmountRequested)
-	builtin.RequireState(rt, amountWithdrawn.GreaterThanEqual(big.Zero()), "negative amount to withdraw: %v", amountWithdrawn)
-	builtin.RequireState(rt, amountWithdrawn.LessThanEqual(availableBalance), "amount to withdraw %v < available %v", amountWithdrawn, availableBalance)
+		// To get the actual amount to be withdrawed
+		amountWithdrawn = big.Min(availableBalance, params.AmountRequested)
+		builtin.RequireState(rt, amountWithdrawn.GreaterThanEqual(big.Zero()), "negative amount to withdraw: %v", amountWithdrawn)
+		builtin.RequireState(rt, amountWithdrawn.LessThanEqual(availableBalance), "amount to withdraw %v < available %v", amountWithdrawn, availableBalance)
```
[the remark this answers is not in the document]
hunjixin: remove  line 2031

actors/builtin/miner/miner_actor.go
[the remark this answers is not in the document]
hunjixin: i see,

ZenGround0:
Landed in builtin actors, repo is deprecated

PR #1575, opened 2022-02-22
PreCommitDeposit is now independent of the verified status

Kubuxu:
Deal data is still fetched to check if deals don't occupy more space than
the room in the sector. This check can be delayed to ProveCommit.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -406,15 +406,32 @@ func (a Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
 	}
 }
 
-//type ActivateDealsParams struct {
-//	DealIDs      []abi.DealID
-//	SectorExpiry abi.ChainEpoch
-//}
-type ActivateDealsParams = market0.ActivateDealsParams
+type ActivateDealsParams struct {
+	Sectors []ActivateDealsParamsInner
+}
+
+//cbor-gen is annoying
+type ActivateDealsParamsInner struct {
+	DealIDs      []abi.DealID
+	SectorExpiry abi.ChainEpoch
+}
+
+type ActivateDealsReturn struct {
+	Sectors []ActivateDealsReturnInner
+}
+
+type ActivateDealsReturnInner struct {
+	Success            bool           // True if activation was a success
+	DealSpace          uint64         // Total space in bytes of submitted deals.
+	DealWeight         abi.DealWeight // Total space*time of submitted deals.
+	VerifiedDealWeight abi.DealWeight // Total space*time of submitted verified deals.
+}
 
 // Verify that a given set of storage deals is valid for a sector currently being ProveCommitted,
 // update the market's internal state accordingly.
-func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *abi.EmptyValue {
+// TODO: ActivateDeals somehow needs to check DealSpace <= SectorSize
+// currently it was done at precommit
+func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *ActivateDealsReturn {
 	rt.ValidateImmediateCallerType(builtin.StorageMinerActorCodeID)
 	minerAddr := rt.Caller()
 	currEpoch := rt.CurrEpoch()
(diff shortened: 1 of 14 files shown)


anorth:
Thanks for pointing out the deal space, and general deal failure cases. I forgot to mention that in the FIP - will update it. Yes, I think we should defer the check until ProveCommit, and hence move all the weight return values to `ActivateDeals` etc as described in the FIP PR.

actors/builtin/miner/miner_actor.go
```diff
@@ -2169,18 +2235,21 @@ func (a Actor) ProveReplicaUpdates(rt Runtime, params *ProveReplicaUpdatesParams
...
 		code := rt.Send(
 			builtin.StorageMarketActorAddr,
 			builtin.MethodsMarket.ActivateDeals,
 			&market.ActivateDealsParams{
-				DealIDs:      update.Deals,
-				SectorExpiry: sectorInfo.Expiration,
+				Sectors: []market.ActivateDealsParamsInner{{
+					DealIDs:      update.Deals,
+					SectorExpiry: sectorInfo.Expiration,
+				}},
```
anorth, line 2246: We should batch this too, and remove the call to `requestDealWeights` below in favour of using this return value. That's a nice saving.

actors/builtin/miner/miner_actor.go
```diff
@@ -2813,6 +2894,25 @@ func requestUnsealedSectorCIDs(rt Runtime, dataCommitmentInputs ...*market.Secto
...
+		&dealWeights,
+	)
+	builtin.RequireSuccess(rt, code, "failed to verify deals and get deal weight")
+	if len(dealWeights.Sectors) != len(sectors) {
+		rt.Abortf(exitcode.ErrIllegalState, "deal weight request returned %d records, expected %d",
+			len(dealWeights.Sectors), len(sectors))
+	}
+}
+
+// TODO remove??
```
anorth, line 2915: Yes I think so. We should only need the weights after activating.

actors/builtin/miner/miner_state.go
```diff
@@ -133,28 +133,29 @@ type WorkerKeyChange struct {
...
-	Info               SectorPreCommitInfo
-	PreCommitDeposit   abi.TokenAmount
-	PreCommitEpoch     abi.ChainEpoch
-	DealWeight         abi.DealWeight // Integral of active deals over sector lifetime
-	VerifiedDealWeight abi.DealWeight // Integral of active verified deals over sector lifetime
+	Info             SectorPreCommitInfo
+	PreCommitDeposit abi.TokenAmount
+	PreCommitEpoch   abi.ChainEpoch
+	//DealWeight         abi.DealWeight // Integral of active deals over sector lifetime
+	//VerifiedDealWeight abi.DealWeight // Integral of active verified deals over sector lifetime
```
anorth, line 158: If we do this, we'll need to migrate existing pre-commits in state. In the FIP I'd suggested just setting to zero.

The migration might be reasonable since there won't be many, and the clarity of not having this field is good. So 👍 for now.

actors/builtin/miner/miner_state.go
```diff
@@ -133,28 +133,29 @@ type WorkerKeyChange struct {
 	EffectiveAt abi.ChainEpoch
 }
 
+// possible states
+// (DealIDs > 0, UnsealedCID != nil) => CommD provided by SP, needs to be checked
+// (DealIDs > 0, UnsealedCID == nil) => Old method or migration, DealIDs checked at PreCommit (Depracated)
+// (DealIDs == 0, UnsealedCID == nil) => NoDeals, CommD of empty sector
+// (DealIDs == 0, UnsealedCID != 0) future approach, fail for now
```
anorth, line 140: Having seen the code now, I would really like to reduce the set of possibilities here by denying the nil unsealed CID. For the CommD of empty sector, we can use a placeholder just to save the bytes. For the old method let's fill it in before persisting this object to state. I'll write more in overall review comment.

anorth (commented):
Thanks for this Kuba. There's nothing like seeing the actual code to reveal all the hidden complexities!

I think we can simplify a bit here, especially to reduce the branching in later parts of the flow. How about something like this:
- For callers of the old method, we call verifyDeals as before, but also for empty sectors, and have that also compute and return the CommDs. Then persist them into the PreCommitInfo.
- For the new method, require CommD to be specified when pre-committing (w/ placeholder for CC sectors). Call the same verification method to verify the deals _and_ validate the resulting CommD matches.
- Then the CommD is always present and already validated at prove-commit, so remove the call to requestUnsealedSectorCIDs in ProveCommitAggregate, and via getVerifyInfo in ProveCommit.

This means the new method is no longer saving on loading the deal metadata during precommit. Upon consideration, I think this is worthwhile to retain the ergonomics of early failure on bad deals. It'll disappear later.

When we later make the architecture changes to specify deals only at prove-commit, we won't be calling the market to compute CommD anyway so we can drop all the pre-commit validation.

actors/builtin/miner/miner_state.go
Kubuxu: We need to migrate the state either way to add the CommD.

actors/builtin/market/market_actor.go
```diff
@@ -406,66 +406,116 @@ func (a Actor) VerifyDealsForActivation(rt Runtime, params *VerifyDealsForActiva
...
+	DealSpace          uint64         // Total space in bytes of submitted deals.
+	DealWeight         abi.DealWeight // Total space*time of submitted deals.
+	VerifiedDealWeight abi.DealWeight // Total space*time of submitted verified deals.
+}
 
 // Verify that a given set of storage deals is valid for a sector currently being ProveCommitted,
 // update the market's internal state accordingly.
-func (a Actor) ActivateDeals(rt Runtime, params *ActivateDealsParams) *abi.EmptyValue {
+// TODO: ActivateDeals somehow needs to check DealSpace <= SectorSize
+// currently it was done at precommit
```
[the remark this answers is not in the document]
Kubuxu: now I remember

actors/builtin/miner/miner_actor.go
```diff
@@ -676,6 +677,40 @@ type PreCommitSectorBatchParams = miner5.PreCommitSectorBatchParams
 // This method calculates the sector's power, locks a pre-commit deposit for the sector, stores information about the
 // sector in state and waits for it to be proven or expire.
 func (a Actor) PreCommitSectorBatch(rt Runtime, params *PreCommitSectorBatchParams) *abi.EmptyValue {
+	newSectors := make([]SectorPreCommitInfo, 0, len(params.Sectors))
+	for _, pi := range params.Sectors {
+		if pi.ReplaceCapacity {
+			rt.Abortf(exitcode.SysErrForbidden, "cc upgrade through precommit discontinued, use lightweight cc upgrade instead")
+		}
+		newSectors = append(newSectors, SectorPreCommitInfo{
```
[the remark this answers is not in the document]
Kubuxu: it is 0 length slice so I have to append

actors/builtin/miner/miner_state.go
Kubuxu: Issue them becomes how do we handle the migration, do we want to compute CommDs during it?
If not we have to allow for nil CommD.

anorth: Yes, we can compute them.

ZenGround0:
Archiving repo

PR #1576, opened 2022-03-01
Actors v8

arajasek:
This is in support of [FIP-0031](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0031.md). Although the FIP replaces the primary usecases of this repo with WASM actors, this change is still needed to support:

- the migration
- Go-based state reads 

This PR:

- Bumps the version and code CIDs
  - Note that the synthetic code CIDs are expected to be replaced by content addressed CIDs
- Introduces the simple migration needed to migrate code CIDs


Stebalien (approved):
The next step is deleting the actors themselves (except the state objects).

gen/gen.go
```diff
@@ -119,7 +117,7 @@ func main() {
 		// method params and returns
 		//paych.ConstructorParams{}, // Aliased from v0
 		paych.UpdateChannelStateParams{}, // Changed in v7
-		paych.SignedVoucher{}, // Changed in v7
+		paych.SignedVoucher{},            // Changed in v7
```
anorth, line 120: Can you alias these (and below) below onto their v7 definitions?

anorth (approved):
Looks thorough, thanks.

anorth:
> deleting the actors themselves (except the state objects)

@Stebalien whoah hold up, can we talk about this please? I understand why you say that, but I think there are better alternatives. cc @ZenGround0 

@Kubuxu after this lands you probably wanna rebase your FIP work onto v8.

anorth:
> Introduces the simple migration needed to migrate code CIDs

Consider doing this in a separate PR to keep this one purely mechanical. I didn't notice the new code when scanning.

gen/gen.go
arajasek: Yes, thank you for the catch!

PR #1580, opened 2022-03-12
Change Label to be a Union of strings and bytes

arajasek:
Implements FIP-0027. Will need a migration.


actors/builtin/market/deal.go
```diff
@@ -12,6 +22,147 @@ import (
...
+// The DealLabel is a "kinded union" -- can either be a String or a byte slice, but not both
+// TODO: What represents the empty label and how do we marshall it?
+type DealLabel struct {
+	labelString string
+	labelBytes  []byte
+}
+
+var EmptyDealLabel = DealLabel{}
+
+func NewDealLabelFromString(s string) (DealLabel, error) {
```
anorth, line 34: Nit: `deal.NewLabelFromString`, don't repeat the word "deal".

anorth (commented):
Thanks, very helpful.

actors/builtin/market/deal.go
```diff
@@ -12,6 +22,147 @@ import (
 //}
 var PieceCIDPrefix = market0.PieceCIDPrefix
 
+// The DealLabel is a "kinded union" -- can either be a String or a byte slice, but not both
+// TODO: What represents the empty label and how do we marshall it?
```
Stebalien, line 26: I'd either put these behind pointers, or store some kind of empty/string/bytes "tag" field.

actors/builtin/market/deal.go
```diff
@@ -12,6 +22,147 @@ import (
...
+	labelBytes  []byte
+}
+
+var EmptyDealLabel = DealLabel{}
+
+func NewDealLabelFromString(s string) (DealLabel, error) {
+	if len(s) > DealMaxLabelSize {
+		return EmptyDealLabel, xerrors.Errorf("provided string is too large to be a label (%d), max length (%d)", len(s), DealMaxLabelSize)
+	}
+	// TODO: Check UTF-8
```
Stebalien, line 38: https://pkg.go.dev/unicode/utf8#ValidString

actors/builtin/market/deal.go
```diff
@@ -12,6 +22,147 @@ import (
...
+	if label.IsBytesSet() {
+		return "", xerrors.Errorf("label has bytes set")
+	}
+
+	return label.labelString, nil
+}
+
+func (label DealLabel) ToBytes() ([]byte, error) {
+	if label.IsStringSet() {
+		return nil, xerrors.Errorf("label has string set")
```
Stebalien, line 72: Why not just convert it?

ZenGround0:
Can't tag you for review since you opened this @arajasek but please review this too

rvagg (approved):
👌 very nice, I'm excited about this

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,188 @@ import (
...
+// The DealLabel is a kinded union of string or byte slice.
+// It serializes to a CBOR string or CBOR byte string depending on which form it takes.
+// The empty label is serialized as an empty CBOR string (maj type 3).
+type DealLabel struct {
+	s  *string
+	bs *[]byte
+}
+
+// Zero value of DealLabel is canonical EmptyDealLabel but three different values serialize the same way:
+// DealLabel{nil, nil}, (*DealLabel(nil), and DealLabel{&"", nil})
```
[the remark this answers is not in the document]
ZenGround0: this is the syntax of typed nil in golang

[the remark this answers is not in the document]
ZenGround0: thanks!

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,184 @@ import (
 //}
 var PieceCIDPrefix = market0.PieceCIDPrefix
 
+// The DealLabel is a kinded union of string or byte slice.
+// It serializes to a CBOR string or CBOR byte string depending on which form it takes.
+// The empty label is serialized as an empty CBOR string (maj type 3).
+type DealLabel struct {
+	s  *string
+	bs *[]byte
+}
```
Stebalien, line 33: personal opinion: I'd write this as

```
type DealLabel struct {
	bs []byte
	notString bool
}
```

Your version is more "go-like", but also requires extra allocations and may be a bit annoying to deal with...

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,184 @@ import (
...
+
+	return DealLabel{
+		bs: &b,
+	}, nil
+}
+
+// True iff neither bytes nor string is set
+func (label DealLabel) IsEmpty() bool {
+	return label.s == nil && label.bs == nil
+}
```
Stebalien, line 64: Is this accurate? I'd assume that `IsEmpty` would mean "is empty". I.e., `""` would also be empty.

Stebalien (approved):
I would have done a few things a bit differently, but this looks correct.

Stebalien:
> Yeah its intentional. Its a type thing not a value thing. Empty means contains no string or bytes and the empty string is a string. EmptyLabel serializes to a string but its not a string. This is all just shorthand to be a bit clearer than comparing both fields to nil in a few places. I'll add a comment but don't think this needs to change. Your suggestion is outdated with the pointer representation I changed to on steb's suggestion. b and s are pointers so never have length.

We shouldn't distinguish between the two kinds of empty. That is, `DealLabel{}` and `DealLabel{s: ""}` should be indistinguishable.

arajasek:
Sorry, I'm still confused about the emptiness thing. Some qs:

- Does the canonical `EmptyDealLabel` change if it's serialized and deserialized? I think it might go from `nil, nil` to `&"", nil`, which seems odd if it's canonical?
- Can the `IsEmpty()` property change if a `DealLabel` is serialized and deserialized? I think it should be a requirement that none of the query-able properties change through the serialization-deserialization process, but I think that's not always the case.

I might be wrong about some of these claims.

actors/builtin/market/deal.go
ZenGround0: Ironically I was following up on your comment [here](https://github.com/filecoin-project/specs-actors/pull/1580#discussion_r826315784) is this not the "put them behind pointers" approach?

ZenGround0: I'm going to move it to the above approach

Stebalien: > Ironically I was following up on your comment https://github.com/filecoin-project/specs-actors/pull/1580#discussion_r826315784 is this not the "put them behind pointers" approach?

Your version is the "put them behind pointers approach". My version here is the "enum thing" approach (which, after looking at this, seems simpler.

> nvm notString is the way it is to make this nice

Welcome to go! zero values are over there :point_right: .

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,167 @@ import (
...
+	} else if label.IsBytes() {
+		if len(label.bs) > cbg.ByteArrayMaxLen {
+			return xerrors.Errorf("labelBytes is too long to marshal (%d), max allowed (%d)", len(label.bs), cbg.ByteArrayMaxLen)
+		}
+
+		if err := cbg.WriteMajorTypeHeaderBuf(scratch, w, cbg.MajByteString, uint64(len(label.bs))); err != nil {
+			return err
+		}
+
+		if _, err := w.Write((label.bs)[:]); err != nil {
```
Stebalien, line 118:
```suggestion
		if _, err := w.Write(label.bs); err != nil {
```

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,167 @@ import (
...
+			return err
+		}
+	} else { // label.IsString()
+		if len(label.bs) > cbg.MaxLength {
```
Stebalien, line 122: Se use `MaxByteArrayLen` in both cases (on further inspection). I'd just lift this case to the top.

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,167 @@ import (
...
+
+func (label DealLabel) Length() int {
+	return len(label.bs)
+}
+
+func (l DealLabel) Equals(o DealLabel) bool {
+	return bytes.Equal(l.bs, o.bs) && l.notString == o.notString
+}
+
+func (label *DealLabel) MarshalCBOR(w io.Writer) error {
```
Stebalien, line 94: I think we can vastly simplify this now:

1. If null, write empty string.
2. If too large, return an error.
3. Set a `majorType` variable depending on the type.
4. Write the value (no need for a conditional).

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,167 @@ import (
...
+
+		if length > cbg.ByteArrayMaxLen {
+			return fmt.Errorf("label bytes was too long (%d), max allowed (%d)", length, cbg.ByteArrayMaxLen)
+		}
+
+		bs := make([]uint8, length)
+		label.bs = bs
+		label.notString = true
+		if _, err := io.ReadFull(br, bs[:]); err != nil {
+			return err
```
Stebalien, line 179: We should be able to unify these codepaths.

1. Make sure we have either a text string or a byte string.
2. Set "not string" if it's a byte string.
3. Check the length.
4. Read the value.
5. If a string, check if it's utf8.

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,145 @@ import (
...
+		if len(label.bs) > cbg.ByteArrayMaxLen {
+			return xerrors.Errorf("label is too long to marshal (%d), max allowed (%d)", len(label.bs), cbg.ByteArrayMaxLen)
+		}
+
+		var majorType byte
+		if label.IsString() {
+			majorType = cbg.MajTextString
+		} else {
+			majorType = cbg.MajByteString
+		}
```
Stebalien, line 116:
```suggestion
		majorType := cbg.MajByteString
		if label.IsString() {
			majorType = cbg.MajTextString
		}
```

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,145 @@ import (
...
+	// on chain structures should never have a pointer to a DealLabel but the case is included for completeness
+	if label == nil {
+		if err := cbg.WriteMajorTypeHeaderBuf(scratch, w, cbg.MajTextString, 0); err != nil {
+			return err
+		}
+		if _, err := io.WriteString(w, string("")); err != nil {
+			return err
+		}
+		return nil
+	} else {
```
Stebalien, line 106: No need for the else.

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,145 @@ import (
...
+	// nil *DealLabel counts as EmptyLabel
+	// on chain structures should never have a pointer to a DealLabel but the case is included for completeness
+	if label == nil {
+		if err := cbg.WriteMajorTypeHeaderBuf(scratch, w, cbg.MajTextString, 0); err != nil {
+			return err
+		}
+		if _, err := io.WriteString(w, string("")); err != nil {
+			return err
+		}
+		return nil
```
Stebalien, line 105: Can just return.

actors/builtin/market/deal.go
```diff
@@ -12,6 +24,145 @@ import (
...
+	buf := make([]byte, length)
+	_, err = io.ReadAtLeast(br, buf, int(length))
+	if err != nil {
+		return err
+	}
+	label.bs = buf
+	label.notString = maj != cbg.MajTextString
+	if !label.notString {
+		s := string(buf)
+		if !utf8.ValidString(s) {
```
Stebalien, line 158:
```suggestion
	if !label.notString && !utf8.ValidString(string(buf)) {
```

Stebalien (approved):
code nits but otherwise LGTM

PR #1582, opened 2022-03-24
Fix bug

ZenGround0:
Fixes bug in #1577 

Please look over and let me know if you agree on the general approach.  If so I'll port this over to rust.


jennijuju:
haha better PR title please!

Stebalien (approved):
This should work.

PR #1583, opened 2022-04-01
Implement nv16 migration according to FIP31

vyzo:
- Adds the manifest data type
- Adds the system actor state
- Implements the migration


actors/migration/nv16/top.go
```diff
@@ -289,17 +310,33 @@ func (job *migrationJob) run(ctx context.Context, store cbor.IpldStore, priorEpo
...
+func (n codeMigrator) migrateState(_ context.Context, _ cbor.IpldStore, in actorMigrationInput) (*actorMigrationResult, error) {
 	return &actorMigrationResult{
 		newCodeCID: n.OutCodeCID,
 		newHead:    in.head,
 	}, nil
 }
 
-func (n nilMigrator) migratedCodeCID() cid.Cid {
-	return n.OutCodeCID
+// System Actor migrator
```
arajasek, line 324: Nit: the historic pattern has been to put this in its own file (eg https://github.com/filecoin-project/specs-actors/blob/release/v7/actors/migration/nv15/verifreg.go)

arajasek (commented):
First-pass review this makes sense, need to look at it more closely later.

Definitely needs tests.

actors/migration/nv16/top.go
vyzo: ok, will do.

vyzo: done.

actors/builtin/manifest/manifest.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"fmt"
+
+	adt8 "github.com/filecoin-project/specs-actors/v8/actors/util/adt"
+
+	"github.com/ipfs/go-cid"
+)
+
+type Manifest struct {
```
ZenGround0, line 12: If you keep this structure this should move to the migration package as that's the only code that uses (ever will use?) this type.

actors/migration/nv16/system.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// System Actor migrator
+type systemActorMigrator struct {
+	OutCodeCID   cid.Cid
+	ManifestData cid.Cid
+}
+
+func (m systemActorMigrator) migrateState(ctx context.Context, store cbor.IpldStore, in actorMigrationInput) (*actorMigrationResult, error) {
+	state := system8.State{BuiltinActors: m.ManifestData}
```
ZenGround0, line 18: nit: If you give the migrator the manifest data itself you can write the data to the store here for extra clarity that we're not writing any unreachable ids

actors/migration/nv16/system.go
```diff
@@ -0,0 +1,29 @@
...
+	stateHead, err := store.Put(ctx, &state)
+
```
ZenGround0, line 20: nit: kill space

actors/migration/nv16/test/miner_migration_test.go
```diff
@@ -146,11 +150,48 @@ func createMinersAndSectorsV7(t *testing.T, ctx context.Context, ctxStore adt.St
...
+
+	manifestCid, err := adtStore.Put(context.Background(), &manifest)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return manifestCid
+}
+
 func TestNv16Migration(t *testing.T) {
```
ZenGround0, line 189: You should remove all the test logic specific to nv15 and replace it with a check that the system actor state looks ok.

actors/migration/nv16/test/miner_migration_test.go
```diff
@@ -146,11 +150,48 @@ func createMinersAndSectorsV7(t *testing.T, ctx context.Context, ctxStore adt.St
 	return append(minersToProve, minerInfos...), v
 }
 
+func makeManifest(t *testing.T, store cbor.IpldStore) cid.Cid {
```
ZenGround0, line 153: `makeFakeManifest` `makeTestCIDsManifest`.. something to indicate its not real

support/vm/testing.go
```diff
@@ -62,7 +63,12 @@ func NewVMWithSingletons(ctx context.Context, t testing.TB, bs ipldcbor.IpldBloc
 	store := adt.WrapBlockStore(ctx, bs)
 	vm := NewVM(ctx, lookup, store)
 
-	initializeActor(ctx, t, vm, &system.State{}, builtin.SystemActorCodeID, builtin.SystemActorAddr, big.Zero())
+	manifestDataCid, err := store.Put(ctx, &manifest.ManifestData{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	initializeActor(ctx, t, vm, &system.State{BuiltinActors: manifestDataCid}, builtin.SystemActorCodeID, builtin.SystemActorAddr, big.Zero())
```
ZenGround0, line 71: instead follow the same pattern: call construct state which you have written to do this empty manifest writing for you and then call initializeActor on the systemState variable output from construcstate

actors/migration/nv16/top.go
```diff
@@ -66,38 +66,62 @@ func ActorHeadKey(addr address.Address, head cid.Cid) string {
...
+func MigrateStateTree(ctx context.Context, store cbor.IpldStore, actorsManifest cid.Cid, actorsRootIn cid.Cid, priorEpoch abi.ChainEpoch, cfg Config, log Logger, cache MigrationCache) (cid.Cid, error) {
 	if cfg.MaxWorkers <= 0 {
 		return cid.Undef, xerrors.Errorf("invalid migration config with %d workers", cfg.MaxWorkers)
 	}
 
+	adtStore := adt8.WrapStore(ctx, store)
+
+	// load manifest
+	var manifest manifest8.Manifest
+	if err := adtStore.Get(ctx, actorsManifest, &manifest); err != nil {
```
ZenGround0, line 78: bike shed: I'm not convinced of the value of the separate versioned manifest object that only comes in during the migration and doesn't make it on chain.  If we want the version for the migration why not on chain?  Even better, if we don't care about the version on chain why include it in the migration at all? I don't see the point of the version and outer wrapper why not just pass the manifest data cid.

actors/migration/nv16/test/miner_migration_test.go
```diff
@@ -146,11 +150,48 @@ func createMinersAndSectorsV7(t *testing.T, ctx context.Context, ctxStore adt.St
...
+	adtStore := adt.WrapStore(context.Background(), store)
+	builder := cid.V1Builder{Codec: cid.Raw, MhType: mh.IDENTITY}
+
+	manifestData := manifest8.ManifestData{}
+	for _, name := range []string{"system", "init", "cron", "account", "storagepower", "storageminer", "storagemarket", "paymentchannel", "multisig", "reward", "verifiedregistry"} {
+		codeCid, err := builder.Sum([]byte(fmt.Sprintf("fil/8/%s", name)))
```
ZenGround0, line 159: paranoia: maybe make it something slightly different to make sure we're not running into usage of the old cid pattern: "fil/8/test/%s"

actors/migration/nv16/top.go
ZenGround0: At the end of the day its a few bytes so nbd

actors/migration/nv16/top.go
```diff
@@ -66,38 +66,62 @@ func ActorHeadKey(addr address.Address, head cid.Cid) string {
...
+		code8Cid, ok := manifest.Get(name)
+		if !ok {
+			return cid.Undef, xerrors.Errorf("code cid for %s actor not found in manifest", name)
+		}
+
+		migrations[code7Cid] = codeMigrator{code8Cid}
 	}
 
-	// Set of prior version code CIDs for actors to defer during iteration, for explicit migration afterwards.
-	var deferredCodeIDs = map[cid.Cid]struct{}{
```
ZenGround0, line 90: Don't remue this please its been helpful in the past and nice to keep easily available for future migrations.

actors/migration/nv16/top.go
```diff
@@ -121,9 +145,6 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, actorsRootIn ci
 		defer close(jobCh)
 		log.Log(rt.INFO, "Creating migration jobs for tree %s", actorsRootIn)
 		if err = actorsIn.ForEach(func(addr address.Address, actorIn *states7.Actor) error {
-			if _, ok := deferredCodeIDs[actorIn.Code]; ok {
```
ZenGround0, line 124: ditto

ZenGround0 (commented):
Once there is a test checking that the system actor state matches the manifest I'll approve, the rest of the comments are all nice to haves.

actors/builtin/manifest/manifest.go
```diff
@@ -0,0 +1,49 @@
...
+	m.entries = make(map[string]cid.Cid)
+	for _, e := range data.Entries {
+		m.entries[e.Name] = e.Code
+	}
+
+	return nil
+}
+
+func (m *Manifest) Get(name string) (cid.Cid, bool) {
+	c, ok := m.entries[name]
```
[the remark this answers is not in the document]
vyzo: We'd have to pass the store in get to do lazy load.

Also, there is no panic, nil map behaves as the empty map!

actors/builtin/system/system_actor.go
```diff
@@ -34,8 +35,11 @@ var _ runtime.VMActor = Actor{}
...
 
-	rt.StateCreate(&State{})
+	empty := rt.StorePut(&manifest.ManifestData{})
+	rt.StateCreate(&State{BuiltinActors: empty})
 	return nil
 }
 
-type State struct{}
+type State struct {
+	BuiltinActors cid.Cid
```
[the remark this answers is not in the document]
vyzo: sure

actors/builtin/manifest/manifest.go
vyzo: I think we want to keep it for completeness.

Note that the constituents (ManifestData, ManifestEntry) are part of the system actor state now, so they have to be here.
So I'd rather keep everything here as oppose to bifurcating and having half the stuff here and the outer struct in migrations.

It also gives us a nice entry point for go code that wants to read manifests.

actors/migration/nv16/system.go
vyzo: Maybe.

But the data should already be in the store, thats where we got them in the first place.

I will add a comment.

actors/migration/nv16/system.go
vyzo: yessir.

actors/migration/nv16/test/miner_migration_test.go
vyzo: ok, will do.

actors/migration/nv16/test/miner_migration_test.go
vyzo: ok.

support/vm/testing.go
vyzo: ok.

actors/migration/nv16/top.go
vyzo: two things here:
- caller doesnt have to parse
- we need to know the version to parse

So i think its much cleaner to pass the cid to the whole manifest to us  as opposed to complicating callers and having to pass the version as well.
And if we ever change the manifest format, user code wont have to change.

actors/migration/nv16/test/miner_migration_test.go
vyzo: that breaks the state invariant checks, we will probably want to use them in future migrations.

actors/migration/nv16/top.go
vyzo: ok, i will keep it.

actors/migration/nv16/top.go
vyzo: ok.

actors/migration/nv16/test/miner_migration_test.go
vyzo: I'll call it `makeTestManifest`.

actors/migration/nv16/test/miner_migration_test.go
vyzo: done.

support/vm/testing.go
vyzo: done.

actors/builtin/manifest/manifest.go
vyzo: Also note, we'll be using it in every subsequent migration from now on.

actors/builtin/manifest/manifest.go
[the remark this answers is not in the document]
ZenGround0: nice TIL

actors/migration/nv16/top.go
ZenGround0: Why we aren't keeping the version on chain too?

actors/migration/nv16/test/miner_migration_test.go
ZenGround0: Not sure I understand I feel like that's exactly the kind of thing I was hoping to flush out.  We should fix the state invariant checks to work with the correct CIDS / parameterize them for testing.  We use state invariant checks on main net state not just in testing.

actors/migration/nv16/top.go
vyzo: because we dont have to?
The data format, and hence the state of the system actor, is imploed by the network version.

actors/migration/nv16/test/miner_migration_test.go
vyzo: wait, this code will never run again after this migration.

Or do you want to keep around the state checking for post migration offline work?

vyzo: so what do you want me to do here?

vyzo: also note, we can read the state of the system actor to find the actual code cids, so no need to do any extra parameterization.

actors/migration/nv16/top.go
raulk: In theory, the version on the manifest is the version of the **exchange, portable format** (i.e. how the manifest is encoded in the bundle). It is entirely possible for the structure on chain to diverge from the bundle structure. The structure on chain would evolve through network versions; the structure in the bundle would evolve through the manifest version.

The code in this PR has some implicit coupling between the two, such as the system actor state constructor forming an empty `ManifestData`, or the getter loading a `ManifestData`, but it's not too harmful given that specs-actors only exists from now onwards to support migrations and provide Go types.

actors/migration/nv16/test/miner_migration_test.go
raulk: Could @vyzo or @ZenGround0 post a link to the state invariant checks we're talking about? In builtin-actors we use [static `/fil/test/<type>` code CIDs](https://github.com/filecoin-project/builtin-actors/blob/5396be01b74f8a90517f31e4db2c0f13f85a3f4c/actors/runtime/src/test_utils.rs#L35-L46) to refer to test actors. This isn't quite the same situation since we're dealing with a CodeCID migration here, so may want to have a "before" and an "after" set? However, without knowing which checks you're talking about, I can't really give an opinion here.

vyzo: @raulk https://github.com/filecoin-project/specs-actors/blob/master/actors/states/check.go#L28

actors/builtin/manifest/manifest.go
ZenGround0: It's confusing to keep it here since the data is not part of the system actor state. It would be clear to put it in the migration package.  The fact that it is in all subsequent migrations strengthens the argument for moving it there.

actors/migration/nv16/test/miner_migration_test.go
ZenGround0: > Or do you want to keep around the state checking for post migration offline work?

@vyzo the state invariant checks are used [here](https://github.com/filecoin-project/ent).  I'm now realizing this should migrate to use what we will build in builtin actors since specs-actors is retiring.  I'll add a comment to the builtin actors issue.

ZenGround0 (approved):
Only thing I still want changed is for the Manifest data type to be moved to the migration package but I won't block the PR on it.

PR #1587, opened 2022-04-07
Fix ManifestData marshalling

vyzo:
The manifest data is a flat vector of tuples, but our data type wraps another tuple around. 

This fixes it to do the right thing; unfortunately I had to write the marshaller/unmarshaller by hand, as cbor-gen doesn't support transparent structs.


vyzo:
rebased on master.

Stebalien (approved):

PR #1589, opened 2022-04-13
fix bug

Edison-Yeah:
When a transaction requiring a large amount of collateral is blocked, transactions requiring only a small amount of collateral from the same client submitted with the transaction are blocked.

### DIFF
--- a/actors/builtin/market/market_actor.go
+++ b/actors/builtin/market/market_actor.go
@@ -225,15 +225,15 @@ func (a Actor) PublishStorageDeals(rt Runtime, params *PublishStorageDealsParams
 		if _, ok := totalClientLockup[client]; !ok {
 			totalClientLockup[client] = abi.NewTokenAmount(0)
 		}
-		totalClientLockup[client] = big.Sum(totalClientLockup[client], deal.Proposal.ClientBalanceRequirement())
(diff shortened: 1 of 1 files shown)


vyzo (commented):
this is consensus breaking and the next upgrade will run the fvm, so this bug should be fixed in bultin-actors repo, if present.

Stebalien (commented):
This fix looks correct to me. cc @jennijuju because it's consensus critical.

anorth:
Please add a test demonstrating the issue fixed

ZenGround0 (commented):
LGTM
As others have pointed out this needs test ~and a corresponding change in builtin actors~ to make it into the next upgrade.
I would also appreciate more separation between names beyond swapping client/provider and total which makes things easy to confuse.  `totalXLockupTmp`, `totalXLockupNew`, `totalXLockupAcc` etc.

arajasek:
Thanks for the fix! I've transported this to the new Rust-based builtin-actors in https://github.com/filecoin-project/builtin-actors/pull/414, along with some tests!

arajasek:
I don't have an opinion on whether or not we should merge this PR, or just close it since this repo is no longer "specs-actors".

ZenGround0:
Archiving repo

PR #1590, opened 2022-04-22
Use new go-state-types state accessors

arajasek:
Needs https://github.com/filecoin-project/go-state-types/pull/36


anorth:
I think I am missing some context for this. I can't see the imported code in go-state-types.

If this is proposing to move the actor message and state types out of this repo, I'd like to discuss that first.

arajasek:
@anorth Are you unable to see [this PR](https://github.com/filecoin-project/go-state-types/pull/36/files)? 

The proposal is to (mostly) leave actors unchanged, but _add_ the state types to go-state-types. Go-based clients like Lotus can then sever their dependency on specs-actors entirely for v8 and onwards, while still having the ability to de/serialize state, message params, etc.

anorth:
Thanks, what was really missing is https://github.com/filecoin-project/lotus/issues/8607. That discussion has started multiple times in different places, thanks for giving it a canonical home (even if Lotus is an odd place for it).

go.sum
```diff
@@ -54,20 +55,21 @@ github.com/filecoin-project/go-state-types v0.0.0-20200928172055-2df22083d8ab/go
...
+github.com/filecoin-project/specs-actors/v2 v2.3.6/go.mod h1:DJMpxVRXvev9t8P0XWA26RmTzN+MHiL9IlItVLT0zUc=
 github.com/filecoin-project/specs-actors/v3 v3.1.0 h1:s4qiPw8pgypqBGAy853u/zdZJ7K9cTZdM1rTiSonHrg=
 github.com/filecoin-project/specs-actors/v3 v3.1.0/go.mod h1:mpynccOLlIRy0QnR008BwYBwT9fen+sPR13MA1VmMww=
 github.com/filecoin-project/specs-actors/v4 v4.0.0 h1:vMALksY5G3J5rj3q9rbcyB+f4Tk1xrLqSgdB3jOok4s=
 github.com/filecoin-project/specs-actors/v4 v4.0.0/go.mod h1:TkHXf/l7Wyw4ZejyXIPS2rK8bBO0rdwhTZyQQgaglng=
-github.com/filecoin-project/specs-actors/v5 v5.0.4 h1:OY7BdxJWlUfUFXWV/kpNBYGXNPasDIedf42T3sGx08s=
-github.com/filecoin-project/specs-actors/v5 v5.0.4/go.mod h1:5BAKRAMsOOlD8+qCw4UvT/lTLInCJ3JwOWZbX8Ipwq4=
-github.com/filecoin-project/specs-actors/v6 v6.0.0 h1:i+16MFE8GScWWUF0kG7x2RZ5Hqpz0CeyBHTpnijCJ6I=
-github.com/filecoin-project/specs-actors/v6 v6.0.0/go.mod h1:V1AYfi5GkHXipx1mnVivoICZh3wtwPxDVuds+fbfQtk=
+github.com/filecoin-project/specs-actors/v5 v5.0.4-0.20220511204328-97fa6fffc45a h1:VBrBfi+7i+IjSDdt7ezdF18ViS5D/Bdc5hRUVOLcfhs=
```
ZenGround0, line 69: What is going on here?

support/agent/blockstore_util.go
```diff
@@ -19,7 +21,7 @@ func BlockstoreCopy(from, to cbor.IpldBlockstore, root cid.Cid) (blocks uint64,
 	cp := func(blk block.Block) error {
 		numBlocks++
 		totalCopySize += uint64(len(blk.RawData()))
-		return to.Put(blk)
+		return to.Put(context.TODO(), blk)
```
ZenGround0, line 24: Why?

ZenGround0 (commented):
Why is cbor gen twiddling lines of code around?

actors/builtin/miner/miner_actor.go
```diff
@@ -330,7 +334,7 @@ type PoStPartition = miner0.PoStPartition
 //	// The ticket randomness on the chain at the chain commit epoch.
 //	ChainCommitRand abi.Randomness
 //}
-type SubmitWindowedPoStParams = miner0.SubmitWindowedPoStParams
+type SubmitWindowedPoStParams = minertypes.SubmitWindowedPoStParams
```
ZenGround0, line 337: Do these param substitutions need to happen here in the same way the proof types change needs to happen?

ZenGround0:
Can we avoid updating cbor-gen? Doing this as part of a non-breaking change scares me a decent amount.  If we need it to match go-state-types I'd prefer waiting until the next upgrade to update.

arajasek:
Yeah, we can avoid the bump -- will just need to downgrade it in go-state-types and cut a new release of that.

support/agent/blockstore_util.go
arajasek: Because go-state-types pulls in the context blockstores (and I don't want to undo that, it should be on the latest thing).

actors/builtin/miner/miner_actor.go
arajasek: No, they don't -- i think. Undoing.

actors/builtin/paych/paych_actor.go
```diff
@@ -150,14 +110,14 @@ func (t *SignedVoucher) SigningBytes() ([]byte, error) {
 //	// Pre-serialized method parameters.
 //	Params []byte
 //}
-type ModVerifyParams = paych0.ModVerifyParams
+type ModVerifyParams = paychtypes.ModVerifyParams
```
ZenGround0, line 113: I take it runtime verification needs these to be the same too?

gen/gen.go
```diff
@@ -15,23 +15,21 @@ import (
...
-		//proof.SectorInfo{},
-		proof.ExtendedSectorInfo{},
-		//proof.SealVerifyInfo{}, // Aliased from v0
-		//proof.PoStProof{}, // Aliased from v0
-		//proof.WindowPoStVerifyInfo{},
-		// proof.WinningPoStVerifyInfo{}, // Aliased from v0
-	); err != nil {
-		panic(err)
-	}
+	//if err := gen.WriteTupleEncodersToFile("./actors/runtime/proof/cbor_gen.go", "proof"); //proof.SectorInfo{},
```
ZenGround0, line 24: remove this section

ZenGround0 (approved):

ZenGround0:
Its worth a paranoid pass over all the cborgen of params moved to go-state-types to make sure nothing has changed.

PR #1591, opened 2022-04-22
FIP-0027: Add JSON un/marshaler for DealLabel

arajasek:
We only marshal DealLabels that are of type string -- we can add support for byte types later.

### DIFF
--- a/actors/builtin/market/deal_test.go
+++ b/actors/builtin/market/deal_test.go
@@ -2,7 +2,12 @@ package market_test
 
 import (
 	"bytes"
+	"encoding/json"
 	"testing"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/ipfs/go-cid"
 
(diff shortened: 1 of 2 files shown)


actors/builtin/market/deal_test.go
```diff
@@ -74,6 +79,41 @@ func TestDealLabel(t *testing.T) {
 	assert.Contains(t, err.Error(), "cannot unmarshal into nil pointer")
 
 }
+func TestDealLabelJSON(t *testing.T) {
```
ZenGround0, line 82: Test empty label too for more confidence.

ZenGround0 (approved):

actors/builtin/market/deal_test.go
ZenGround0: i.e. Proposal with Label field unset and defaulting to zero value

arajasek: Yessir

PR #1598, opened 2022-06-10
chore: migration: don't panic during migration

geoff-vball:


arajasek (approved):
Thanks! Can you please cut a new release of specs-actors v8 with this, and bubble that into Lotus?

PR #1602, opened 2022-09-21
fix: Use the can-proof-type funcs from the go-state-types package

diwufeiwen:
The custom proof-types of venus/lotus are registered to `go-state-types`, so the functions defined in `go-state-types` should also be used here:
lotus/chain/actors/policy/policy.go
```
import(
     miner8 "github.com/filecoin-project/go-state-types/builtin/v8/miner"
)

func AddSupportedProofTypes(types ...abi.RegisteredSealProof) {
	for _, t := range types {
		// other ...

		miner8.PreCommitSealProofTypesV8[t+abi.RegisteredSealProof_StackedDrg2KiBV1_1] = struct{}{}
		wpp, err = t.RegisteredWindowPoStProof()
		if err != nil {
			// Fine to panic, this is a test-only method
			panic(err)
		}

		miner8.WindowPoStProofTypes[wpp] = struct{}{}

	}
}

```

### DIFF
--- a/actors/builtin/miner/miner_actor.go
+++ b/actors/builtin/miner/miner_actor.go
@@ -12,6 +12,7 @@ import (
 	"github.com/filecoin-project/go-bitfield"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/big"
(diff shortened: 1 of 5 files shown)


ZenGround0:
Deprecating repo

PR #1605, opened 2023-06-05
update go-ipld-format to remove globals, and use go-ipld-cbor explicitly

aschmahmann:
Coming along the same set of changes as https://github.com/filecoin-project/lotus/pull/10921.

The change to go-ipld-format hasn't been merged + released yet but should soon so please let me know if there's a problem

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/ipfs/go-block-format v0.0.3
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-ipld-cbor v0.0.5
-	github.com/ipfs/go-ipld-format v0.0.2
+	github.com/ipfs/go-ipld-format v0.4.1-0.20230530195241-c3da01c74a06
 	github.com/ipld/go-car v0.1.0
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
 	github.com/minio/sha256-simd v0.1.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/ipfs/go-block-format v0.0.3
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-ipld-cbor v0.0.5
-	github.com/ipfs/go-ipld-format v0.0.2
+	github.com/ipfs/go-ipld-format v0.4.1-0.20230530195241-c3da01c74a06
```
aschmahmann, line 21: TODO: use released version

support/vm/vector.go
```diff
@@ -334,7 +335,7 @@ func (a *adtNodeGetter) Get(ctx context.Context, c cid.Cid) (format.Node, error)
 	if err != nil {
 		return nil, err
 	}
-	return format.Decode(b)
+	return format.Decode(b, ipldcbor.DecodeBlock)
```
aschmahmann, line 338: 1. It seems like you were only using dag-cbor here, is that true?
2. Would you find it more helpful to either
     a) just do `ipldcbor.DecodeBlock(b)` instead of using the helper in go-ipld-format for checking if the Block is already a Node?
      b) have a registry object you can create and use (either as a global in this file or passed into the constructor) so that you get `myRegistry.Decode(b)`

ZenGround0:
Closing down because we are archiving

